PR #80, opened 2014-04-14
Add SkipFlagParsing option to Command

wendorf:
We would like to be able to have command arguments that begin with a dash, but that currently triggers the flag parsing and results in an error that our flag was not defined.

We solved this problem by adding an optional Command option called SkipFlagParsing which, if present, disables all flag parsing and allows "normal" (e.g. non-flag) arguments to begin with a dash.

### DIFF
--- a/command.go
+++ b/command.go
@@ -22,6 +22,8 @@ type Command struct {
 	Action func(context *Context)
 	// List of flags to parse
 	Flags []Flag
+	// Treat all flags as normal arguments if true
+	SkipFlagParsing bool
 }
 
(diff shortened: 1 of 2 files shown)


jszwedko:
I feel that the default flag parsing behavior is correct and consistent with other command line interfaces (e.g. consider `grep --foo *`). If you need to specify an argument with a flag, you should use `--` to indicate the end of the flag list. I'm going to open a PR to revert this behaviour, but we'll need to be careful about merging it since it will break compatibility.

PR #82, opened 2014-04-16
A better version of infinite subcommands

smousa:
This version is more consistent with the rest of the application.  To add recursively add a sub-sub-command, populate the Subcommands field in the Command object.


sthulb:
There's a bug that was introduced with this PR.

The `if len(c.Subcommands) > 0 || c.Before != nil {` line causes `startApp` to be run if `Before` is specified, but the `startApp` func causes the `app.Name` to be altered for the `Action`. This is somewhat problematic if you don't have subcommands and still use the `Before` option.

PR #127, opened 2014-08-10
correct issue #103

### LINKED ISSUE #103: Reordering of args to Command if BoolFlag is after args
if a `Command` is invoked  as `cli cmd arg1 -f arg2` where `-f` is a `BoolFlag` the result of `ctx.Args()` in the command function will be `[arg2 arg1]`. 

This appears to be related to #36 

I have made the test case below (which currrently fails), but I am not really able to find a decent fix.

```
func TestApp_CommandWithBoolFlagAfterArgs(t *testing.T) {

	var args []string

	app := cli.NewApp()
	command := cli.Command{
		Name: "cmd",
		Flags: []cli.Flag{
			cli.BoolFlag{
				Name:  "f",
				Usage: "foo",
			},
		},
			args = c.Args()
		},
	}
	app.Commands = []cli.Command{command}
	app.Run([]string{"", "cmd", "1", "2", "-f", "3"})

	expect := []string{"1", "2", "3"}
	if !reflect.DeepEqual(args, expect) {
		t.Errorf("Got %v (type %T) - expected %v (type %T)", args, args, expect, expect)
	}
}

```
Fails with
```
	app_test.go:216: Got [3 1 2] (type []string) - expected [1 2 3] (type []string)
```

Hope you can come up with something good.

ghigt:
Closes #103

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -3,6 +3,7 @@ package cli_test
 import (
 	"fmt"
 	"os"
+	"reflect"
 	"testing"
 
 	"github.com/codegangsta/cli"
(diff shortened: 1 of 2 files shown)


jszwedko:
Indeed, I'd almost want to pull out the argument re-ordering before introducing this sort of complexity, but this definitely does need to be fixed.

jszwedko:
Removed the reordering in #398 (part of the `v2` branch -- to be released).

PR #128, opened 2014-08-11
Fixed ShowSubcommandHelp

ghost:
`ShowSubcommandHelp` should show help for the subcommand not the app


tiborvass:
@codegangsta this introduced a bug: if there's a subcommand `app command -h` will show the help for `app command subcommand`. Working on a fix

Issue #66, opened 2014-02-11
allow actions to return its execution status.


AntonioMeireles:
Hi, 

Like what is already possible for 'Before' actions, i found useful and somewhat logical to  have 'Actions' return something on exit, so that app.Run() gets to catch  all errors, if any, underneath it. Please look at 52496bf2b67c04301dc20538a8a7736c11bc8bf6 in case you care. 

All the best

jszwedko:
Will try to get #66 in soon.

PR #135, opened 2014-08-31
Return error from app and command actions. See #66.

kytrinyx:
This introduces a significant breaking change, and I don't expect it to
be merged off-hand. I do think that it's worth discussion, since it
seems like a very idiomatic choice in terms of how errors would be handled.

A similar backwards-incompatible change was introduced in e6e64114,
allowing the app's Run method to return an error.


jszwedko:
I like this change a lot. Since it is not backwards compatible, I'd like to wrap it up with some other incompatible changes.

PR #151, opened 2014-11-18
After: companion action for Before

imdario:
I patched cli in order to add a deferred action to run after a subcommand (even if Before fails).

I needed this in a project of my own, to properly close resources.


app.go
```diff
@@ -197,6 +210,16 @@ func (a *App) RunAsSubcommand(ctx *Context) error {
 		}
 	}
 
+	if a.After != nil {
+		defer func() {
+			aferr := a.After(context)
+			// check Action error to avoid shadowing
+			if err == nil {
```
jszwedko: I don't believe `err` will ever be not `nil` here. The only time it is set is on line 178 above and there is a check on 193 to see if it is non-nil.

jszwedko:
Thanks! This seems reasonable -- just one comment and it needs to be brought up to speed with master. :+1:

app.go
imdario: So, just doing this...

```go
defer func() {
        err = a.After(context)
}()
```

... would do the trick?

jszwedko: Indeed, maybe with a comment that `err` is always `nil` there. I think we also need to change line 246 to `return err` to get it to propogate.

imdario: Done! It builds and passes tests, so it looks good. If you find anything to change, don't hesitate to contact me back.

jszwedko:
Thanks for the contribution!

Issue #85, opened 2014-04-29
Required Flags


slantview:
It would be really nice to have required flags and required arguments.  I'm mostly putting this here as a place holder to discuss although I would be willing to put some work in on this if we can talk about how we would like to implement it.

I was thinking the easiest would just add a new field to the struct called "Required".  Internally we could have some checking and fail if the required flag is not set.

``` Go
type Flag interface {
    fmt.Stringer
    Required bool
    // Apply Flag settings to the given flag set
    Apply(*flag.FlagSet)
    getName() string
}

func (f *Flag) IsRequired() bool {
  return f.Required
}
```

jhiemer:
+1

assimovt:
:+1:

radeksimko:
:+1:

bbiskup:
+1

stevvooe:
@codegangsta I am going to go against the tide here: if a flag is required, it should be a positional argument. Having required flags in the package promotes this anti-pattern. The suggestion in https://github.com/codegangsta/cli/issues/85#issuecomment-48777510 covers the use case. Perhaps, some helper functions could be added to cover the required use case, as well as cases where certain flags are required when others are set.

szank:
I did check comment https://github.com/codegangsta/cli/issues/85#issuecomment-48777510, and for a StringFlag named "comment" I get : 
```go
fmt.Printf("comment : %#v\n", c.Generic("comment"))
```
```sh
#run app
./go run main.go
#result
comment : (*flag.stringValue)(0xc20802a9f0)
```
Is there any other way to check if flag is set ?
I wouldn't mind forking the code, adding 'required

frodopwns:
I'd like to see a required flag that either automatically prints the command help when not present or prompts the user for the value.

bwiggs:
Also looking for this feature. 

As for @stevvooe positional arguments: This becomes unfriendly to the user if there are more than a few required parameters. Imaging a cli app that needs to connect to a database, like `psql`. That would require a bunch of required parameters like username, password, host, port. If these were to be positional arguments, then the user would have to remember what order to append these when running the CLI. Using required params has the added benefit of providing labels for the user. This is more user friendly.

ex: 
```
$ psql localhost postgres postgres 5434
vs
$ psql -h localhost -d postgres -U postgres -p 5434
```

jszwedko:
[...]

PR #155, opened 2014-12-02
Required flags

ivey:
Implements required flags as originally discussed in #85 

If we _really_ don't want to change the interface, we have some other ideas of ways to do it, but are submitting this as the clean approach.
 -- @ivey @jhowarth


jhowarth:
@jszwedko I updated the pull request per your comments. Could you re-review when you have a moment?

command.go
```diff
@@ -156,5 +166,8 @@ func (c Command) startApp(ctx *Context) error {
 		app.Action = helpSubcommand.Action
 	}
 
+	// set the writer to the original App's writer
```
jhowarth: This is a separate bug that I noticed when I tried to integrate the latest code with our codebase. The `Writer` wasn't copied to the sub command, so subcommands got a new `os.Stdout` `Writer` even though I set the `App`'s `Writer` to something else to capture output in my tests.

jhowarth:
@codegangsta Is it possible to get some eyes on this? We've addressed the previous feedback.

required_flags_test.go
```diff
@@ -0,0 +1,60 @@
+package cli
```
jszwedko: I think these tests could just live in `flag_test.go`.

jszwedko:
Hey, sorry @jhowarth, was busy the last couple of months and was putting off this PR because it is a little more involved. Overall I'm pretty happy with the implementation (though I want to look at it a little more closely). This still appears to not handle the case of setting flags via the environment though.

Example test case:
```go
func ExampleAppWithRequiredEnvFlags() {                                                                                                                                                           
  os.Clearenv()                                                                                                                                                                                   
  os.Setenv("NAME", "Jeremy")                                                                                                                                                                     
  app := cli.NewApp()                                                                                                                                                                             
  app.Name = "greet"                                                                                                                                                                              
  app.Flags = []cli.Flag{                                                                                                                                                                         
    cli.StringFlag{                                                                                                                                                                               
      Required: true,                                                                                                                                                                             
      EnvVar:   "NAME",                                                                                                                                                                           
      Name:     "name",                                                                                                                                                                           
      Value:    "bob",                                                                                                                                                                            
      Usage:    "a name to say",                                                                                                                                                                  
    },                                                                                                                                                                                            
  }                                                                                                                                                                                               
  app.Action = func(c *cli.Context) {                                                                                                                                                             
    fmt.Printf("Hello %v\n", c.String("name"))                                                                                                                                                    
  }                                                                                                                                                                                               
  app.Run([]string{"say"})                                                                                                                                                                        
  // Output:                                                                                                                                                                                      
  // Hello Jeremy                                                                                                                                                                                 
} 
```

jszwedko:
Should be able to do something with `os.Environ` though.

jszwedko:
Or actually a better solution is probably to update the `Apply` methods to use `.Set` when setting the value from the environment.

digitalcrab:
Any news here? Or maybe I can take care of it?

jszwedko:
Hey @digitalcrab,

I haven't had much time to dig into this myself, but if you want to take a stab at addressing the above comments, I'm happy to review an additional PR.

brunson:
How about if we don't care if it can be overridden with an environment variable. The ask was for a way a flag could be specified as required. If it's required, then it must be supplied unless -h is given, in which case only help its displayed and no other action is taken.

jszwedko:
@brunson I believe required options should be specifiable via any of the sources (command line, environment variable, file source). Anything else is likely to be confusing / surprising to the developer and end-user.

lynncyrin:
I branched off this PR and refined it via => https://github.com/urfave/cli/pull/819

Please feel free to close this PR at your leisure ^^

PR #156, opened 2014-12-08
Strip unneeded string in Generic flag during usage invocation

sudharsh:
String() in GenericFlag returns some extra text that makes the usage output of subcommands with custom GenericFlag look a bit ugly.
For example,
``` --topology 'topology'	`-topology option -topology option` Path to the topology file```

The extraneous text in this case is `-topology option -topology option` whereas the expected usage output is,

``` --topology 'topology'   Path to the topology file```


jszwedko:
Thanks for pointing this out @sudharsh; I've addressed this in #166 so I'll close this one.

PR #159, opened 2014-12-15
Add option to make categories with command, to display a more structured help.

Soulou:
I'm sure the template could be nicer, if you have any idea how to improve this poc, I'd be glad. It's mergeable as-is, there are improvements to be done.

Tests will be added tomorrow, I wanted to share that already ;-)

### DIFF
--- a/app.go
+++ b/app.go
@@ -27,6 +27,10 @@ type App struct {
 	HideHelp bool
 	// Boolean to hide built-in version flag
 	HideVersion bool
+	// Display commands by category
+	CategorizedHelp bool
+	// Populate when displaying AppHelp
+	Categories map[string]Commands
 	// An action to execute when the bash-completion flag is set
 	BashComplete func(context *Context)
 	// An action to execute before any subcommands are run, but after the context is ready
@@ -80,6 +84,13 @@ func NewApp() *App {
 func (a *App) Run(arguments []string) (err error) {
 	if a.Author != "" || a.Email != "" {
 		a.Authors = append(a.Authors, Author{Name: a.Author, Email: a.Email})
+	}
+
+	if a.CategorizedHelp {
+		a.Categories = make(map[string]Commands)
+		for _, command := range a.Commands {
+			a.Categories[command.Category] = append(a.Categories[command.Category], command)
+		}
 	}
 
 	// append help to commands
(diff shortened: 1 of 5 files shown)


app.go
```diff
@@ -29,6 +29,10 @@ type App struct {
 	HideHelp bool
 	// Boolean to hide built-in version flag
 	HideVersion bool
+	// Display commands by category
+	CategorizedHelp bool
+	// Populate when displaying AppHelp
+	Categories map[string]Commands
```
jszwedko: Could we maybe make this a function instead of a property?

help.go
```diff
@@ -18,9 +18,12 @@ AUTHOR:{{if .Author}}
   {{.Author}}{{if .Email}} - <{{.Email}}>{{end}}{{else}}
   {{.Email}}{{end}}{{end}}
 
-COMMANDS:
-   {{range .Commands}}{{.Name}}{{with .ShortName}}, {{.}}{{end}}{{ "\t" }}{{.Usage}}
-   {{end}}{{if .Flags}}
+COMMANDS:{{if .CategorizedHelp}}{{range $category, $commands := .Categories}}{{if $category}}
```
jszwedko: The sort order of maps isn't guaranteed to be consistent, it might be nice to sort the keys alphabetically to keep it the same across builds.

jszwedko:
@Soulou thank you for the contribution! Just a few nitpiks. It would be nice to add tests if you get a moment too.

app.go
jszwedko: Just `Categories` (just since it is a calculated field which we wouldn't want to risk getting out of sync).

Soulou:
Rebased, fix tests and additional test about the `categories` structure  generation from all the commands

beevik:
Is this pull request dead? Would love to see this change make it into cli.

jszwedko:
:+1: this looks good, apologies, this fell off my radar. I'll see about resolving the conflicts and rebasing this in.

jszwedko:
Just realized that this only works for top level commands and not subcommands, so there may be some more work to be done here.

jszwedko:
I think we also need to handle the case where some commands are not placed into a category (perhaps placing them in a "default" category) which would alleviate the need for `CategorizedHelp`.

jszwedko:
Closing this in lieu of #346 which is rebased on master and includes a few changes.

Thank you for getting this started @Soulou! Sincere apologies for losing track of it.

PR #167, opened 2015-01-08
Use double quotes in output for defaults

### LINKED ISSUE #136: quotes on windows
Autogenerated help for commands (-h) wraps string flag example in single quotes
eg 'english':
GLOBAL OPTIONS:
   --lang, -l 'english' language for the greeting

Windows handle quotes differently so running cli.exe with -l 'english' does not work, but -l "english" does.

jszwedko:
Windows doesn't recognize 's for wrapping values. "s should work on all
systems.

Fixes #136

### DIFF
--- a/flag.go
+++ b/flag.go
@@ -73,7 +73,7 @@ type GenericFlag struct {
 // help text to the user (uses the String() method of the generic flag to show
 // the value)
 func (f GenericFlag) String() string {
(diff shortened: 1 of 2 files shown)


shawnzhu:
I wonder if it's the case that `%q` fits here which should be used in function [`fmt.Sprintf()`](https://golang.org/pkg/fmt/):

> %q	a double-quoted string safely escaped with Go syntax

PR #168, opened 2015-01-09
strconv.ParseInt should be used instead of strconv.ParseUint when readin...

szaydel:
It looks like when IntFlag is obtained from reading and Environment Variable, instead of directly from command line, `ParseUint` is being used, yet this should be a Signed Int flag, so negative values will likely result in 0 being read from the Environment, since the `ParseUint` will not work correctly with a `0<` number. I believe we really want `ParseInt` here instead.


jszwedko:
You are correct, thank you far catching that :+1:

szaydel:
I am glad to help in a small way. :) Thank you for building such an excellent library!

PR #169, opened 2015-01-09
Use 0 as the base when parsing ints

jszwedko:
To be consistent with what the stdlib flag package does (allows for users to use numbers in other bases as flags).

http://golang.org/src/flag/flag.go?s=14139:14203#L479


meatballhat:
whoops!

PR #171, opened 2015-01-23
Add application name to subcommand help template

radeksimko:
This will make the help text more helpful for users.

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -89,7 +89,7 @@ func ExampleAppHelp() {
 	//    describeit - use it to see a description
 	//
 	// USAGE:
-	//    command describeit [arguments...]
+	//    greet describeit [arguments...]
 	//
 	// DESCRIPTION:
 	//    This is how we describe describeit the function
--- a/command.go
+++ b/command.go
@@ -10,6 +10,8 @@ import (
 type Command struct {
 	// The name of the command
 	Name string
+	// Application name
+	AppName string
 	// short name of the command. Typically one character
 	ShortName string
 	// A short description of the usage of this command
--- a/help.go
+++ b/help.go
@@ -33,7 +33,7 @@ var CommandHelpTemplate = `NAME:
    {{.Name}} - {{.Usage}}
 
 USAGE:
-   command {{.Name}}{{if .Flags}} [command options]{{end}} [arguments...]{{if .Description}}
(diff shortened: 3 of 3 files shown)


vieux:
:+1:

jszwedko:
I don't think this would work as expected for nested subcommands (you would just get the "parent" subcommand in the output as the `AppName`).

radeksimko:
@jszwedko Good point, do you have any example of nested subcommands, so we can come up with appropriate solution for that as well?

jszwedko:
@radeksimko sorry for the delay.

Example:
```go
package main

import (
  "os"
  "github.com/codegangsta/cli"
)

func main() {
  app := cli.NewApp()
  app.Name = "greet"
  app.Usage = "fight the loneliness!"
  app.Commands = []cli.Command{
  {
    Name:      "add",
    ShortName: "a",
    Usage:     "add a task to the list",
      println("added task: ", c.Args().First())
    },
  },
  {
    Name:      "complete",
    ShortName: "c",
    Usage:     "complete a task on the list",
      println("completed task: ", c.Args().First())
    },
  },
  {
    Name:      "template",
    ShortName: "r",
    Usage:     "options for task templates",
    Subcommands: []cli.Command{
      {
        Name:  "add",
        Usage: "add a new template",
            println("new task template: ", c.Args().First())
        },
      },
      {
        Name:  "remove",
        Usage: "remove an existing template",
          println("removed task template: ", c.Args().First())
        },
      },
    },
  },     
}

  app.Run(os.Args)
}
```

jszwedko:
Note that I just frakenstein'd this from the README so it may not compile, but hopefully it gives you an idea.

JohnMaguire:
@jszwedko @radeksimko Err, not quite. I had almost the exact same solution at first. And it actually would work, because, as you'll see [here](https://github.com/codegangsta/cli/blob/30b361bf66a60174a9589458afbf494478bafd00/command.go#L142-L146), the Name gets set when `startApp()` is called on the Command.

Which means the real bug is that the name from the previous command doesn't get prepended if "help" is called, because "help" is a different sub-command (thus, `startApp()` isn't called, and the Name isn't set.)

I'm about to submit a pull request in a moment with a modified patch.

jszwedko:
I believe this is addressed by #246, let me know what you think @radeksimko

radeksimko:
Looks :ok_hand:

PR #173, opened 2015-01-27
Add new func GlobalArgs in context, and new SkipFlagNotDefined in command

smallfish:
`SkipFlagNotDefined` is used for when raise flag not defined error,
continue to execute, not print error/help message.


jszwedko:
I'm not crazy about the string matching for that particular error, but I like the overall idea. I wish the `flag` stdlib was better about returning concrete error values.

I'm going to leave this open for now until I can think of whether it is worth introducing the possible brittleness.

jszwedko:
Closing in lieu of #387 -- see discussion over there.

PR #178, opened 2015-01-30
Add support for multiple authors

paked:
Through the addition of a new Author struct, I have added support for multiple Authors. This, however, is change to the API (although, it could be averted if wanted?)

### DIFF
--- a/app.go
+++ b/app.go
@@ -40,10 +40,8 @@ type App struct {
 	CommandNotFound func(context *Context, command string)
 	// Compilation date
 	Compiled time.Time
-	// Author
-	Author string
-	// Author e-mail
-	Email string
+	// Authors
+	Authors []Author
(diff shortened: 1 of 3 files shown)


jszwedko:
I like this addition. It is a change to the API though, so I'll mark it to be included in the next round of API breaking changes. We could get this change in sooner if we maintained the old API, but also allowed for a list to be specified.

jszwedko:
i.e. keep `Author` and `Email`, but also add `Authors` with a note in the doc comments warning people away from using `Author` and `Email`.

paked:
Just added support for this. If ```Author``` and ```Email``` are set app.Run will append an ```Author``` object onto the ```Authors``` list.

jszwedko:
I branched off this to merge in master to eliminate conflicts in #192. Thank you for this!

PR #180, opened 2015-02-06
Allow commands to be hidden from help and autocomplete

psmit:
It would be great if commands can be hidden from the user. This PR does exactly that.

### DIFF
--- a/command.go
+++ b/command.go
@@ -31,6 +31,8 @@ type Command struct {
 	SkipFlagParsing bool
 	// Boolean to hide built-in help command
 	HideHelp bool
+	// Boolean to hide this command from help or completion
+	Hidden bool
 }
 
 // Invokes the command given the context, parses ctx.Args() to generate command-specific flags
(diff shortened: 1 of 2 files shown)


jszwedko:
I understand the motivation for a feature like this, but this could also be accomplished by checking for an environment variable and conditionally adding commands in the source (e.g. looking for `HIDDEN_COMMANDS=1` and adding the commands if it is set).

phsmit:
Sorry for getting back so late, I have been sick this week.

Actually, it is very hard to set enviroment variables sometimes. For example, in our use case, gogits/gogs, we need to call a subcommand as entry point for persons that connect through ssh. This means we add keys to ~/.ssh/authorized_keys with an appropriate command="". It is quite a hassle to get environment variables working well in such case.

I understand to keep the codebase clean and tidy, but this change is very small and I would very much appreciate if you could consider again merging this request.

meatballhat:
Especially given the recent merge (by me!) of the hidden flags feature, I'd like to get this in.  I'm working on the merge resolution right now. :metal:

jszwedko:
:+1: the use case makes sense to me -- apologies for losing track of this issue.

PR #181, opened 2015-02-12
Add hidden flags

hatchan:
This is similar to #180, but then applied to flags. Implementation is done similar to #155

### DIFF
--- a/flag.go
+++ b/flag.go
@@ -36,6 +36,7 @@ type Flag interface {
 	// Apply Flag settings to the given flag set
 	Apply(*flag.FlagSet)
 	getName() string
+	IsHidden() bool
 }
 
 func flagSet(name string, flags []Flag) *flag.FlagSet {
@@ -67,6 +68,7 @@ type GenericFlag struct {
 	Value  Generic
 	Usage  string
 	EnvVar string
+	Hidden bool
 }
 
 // String returns the string representation of the generic flag to display the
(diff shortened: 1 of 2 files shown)


termie:
This feature is definitely a help to us for shipping products where some features are unlikely to be used by the end user (but we use them internally), ideally there would be a `help all` added after this lands that would allow displaying those flags as well

phsmit:
:+1:

jszwedko:
I understand the motivation for a feature like this, but this could also be accomplished by checking for an environment variable and conditionally adding flags in the source (e.g. looking for `INTERNAL_FLAGS=1`).

termie:
@jszwedko perhaps this is not how you design your applications, but I would assume for most the app.Run() is their first entry point, to establish the environment within which the application will execute, creating a separate pre-run step that _also_ establishes the environment seems like duplication of effort

jszwedko:
@termie sure, but it doesn't _need_ to be. The main entry point is always `main` and I don't think it is improper to vary the commands, flags, etc. that are allowed based on the environment the application is running under. Every feature adds cost to the library, so I'm just trying to consider the alternatives.

harshavardhana:
Oh wow, i should have seen this PR before i just submitted - https://github.com/codegangsta/cli/pull/201 .. damn!

harshavardhana:
Improved upon #201 with filtering flags at higher level than 'templates' so that we don't end up with odd formatting issues.  This way we don't also have to keep ``IsHidden()`` or ``IsNotHidden()`` exported functions.  Please review.. 

Also by default ``--generate-bash-completion`` flag is never printed along with help.

ravigadde:
+1

PR #186, opened 2015-02-23
Fixing the issue with a command with subcommands not showing help message

nrdufour:
I think it might close a bunch of issues about it.

It happens if you define a command (let's say `foo`) with subcommands (let's `say` bar and `baz`).
And if you type `mytool foo -h`, only the first subcommand help is shown (so here, `bar`).
I've seen that a new App structure is created based on the first command, and so the following issues:

- the command name is set to "" and HasName was returning true for an empty ShortName, therefore returning the first subcommand all the time (providing your subcommand has no shortname).
- the ShowCommandHelp method wasn't aware that command name was just "" and returned the first subcommand.

### DIFF
--- a/command.go
+++ b/command.go
@@ -119,7 +119,7 @@ func (c Command) Run(ctx *Context) error {
 
 // Returns true if Command.Name or Command.ShortName matches given name
 func (c Command) HasName(name string) bool {
-	return c.Name == name || c.ShortName == name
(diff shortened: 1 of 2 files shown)


wricardo:
I'm hoping to see this issue fixed.

edgard:
:+1: thanks for that, was driving me crazy :D

jszwedko:
Awesome, thank you for this :+1:

PR #187, opened 2015-02-26
Added an API for number of flags set

ravigadde:
Use case - A global flag and per command flags are not compatible. Want to check if global flag and any command flag are set together. Without this API, the check is very tedious as every per command flag needs to be checked explicitly.

### DIFF
--- a/context.go
+++ b/context.go
@@ -106,6 +106,11 @@ func (c *Context) GlobalGeneric(name string) interface{} {
 	return lookupGeneric(name, c.globalSet)
 }
 
+// Returns the number of flags set
+func (c *Context) NumFlags() int {
+	return c.flagSet.NFlag()
+}
+
 // Determines if the flag was actually set
 func (c *Context) IsSet(name string) bool {
 	if c.setFlags == nil {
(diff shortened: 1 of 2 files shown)


toqueteos:
Does `len(c.Args())` not work?

kytrinyx:
I use `len(c.Args())`, it works fine.

ravigadde:
@toqueteos @kytrinyx 

Args and flags are different. Flags are command line options that are specified with - or -- ahead of args. Lets look at a few examples:
1) cli test foo
test is the command. foo is the only argument. len(c.Args()) gives 1.
2) cli test --verbose=1 foo
verbose is a flag. len(c.Args()) still gives 1. c.NumFlags() returns 2 (it counts both verbose and 1).
3) cli test --verbose=1 --verbosity_level=5
len(c.Args()) is 0. c.NumFlags() returns 4.

kytrinyx:
@ravigadde Ah, right. Thanks for elaborating, that makes sense.

jszwedko:
Seems reasonable :+1: . I know the logic is trivial, but would you mind adding a test for completeness?

ravigadde:
@jszwedko Added a test case, let me know if this looks ok to be merged

jszwedko:
Awesome, thanks!

Issue #183, opened 2015-02-16
Command alias'?


ianneub:
Is it possible to have a single command with multiple names or multiple shortnames? For example:

```go
  app.Commands = []cli.Command{
    {
      Name:      "start, up",
      ShortName: "s, u",
      Usage:     "Start the docker engine",
      Action: func(c *cli.Context) {
        ...
      },
    },
  }
```

The use case is that some users can't seem to remember `start`, or would rather type `up` or vice versa.

jszwedko:
Currently this is not directly supported, but you can emulate the behavior with something like:
```go
  upAction := func(c *cli.Context) {
  }
  app.Commands = []cli.Command{
    {
      Name:      "up",
      ShortName: "u",
      Usage:     "Start the docker engine",
      Action: upAction,
      {
      Name:      "start",
      ShortName: "s",
      Usage:     "Start the docker engine",
      Action: upAction,
    },
  }
```

jszwedko:
I wouldn't mind just having a `Names` attribute that is a list of possible names rather than `Name` and `ShortName` though.

ianneub:
> I wouldn't mind just having a Names attribute that is a list of possible names rather than Name and ShortName though.

:+1:

jszwedko:
@ianneub added `Aliases` in #193.

ianneub:
Right on. Thanks!

PR #193, opened 2015-03-10
Add `Command.Aliases` and deprecate `Command.ShortName`

jszwedko:
`Aliases` will be more flexible while still allowing "ShortName"
behaviour via `Aliases`.

Addresses #183


meatballhat:
:clap: woop!

PR #194, opened 2015-03-10
fix Run()

fugr:


jszwedko:
Thank you for catching this -- I should have made sure there was a test around this. I've opened #196 because I think that is closer to the existing behavior. I'll try to get some tests around this later today.

jszwedko:
Supercedede by #196

PR #195, opened 2015-03-10
Author(s) default fixups.

thorduri:
Didn't bother with the 'real' domains in the tests though.


jszwedko:
Thank you for this -- I missed that #192 added default values. I removed them in #196 to be closer to what the behavior used to be.

thorduri:
@jszwedko Didn't want to kill the example stuff right off. But #196 looks better.

I'll close.

PR #196, opened 2015-03-10
Fix support for deprecated author fields

jszwedko:
Should add an author if either name or email is specified.


jszwedko:
Supercedes #194 #195.

jszwedko:
Merging since Travis doesn't seem to be picking this commit up -- I'll watch the merge build.

PR #198, opened 2015-03-12
replace tab before authors with two spaces

bontibon:
this makes authors consistent with everything else


jszwedko:
Superseded by #200

jszwedko:
Thank you though!

PR #199, opened 2015-03-12
fix panic on broken pipe

bontibon:
If the help printer tries writing to a broken pipe (e.g. `myprog --help | exec-that-does-not-exist`), a runtime panic would occur. This patch removes the panic.


jszwedko:
I agree that we should address this issue, but there are many other places this library writes to `stdout`. I think a better solution would be to wrap the injected `io.Writer` and ignore the broken pipe issue.

We could push this responsibility back to the user, but I don't think it is unreasonable to expect this library to handle this since it writes to `stdout` so much.

jszwedko:
I think I've decided that the broken pipe error is OK and is common in other commands where the thing being piped to closes unexpectedly.

Issue #163, opened 2014-12-18
Sub-command showing help message for sub-sub-command


psanford:
Consider the following program:

```
package main

import (
	"os"

	"github.com/codegangsta/cli"
)

func main() {
	app := cli.NewApp()
	app.Commands = []cli.Command{
		{
			Name: "a",
			Subcommands: []cli.Command{
				{
					Name: "a1",
				},
				{
					Name: "a2",
				},
			},
		},
		{
			Name: "b",
			Subcommands: []cli.Command{
				{
					Name: "b1",
				},
				{
					Name: "b2",
				},
			},
		},
	}
	app.Run(os.Args)
}
```

I would expect the output for `./bin/cliexample a` to show the help text of the `a` command. But instead it shows the help text of `a1`: 

```
 $ ./bin/cliexample a
NAME:
   a1 - 

USAGE:
   command a1 [arguments...]
```

However if you give it the help command it shows the correct help text:
```
 ./bin/cliexample a help
NAME:
   cliexample a - 

USAGE:
   cliexample a [global options] command [command options] [arguments...]

VERSION:
   0.0.0

AUTHOR:
  Author - <unknown@email>

COMMANDS:
   a1
   a2
   help, h      Shows a list of commands or help for one command
   
GLOBAL OPTIONS:
   --help, -h   show help
```

frellus:
I get the same behavior. I'm just starting to use this code, but will try to figure out a solution if I can and send a fix up

jszwedko:
This seems to be working correctly in master, can you verify @psanford ?

psanford:
I still experience the problem on master as of bf4a526f, what commit do you think fixed it?

nrdufour:
I have the same issue. I've been trying to figure out a good way to solve it for a while.
It seems that the template `SubcommandHelpTemplate` is never used anywhere.
And I think that the implementation of `ShowCommandHelp` is rather strange or wrong.

oleksandr:
Confirming the same issue form the recent master.

pandrew:
Ran into this right know.

jszwedko:
Apologies for the delay!

I'm not sure why I thought I was seeing the correct behaviour, but I was able to reproduce this on master. I'm still trying to wrap my head around the app/command/subcommand differences, but #172 appears to fix this issue. I'd like to refactor some of the duplication between `App` and `Command` away.

ehazlett:
[...]

PR #202, opened 2015-03-19
Add regression test for missing help topics

aybabtme:
Hi there,

I was working on fixing missing help topics on subcommands when I realized a PR fixing that was merged 9 days ago. Since I wrote a regression test as part of my work, I thought I could provide it to the project so that the issue doesn't appear again.

Also it sets the `Version`/`Author`/etc fields in the child app created by the command.

related to #186 and #163

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -1,9 +1,11 @@
 package cli_test
 
 import (
+	"bytes"
 	"flag"
 	"fmt"
 	"os"
+	"strings"
 	"testing"
 
 	"github.com/codegangsta/cli"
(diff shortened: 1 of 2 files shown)


aybabtme:
ping @codegangsta =]

jszwedko:
@aybabtme thank you for these tests -- we definitely need more around the help output!

Case in point, I think it's a bug that the subcommand help outputs the app information at all (version, author, etc.) especially since it isn't consistent between `-h` and `help`.

Would you mind removing that portion from this PR? I'll see about addressing the subcommand help with respect to outputting application information.

jszwedko:
I changed my mind about setting the additional app information (even though I still believe it shouldn't appear in the help output, it should still be available on the `App` that is created). I re-added that as a889873.

PR #209, opened 2015-04-04
Remove SkipFlagParsing behavior

jszwedko:
Requiring `--` before flags that start with a `-` is consistent with
typical CLI interfaces (consider that stdlib `flag` has this behavior).

We'll want to be careful about merging this before releasing it because it will break existing implementations. We may want to consider creating a branch with backwards incompatible changes that we can merge all at once.

@codegangsta @meatballhat any opinions?


meatballhat:
Does it make sense to target v2.0.0 here, maybe?

jszwedko:
I think I actually changed my mind on this -- I can see some value in letting users disable flag parsing entirely and handling them themselves in certain cases. Closing this out!

PR #213, opened 2015-04-16
Clean Shutdown

jbriggs-zvelo:
- Watch for signals
- Call a Shutdown action when a signal is received

I thought this would be a nice addition. Gives you a chance to call server.Stop() or just os.Exit(0). Putting it in cli means writing less identical code even if it is just a channel and a call to Notify.

### DIFF
--- a/app.go
+++ b/app.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"os/signal"
 	"strings"
 	"text/tabwriter"
 	"text/template"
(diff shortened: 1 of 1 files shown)


jszwedko:
Hey! So, this traps all signals which is probably not what most users want (consider `syscall.SIGWINCH` e.g.). I could see maybe trapping just `SIGTERM` here, but I'm not sure we are gaining much over just having the user trap the signal in their own code if they desire to (especially since they will likely want to trap within each subcommand/command to do specific cleanup if they have subcommands).

Thank you for suggesting this though!

PR #214, opened 2015-04-23
do not print 'AUTHOR(S):' if there is no author

vieux:
the refactor from `Author` to `Authors` introduced a regression.


JohnMaguire:
:+1:

jszwedko:
Thanks for catching this! I rebased as 7c041f5 and added a test in 96ad929.

PR #215, opened 2015-04-25
Add parent name to subcommand name when help is called

JohnMaguire:
Normally the subcommand's name is updated when startApp is called, but
we have to make a special exception for the help file, since it's a
different sub-command that's getting started.

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -89,10 +89,10 @@ func ExampleAppHelp() {
 	app.Run(os.Args)
 	// Output:
 	// NAME:
-	//    describeit - use it to see a description
+	//    greet describeit - use it to see a description
 	//
 	// USAGE:
-	//    command describeit [arguments...]
+	//    greet describeit [arguments...]
 	//
 	// DESCRIPTION:
 	//    This is how we describe describeit the function
--- a/help.go
+++ b/help.go
@@ -14,7 +14,7 @@ USAGE:
 VERSION:
    {{.Version}}
 
-AUTHOR(S): 
+AUTHOR(S):
    {{range .Authors}}{{ . }}
    {{end}}
 COMMANDS:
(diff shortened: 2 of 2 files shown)


JohnMaguire:
This patch fixes the same bug as targeted here: https://github.com/codegangsta/cli/pull/171

jszwedko:
Thank you for taking a look at this!

However, I'm not crazy about changing the command's name to be a concatenation of the app name and the command's names. If we desire this, I think a better approach may be to have a bidirectional tree where the command also knows its parent command and walk that tree upwards to print the leading commands. What do you think?

JohnMaguire:
@jszwedko Neither am I. I did it this way due to the fact that it's already done this way in a different location in the code. Per commit message:

> Normally the subcommand's name is updated when startApp is called, but
> we have to make a special exception for the help file, since it's a
> different sub-command that's getting started.

If you take a look at `startApp`, on [line 146 of command.go](https://github.com/codegangsta/cli/blob/master/command.go#L146), you can see where that happens.

If you want to change it in the pull request, we should probably update it there too. Would you like me to look into updating both (and any other possible) locations where the name is getting sent and see about converting it to a tree-based name generation?

jszwedko:
Ah, and so it is!

I thought about the concept of a command knowing its parent and decided that that feels a bit confusing because only the 'active' command would have that information populated. I considered attaching this information to the `Context`, but given that most of the information used in the template is on the `App` or `Command` (and this is what is passed into the template execution) I'm not sure this would work either. Thus what you have may be a reasonable compromise, but I'd like to think about storing this information on the context a bit more.

JohnMaguire:
I thought that may have been the reason. :) Let me know if you'd like me to make any changes, unless you feel like going ahead and making them yourself.

jszwedko:
I think this has been superseded by #246 , let me know if that doesn't work for you though!

JohnMaguire:
Thanks @jszwedko and @cpuguy83!

PR #225, opened 2015-05-13
Update string slice usage

vieux:
Feel free to close I guess it's a change some ppl won't like :)

I like how the `StringSliceFlag` usage is constructed.

Let's take `--help [--help option --help option]`

* It's very wide
* The first --help don't have option but the 2 others have
* Why limit to 2 ?

It doesn't feel very intuitive for me.

I propose something simpler: `--help [--help ...]`


meatballhat:
I agree with @vieux.  Whaddya think, @jszwedko?

PR #226, opened 2015-05-13
fixes #204

### LINKED ISSUE #204: App.After breaks error checking from App.Before
Hello,

Using `App.After` resets the `err` variable (in `App.Run`) used to capture the error from `App.Before`, so no error is reported.

Here is a sample: 

```go
package main

import (
	"errors"

	"github.com/codegangsta/cli"
)

func main() {

	app := cli.NewApp()

	app.Name = "test"
	app.Usage = "test"
	app.Action = func(c *cli.Context) {}

	app.After = func(c *cli.Context) error { return nil }
	app.Before = func(c *cli.Context) error { return errors.New("test") }

	app.RunAndExitOnError()

}
```

I would expect the error from `App.Before` to get catched by `App.RunAndExitOnError`, which it also does if i remove the `App.After` function.

apriendeau:
This does the before check before after so that the error is properly reported back fails.

### DIFF
--- a/app.go
+++ b/app.go
@@ -130,6 +130,13 @@ func (a *App) Run(arguments []string) (err error) {
 		return nil
 	}
 
+	if a.Before != nil {
+		err := a.Before(context)
+		if err != nil {
+			return err
+		}
+	}
+
 	if a.After != nil {
 		defer func() {
 			// err is always nil here.
@@ -137,13 +144,6 @@ func (a *App) Run(arguments []string) (err error) {
 			// just few lines before.
 			err = a.After(context)
 		}()
-	}
-
-	if a.Before != nil {
-		err := a.Before(context)
-		if err != nil {
-			return err
-		}
 	}
 
 	args := context.Args()


jszwedko:
Thank you for the PR! This appears to have the side-effect of not running `After` if `Before` errors which I'm not sure is desired -- I'll have to think about this a little more.

apriendeau:
@jszwedko I think that if you panic on the before, more than likely the after will fail too. I would rather have this because it was swallowing the error for me.

jszwedko:
That makes sense, I was just trying to think about the case where someone might be using `After` in a manner similar to `defer` to clean up resources initialized in `Before`. They'd likely still want the cleanup to happen even if they return an error in `Before`.

In a perfect world I think that it would just pass in `err` to the `After` function, but without changing the interface, I think the best solution may be to aggregate the errors from `Before` and `After` and return that.

jszwedko:
Something like #230, what do you think @apriendeau ?

apriendeau:
That is much better and should solve both use cases. My last comment is just a style one that it makes sense for readability to have the `Before` before the `After.` That is just semantics though.

Should I close this pull request then?

jszwedko:
@apriendeau yep, I'll close this one, thanks for bringing attention to this issue!

Issue #142, opened 2014-09-30
Nested subcommands with flags. Intermediate flags disappear.


sgirones:
When having more than one level of subcommands, flags are not accumulated.
Simple example:
```
package main

import (
	"fmt"
	"os"

	"github.com/codegangsta/cli"
)

func main() {
	app := cli.NewApp()
	app.Name = "redbooth"
	app.Flags = []cli.Flag{
		cli.StringFlag{"global-flag", "global-flag", "", ""},
	}
	app.Commands = []cli.Command{
		command1,
	}
	app.Run(os.Args)
}

var command1 = cli.Command{
	Name: "command1",
	Flags: []cli.Flag{
		cli.StringFlag{"command1-flag", "command1-flag", "", ""},
	},
	Subcommands: []cli.Command{
		subcommand1,
	},
}

var subcommand1 = cli.Command{
	Name: "subcommand1",
	Flags: []cli.Flag{
		cli.StringFlag{"subcommand1-flag", "subcommand1-flag", "", ""},
	},
	Action: func(c *cli.Context) {
		fmt.Println(c.GlobalString("global-flag"))
		fmt.Println(c.String("global-flag"))
		fmt.Println(c.GlobalString("command1-flag"))
		fmt.Println(c.String("command1-flag"))
		fmt.Println(c.GlobalString("subcommand1-flag"))
		fmt.Println(c.String("subcommand1-flag"))
	},
}

---

$ go run example.go --global-flag "global" command1 --command1-flag "cmd1" subcommand1 --subcommand1-flag "subcmd1"
global




subcmd1
```

I would expect command1-flag to be part of the globalSet, but currently is not appended.

jszwedko:
This is an interesting question -- I /think/ the behaviour you describe is expected, but I'll have to ponder this a bit more.

thorduri:
Didn't expect this behaviour, but came looking for methods to properly pass command flags to subcommands globally if you will.

I'd enjoy this behaviour from the package.
+1

jszwedko:
Addressed by #227, let me know what you think!

sgirones:
Looks good. Thanks for the PR!

PR #227, opened 2015-05-18
Collect global flags from the full command hierarchy

databus23:
This pull request tries to solve the problem reported in #142, which is that flags from parent commands are not accessible from a subcommands context.
I'm not sure if this is by design, one reply in the above issue suggest that it might be (I hope not). I think it is useful to define flags on a command level that are accessible in subcommands.

This pull request changes the internal structure of the `Context` struct. Instead of  a pointer to the global `FlagSet`a pointer to the immiediate parent context is kept.
When looking up global flags from a context the flag sets of all ancestors are considered. This means from a command context perspective any flag defined at any parent command is *global*.

### DIFF
--- a/app.go
+++ b/app.go
@@ -104,12 +104,12 @@ func (a *App) Run(arguments []string) (err error) {
 	nerr := normalizeFlags(a.Flags, set)
 	if nerr != nil {
 		fmt.Fprintln(a.Writer, nerr)
-		context := NewContext(a, set, set)
+		context := NewContext(a, set, nil)
 		ShowAppHelp(context)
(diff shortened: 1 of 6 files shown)


jszwedko:
@databus23 awesome, thank you for putting this together. It feels much more intuitive to me.

Eventually I'd like to remove the "Global" flags altogether and just have the "normal" flag commands walk up the chain (though this introduces the possibility of shadowing). I haven't thought it completely through yet though.

Reviewing now.

jszwedko:
Looks great, I think the having pointers up the chain of contexts makes for an elegant solution.

Issue #179, opened 2015-02-05
Using IntSliceFlag results in panic


mrmylanman:
I am very new to Go, but am trying to convert a few scripts we run in our business to Go as a test to see if it is something I can get on board with.

Your cli library is working out pretty well, and I decided to try out the `IntSliceFlag` functionality so that I can pass in a slice of integers. When I do, I get a panic:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x55e0b]

goroutine 1 [running]:
github.com/codegangsta/cli.(*IntSlice).String(0x0, 0x0, 0x0)
	/Users/mylanc/Go/src/github.com/codegangsta/cli/flag.go:169 +0x11b
flag.(*FlagSet).Var(0xc208044180, 0x53be50, 0x0, 0x31d790, 0xa, 0x373090, 0x3c)
	/usr/local/Cellar/go/1.4.1/libexec/src/flag/flag.go:669 +0x3c
github.com/codegangsta/cli.func·010(0x31d790, 0xa)
	/Users/mylanc/Go/src/github.com/codegangsta/cli/flag.go:209 +0x84
github.com/codegangsta/cli.eachName(0x31d790, 0xa, 0xc20807b618)
	/Users/mylanc/Go/src/github.com/codegangsta/cli/flag.go:54 +0x11e
github.com/codegangsta/cli.IntSliceFlag.Apply(0x31d790, 0xa, 0x0, 0x373090, 0x3c, 0x344470, 0x1a, 0xc208044180)
	/Users/mylanc/Go/src/github.com/codegangsta/cli/flag.go:210 +0x403
github.com/codegangsta/cli.(*IntSliceFlag).Apply(0xc208036340, 0xc208044180)
	<autogenerated>:24 +0xaa
github.com/codegangsta/cli.flagSet(0x322bb0, 0xa, 0xc208044120, 0x4, 0x6, 0xc208044120)
	/Users/mylanc/Go/src/github.com/codegangsta/cli/flag.go:45 +0x133
github.com/codegangsta/cli.Command.Run(0x322bb0, 0xa, 0x0, 0x0, 0x344570, 0x19, 0x0, 0x0, 0x0, 0x0, ...)
	/Users/mylanc/Go/src/github.com/codegangsta/cli/command.go:55 +0x3bf
[...]

PR #229, opened 2015-05-31
Fix panic if Valus in Int/StringSliceFlasg is missing

saromanov:
Hi!
I found that we get panic if parameter Value in ```IntSliceFlag``` or in ```StringSliceFlag``` was missed. For example:
```go
cli.StringSliceFlag {
   	Name: "Files",
   	Usage: "Get all of your files",
  },
```
And output:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x1 pc=0x80823ac]

goroutine 1 [running]:
github.com/codegangsta/cli.(*StringSlice).String(0x0, 0x0, 0x0)
	/home/haunted/gopath/src/github.com/codegangsta/cli/flag.go:110 +0xdc
flag.(*FlagSet).Var(0x1843a090, 0xb77c6b60, 0x0, 0x8155978, 0x5, 0x8167788, 0x15)
```
This behavior is identical as in #179, however, this issue was closed, but problem wasn't solved. I just fixed it and provide additional test.

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -342,6 +342,38 @@ func TestApp_ParseSliceFlags(t *testing.T) {
 	}
 }
 
+func TestApp_ParseSliceFlagsWithMissingValue(t *testing.T) {
+	var parsedIntSlice []int
+	var parsedStringSlice []string
+
+	app := cli.NewApp()
+	command := cli.Command{
+		Name: "cmd",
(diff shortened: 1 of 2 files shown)


jszwedko:
@saromanov awesome, thank you!

PR #234, opened 2015-06-08
Allow context value to be set after parse

bryanl:
This change allows a context value to be set after parsing. The use case is updating default settings in a Before func. 

An example usage:

```go 
f, err := os.Open(configPath)
if err == nil {
  config, err := docli.NewConfig(f)
  if err != nil {
    panic(err)
  }

  // set token and children contexts will never know where it came from
  c.Set("token", config.APIKey)
}
```

### DIFF
--- a/context.go
+++ b/context.go
@@ -132,6 +132,11 @@ func (c *Context) NumFlags() int {
 	return c.flagSet.NFlag()
 }
 
+// Set sets a context flag to a value.
+func (c *Context) Set(name, value string) error {
+	return c.flagSet.Set(name, value)
+}
+
 // Determines if the flag was actually set
 func (c *Context) IsSet(name string) bool {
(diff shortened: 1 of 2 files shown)


jszwedko:
@bryanl thanks for this! It feels like we should have a `GlobalSet` too to keep parity with the other methods, perhaps walking up the context list until it finds a flagset with the given flag defined (using `.Lookup`). What do you think?

jfarrell:
+1, would love to see this make it in for the exact case outlined

munkyboy:
+1 I have the same use case (commands utilizing values computed in the Before hook)

zpatrick:
+1

solher:
+1

zmarcantel:
+1

Issue #237, opened 2015-06-09
Add Copyright option


polds:
Wondering if a Copyright option could be added to the app settings. Certain license types especially BSD require the ability for binaries to reproduce their own license.

polds:
Resolved in #238

PR #238, opened 2015-06-09
Add the ability to add a copyright

polds:
References #237 . Some license types such as BSD require the ability for binaries to reproduce their copyright. 

Adds:
```
app.Copyright = "(c) 2015"
```

Which yields:

```
COPYRIGHT:
   (c) 2015
```

Following the `GLOBAL OPTIONS` section.

### DIFF
--- a/help.go
+++ b/help.go
@@ -28,7 +28,10 @@ COMMANDS:
    {{end}}{{if .Flags}}
 GLOBAL OPTIONS:
    {{range .Flags}}{{.}}
-   {{end}}{{end}}
+   {{end}}{{end}}{{if .Copyright }}
+COPYRIGHT:
+   {{.Copyright}}{{end}}
+
 `
 
 // The text template for the command help topic.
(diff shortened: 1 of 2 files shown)


jszwedko:
@polds seems reasonable, thanks for the contribution!

jszwedko:
@polds looks like this adds an additional blank line if the copyright isn't set -- I think you just need to drop line 34.

polds:
ok @jszwedko , I have removed the whitespace.

jszwedko:
Thank you!

help.go
```diff
@@ -28,7 +28,9 @@ COMMANDS:
    {{end}}{{if .Flags}}
 GLOBAL OPTIONS:
    {{range .Flags}}{{.}}
-   {{end}}{{end}}
+   {{end}}{{end}}{{if .Copyright }}
+COPYRIGHT:
+   {{.Copyright}}{{end}}
 `
```
shurcooL: An observation, the above could be rewritten to make it a little simpler by using `with` [action](https://godoc.org/text/template#hdr-Actions) instead of `if`:

```
{{with .Copyright}}
COPYRIGHT:
   {{.}}{{end}}
```

jszwedko: Interesting, I didn't realize that `with` did that (or that it had an `else` clause). That is nicer.

polds: I can put together another PR with these changes thanks @shurcooL

PR #242, opened 2015-06-11
Add Required parameter to flag structs

szank:
This change add a 'Required' boolean variable to all Flag structs. 
When this parameter is set to true while defining flags in the app/context then cli will return an error if user didn't provide such flag in command line parameters. 

Unit test included.


szank:
I just read issue #85 , closing this request as it breaks Flags interface.

PR #245, opened 2015-06-19
Add func to get context parent

cpuguy83:
Need this for being able to show correct help output for subcommands.
Ie, instead of just `{{.Name}} - {{.Usage}}` we need to be able to do `{{.Parent.Name}} {{.Name}} - {{.Usage}}`


jszwedko:
:+1: thanks @cpuguy83

PR #246, opened 2015-06-21
Sets a subcommand's parent cmd

cpuguy83:
This allows the help output to show the correct/full command path to the
subcommand.
For instance:
```
BaseCommand
    -> SubCommandA
        -> SubCommandA1
```

Currently the "usage" help output for `SubCommandA1` would look like `BaseCommand SubCommandA1`
When really what it should be is `BaseCommand SubCommandA SubCommandA1`

#245 was originally intended to help solve this problem, unfortunately I submitted that PR a bit too early in the process (so sorry!)

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -735,6 +735,36 @@ func TestApp_Run_CommandWithSubcommandHasHelpTopic(t *testing.T) {
 	}
 }
 
+func TestApp_Run_SubcommandFullPath(t *testing.T) {
+	app := cli.NewApp()
+	buf := new(bytes.Buffer)
+	app.Writer = buf
+
+	subCmd := cli.Command{
+		Name:  "bar",
+		Usage: "does bar things",
+	}
+	cmd := cli.Command{
+		Name:        "foo",
+		Description: "foo commands",
+		Subcommands: []cli.Command{subCmd},
+	}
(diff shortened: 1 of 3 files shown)


jszwedko:
Awesome, thank you @cpuguy83 -- do you think you could throw in a high level test? If not, I should be able to get to it.

cpuguy83:
Sure thing.

cpuguy83:
Tests pushed. Sorry it took so long.
Was banging my head against the desk wondering why tests weren't working...
Turns out I was running tests from my local checkout, but it was importing code from some other version of cli in my gopath.

Any particular reason the tests use a different package name?

jszwedko:
@cpuguy83 no worries! Having them in a separate package is nice in that it forces you to only test the public interface, but I agree that it isn't idiomatic and can cause problems like the one you ran into.

jszwedko:
I think we can actually take advantage of the fact that the context knows its parent now rather than maintaining this list ourselves. Let me see what I can come up with.

jszwedko:
Never mind, I forgot we are passing the command, not the context, to the template. In that case, this implementation seems reasonable.

cpuguy83:
@jszwedko :) I was originally assuming context was being passed in, hence the first PR

Issue #248, opened 2015-06-25
If the Version string is empty, the version section should not appear in usage output


MartyMacGyver:
If the Version string is empty, then the version section should not be displayed at all in the usage output. Currently if the version string is empty, the usage output prints "VERSION:" followed by an empty line.

It would be up to the user to set HideVersion=true to remove the now-superfluous option when they define Version=""

(I was originally going to suggest that the default be an empty string as well, but that could have unwanted side-effects on downstream projects. Instead, one would simply empty the Version string to cause it to be hidden from the usage output, which should not have side-effects).

MartyMacGyver:
I've submitted pull request #250 to address this issue (note that this pull request contains the branch for pull request #249 as well)

jszwedko:
Closed via #251; thank you!

tonglil:
How come when I use this package it still shows the version?

MartyMacGyver:
Are you seeing the VERSION: info in the help output or elsewhere?

tonglil:
Yes, in the help output. On second look, I am assuming that PR only affects it not in the help output?

MartyMacGyver:
After getting the pointer from NewApp (e.g., "app"), did you define app.Version as ""?

MartyMacGyver:
In addition, if you don't want the *option* named "version" working nor displayed in the help, you should also set app.HideVersion = true (as noted in the PR above). Then -v will throw an error and -h will not show a version number nor a version global option.

PR #251, opened 2015-06-25
More cleanups

MartyMacGyver:
This adds some formatting cleanups based on issue #248 and handles the case where there are no Commands (don't show the Commands block *and* don't show "command" in the Usage example.)


jszwedko:
Awesome, thank you for these clean-ups @MartyMacGyver.

MartyMacGyver:
Glad I could help!

PR #253, opened 2015-06-29
Fix global flags processing on top level

### LINKED ISSUE #252: Regression: global flags broken for app.Action
I've managed to bisect to 65d50017d4f34772b8d767fb7478b9416b5d30c5 (#227), which broke my usage of global flags, and caused quite a headache trying to determine what went wrong!

Consider the following program, which has one global flag, "foo". On 942282e, the commit before the regression, it correctly outputs `value=blah` when invoked as `./test --foo=blah`. After that commit, it just says `isset=false value=`.

```go
package main

import (
	"log"

	"github.com/codegangsta/cli"
)

func main() {
	a := cli.NewApp()

	a.Flags = []cli.Flag{
		cli.StringFlag{
			Name:  "foo",
			Value: "unset",
		},
	}

	a.Action = ActionMain

	a.RunAndExitOnError()
}

func ActionMain(c *cli.Context) {
	log.Printf("isset=%v value=%v", c.GlobalIsSet("foo"), c.GlobalString("foo"))
}
```

/cc @databus23

databus23:
This fixes a regression introduced by #227. When looking up global flags by walking up the parent context's we need to consider the special case when we are starting at the very top and there is no parent context to start the traversal.

Fixes #252

Ping @jszwedko

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -627,6 +627,23 @@ func TestAppCommandNotFound(t *testing.T) {
 	expect(t, subcommandRun, false)
 }
 
+func TestGlobalFlag(t *testing.T) {
+	var globalFlag string
+	var globalFlagSet bool
+	app := cli.NewApp()
+	app.Flags = []cli.Flag{
+		cli.StringFlag{Name: "global, g", Usage: "global"},
+	}
+	app.Action = func(c *cli.Context) {
+		globalFlag = c.GlobalString("global")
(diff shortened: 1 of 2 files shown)


jszwedko:
Ah! I'd actually did #232 to fix a form of this issue for help/version flags, but for some reason I was thinking that the old behaviour was that you'd use the non-global version when you were at the top level, but I see that is not the case. Thanks for fixing this!

PR #256, opened 2015-07-03
Add website info to the output

saromanov:
Hi!
I think, it would be great to provide information about website on the output
Usage is a trivial
```go
app := cli.NewApp()
app.Name = "greet"
app.Usage = "fight the loneliness!"
app.Website = "https://github.com/codegangsta/cli/"
```
And output with flag ```-h``` will looks like:
```
NAME:
   greet - fight the loneliness!

USAGE:
   greet [global options] command [command options] [arguments...]
   
VERSION:
   0.0.0
   
WEBSITE:
   https://github.com/codegangsta/cli/
```
What do you think?

### DIFF
--- a/app.go
+++ b/app.go
@@ -17,6 +17,8 @@ type App struct {
 	Usage string
 	// Version of the program
 	Version string
+	// Website of the program
+	Website string
 	// List of commands to execute
 	Commands []Command
 	// List of flags to parse
--- a/help.go
+++ b/help.go
@@ -22,6 +22,9 @@ VERSION:
    {{end}}{{if len .Authors}}
 AUTHOR(S): 
    {{range .Authors}}{{ . }}{{end}}
+   {{end}}{{if .Website}}
+WEBSITE:
+   {{.Website}}
    {{end}}{{if .Commands}}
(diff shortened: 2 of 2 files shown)


jszwedko:
@saromanov thanks for this! There have been an influx of these sorts of requests over the last few weeks (licence, copyright, and website) -- I think I may actually prefer to come up with a general key/value place where users can set additional fields like this rather than polluting the `App` struct further (I just worry that we'll see a desire for more of these in the future).

saromanov:
From other side, it would be better, because of amount of fields. But at the same time, "keys" might be different and for every key, user need to write additional documentation. Also, this approcah has a problem about how to locate this fields in console output? Now, every program with ```cli``` has a same output
```
NAME
... 
USAGE
...
```
And this is great. But with a key-value case, there will be not structured. So, probably better way is stay the same, but after some time, is just "freeze" requests for additional of new fields.

meatballhat:
@saromanov This new section in the docs aims to address your concern, fwiw: https://github.com/codegangsta/cli/blob/master/README.md#generated-help-text-customization 😺

PR #257, opened 2015-07-09
Allow hiding default values.

jgeewax:
Also changes how defaults are shown.

### DIFF
--- a/app.go
+++ b/app.go
@@ -51,6 +51,8 @@ type App struct {
 	Email string
 	// Writer writer to write output to
 	Writer io.Writer
+	// The specific usage of arguments
+	ArgumentUsage string
 }
 
 // Tries to find out when this binary was compiled.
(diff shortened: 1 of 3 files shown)


jacobsa:
@codegangsta: Are you able to take a look at these changes? They would be quite helpful for GoogleCloudPlatform/gcsfuse. In particular, we need to hide the default for a flag (mount options, `-o`) that can be repeated and doesn't really have a default per se.

jgeewax:
It's fine to depend on my fork while we wait though, right ?

jacobsa:
Right, it's just the usual issue of drift from upstream; I'd like to depend on the official one sooner rather than later.

jszwedko:
@jgeewax thanks! I like the idea, but I'm not crazy about adding more conditional logic there to support a different "format". I think what may be a more flexible approach would be to have a configurable template that would be executed that could be set at the `App` or maybe even the `Flag` level that way we can support your use case as well as future cases where users want to change the flag usage format.

jszwedko:
If I find some time I'll try to spike out that approach.

jgeewax:
Cool - I'm +1 to that. Letting people specify a template for each item sounds like a great plan.

PR #264, opened 2015-07-20
Improve vendoring options by removing self-referential imports in tests.

wardn:


jszwedko:
@wardn thanks for this, this is definitely more idiomatic.

PR #266, opened 2015-07-28
Added exit code support

### LINKED ISSUE #66: allow actions to return its execution status.

### LINKED ISSUE #164: How best to propagate exit codes to main()?
We've found that using os.Exit(code) in our command functions makes them impossible to test reliably. But we do want to allow our command functions the ability to set the process exit code.

Our current hack is to just use a global variable, set it in the command function when non-zero, and call os.Exit() at the end of main(). But this doesn't seem very elegant.

It seems that there should be a way to pass the exit code back up to the app somehow. Maybe via the Context? We'd then still need a way to get it from main() since the Context is not available there, so maybe adding an App.After() function that could examine the context.ExitCode and decide what to do.

Thoughts?

txgruppi:
Now the exit code can be returned by BeforeFn, ActionFn and AfterFn.

The `os.Exit` function is not called by this packaged

This closes #66 and closes #164

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ export PATH=$PATH:$GOPATH/bin
 ```
 
 ## Getting Started
-One of the philosophies behind cli.go is that an API should be playful and full of discovery. So a cli.go app can be as little as one line of code in `main()`. 
+One of the philosophies behind cli.go is that an API should be playful and full of discovery. So a cli.go app can be as little as one line of code in `main()`.
 
 ``` go
 package main
@@ -57,7 +57,7 @@ func main() {
   app.Action = func(c *cli.Context) {
     println("boom! I say!")
   }
-  
+
   app.Run(os.Args)
 }
 ```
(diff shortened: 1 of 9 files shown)


pjvds:
Is there any reason this has not been pulled yet?

jszwedko:
Apologies for the delay in looking at this. My main hesitation is that this is a backwards incompatible change to the API, I would love to have it though. I'll tag this for version 2.0, thank you for the contribution!

dmathieu:
Any idea what the ETA is for this 2.0 version?

jszwedko:
Hi @dmathieu! Unfortunately there isn't currently an ETA. I don't currently have the time to dedicate to drive forward the changes that would be part of a 2.0 release, but I'm hopeful that at some point I will or that we can bring on more active maintainers to share that load.

meatballhat:
👋  **NOTE** this PR has been superseded by #361 and will close this PR automatically when merged

PR #267, opened 2015-08-03
Add ArgsUsage field to document the use of arguments

tristanz:
This pull request adds an `ArgsUsage` field which simply allows the user to override the `[arguments...]` message.  This is 100% backward compatible.

Right now there is no easy way to document arguments.  For instance:

```
cp src dest
```

would currently show up as:

```
cp [arguments...]
```

That's less than ideal...  This pull requests allows you to write much nicer help messages.

#140 proposes a full blown argument specification, but this is very difficult to do, given how flexible arguments can be.  Rather than trying to come up with the formal grammar for arguments, this pull request just adds a simple description field.  Users can choose how best to format their argument help text.

### DIFF
--- a/app.go
+++ b/app.go
@@ -13,7 +13,7 @@ import (
 type App struct {
 	// The name of the program. Defaults to os.Args[0]
 	Name string
-	// Name of command for help, defaults to Name
+	// Full name of command for help, defaults to Name
 	HelpName string
 	// Description of the program.
 	Usage string
--- a/app.go
+++ b/app.go
@@ -89,7 +89,9 @@ func (a *App) Run(arguments []string) (err error) {
 
 	newCmds := []Command{}
 	for _, c := range a.Commands {
-		c.HelpName = fmt.Sprintf("%s %s", a.HelpName, c.Name)
+		if c.HelpName == "" {
+			c.HelpName = fmt.Sprintf("%s %s", a.HelpName, c.Name)
+		}
 		newCmds = append(newCmds, c)
 	}
 	a.Commands = newCmds
@@ -199,7 +201,9 @@ func (a *App) RunAsSubcommand(ctx *Context) (err error) {
 
 	newCmds := []Command{}
 	for _, c := range a.Commands {
-		c.HelpName = fmt.Sprintf("%s %s", a.HelpName, c.Name)
+		if c.HelpName == "" {
+			c.HelpName = fmt.Sprintf("%s %s", a.HelpName, c.Name)
+		}
 		newCmds = append(newCmds, c)
 	}
 	a.Commands = newCmds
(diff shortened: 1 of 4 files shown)


jszwedko:
@tristanz awesome, thank you for the contribution! I think this is a reasonable step forward until an argument parser is introduced. Mind taking a peek at the test failures when you get a moment?

tristanz:
Will look into this asap.

tristanz:
Fixed.

This also includes better subcommand help messages that displays the full command name.  Users can override this behavior by specifying a custom/shorter HelpName if they want.  The prior behavior was to drop subcommands and display a generic `command`, so I believe this behavior is more consistent and prettier.

If you want me to pull these out into another pull request, let me know.  :disappointed:

jszwedko:
@tristanz awesome, thanks for address the the test failures!

With respect to the subcommand names in the help, I thought that was already addressed by #246, were you seeing something different? Or maybe your branch was based off an earlier commit?

ghicks-rmn:
@jszwedko @tristanz Hopefully I'm not speaking out of turn here but I stumbled upon this module today, and Tristan's changes address the issues that I came across.

#246 updates the help parsing so that a subcommand will show its parent command in the help text. Taking the example of the AWS CLI, say you have an application named `aws`, a command named `ec2`, and a subcommand named `run-instances`. The changes in #246 ensure that the `run-instances` USAGE will show up as `command ec2 run-instances...` in the help text, preserving the command/subcommand chain; but the generic word `command` in the text doesn't reflect the actual command you have to enter to invoke it.

@tristanz's new change updates the command and subcommand templates to reference the original app name in the usage data. In the above example, `run-instances` will have help text showing `aws ec2 run-instances...`, which reflects the real command-line usage.

Definitely +1 on this pull request, both changes greatly improve the generated help text.

tristanz:
Yes, I believe @ghicks-rmn hits the difference on the head.  I was slightly confused by what happened since #246 landed in the middle of this pull request.  But the currently code has what I believe is the desired behavior.  The help text will be

```
aws ec2 run-instances
```

rather than

```
command ec2 run-instances
```

Given how subcommands are structured as apps, I couldn't find a great way to add the base command name without adding a `HelpName` property.  So I added `HelpName` which allows you to override the default behavior.  For instances if you don't want the full command names you can just specify `run-instances`.  I believe full names are better and certainly better than the current generic `command` help text.

The original motivation of this pull request was to override the default `arguments...` using `ArgsUsage`.  So the second part of this pull request allows you do have help text like:

```
cp src dest
```

vs

```
cp arguments...
```

So I think we're good to merge if there are no code refactoring issues.

tboerger:
Any update for that pull request? I really would like to see that change.

gravis:
Would love to see this merged please

tristanz:
Folks interested in this pull request should also investigate https://github.com/jawher/mow.cli.

jszwedko:
This looks good, thank you @tristanz! Apologies for the delay, been a hectic month.

PR #274, opened 2015-08-29
Set default PROG in bash_autocomplete

klueska:
By setting the default value of PROG to the basname of whatever the
filename is, we allow bash_autocomplete to be copied into
/etc/bash_completion.d without modification.


jszwedko:
@klueska I like this, thanks!

PR #278, opened 2015-09-25
Do not return error when flag parsing should be skipped

### LINKED ISSUE #208: SkipFlagParsing doesn't work if first argument is a flag
It prints 'Incorrect usage' if the first argument is a flag:

```go
app := cli.NewApp()
app.Commands = []cli.Command{
  {
    Name: "test",
    Usage: "cli testing",
      for _, x := range c.Args() {
        fmt.Fprintln(os.Stdout, x)
      }
    },
    SkipFlagParsing: true,
  },
}
app.Run(os.Args)
```

```
$ go run main.go test arg --flag
arg
--flag

$ go run main.go test --flag arg
Incorrect Usage.

NAME:
   test - cli testing

USAGE:
   command test [arguments...]
```

nathanleclaire:
Closes #208 

cc @codegangsta @jszwedko 

FWIW I'm not in favor of removing `SkipFlagParsing` in general as it's proven useful in a couple of situations for us so far.  Granted, that may not be the best thing for all users.

Signed-off-by: Nathan LeClaire <nathan.leclaire@gmail.com>

### DIFF
--- a/command.go
+++ b/command.go
@@ -98,6 +98,12 @@ func (c Command) Run(ctx *Context) error {
 		err = set.Parse(append(flagArgs, regularArgs...))
 	} else {
 		err = set.Parse(ctx.Args().Tail())
+
+		// Work around issue where if the first arg in ctx.Args.Tail()
+		// is a flag, set.Parse returns an error
+		if c.SkipFlagParsing {
+			err = nil
+		}
 	}
 
 	if err != nil {
--- a/command_test.go
+++ b/command_test.go
@@ -45,3 +45,25 @@ func TestCommandIgnoreFlags(t *testing.T) {
 
 	expect(t, err, nil)
 }
+
+// Fix bug with ignoring flag parsing that would still parse the first flag
+func TestCommandIgnoreFlagsIncludingFirstArgument(t *testing.T) {
+	app := NewApp()
+	set := flag.NewFlagSet("test", 0)
+	test := []string{"blah", "-break"}
+	set.Parse(test)
+
+	c := NewContext(app, set, nil)
+
+	command := Command{
+		Name:            "test-cmd",
+		Aliases:         []string{"tc"},
+		Usage:           "this is for testing",
+		Description:     "testing",
+		Action:          func(_ *Context) {},
+		SkipFlagParsing: true,
+	}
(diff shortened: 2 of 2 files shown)


nathanleclaire:
I know it may not be the most elegant implementation, but this seems to be the most non-intrusive solution I've come up with so far and includes a test case which fails on `master` but passes in this branch.

jszwedko:
Hi @nathanleclaire, thanks for the submission!

I took a look at the PR that introduced this (https://github.com/codegangsta/cli/pull/80) and am a bit confused about its intention. The naming, `SkipFlagParsing` makes me think that it should skip _all_ flag parsing, but it seems like it is just skipping reordering of flags and args and still ultimately calls `set.Parse`. I think it'd be better to skip the flag parsing in general when this flag is specified, but it will break backwards compatibility so it is worth a second thought. Are you relying on that behavior?

nathanleclaire:
>The naming, SkipFlagParsing makes me think that it should skip all flag parsing, but it seems like it is just skipping reordering of flags and args and still ultimately calls set.Parse

Yeah, I was really confused by that as well.

> I think it'd be better to skip the flag parsing in general when this flag is specified, but it will break backwards compatibility so it is worth a second thought. Are you relying on that behavior?

We're not relying on flag parsing (i.e. the results of `c.Args` or `c.String` etc. populated) when we have `SkipFlagParsing` set to `true`.  In that case we handle any needed querying of the CLI args manually.

nathanleclaire:
@jszwedko How are you feeling on this?  OK to merge it in its current form?  It should have minimal impact on existing use cases outside of `SkipFlagParsing`.

jszwedko:
@nathanleclaire apologies for the delay! I think it'd be least confusing to just skip flag parsing in its entirety if the field is set (even if this breaks things depending on this strange behavior). To this end I put together #283 with your commit and then one more just removing flag parsing if `SkipFlagParsing` is set.

nathanleclaire:
I'll test #283 and see how it works for our use case.  Thanks!

nathanleclaire:
Commented on one issue I have with the proposed PR in that thread.

nathanleclaire:
This is covered by the referenced issue.

PR #279, opened 2015-09-28
Honor HideHelp and HideVersion

nodirt:
App.Run does not check a.Hide{Help,Version} before checking help or version,
as a result customers cannot define their own -v flag (e.g. for verbose)


jszwedko:
@nodirt good catch, thank you!

PR #281, opened 2015-10-06
Remove 'bitchass' from README

thiderman:
Having needless derogatory words in the description of the project is just juvenile, and this stood out as a blemish on an otherwise pretty flawless introduction.


jszwedko:
@thiderman I agree :+1: :+1: , thank you!

PR #283, opened 2015-10-17
Really skip flag parsing

jszwedko:
See #278 for where this was brought up.

### DIFF
--- a/command.go
+++ b/command.go
@@ -102,6 +102,12 @@ func (c Command) Run(ctx *Context) error {
 		err = set.Parse(append(flagArgs, regularArgs...))
 	} else {
 		err = set.Parse(ctx.Args().Tail())
+
+		// Work around issue where if the first arg in ctx.Args.Tail()
+		// is a flag, set.Parse returns an error
+		if c.SkipFlagParsing {
+			err = nil
+		}
 	}
 
 	if err != nil {
--- a/command.go
+++ b/command.go
@@ -86,27 +86,23 @@ func (c Command) Run(ctx *Context) error {
 	}
 
 	var err error
-	if firstFlagIndex > -1 && !c.SkipFlagParsing {
-		args := ctx.Args()
-		regularArgs := make([]string, len(args[1:firstFlagIndex]))
-		copy(regularArgs, args[1:firstFlagIndex])
+	if !c.SkipFlagParsing {
+		if firstFlagIndex > -1 {
+			args := ctx.Args()
+			regularArgs := make([]string, len(args[1:firstFlagIndex]))
+			copy(regularArgs, args[1:firstFlagIndex])
 
-		var flagArgs []string
-		if terminatorIndex > -1 {
-			flagArgs = args[firstFlagIndex:terminatorIndex]
-			regularArgs = append(regularArgs, args[terminatorIndex:]...)
+			var flagArgs []string
+			if terminatorIndex > -1 {
+				flagArgs = args[firstFlagIndex:terminatorIndex]
+				regularArgs = append(regularArgs, args[terminatorIndex:]...)
+			} else {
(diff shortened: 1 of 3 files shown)


nathanleclaire:
So I think I tried this solution and the issue I would always run into is that `c.Args` stops working completely when `SkipFlagParsing` is set to `true`.  Now, granted, I may be misunderstanding how it's meant to work, but I would assume `c.Args()` would still return values (including the literal flag args) if `SkipFlagParsing` is set to `true`.

For instance consider this example:

```golang
package main

import (
	"fmt"
	"os"

	"github.com/codegangsta/cli"
)

func main() {
	app := cli.NewApp()
	app.Name = "foo"
	app.Commands = []cli.Command{
		{
			Name:            "bar",
			SkipFlagParsing: true,
				fmt.Println(c.Args())
			},
		},
	}

	app.Run(os.Args)
}
```

I compiled your PR into separate binaries with and without the `SkipFlagParsing` option.

```console
$ ./skip bar a b c
[]
$ ./noskip bar a b c
[a b c]
```

Whereas compiled with master today it _does_ work:

```console
$ ./skip bar a b c
[a b c]
$ ./skip bar a b --foo c
[a b --foo c]
```

If this is expected behavior that's OK, and maybe I should just use `os.Args` directly in this case, but I wanted to make sure you're advised of this change in behavior before merging in this PR.

jszwedko:
@nathanleclaire mmm, you are correct. I think it would be best to try to keep the arguments in `.Args()`, but this may prove difficult as the call on `Context` currently just delegates to the `FlagSet`. Let me give it a shot.

jszwedko:
Not crazy about the implementation, but it seemed like the least invasive to do as I did in https://github.com/codegangsta/cli/commit/845c83453c000f2fdf729e7f7bad7e2b1703ce4e. Thanks bringing this up!

nathanleclaire:
Cool.  I'll try the code out and get back to you on how it looks from our end.

nathanleclaire:
Everything looks good with this patch that I've tried so far.  It fixes an issue I was running into where `--help` was being parsed in spite of `SkipFlagParsing` as well (maybe you want to add a test for that too?).  Thanks a bunch for taking the time to work on this @jszwedko !

nathanleclaire:
I added a test for `--help` case and also to revise to use a test table (cleaner) here: https://github.com/codegangsta/cli/pull/291 if you are interested @jszwedko

nathanleclaire:
I think I'm still running into some issues with e.g. `-h` with this patch, strangely. I would expect `-h` to be skipped as well, and I though the additional test covered that, but maybe not.

nathanleclaire:
@jszwedko Any update on this?

jszwedko:
Hey @nathanleclaire, apologies for the delay! I wasn't able to reproduce this using the above snippet and this branch, i.e:

```
package main

import (
    "fmt"
    "os"

    "github.com/codegangsta/cli"
)

func main() {
    app := cli.NewApp()
    app.Name = "foo"
    app.Commands = []cli.Command{
        {
            Name:            "bar",
            SkipFlagParsing: true,
                fmt.Println(c.Args())
            },
        },
    }

    app.Run(os.Args)
}
```

```
$ go run -a /tmp/tmp.go bar -h
[-h]
```

Is there something I'm missing?

nathanleclaire:
Oh boy... I'm sorry, this is actually entirely on me: https://github.com/docker/machine/blob/master/commands/ssh.go#L12

:blush:

nathanleclaire:
But at any rate, any ETA on merging this PR?

jszwedko:
Haha :smile: , I think this is good to go now then! Merging.

nathanleclaire:
:confetti_ball: :dancer: :dancers: :+1: :+1: :+1:

PR #286, opened 2015-10-19
only display app version and help message once

### LINKED ISSUE #285: version and help are double printed
This as expected:
 - `app help`

These double the help output or version string, respectively:
 - `app -h`
 - `app --help`
 - `app -h help`
 - `app --help help`
 - `app -v`
 - `app --version`

rmg:
When processing the flags for -h/--help and -v/--version, only check
the flags in checkVersion() and checkHelp() instead of also printing
the associated message.

This fixes the problem of `app -h` and `app -v` printing their output
twice. The doubling was caused by printing the message once for each
registred alias for the given flags (-h/--help and -v/--version).

Resolves #285

### DIFF
--- a/app.go
+++ b/app.go
@@ -138,10 +138,12 @@ func (a *App) Run(arguments []string) (err error) {
 	}
 
 	if !a.HideHelp && checkHelp(context) {
+		ShowAppHelp(context)
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


jszwedko:
:fearful: oops, thanks for catching this @rmg !

PR #289, opened 2015-10-25
Added command to have a custom text on the USAGE section of help

KSubedi:
Setting app.UsageText will override the USAGE section text of help menu with the set string.

### DIFF
--- a/app.go
+++ b/app.go
@@ -17,6 +17,8 @@ type App struct {
 	HelpName string
 	// Description of the program.
 	Usage string
+	// Text to override the USAGE section of help
+	UsageText string
 	// Description of the program argument format.
 	ArgsUsage string
 	// Version of the program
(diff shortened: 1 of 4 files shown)


jszwedko:
@KSubedi I think this is addressed by https://github.com/codegangsta/cli/pull/267, but let me know if you think otherwise!

KSubedi:
@jszwedko I checked that PR yesterday, that will only allow you to override the `[arguments...]` message of the USAGE section only, however this will allow you to override the whole message. It might be useful to create more precise usage messages since you will be able to override all of the text, including the command name.

I had to implement this for a project I was working on, and thought it would be a good idea to submit a PR since it might be handy to others too.

jszwedko:
:+1: I think this is reasonable.

PR #291, opened 2015-10-27
Use a test table and add --help test

nathanleclaire:
I cleaned up the separate tests into just one which uses a [test table](http://dave.cheney.net/2013/06/09/writing-table-driven-tests-in-go) (since I noticed there is a lot of duplication in them) and added a case for `--help` (another issue I had encountered that this patch fixes).

Signed-off-by: Nathan LeClaire <nathan.leclaire@gmail.com>

### DIFF
--- a/command_test.go
+++ b/command_test.go
@@ -1,70 +1,43 @@
 package cli
 
 import (
+	"errors"
 	"flag"
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


jszwedko:
Awesome, thank you for this! :clap: :100:

PR #304, opened 2015-11-28
test: avoid output from "go test" when tests pass

osocurioso:
Some tests where printing to os.Stdout as a side effect even if the
output was not used/checked in the test.


jszwedko:
:+1: thanks!

PR #306, opened 2015-12-09
Update to add yaml support

ChrisPRobinson:


ChrisPRobinson:
Initial pull request to add support for flags to get values from yaml files. Still some work to do to get this working on older versions of go. Open to review comments.

ChrisPRobinson:
I attempted to add build constraints to the file that has yaml dependencies, but its not working yet. Hopefully this will be ok. Took a survey of yaml files and all of them use the encoding package which isn't in go version 1.0 and 1.1.

jszwedko:
Thank you for putting this together @ChrisPRobinson! I'm still trying to find a solid block of time to give this the review it deserves, but I haven't forgotten it.

I think not supporting this feature in Go 1.0 and 1.1 is OK though.

ChrisPRobinson:
@jszwedko: Sounds good. Can't wait for the review. 

I think one area that likely has a bug is requesting for a GenericValue from a yaml source, this likely doesn't work. I'm not sure what the correct scenario is there, on how to decode the yaml value. So I kind of passed on figuring it out.

The other likely issue is with errors. How should I report them when there is a conversion error as an example? Just thought I'd point those out ahead of time.

jszwedko:
@ChrisPRobinson I regret putting this review off for so long, I was intimidated by the SLOC, but this is really well written and easy to follow. I like that you implemented it in an extensible fashion (and provided two initial implementations), this should make it easy to add an INI loader for example.

I agree with you regarding errors -- it looks like conversion errors are only correctly handled in this library when parsed from a flag (via `flagset.Parse`). See https://github.com/codegangsta/cli/blob/master/flag.go#L299-L305 for example where if the value can't be parsed as a boolean it is just ignored. This pattern is repeated for the other types. It feels like `cli.Flag.Apply` should also be able to return an error, but this would break the current API -- I'll note for that`.

In your specific case, it seems like you should be able to have the `ApplyInputSourceValue` interface function have `error` in its return signature and bubble up the errors from https://github.com/codegangsta/cli/pull/306/files#diff-09e57851363e51591d63294f8e7ebd00R38. You'd also want to have all of the methods on `InputSourceContext` be able to return `error`. Unless I'm missing something.

For the generic flag it feels like you shouldn't do any type checking here: https://github.com/codegangsta/cli/pull/306/files#diff-e3d69d75afd9528121339b5ce88daa8dR99 and just return whatever the value is, letting https://github.com/codegangsta/cli/pull/306/files#diff-c614971d7e599304e5f08cbf7fcdff45R39 pass it along to `.Set` which should call the method they defined to unmarshal the value. Come to think of it, it feels like we could delegate all of the type marshalling to the `flag` library and just have the YAML file loader return everything as a `string`.

Thank you again for taking the time to put this together! I'll be much more active around this until we get it merged in.

ChrisPRobinson:
@jszwedko - Sounds good, I'll work on getting those changes in and updated. Thanks for the review and comments.

ChrisPRobinson:
@jszwedko - made the updates you pointed out, one more step to possible completion.

jszwedko:
Awesome @ChrisPRobinson -- this looks great!

Let me know if there is any more touch-ups that you wanted to do, but otherwise I think the only remaining piece is an update to the README.

ChrisPRobinson:
@jszwedko - Yes the README needs updating, updated it.

I think there is one thing that definitely needs to be fixed. I put all of the other packages in "inputfilesource". Looking at this now I don't like this name. Technically speaking the interface allows data to come from really anything not just a file. I think I should switch this to "altinputsource" or perhaps "altsrc"? Let me know what you think there.

The other thing I'm concerned about is the patterns for the InputSourceContext initialization patterns. Right now I implemented one func for the yaml one, but what about the case where the file name is always hardcoded and not pulled from a context variable. Also would we want each alternate directory/package for each alternate input source?

ChrisPRobinson:
@jszwedko - I made a couple of updates, I change the package name to altsrc for alternate source context. I also tweaked the yaml source context and made two new functions that are used to create the source. One where a file is provided directly. Another that uses the cli.Context to get a file name and then  creates the source.

I then made two general functions that can be used for any input source context to load them for the Before method. The idea is that people implementing new sources can just create the required New funcs for their Input source. Then it integrates with the existing common code path that was used for the yaml one. Maybe I should have done this as a separate commit so you could see the difference rather than squashing them. 

The readme is updated with samples

jszwedko:
:+1: I agree that `altsrc` is a better package name and the rest of these changes seem sane. Thank you for updating the README too!

Looks like there is a build failure for Go 1.0 -- I'm also ok with just dropping 1.0 support.

ChrisPRobinson:
@jszwedko - Good, glad you like the changes. I did attempt to optionally compile it for go 1.0 but it worked for go 1.1 but not go 1.2. Let me know if any other changes are required.

jszwedko:
This looks great @ChrisPRobinson, thank you again! I'll merge and drop support for Go 1.0 as I don't see a compelling reason to continue that support.

ChrisPRobinson:
@jszwedko Awesome, glad to help!

PR #311, opened 2015-12-15
Export cli.Flag.GetName (previously cli.Flag.getName)

Jille:
Shouldn't cause BC problems as the previous one wasn't exported.

I need this to write some code that generates shell completions code. (Our initialization is so slow we can't use the normal completion, and we need some caching anyway.)


jszwedko:
:+1: I'd been meaning to do this anyway so that people can implement their own structs that satisfy the interface. Thanks!

Issue #265, opened 2015-07-20
panic on help when redirecting output and the target halts


yichengq:
I use the guide example and cli on latest master branch:
```
package main

import (
	"os"

	"github.com/codegangsta/cli"
)

func main() {
	app := cli.NewApp()
	app.Name = "greet"
	app.Usage = "fight the loneliness!"
	app.Action = func(c *cli.Context) {
		println("Hello friend!")
	}

	app.Run(os.Args)
}
```

If you run it in this way, it works fine:
```
$ ./test | grep --peers 
grep: unrecognized option `--peers'
usage: grep [-abcDEFGHhIiJLlmnOoqRSsUVvwxZ] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]
Hello friend!
```

But when you print help message, it panics:
```
$ ./test -h | grep --peers 
grep: unrecognized option `--peers'
usage: grep [-abcDEFGHhIiJLlmnOoqRSsUVvwxZ] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]
panic: template: help:1:0: executing "help" at <NAME:
   >: write /dev/stdout: broken pipe

goroutine 1 [running]:
github.com/codegangsta/cli.func·004(0x1a0210, 0x1e7, 0x1428c0, 0x20827eb60)
	/Users/unihorn/gopath/src/github.com/codegangsta/cli/app.go:88 +0x1ef
github.com/codegangsta/cli.ShowAppHelp(0x2082ce000)
	/Users/unihorn/gopath/src/github.com/codegangsta/cli/help.go:100 +0x60
github.com/codegangsta/cli.checkHelp(0x2082ce000, 0x2082aa000)
	/Users/unihorn/gopath/src/github.com/codegangsta/cli/help.go:170 +0x88
github.com/codegangsta/cli.(*App).Run(0x20827eb60, 0x20826a000, 0x2, 0x2, 0x0, 0x0)
	/Users/unihorn/gopath/src/github.com/codegangsta/cli/app.go:134 +0x9e8
main.main()
	/Users/unihorn/gopath/src/github.com/yichengq/test/main.go:17 +0xae
```

[...]

PR #320, opened 2016-01-20
Remove panic from help.

technosophos:
There is a panic in printHelp that can be trivially triggered when the
shell closes os.Stdout. This happens, for example, when data is piped
between a cli app and something else.

See https://github.com/helm/helm/issues/387


jszwedko:
Thank you!

I've come around on this from my opinion in #199 -- I still think we should take a look at the other places that we are writing to `stdout`, but this is a good incremental step.

Fixes #265

PR #321, opened 2016-01-22
Improve help messages for empty strings and do os specific envVar handling on windows

blaubaer:
* Improve GenericFlag.String() by suppressing empty "" for GenericFlags on nil or empty Generic.String()
* Cleanup StringFlag.String()
* Add os specific envHint handling for Windows (%ENV_VAR% instead of $ENV_VAR on posix systems)

### DIFF
--- a/flag.go
+++ b/flag.go
@@ -350,9 +350,8 @@ func (f StringFlag) FormatValueHelp() string {
 	s := f.Value
 	if len(s) == 0 {
 		return ""
-	} else {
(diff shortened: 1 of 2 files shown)


jszwedko:
:+1: thank you for this! We should probably drop support for 1.0 at some point (where the compiler isn't able to recognize that all logic paths result in a return).

It'd be nice to be able to test the Windows formatting of env vars outside of running on a Windows system (especially since CI is just Linux), but this is a good step.

PR #323, opened 2016-01-26
Fixed mishandling of a "-"(dash)-argument causing reordering of cli n…

andreas-kupries:
…on-flag arguments.

Added test demonstrating issue (PASS with fix, FAIL without).


jszwedko:
This looks great, thank you for the fix and test case!

PR #328, opened 2016-01-31
fix simple type error.

soarpenguin:


app.go
```diff
@@ -119,13 +119,12 @@ func (a *App) Run(arguments []string) (err error) {
...
 	err = set.Parse(arguments[1:])
 	nerr := normalizeFlags(a.Flags, set)
+	context := NewContext(a, set, nil)
 	if nerr != nil {
 		fmt.Fprintln(a.Writer, nerr)
-		context := NewContext(a, set, nil)
 		ShowAppHelp(context)
 		return nerr
 	}
-	context := NewContext(a, set, nil)
```
jszwedko: I actually just merged #329 which had this same change.

command.go
```diff
@@ -189,7 +189,7 @@ func (c Command) startApp(ctx *Context) error {
 
 	// set the name and usage
 	app.Name = fmt.Sprintf("%s %s", ctx.App.Name, c.Name)
-	if c.HelpName == "" {
+	if c.HelpName != "" {
```
jszwedko: Oops, that is definitely a bug.

jszwedko: Actually, I take that back, that logic appears to be intentional -- was it manifesting in an issue for you?

meatballhat:
@soarpenguin Please yell if you're still having issues related to this 😺

PR #329, opened 2016-02-04
optimization of some places

leonardyp:
repeat context statement because of a is a pointer
performance  optimization
gofmt  code


jszwedko:
Looks good, thank you @leonardyp !

PR #334, opened 2016-02-17
Fix semantic typo

djui:


jszwedko:
Hah, good catch, thank you!

PR #336, opened 2016-02-21
Add NArg method to context

muraty:
This PR adds ```NArg``` method to ```Context``` struct.

IMHO, it would be better to use ```c.NArg()``` just like built-in [flag.NArg](https://golang.org/pkg/flag/#NArg)

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ app.Flags = []cli.Flag {
 }
 app.Action = func(c *cli.Context) {
   name := "someone"
-  if len(c.Args()) > 0 {
+  if c.NArg() > 0 {
     name = c.Args()[0]
   }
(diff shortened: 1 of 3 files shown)


jszwedko:
This is a nice addition -- do you mind adding a test for it? Otherwise, this looks great!

muraty:
Hi @jszwedko, I already added a test case for this new method; https://github.com/codegangsta/cli/pull/336/files#diff-bf503e77c0e6429882a1ec6525e4cc95

jszwedko:
Oops! I must have missed that.

shurcooL:
I wonder if I'm the only one who doesn't like standard library's `func NArg() int`. I wish it just exposed a slice, and user code can check its size with `len`. Simpler and more readable IMO.

PR #339, opened 2016-03-02
Add App.Metadata

jack230230:
Sometimes, we need call cli with other enviroment. So we need pass some params to the app.

This is a example, a interactive console app. I want to read some runes from readline instance to handle the user's input. 

```
func (s *CLIServer) handlePty(l *readline.Instance, channel ssh.Channel, sid string) {
	defer channel.Close()
	defer l.Close()

	for {
		line, err := l.Readline()
		if err == io.EOF {
			break
		}
		line = strings.TrimSpace(line)
		// quit
		lineLower := strings.ToLower(line)
		if lineLower == "exit" || lineLower == "quit" || lineLower == "bye" {
			l.Stdout().Write([]byte("exit...\n"))
			break;
		}
		handleLine(l, line, sid)
	}
}

func handleLine(l *readline.Instance, line string, sid string) {

	a := cli.NewApp()
	a.Commands = appCommands
	a.Name = "example"
	a.Usage = "example command line tool"
	a.UsageText = "command [subcommand] [command opstions] [arguments...]"
	a.Version = "0.1.1"
	a.Author = "wzw"
	a.Writer = l.Stdout()
	a.Extras = make(map[string]interface{})

	a.Extras["readline"] = l
	a.Extras["sid"] = sid
	
	cmds := strings.Split(line, " ")

	if cmds[0] != "example" {
		cmds = append([]string{"example"}, cmds...)
	}
	a.Run(cmds)
}
```

Here is my Command Action.

```
func MyAction(c *cli.Context) {
	page := 1
	pagesize := 10

	var getData = func() (data []model.Task, err error) {
		data, err = api.TaskList(page, pagesize, "")
		page++
		return
	}

	list, err := getData()
	if err != nil {
		log.Fatalln(err.Error())
	}

	printTasks(list, c.App.Writer)

	l, ok := c.App.Extras["readline"].(*readline.Instance)
	if !ok {
		log.Println("c.App.Extras[\"readline\"] is not a readline.Instance")
		return
	}

	l.Terminal.EnterRawMode()
	defer l.Terminal.ExitRawMode()
	buf := bufio.NewReader(l.Config.Stdin)

	for {
		ru, _, err := buf.ReadRune()
		if err != nil {
			break
		}
		if ru == 27 {
			fmt.Fprintln(l.Stdout(), "ESC pressed quiting...")
			break
		} else if ru == 32 {
			list, err := getData()
			if err != nil {
				log.Println(err)
				fmt.Fprintln(l.Stdout(), "getData err")
			}
			printTasks(list, l.Stdout())
		}
	}
}
```

### DIFF
--- a/app.go
+++ b/app.go
@@ -63,7 +63,7 @@ type App struct {
 	// Writer writer to write output to
 	Writer io.Writer
 	// Other custom info
-	Extras map[string]interface{}
+	Metadata map[string]interface{}
 }
 
 // Tries to find out when this binary was compiled.
--- a/app.go
+++ b/app.go
@@ -62,6 +62,8 @@ type App struct {
 	Email string
 	// Writer writer to write output to
 	Writer io.Writer
+	// Other custom info
+	Extras map[string]interface{}
 }
 
 // Tries to find out when this binary was compiled.
--- a/command.go
+++ b/command.go
@@ -197,7 +197,7 @@ func (c Command) HasName(name string) bool {
 
 func (c Command) startApp(ctx *Context) error {
 	app := NewApp()
-
+	app.Extras = ctx.App.Extras
 	// set the name and usage
 	app.Name = fmt.Sprintf("%s %s", ctx.App.Name, c.Name)
 	if c.HelpName == "" {
--- a/command.go
+++ b/command.go
@@ -197,7 +197,7 @@ func (c Command) HasName(name string) bool {
 
 func (c Command) startApp(ctx *Context) error {
 	app := NewApp()
-	app.Extras = ctx.App.Extras
+	app.Metadata = ctx.App.Metadata
 	// set the name and usage
 	app.Name = fmt.Sprintf("%s %s", ctx.App.Name, c.Name)
 	if c.HelpName == "" {


jszwedko:
Hi @jack230230!

I think this could also be accomplished by composition, e.g.:

```
type MyApp struct {
  *cli.App
  Extras map[string]interface{}
}
```

Would that satisfy your use case?

jack230230:
@jszwedko 

Can not read the Extras map from a Context which the Action method define.  Because Extras map is a member of MyApp, and cli.Context.App is a cli.App type pointer which means the cli.Context.App.Extras is not exists.

satoshikumano:
I like the approach using embedded struct to add extra properties.
I can initiate custom type as App and run but I could not refer it as custom type from the context.
I used type assertion and it doesn't allowed.

may need to change type of Context.App to interface{} ?
https://github.com/codegangsta/cli/blob/master/context.go#L16

jack230230:
@satoshikumano That's a good idea.

enkoder:
:+1: for this PR getting merged.

meatballhat:
@jack230230 @satoshikumano Thanks for the feedback!  I'm strongly :-1: to making `Context.App` type `interface{}`, and I'd like to avoid the use of `interface{}` whenever possible.  Switching to an interface type seems like a reasonable first step, although the surface area of `*App` would mean a type something like the following:

``` go
type CategoriesCommandRunner interface {
    Categories() CommandCategories
    Command(string) *Command
    Run([]string) error
    RunAndExitOnError()
    RunAsSubcommand(*Context) error
}

// ...

type Context {
    App CategoriesCommandRunner
    // ...
}
```

If we were to make such a change, then you could use your composite type and cast to your concrete type like so:

``` go
type MyApp struct {
    *cli.App
    Extras map[string]interface{}
}

// ...
    app := &MyApp{}
    app.Action = func(ctx *cli.Context) {
        myApp := ctx.App.(*MyApp)
        if foo, ok := myApp.Extras["foo"]; ok {
            // bar the foo
        }
    }
```

The type casting is a bit of a pain, and I'm not a huge fan of this idea given that it would require type casting internally to access `*App` struct members, but I *think* it might do the trick.

All of that being said, your patch is 2 lines, and what I'm thinking of here is much bigger 😺 , which means that I much prefer your solution, but with one change.  Do you feel strongly about `Extras`, or would `State` or `Metadata` be an acceptable name?

jack230230:
@meatballhat I agree that `Metadata` is a nice choice.

meatballhat:
Woop!

jszwedko:
:shipit:

PR #343, opened 2016-03-11
Never show version if HideVersion=true

radeksimko:
Because the default version is `0.0.0`, it is actually impossible to hide the version just by setting `HideVersion` to `true`.

This PR is fixing it.

### DIFF
--- a/app.go
+++ b/app.go
@@ -32,7 +32,7 @@ type App struct {
 	EnableBashCompletion bool
 	// Boolean to hide built-in help command
 	HideHelp bool
-	// Boolean to hide built-in version flag
+	// Boolean to hide built-in version flag and the VERSION section of help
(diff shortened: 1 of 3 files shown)


jszwedko:
@radeksimko thanks for this!

It looks like the original intention of this struct field is to hide the `-v` flag, but I can definitely see the the usefulness of expanding its control to also hide it from the `-h` output. Given the naming of the field, I think it's ok to introduce this -- even though it isn't strictly backwards compatible -- since it feels to be more in the spirit of the field.

Do you mind updating the comment on the field declaration to include this new behavior?

radeksimko:
> Do you mind updating the comment on the field declaration to include this new behavior?

Done.

jszwedko:
Perfect, thank you!

PR #345, opened 2016-03-19
Fix yaml file loader

ykanda:
* fix typo (yamlSourceLoader -> yamlSourceContext)
* fix function return type (NewYamlSourceFromFlagFunc)


ykanda:
I tried to read from the yaml of the flag, but it was not an error such as the following at the time of build.

```
./my_app.go:103: undefined: altsrc.NewYamlSourceFromFlagFunc
```

I sent a pull request to fix this

jszwedko:
@ykanda oops, thank you for this! I see now that the build flags aren't correct (which is why it wasn't building in CI where we would have seen this failure). I took your changes and also updated the build tags in https://github.com/codegangsta/cli/commit/ad7ed77a7e4434b280c11c9c83d4c3982363a571

PR #346, opened 2016-03-20
Add option to make categories with command, to display a more structured help

jszwedko:
Extension of #159 to add categories to commands.

Changes from that PR:

* Subcommands can have categories now too
* Removed `CategorizedHelp` in lieu of just always categorizing (with the default category being `""`) so that you can have a mix of categorized and uncategorized subcommands.
* Added notes to README about usage

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,6 +328,45 @@ app.Commands = []cli.Command{
 ...
 ```
 
+### Subcommands categories
+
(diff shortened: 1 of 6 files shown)


jszwedko:
Will let this bake for a few -- would love any feedback from @Soulou and @beevik if you get a moment!

beevik:
Tried it out. Looks good!

PR #348, opened 2016-03-30
yaml file laoder has so as to expand the tilde

ykanda:
Currently, we only take in the Unix-like OS

__altsrc is not to expand the tilde__

From the path, which includes the tilde ( "~ / .mysrc") attempted to load a YAML file, but failed.
altsrc does not expand the tilde as "$HOME".

Use the https://github.com/mattes/go-expand-tilde, we want to improve this.
We prepare to commit to pull request.

However, because dependent external library increases, is worried.
If you are satisfied, please let the opinion of the committers.

### DIFF
--- a/altsrc/yaml_command_test.go
+++ b/altsrc/yaml_command_test.go
@@ -12,6 +12,8 @@ import (
 	"testing"
 
 	"github.com/codegangsta/cli"
+
+	"gopkg.in/mattes/go-expand-tilde.v1"
 )
 
 func TestCommandYamlFileTest(t *testing.T) {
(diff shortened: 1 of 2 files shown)


jszwedko:
Hi @ykanda!

`~` is a shell (specifically `bash`, but other shells support this) construct rather than a general file path component and is expanded before passing as an argument to the program. E.g.:

`test-cmd --load ~/.mysrc` should pass something like `/home/jesse/.mysrc` as the value of the `load` flag. Note that if you quote the flag value, the shell will not expand it, which is expected.

Let me know if this helps!

ykanda:
Thanks @jszwedko !!


`"~/.mysrc"` Was writing not good.I as the default value of `--load` option, wanted to set a value, such as `~/.mysrc`.

However, I was a big misunderstanding. I wrote a Go code such as the following.Because it does not pass through the shell, not to expand the `~` is of course.

```
subCommand := new(cli.Command)
subCommand.Flags = []cli.Flag{
	cli.StringFlag{
		Name:   "load",
		Value:  "~/.mysrc",  // default value
	},
}
subCommand.Before = altsrc.InitInputSourceWithContext(
	subCommandCommand.Flags,
	altsrc.NewYamlSourceFromFlagFunc("load"),
)
```

The configuration file is common to be placed in the following `$HOME`. However, the actual path, because it depends on, such as user names, it can not be a specific value on the Go code. So, it was I think the good us to expand the `~` and `$HOME`.And, we want to set it as the default value of `--load` options.

`--load` option without setting the default value, read the implicit configuration file What is better to do on the application side to use the codegangsta/cli?

jszwedko:
@ykanda ah I see!

For that I would recommend using `os/user` to find the home directory and do something like this:

```
u, err := user.Current()
// do something if err != nil
subCommand := new(cli.Command)
subCommand.Flags = []cli.Flag{
    cli.StringFlag{
        Name:   "load",
        Value:  filepath.Join(user.HomeDir, ".mysrc"),  // default value
    },
}
subCommand.Before = altsrc.InitInputSourceWithContext(
    subCommandCommand.Flags,
    altsrc.NewYamlSourceFromFlagFunc("load"),
)
```

ykanda:
@jszwedko

Well, your way is more good. Without undue expansion, and, we use a standard package. And what I wanted is probably this. Thank you for letting me know a good way.

This ticket will be closed. Thank you.

jszwedko:
Happy to help!

PR #349, opened 2016-04-03
use filepath.Base instead of path.Base in Name & HelpName as default.

Felamande:
path.Base does not support backslash on windows.


jszwedko:
:+1: thank you!

PR #350, opened 2016-04-05
Overwrite slice flag defaults when set

### LINKED ISSUE #160: cli.StringSliceFlag and default value
Hello,
I define my flag as follow:

```go
flFilter = cli.StringSliceFlag{
        Name:  "filter",
        Value: &cli.StringSlice{"flag1", "flag2", "flag3"},
}
```

If I start my binary without arguments, `c.StringSlice("filter")` returns `["flag1", "flag2", "flag3"]` as expected.
If I start my binary with `--filter flag4`,  `c.StringSlice("filter")` returns `["flag1", "flag2", "flag3", "flag4"]`. I find this behaviour odd, I expected it to erase the default value.
Even worst, if I start my binary with `--filter flag1`,  `c.StringSlice("filter")` returns `["flag1", "flag2", "flag3", "flag1"]`.

Is this the expected behavior ? I know I have 3 default value so I can do `c.StringSlice("filter")[3:]` if set, but still it's not perfect.

### LINKED ISSUE #307: EnvVar shadows command line parameters
When EnvVar value is set on global flags, command line parameters has no effective.

### LINKED ISSUE #314: Context.IsSet returns false even if StringSlice flag is set
On my environment (mac OS X 10.10.5), `Context.IsSet` does not work when using `StringSliceFlag` .

```go
// `cmd` is a "sub_command" object
cmd.Flags = []cli.Flag{
	cli.StringFlag{
		Name:   "string_arg, s",
		Value:  "",
		Usage:  "snip",
	},
	cli.StringSliceFlag{
		Name:  "string_slice_arg, a",
		Value: &cli.StringSlice{},
		Usage: "snip",
	},
}

// ...snip

fmt.Println(c.IsSet("string_arg"))
fmt.Println(c.IsSet("string_slice_arg"))
```

```bash
./go_binary sub_command -s str1 -a str2
true
false
```

### LINKED ISSUE #319: values get added to cli.StringSlice if both flag and environment values are given
I encountered this case when I gave a command flag value while the environment variable was also set for the same. I am not sure it is a bug/feature. but decided to log it.

meatballhat:
Closes #160 
Closes #319 
Closes #314 
Closes #307 

*NOTE*: this includes an API change:

### old way, no longer valid:

``` go
s := &StringSlice{"foo", "bar"}
i := &IntSlice{1, 2}
```

### new way:

``` go
s := NewStringSlice("foo", "bar")
i := NewIntSlice(1, 2)
```

### DIFF
--- a/cli-migrate-slice-types
+++ b/cli-migrate-slice-types
@@ -1,0 +1,75 @@
+#!/usr/bin/env python
+from __future__ import print_function, unicode_literals
+
+import argparse
+import os
+import re
+import sys
+
+
+DESCRIPTION = """\
+Migrate arbitrary `.go` sources from the pre-v2.0.0 API for StringSlice and
+IntSlice types, optionally writing the changes back to file.
+"""
+SLICE_TYPE_RE = re.compile(
+    '&cli\\.(?P<type>IntSlice|StringSlice){(?P<args>[^}]*)}',
+    flags=re.DOTALL
+)
+
+
+def main(sysargs=sys.argv[:]):
+    parser = argparse.ArgumentParser(
+        description=DESCRIPTION,
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument('basedir', nargs='?', metavar='BASEDIR',
+        type=os.path.abspath, default=os.getcwd())
+    parser.add_argument('-w', '--write', help='write changes back to file',
+        action='store_true', default=False)
+
+    args = parser.parse_args(sysargs[1:])
+
+    for filepath in _find_candidate_files(args.basedir):
+        updated_source = _update_source(filepath)
+        if args.write:
+            print('Updating {!r}'.format(filepath))
+
+            with open(filepath, 'w') as outfile:
+                outfile.write(updated_source)
+        else:
+            print('// -> Updated {!r}'.format(filepath))
+            print(updated_source)
+
+    return 0
+
+
+def _update_source(filepath):
+    with open(filepath) as infile:
+        source = infile.read()
+        return SLICE_TYPE_RE.sub(_slice_type_repl, source)
+
+
+def _slice_type_repl(match):
+    return 'cli.New{}({})'.format(
+        match.groupdict()['type'], match.groupdict()['args']
+    )
+
+
+def _find_candidate_files(basedir):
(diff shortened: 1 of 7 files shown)


jszwedko:
:+1: :100: 

I thought about this a bit more and it does feel like it would be disruptive to backwards compatibility (since the only people who would have been bitten by this must already be specifying a value), but I can't think of another way to address this bug, so I would like to:

* Tag the current master
* Create a CHANGELOG

before merging this just to ease the pain a little.

meatballhat:
@jszwedko SGTM!  I'm happy to do the tagging and CHANGELOG bits.

meatballhat:
Tagging is done.  See #360 for CHANGELOG :smiley_cat: :metal:.

cli-migrate-slice-types
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env python
```
jszwedko: :clap: this is above, beyond, and awesome.

flag.go
```diff
@@ -10,6 +11,10 @@ import (
 	"time"
 )
 
+var (
+	slPfx = fmt.Sprintf("sl:::%d:::", time.Now().UTC().UnixNano())
```
jszwedko: Why is this prefix needed?

meatballhat: This is me being paranoid.  I'm trying to minimize the risk that we'd deserialize something coming out of the stdlib `flag` layer that's not "ours" to deserialize.  My concern isn't so much about safety, but rather aiming for clarity, avoiding unexpected bugs, making it super easy to spot when a string value from the cli layer isn't being handled correctly.

jszwedko: Gotcha, that makes sense to me, do you mind adding a comment to that effect for _enhanced_ clarity?

jszwedko: Otherwise :shipit:

jszwedko:
@meatballhat what do you think about targeting a `2.0` branch with this so that we can continue to iterate against master while we prepare the 2.0 release? Alternatively, we could have a `1.x` branch that we backport to until 2.0 is officially released.

meatballhat:
I'm a fan of a v2 branch, assuming the majority of humans who use cli are not using version pinning.

meatballhat:
Superseded by #392

PR #357, opened 2016-04-22
Parses usage placeholders via back quotes. Resolves #333

### LINKED ISSUE #333: Insufficient option usage help
It's possible to write this:

```go
		cli.StringFlag{
			Name:  "config, c",
			Usage: "Load configuration from this file",
		},
```

Now the help says:

```
GLOBAL OPTIONS:
   --config, -c 	Load configuration from this file
  [...]
```

But this doesn't tell the end user that the flag takes a value. The usual way is to show that it is:

```
   --config FILE, -c FILE	Load configuration from FILE
```

As far as I can see, you can't do this. The `flag` package, of course, has its own way:

```go
flag.String("c", "", "Load configuration from `file`")
```

Here, the backtick-enclosed part of the help will be used to infer what the name of the value is.

kevin-cantwell:
Addresses https://github.com/codegangsta/cli/issues/333

This change adds support for placeholders in usage text. It parses back-quoted names to show that flags take a particular value. 

For example this:
```go
cli.StringFlag{
  Name:  "config, c",
  Usage: "Load configuration from `FILE`",
}
```

Will result in help output like:

```
--config FILE, -c FILE   Load configuration from FILE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,27 @@ app.Action = func(c *cli.Context) {
 
 See full list of flags at http://godoc.org/github.com/codegangsta/cli
 
+#### Placeholder Values
+
+Sometimes it's useful to specify a flag's value within the usage string itself. Such placeholders are
+indicated with back quotes.
+
+For example this:
+```go
+cli.StringFlag{
+  Name:  "config, c",
+  Usage: "Load configuration from `FILE`",
+}
+```
+
+Will result in help output like:
+
+```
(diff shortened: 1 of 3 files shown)


jszwedko:
:+1: looks good, thank you for this @kevin-cantwell! Do you mind updating the README with this usage?

kevin-cantwell:
@jszwedko Added!

jszwedko:
:clap:

PR #361, opened 2016-04-25
Optional exit code support

meatballhat:
This began as a merge resolution of #266 with some whitespace formatting fixes in README, exposed definition of default "success" and "error" exit codes, and style change from `.*Fn$` to `.*Func$` func type names.  Based on discussion below, the patch is evolving:

- [x] switch from `int` in return signature to optional custom `error` type
- [x] does it blend?

See changes to `CHANGELOG.md` for more details.

Closes #266


README.md
```diff
@@ -265,20 +265,20 @@ altsrc.InputSourceContext for their given sources.
...
-		Action: func(c *cli.Context) {
-			// Action to run
-		},
-		Flags: []cli.Flag{
-			NewIntFlag(cli.IntFlag{Name: "test"}),
-			cli.StringFlag{Name: "load"}},
-	}
-	command.Before = InitInputSourceWithContext(command.Flags, NewYamlSourceFromFlagFunc("load"))
-	err := command.Run(c)
+  command := &cli.Command{
```
jszwedko: :+1:

funcs.go
```diff
@@ -0,0 +1,24 @@
+package cli
```
jszwedko: :clap:

jszwedko:
@meatballhat overall :+1:, I like the idea of allowing the centralization of error handling.

Did you also consider this approach in comparison to having `Action`s return an `err`? I don't have a strong opinion yet, but I was just curious if you had any thoughts.

meatballhat:
tbh I feel weird about `Action` returning an `int`, and the switch to multi-returns in general.  I'd rather keep single-return `error` signatures everywhere and provide a helper func to allow folks to bubble up their own exit code, but only if they want to, e.g.:

``` go
// ...
    // ... bleep bloop
    return cli.NewExitError("invalid blooping", 3)
}
```

... which we would handle internally a la:

``` go
err := app.Action(ctx)
if exitErr, ok := err.(*ExitError); ok {
    // do stuff maybe
    os.Exit(exitErr.Code)
}
// else don't os.Exit...
```

whaddya think?

jszwedko:
I am a big fan of this approach over the `int`

meatballhat:
@txgruppi Please jump in and let us know your thoughts! 😻

txgruppi:
@meatballhat I like the idea of having an error that can bubble up the exit code. 😄

meatballhat:
@txgruppi @jszwedko what are your thoughts on using reflection to ease migration such that folks can continue to use `Action = func(ctx *cli.Context) { }` but will get a deprecation warning?  I'm thinking about anyone who's using this library *without* version pinning tools.

txgruppi:
@meatballhat for sure it will be great, I believe the great majority of the users don't use any type of versioning for Go packages. Sadly I am one of them 😞 

I'm not sure about performance costs of using reflection but I believe it doesn't apply to this package

meatballhat:
@txgruppi yep, I have a few projects kicking around that don't version pin 😺   I seriously doubt this will have any performance impact 🔬

meatballhat:
@txgruppi @jszwedko ready for review!

meatballhat:
ftr I removed the v2.0.0 milestone as this PR no longer introduces backward-incompatible API changes.

.travis.yml
```diff
@@ -13,6 +13,10 @@ matrix:
   allow_failures:
     - go: tip
 
+before_script:
+- go get github.com/meatballhat/gfmxr/...
+
 script:
 - go vet ./...
 - go test -v ./...
+- gfmxr -c $(grep -c 'package main' README.md) -s README.md
```
jszwedko: 👏

jszwedko:
@meatballhat nice, I really like the use of reflection to support backwards compatibility.

app.go
```diff
@@ -38,21 +39,22 @@ type App struct {
...
+	Before BeforeFunc
 	// An action to execute after any subcommands are run, but after the subcommand has finished
 	// It is run even if Action() panics
-	After func(context *Context) error
+	After AfterFunc
 	// The action to execute when no subcommands are specified
-	Action func(context *Context)
+	Action interface{}
+	// TODO: replace `Action: interface{}` with `Action: ActionFunc` once some kind
+	// of deprecation period has passed, maybe?
```
jszwedko: :+1:

app.go
```diff
@@ -198,12 +204,19 @@ func (a *App) Run(arguments []string) (err error) {
...
+	if err != nil {
+		HandleExitCoder(err)
+	}
+	return err
 }
 
-// Another entry point to the cli app, takes care of passing arguments and error handling
+// DEPRECATED: Another entry point to the cli app, takes care of passing arguments and error handling
 func (a *App) RunAndExitOnError() {
+	fmt.Fprintln(os.Stderr,
```
jszwedko: I think deprecating this is great, but I'm a little worried that the end-user (consumer of whatever CLI app uses this library) will see this warning instead of the programmer if they blindly update the library and release. Hopefully this would rarely happen, but maybe we include a note to inform the distributor in case an end-user sees this?

app.go
```diff
@@ -361,3 +380,32 @@ func (a Author) String() string {
 
 	return fmt.Sprintf("%v %v", a.Name, e)
 }
+
+// HandleAction uses ✧✧✧reflection✧✧✧ to figure out if the given Action is an
+// ActionFunc, LegacyActionFunc, or some other invalid thing.  If it's an
+// ActionFunc or LegacyActionFunc, the func is run!
```
jszwedko: :heart:

errors.go
```diff
@@ -0,0 +1,75 @@
...
+	for i, err := range m.Errors {
+		errs[i] = err.Error()
+	}
+
+	return strings.Join(errs, "\n")
+}
+
+// ExitCoder is the interface checked by `App` and `Command` for a custom exit
+// code
+type ExitCoder interface {
```
jszwedko: I really like this.

jszwedko:
I left one comment that may need action taken, but otherwise this is great. I like that you were able to preserve backwards compatibility (at the cost of type safety, but I think it is acceptable in this case). We can remove support for the deprecated signature in the future.

The only other way I can think of preserving backwards compatibility is to add another field on the struct, but I think this would introduce a more complexity into the code (thus causing bugs) so I think your approach is appropriate.

app.go
```diff
@@ -361,3 +380,32 @@ func (a Author) String() string {
...
+func HandleAction(action interface{}, context *Context) error {
+	if reflect.TypeOf(action).Kind() != reflect.Func {
+		panic("given Action must be a func")
```
txgruppi: Can't this be a returned error?

PS: I just don't like to use panic

app.go
```diff
@@ -361,3 +380,32 @@ func (a Author) String() string {
...
+	if len(vals) == 0 {
+		fmt.Fprintln(os.Stderr,
+			"DEPRECATED Action signature.  Must be `cli.ActionFunc`")
+		return nil
+	}
+
+	if len(vals) > 1 {
+		fmt.Fprintln(os.Stderr,
+			"ERROR invalid Action signature.  Must be `cli.ActionFunc`")
+		panic("given Action has invalid return signature")
```
txgruppi: Can't this be a returned error?

app.go
```diff
@@ -404,16 +417,16 @@ func HandleAction(action interface{}, context *Context) error {
 	if len(vals) == 0 {
 		fmt.Fprintln(os.Stderr,
 			"DEPRECATED Action signature.  Must be `cli.ActionFunc`")
-		return nil
+		return err
```
jszwedko: I kinda like `return nil` here -- `return err` visually implies to me that it it is not `nil`.

jszwedko: Aaah, nvm, I think I see now that this would end up being a real error if there is a panic.

jszwedko: Wait, maybe, now I've confused myself. Is there a case where this wouldn't return `nil`?

app.go
```diff
@@ -11,6 +11,15 @@ import (
 	"time"
 )
 
+var (
+	errNonFuncAction = NewExitError("ERROR invalid Action type.  "+
```
jszwedko: I like the addition of the hyperlink, but I still think it'd be nice to include something like: `This is an error in the application, please contact the distributor of this application if this is not you` just in case I, as a non-programmer, install a program using `codegangsta/cli` and receive this error message.

Harkening back to `Please contact your system administrator` which was always amusing when that was indeed you.

app.go
meatballhat: I think this branch will only allow `nil`, yep.  I think I was being overly zealous with the `defer` impl above.

app.go
```diff
@@ -14,11 +14,13 @@ import (
 var (
 	appActionDeprecationURL = "https://github.com/codegangsta/cli/blob/master/CHANGELOG.md#deprecated-cli-app-action-signature"
 
+	contactSysadmin = "This is an error in the application.  Please contact the distributor of this application if this is not you."
```
jszwedko: :clap:

jszwedko:
:shipit:

PR #364, opened 2016-04-28
Revising/adding tests that assert order of operations inside App.Run

### LINKED ISSUE #337: App.Before looks like it executes before sub commands but comments say it executes after
Which is correct. If after I'd like a good way to set global options on my app before my subcommands execute like for instance:

command -v subcommand --flag value

so that I can take -v and turn on verbose logging for all of the subcommands without reepeating the code. Or is there a way to do this already?

meatballhat:
Closes #337

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -13,6 +13,10 @@ import (
 	"testing"
 )
 
+type opCounts struct {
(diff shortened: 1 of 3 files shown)


jszwedko:
:+1: :+1:

PR #365, opened 2016-04-29
altsrc: allow nested defaults in yaml files

roboll:
Ran into some issues with nested default keys in yaml - this patch should allow nested values in yaml to be used in all cases. All the existing tests still pass and I included variations of them for nested keys. The new behavior is only a fallback, so shouldn't change the public api / expected behaviors at all.

    Previously, defaults specified as nested keys in a yaml file would not
    be recognized, i.e. `top: \n  bottom: key` would not be accessible using
    the name `top.bottom`, but `top.bottom: key` would. These changes
    support using nested keys by traversing the configuration tree if the
    key name uses '.' as a delimiter.

### DIFF
--- a/altsrc/flag_test.go
+++ b/altsrc/flag_test.go
@@ -296,7 +296,7 @@ func TestFloat64ApplyInputSourceMethodEnvVarSet(t *testing.T) {
 }
 
 func runTest(t *testing.T, test testApplyInputSource) *cli.Context {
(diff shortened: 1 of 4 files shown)


meatballhat:
@roboll Nice!  Good to merge as-is?

meatballhat:
LGTM btw 😺

roboll:
:+1: Not sure if the question was to me, but as far as I'm concerned this is complete. I've been using it in my fork and haven't run into an issues.

PR #367, opened 2016-04-30
Merging #234 set context values

meatballhat:
Closes #234

### DIFF
--- a/context.go
+++ b/context.go
@@ -146,6 +146,11 @@ func (c *Context) Set(name, value string) error {
 	return c.flagSet.Set(name, value)
 }
 
+// GlobalSet sets a context flag to a value on the global flagset
+func (c *Context) GlobalSet(name, value string) error {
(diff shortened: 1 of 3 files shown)


context.go
```diff
@@ -141,6 +141,16 @@ func (c *Context) NumFlags() int {
...
 }
 
+// Set sets a context flag to a value.
+func (c *Context) Set(name, value string) error {
+	return c.flagSet.Set(name, value)
+}
+
+// GlobalSet sets a context flag to a value on the global flagset
+func (c *Context) GlobalSet(name, value string) error {
+	return globalContext(c).flagSet.Set(name, value)
```
jszwedko: I think this may only want to walk up the tree as far as needed to find where the flag is defined (see `lookupGlobalFlagSet` e.g.).

meatballhat: Oh!  I clearly wasn't thinking about it that way 😸

PR #375, opened 2016-05-02
	Assert type against actual return val's interface.

PaulMakepeace:
Exit code example produces now correctly,
	https://github.com/codegangsta/cli#exit-code

	```
	$ ./ec --ginger-crouton=false
	it is not in the soup
	$ echo $?
	86
	$
	```


meatballhat:
Egad shame on me for not checking PRs first.  Sorry!

PR #381, opened 2016-05-02
Allow for pluggable flag-level help text formatting

meatballhat:
by defining `cli.DefaultFlagStringFunc` with a default value that uses
`withEnvHint`, conditionally running a given flag's `FormatValueHelp` if
present.

Closes #257
Closes #225

### DIFF
--- a/flag.go
+++ b/flag.go
@@ -31,7 +31,7 @@ var HelpFlag = BoolFlag{
 	Usage: "show help",
 }
 
-var DefaultFlagStringFunc FlagStringFunc = flagStringer
+var FlagStringer FlagStringFunc = stringifyFlag
 
 // Flag is a common interface related to parsing flags in cli.
 // For more advanced flag parsing techniques, it is recommended that
@@ -79,7 +79,7 @@ type GenericFlag struct {
 // help text to the user (uses the String() method of the generic flag to show
 // the value)
 func (f GenericFlag) String() string {
-	return DefaultFlagStringFunc(f)
+	return FlagStringer(f)
 }
 
 func (f GenericFlag) FormatValueHelp() string {
@@ -147,7 +147,7 @@ type StringSliceFlag struct {
 
 // String returns the usage
 func (f StringSliceFlag) String() string {
-	return DefaultFlagStringFunc(f)
+	return FlagStringer(f)
 }
 
 // Apply populates the flag given the flag set and environment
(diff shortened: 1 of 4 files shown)


flag.go
```diff
@@ -595,3 +596,27 @@ func withEnvHint(envVar, str string) string {
 	}
 	return str + envText
 }
+
+func stringifyFlag(f Flag) string {
+	fv := reflect.ValueOf(f)
+	placeholder, usage := unquoteUsage(fv.FieldByName("Usage").String())
```
jszwedko: I'm not crazy about the implicit contract here -- e.g. considering if a user wants to create their own flag. Is it possible to add the needed methods to the `Flag` interface to extract this data?

jszwedko: Actually, didn't realize this is package private, that will change my opinion.

meatballhat: Has that opinion changed yet, eh? 😁

jszwedko: Hehe, I'm still curious what you think of using an interface here. Maybe creating a:

```go
type StringableFlag interface {
  NameString() string
  UsageString() string
  ValueString() string
  FormatValueHelp() function ...
}
```

Or something similar. Perhaps also adding additional methods to the interface to guard against other info the user may want to extract

I'm just wary of forcing the user to use reflection if they want to provide their own `FlagStringer` to do anything meaningful.

jszwedko: Though this will break existing user created flags which will immediately fail to satisfy that interface (as well as only storing a list of `Flag`s) internally.

flag.go
```diff
@@ -595,3 +596,27 @@ func withEnvHint(envVar, str string) string {
...
+
+	defaultValueString := ""
+	if val := fv.FieldByName("Value"); val.IsValid() {
+		defaultValueString = fmt.Sprintf("%v", val.Interface())
+	}
+
+	formatFunc := fv.MethodByName("FormatValueHelp")
+	if formatFunc.IsValid() {
+		defaultValueString = formatFunc.Call([]reflect.Value{})[0].String()
```
jszwedko: At the very least, we should assert the signature of this method before calling it.

flag.go
jszwedko: Could we make `FlagStringer` private? That would assuage my fears a little in that at least we keep the reflection internal to the library.

jszwedko: Never mind, that would mean this wouldn't address #257. Let's keep it as is -- this would be an advanced feature anyway.

jszwedko:
:shipit:

flag.go
jszwedko: Never mind! I see that this was removed.

PR #387, opened 2016-05-03
Merging #173 @smallfish SkipFlagNotDefined & GlobalArgs

meatballhat:
Closes #173

### DIFF
--- a/command.go
+++ b/command.go
@@ -7,6 +7,12 @@ import (
 	"strings"
 )
 
+const (
+	// Hacky hack string error shenanigans.
+	// See https://github.com/golang/go/blob/0104a31/src/flag/flag.go#L858
+	knownStdlibFlagNotDefinedErrPrefix = "flag provided but not defined: -"
+)
+
 // Command is a subcommand for a cli.App.
 type Command struct {
 	// The name of the command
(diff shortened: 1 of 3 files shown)


command.go
```diff
@@ -7,6 +7,12 @@ import (
 	"strings"
 )
 
+const (
+	// Hacky hack string error shenanigans.
+	// See https://github.com/golang/go/blob/0104a31/src/flag/flag.go#L858
+	knownStdlibFlagNotDefinedErrPrefix = "flag provided but not defined: -"
```
jszwedko: Sometimes I wish the `stdlib` used more structured errors.

jszwedko:
@meatballhat I missed this before when I saw #173, but it looks like the `flag` stdlib immediately stops parsing when it encounters a flag that it is not defined (https://github.com/golang/go/blob/0104a31/src/flag/flag.go#L896-L917). I feel like the spirit of a `SkipFlagNotDefined` field would be to parse all of the possible flags (ignoring the undefined ones -- or possibly returning them as arguments?), but I don't believe that is feasible with `flag`.

Thoughts? I'm just not sure how useful the behavior introduced here would be.

meatballhat:
@jszwedko Yep, you're totally right.  I don't think this behavior can be (meaningfully) introduced while still depending on stdlib flag.

meatballhat:
@smallfish Please feel free to continue the discussion here.  It may be that we should consider dropping the stdlib flag layer entirely, or perhaps maintain a vendored fork.

PR #389, opened 2016-05-04
Include directories in shell completion

### LINKED ISSUE #388: bash auto-completion with directories as argument
Hey,
I'm trying to build something that uses a directory as first argument. Something like:
`command sub-command [flags] directory-argument` 

If I enable bash auto-completion, it doesn't show up the directories that I can use as argument.

I was looking that `command sub-command [flags] fo[TAB]` would present
`foo/ foo2/ foo3/` and `command sub-command [flags] foo/[TAB]` would present `bar/ bar2/ bar3/` (being `bar*` the directories inside `foo`.

Is this possible or is it too hard to implement?

meatballhat:
Closes #388

### wait for it 🐉 
- [ ] port to `./autocomplete/zsh_autocomplete`
- [ ] add some tests
- [ ] does it blend?

### example
Here's some real interaction with an example app:

``` go
package main

import (
        "fmt"
        "os"

        "github.com/codegangsta/cli"
)

func main() {
        app := cli.NewApp()

        app.Commands = []cli.Command{
                cli.Command{
                        Name: "wheespurr",
                                fmt.Println("shhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh")
                                return nil
                        },
                },
                cli.Command{
                        Name: "do",
                        Flags: []cli.Flag{
                                cli.StringSliceFlag{
                                        Name:  "hats",
                                        Value: &cli.StringSlice{"top", "fruit"},
                                },
                        },
                        Subcommands: cli.Commands{
                                cli.Command{
                                        Name: "snore",
                                                fmt.Println("snxxxxxxxx")
                                                return nil
                                        },
                                },
                        },
                                defer func() {
                                        if r := recover(); r != nil {
                                                fmt.Printf("%s\n    -> %#v\n", r, r)
                                        }
                                }()
                                fmt.Printf("%#v\n", ctx.GlobalStringSlice("hats"))
                                return nil
                        },
                },
        }

        app.Run(os.Args)
}
```

I've compiled it to `./.t/example`, and here's a session with the `./autocomplete/bash_autocomplete` from this PR:

``` shell-session
failb0wl:.t me$ tree .
.
├── example
├── example.go
├── a
│   ├── b1
│   │   └── c
│   │       └── d
│   │           └── exclaim
│   ├── b2
│   ├── b3
│   └── b4
├── a.txt
└── b.txt
failb0wl:.t me$ PROG=example source ../autocomplete/bash_autocomplete ; export PATH="$PWD:$PATH"
failb0wl:.t me$ example <TAB>
do         h          help       wheespurr
failb0wl:.t me$ example a<TAB>/<TAB>
a/b1/  a/b2/  a/b3/  a/b4/
failb0wl:.t me$ example a/b1<TAB>/<TAB>c/<TAB>d/<TAB>
```

### DIFF
--- a/autocomplete/bash_autocomplete
+++ b/autocomplete/bash_autocomplete
@@ -1,14 +1,27 @@
-#! /bin/bash
+#!/bin/bash
 
 : ${PROG:=$(basename ${BASH_SOURCE})}
 
 _cli_bash_autocomplete() {
-     local cur opts base
-     COMPREPLY=()
-     cur="${COMP_WORDS[COMP_CWORD]}"
-     opts=$( ${COMP_WORDS[@]:0:$COMP_CWORD} --generate-bash-completion )
-     COMPREPLY=( $(compgen -W "${opts}" -- ${cur}) )
-     return 0
- }
-  
- complete -F _cli_bash_autocomplete $PROG
+    local cur opts base replylen
+
+    cur="${COMP_WORDS[COMP_CWORD]}"
+
+    if [ -n "${cur}" ] ; then
+      COMPREPLY=( $(compgen -S / -d -- ${cur}) )
+    else
+      COMPREPLY=()
+    fi
+
+    replylen=${#COMPREPLY}
+    opts=$( ${COMP_WORDS[@]:0:$COMP_CWORD} --generate-bash-completion )
+    COMPREPLY=( "${COMPREPLY[@]}" $(compgen -W "${opts}" -- ${cur}) )
+
+    if [ ${#COMPREPLY} -eq ${replylen} ] ; then
+      compopt -o nospace
+    fi
+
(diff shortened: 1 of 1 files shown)


jszwedko:
@meatballhat does this only autocomplete directories? I feel like better behavior would be to include both files and directories.

meatballhat:
@jszwedko Yes, this only autocompletes directories.  I agree this is unexpected behavior, but it was intentionally done that way to address #388.  Given that I don't believe the autocomplete files that ship with cli should be trying to address any and all completion behavior, I'd say that including dirs *and* files **might** be an acceptable addition.

meatballhat:
I'm closing the PR, but I'm happy to continue discussing on #388

PR #393, opened 2016-05-05
Provide a variable for writing output (the writer) with a default of os.Stderr

mattfarina:
This allows:

- Users to set their own writer (e.g., to wrap the text or send it elsewhere such as a log)
- Allows tests to capture the output and check responses.

### DIFF
--- a/app.go
+++ b/app.go
@@ -228,11 +228,11 @@ func (a *App) Run(arguments []string) (err error) {
 
 // DEPRECATED: Another entry point to the cli app, takes care of passing arguments and error handling
 func (a *App) RunAndExitOnError() {
-	fmt.Fprintf(os.Stderr,
+	fmt.Fprintf(OutWriter,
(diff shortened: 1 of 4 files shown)


meatballhat:
@mattfarina Thanks for the PR!  I would support this if it were implemented similarly to `App.Writer`.  Perhaps `App.ErrorWriter`?

mattfarina:
@meatballhat There are several places a change like that would work. I'm not sure how to apply it to the `Apply` method on `IntSliceFlag` or inside `HandleExitCoder`. Do you have any guidance on those cases?

meatballhat:
How about defining ErrorWriter as both a package level var and a member of
App?  nil at the App level falls back to package level value.  Sound OK?

mattfarina:
@meatballhat I used `ErrWriter` (shortening Error because it's the Go pattern). Is the nil check an acceptable pattern?

jszwedko:
:+1: this seems reasonable to me

meatballhat:
Lovely! 👍

PR #396, opened 2016-05-07
Exit nonzero on unknown command. Resolves #276.

### LINKED ISSUE #276: Exit code != 0 in case an invalid command is called?
It seems logical to expect a non zero exit code in case a non existing command is specified.

At the moment it just prints the info/warning message: `No help topic for 'asdf'`
But the exit code is 0

danslimmon:
When an unknown command is given (iff the app has custom commands at
all!) the usage message is printed and we exit with status 3.

### DIFF
--- a/app.go
+++ b/app.go
@@ -217,6 +217,18 @@ func (a *App) Run(arguments []string) (err error) {
 		if c != nil {
 			return c.Run(context)
 		}
+
+		if len(a.Commands) != 1 && a.CommandNotFound == nil {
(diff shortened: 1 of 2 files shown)


jszwedko:
Hi @danslimmon, thanks for getting the ball rolling on this!

This may not cover all of the cases where that help message is printed (specifically for subcommands), but I have an idea for something that might be more generic. I'll follow up with another PR to get your feedback.

jszwedko:
Proposed alternate approach in #399 which handles the unknown command under a subcommand case, let me know what you think!

PR #398, opened 2016-05-07
Remove reordering of flags and arguments

### LINKED ISSUE #103: Reordering of args to Command if BoolFlag is after args

### LINKED ISSUE #355: cli.Command: Argument order is changed depending on the flag set
Hello!

I'm not sure if this is the expected behavior, however the following code:

```go
package main

import (
	"fmt"
	"os"

	"github.com/codegangsta/cli"
)

func main() {
	app := cli.NewApp()

	app.Commands = []cli.Command{
		cli.Command{
			Name:  "update",
			Usage: "update a report",
			Flags: []cli.Flag{
				cli.StringFlag{
					Name:  "filename",
					Usage: "filename",
				},
			},
				fmt.Println(c.Args())
			},
		},
	}

	app.Run(os.Args)
}
```

Runs differently depending on the position of the arguments passed:

```bash
$ ./main update Arg1 Arg2 Arg3 --filename test.json
[Arg1 Arg2 Arg3]

$ ./main update Arg1 --filename test.json Arg2 Arg3
[Arg2 Arg3 Arg1]
```

Expected behavior for me would be that when running the second command, the following should output:

```bash
$ ./main update Arg1 --filename test.json Arg2 Arg3
[Arg1 Arg2 Arg3]
```

It seems that the fact that the flag is in the middle is breaking the Args() order.

Thanks

jszwedko:
This was introduced by #36, but only worked in the specific case of all
arguments being passed before all flags. If the user mixed them, they
ended up with odd parsing behavior where the arguments were reordered
(causing #103 and #355).

Given the tradeoffs I think we should remove support for flag
reordering.

Fixes #103
Fixes #355

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,15 @@
 
 ### Removed
 - the ability to specify `&StringSlice{...string}` or `&IntSlice{...int}`.
-To migrate to the new API, you may choose to run [this helper
-(python) script](./cli-migrate-slice-types).
+  To migrate to the new API, you may choose to run [this helper
+  (python) script](./cli-migrate-slice-types).
+- The optimistic reordering of arguments and flags introduced by
+  https://github.com/codegangsta/cli/pull/36. This behavior only worked when
+  all arguments appeared before all flags, but caused [weird issues with boolean
+  flags](https://github.com/codegangsta/cli/issues/103) and [reordering of the
+  arguments](https://github.com/codegangsta/cli/issues/355) when the user
+  attempted to mix flags and arguments. Given the trade-offs we removed support
+  for this reordering.
 
 ## [Unreleased] - (1.x series)
 ### Added
--- a/app_test.go
+++ b/app_test.go
@@ -189,29 +189,6 @@ func TestApp_Command(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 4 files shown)


meatballhat:
yisssss

PR #399, opened 2016-05-08
Exit non-zero if a unknown subcommand is given

### LINKED ISSUE #276: Exit code != 0 in case an invalid command is called?

jszwedko:
Currently it just prints the help message and exits 0.

We do this by modifying the helpCommand and helpSubcommand cli.Commands
to return an error if they are called with an unknown subcommand. This
propogates up to the app which exits with 3 and prints the error.

Thanks to @danslimmon for the initial approach!

Fixes #276

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,15 @@
 
 ### Changed
 - `Float64Flag`, `IntFlag`, and `DurationFlag` default values are no longer
-quoted in help text output.
+  quoted in help text output.
 - All flag types now include `(default: {value})` strings following usage when a
-default value can be (reasonably) detected.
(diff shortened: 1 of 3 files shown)


jszwedko:
cc/ @meatballhat @danslimmon

danslimmon:
@jszwedko Nice. It's a new codebase for me, so I was trying to be as non-invasive as possible with #396. But your way is better.

jszwedko:
I know that feeling, thank you for the feedback!

PR #400, opened 2016-05-09
Add a vet/test runner script with coverage on by default

meatballhat:
The script is written in Python in the hope that it is available "everywhere", and is slightly more available to Windows folks than bash, although it's been years since I developed primarily on Windows, so I'm happy to port it to something else.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+*.coverprofile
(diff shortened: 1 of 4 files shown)


jszwedko:
:clap: :100:

jszwedko:
Good call on using Python to be more Windows friendly.

PR #406, opened 2016-05-10
Fix command alias printing in help text

### LINKED ISSUE #405: Help does not display command aliases
Hello,

First I would like to say that your project is awesome and it is the best CLI library I have used! Keep up the good work! Thank you for that.

I have a question related to latest version. Recently I have noticed that when displaying help it does no longer show aliases next to the command names. 

When command is defined in the following way, no aliases are shown in the help output
`{Name:     "nginx-start",
 Aliases:  []string{"ngstart", "nstart"},
 Category: cat(NginxCategory),}`

When deprecated ShortName is used instead Aliases, help does display alias properly.
`{
    Name:     "nginx-start",
    ShortName: "ngstart",
    Category: cat(NginxCategory),
    ...
}`

Is this change in behaviour intentional and we should define a custom help template or did it simply slipped during refactoring?

meatballhat:
Closes #405

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 ### Added
 - `./runtests` test runner with coverage tracking by default
 
+### Fixed
+- Printing of command aliases in help text
+
 ## [1.17.0] - 2016-05-09
 ### Added
 - Pluggable flag-level help text rendering via `cli.DefaultFlagStringFunc`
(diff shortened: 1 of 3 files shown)


jszwedko:
Doh, thanks for catching that!

tianon:
Would you be open to a knob for disabling this (for only printing `.Name` instead), or should I just plan to roll our own help template when we get this change pulled in?  Or perhaps a new `HiddenNames`/`HiddenAliases` field (analogous to hidden flags)?

PR #408, opened 2016-05-14
Support hidden flags

dhiltgen:
Sometimes you might want to have command flags that aren't displayed
in the help output.  This change adds a new `GetUsage` to the `Flag`
interface so that help templates can be constructed to easily filter
out any flags that don't have usage information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -571,6 +571,55 @@ VERSION:
   cli.NewApp().Run(os.Args)
 }
 ```
+### Hidden flags
+
+Sometimes its useful to have some flags that aren't displayed in the help
+output.
+
+```go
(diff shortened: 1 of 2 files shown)


meatballhat:
@dhiltgen Hey there :wave: :smiley:  Is the existing support via the `Hidden` struct member not sufficient?  It was [recently merged](https://github.com/codegangsta/cli/commit/fed78b8bab633ee554cfa229e9975ae731d40f5d), fwiw.

dhiltgen:
Nice!  Hadn't noticed when I rebased to top of tree.  That should do the trick for my usecase.

PR #410, opened 2016-05-16
Remove all Context.Global* methods

### LINKED ISSUE #385: Remove support for `Global*` methods.
Instead, have the normal flag fetch methods always walk up the context hierarchy until they find their matching flag (or hit the root).

meatballhat:
and change the behavior of the non-Global variants to always search up the
context lineage.

Closes #385

### DIFF
--- a/context.go
+++ b/context.go
@@ -28,129 +28,76 @@ func NewContext(app *App, set *flag.FlagSet, parentCtx *Context) *Context {
 
 // Int looks up the value of a local int flag, returns 0 if no int flag exists
 func (c *Context) Int(name string) int {
-	return lookupInt(name, c.flagSet)
-}
-
-// Duration looks up the value of a local time.Duration flag, returns 0 if no
-// time.Duration flag exists
-func (c *Context) Duration(name string) time.Duration {
-	return lookupDuration(name, c.flagSet)
-}
-
-// Float64 looks up the value of a local float64 flag, returns 0 if no float64
-// flag exists
-func (c *Context) Float64(name string) float64 {
-	return lookupFloat64(name, c.flagSet)
-}
-
-// Bool looks up the value of a local bool flag, returns false if no bool flag exists
-func (c *Context) Bool(name string) bool {
-	return lookupBool(name, c.flagSet)
-}
-
-// BoolT looks up the value of a local boolT flag, returns false if no bool flag exists
-func (c *Context) BoolT(name string) bool {
-	return lookupBoolT(name, c.flagSet)
-}
-
-// String looks up the value of a local string flag, returns "" if no string flag exists
-func (c *Context) String(name string) string {
-	return lookupString(name, c.flagSet)
-}
-
-// StringSlice looks up the value of a local string slice flag, returns nil if no
-// string slice flag exists
-func (c *Context) StringSlice(name string) []string {
-	return lookupStringSlice(name, c.flagSet)
-}
-
-// IntSlice looks up the value of a local int slice flag, returns nil if no int
-// slice flag exists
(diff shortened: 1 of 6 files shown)


context.go
```diff
@@ -166,68 +113,63 @@ func (c *Context) Set(name, value string) error {
...
 	return c.parentContext
 }
 
+// Lineage returns *this* context and all of its ancestor contexts in order from
+// child to parent
+func (c *Context) Lineage() []*Context {
+	lineage := []*Context{}
+	cur := c
+
+	for {
```
jszwedko: Feels like you could shrink this up by doing something like:

```
for cur := c; cur != nil; cur = cur.parentContext {
  lineage = append(lineage, cur)
}
```

context.go
```diff
@@ -445,3 +385,12 @@ func normalizeFlags(flags []Flag, set *flag.FlagSet) error {
 	}
 	return nil
 }
+
+func makeFlagNameVisitor(names *[]string) func(*flag.Flag) {
+	return func(f *flag.Flag) {
+		name := strings.Split(f.Name, ",")[0]
+		if name != "help" && name != "version" {
```
jszwedko: Can you elaborate why we don't want to append if the name is `help` or `version`?

jszwedko:
:clap: I'm super happy to see this.

Do you think you could add a couple of tests testing:

* `context.Lineage`
* That the context accessor functions properly traverse up (or maybe just test `lookupFlagSet` since that is where most of the logic is)
* That `IsSet` traverses up

:heart:

context.go
meatballhat: I dunno!  This behavior came along for the ride :smile_cat: https://github.com/codegangsta/cli/blob/3a52a3b8124e194986964c547054f0a7b8f56892/context.go#L202-L224

meatballhat: Taking these exclusions out breaks no tests, so I'm taking them out until we can find a good reason to have them :smiley_cat:

context.go
```diff
@@ -13,12 +13,11 @@ import (
 // can be used to retrieve context-specific Args and
 // parsed command-line options.
 type Context struct {
-	App            *App
-	Command        Command
-	flagSet        *flag.FlagSet
-	setFlags       map[string]bool
-	globalSetFlags map[string]bool
-	parentContext  *Context
+	App     *App
```
jszwedko: :clap: This reduced signature feels much better.

context.go
```diff
@@ -138,33 +137,19 @@ func (c *Context) LocalFlagNames() []string {
...
-}
-
 // Lineage returns *this* context and all of its ancestor contexts in order from
 // child to parent
 func (c *Context) Lineage() []*Context {
 	lineage := []*Context{}
-	cur := c
 
-	for {
+	for cur := c; cur != nil; cur = cur.parentContext {
```
jszwedko: :heart_eyes:

context_test.go
```diff
@@ -84,18 +84,22 @@ func TestContext_NArg(t *testing.T) {
...
+	parentCtx := NewContext(nil, parentSet, nil)
+	ctx := NewContext(nil, set, parentCtx)
+
+	set.Parse([]string{"--one-flag", "--two-flag", "--three-flag", "frob"})
+	parentSet.Parse([]string{"--top-flag"})
+
+	expect(t, ctx.IsSet("one-flag"), true)
+	expect(t, ctx.IsSet("two-flag"), true)
+	expect(t, ctx.IsSet("three-flag"), true)
+	expect(t, ctx.IsSet("top-flag"), true)
```
jszwedko: :+1:

jszwedko:
Beautiful.

PR #417, opened 2016-05-21
Trim go1.1 support

meatballhat:
as the upstream yaml dependency in `./altsrc` depends on features present in go1.2+.


jszwedko:
:+1: Good catch

PR #418, opened 2016-05-21
Flag re-implementation

jszwedko:
This PR is a WIP of rethinking the abstraction around flags

PR #421, opened 2016-05-22
Minimize copying while keeping things safe-ish

### LINKED ISSUE #420: Minimize struct copying
aka make func receivers more consistent.

meatballhat:
by using pointer func receivers and slices of struct pointers where possible.

Closes #420 

Extracted from #418

### DIFF
--- a/app.go
+++ b/app.go
@@ -36,8 +36,8 @@ type App struct {
 	HideHelp bool
 	// Boolean to hide built-in version flag and the VERSION section of help
 	HideVersion bool
-	// Populate on app startup, only gettable through method Categories()
-	categories *CommandCategories
+	// Categories contains the categorized commands and is populated on app startup
+	Categories CommandCategories
 	// An action to execute when the bash-completion flag is set
 	BashComplete BashCompleteFunc
 	// An action to execute before any subcommands are run, but after the context is ready
@@ -113,11 +113,11 @@ func (a *App) Setup() {
 	}
 	a.Commands = newCmds
 
-	a.categories = NewCommandCategories()
+	a.Categories = newCommandCategories()
 	for _, command := range a.Commands {
-		a.categories.AddCommand(command.Category, command)
+		a.Categories.AddCommand(command.Category, command)
 	}
-	sort.Sort(a.categories)
+	sort.Sort(a.Categories.(*commandCategories))
 
 	// append help to commands
 	if a.Command(helpCommand.Name) == nil && !a.HideHelp {
@@ -184,7 +184,7 @@ func (a *App) Run(arguments []string) (err error) {
 		defer func() {
 			if afterErr := a.After(context); afterErr != nil {
 				if err != nil {
-					err = NewMultiError(err, afterErr)
+					err = newMultiError(err, afterErr)
 				} else {
 					err = afterErr
 				}
@@ -296,7 +296,7 @@ func (a *App) RunAsSubcommand(ctx *Context) (err error) {
 			if afterErr != nil {
 				HandleExitCoder(err)
 				if err != nil {
-					err = NewMultiError(err, afterErr)
+					err = newMultiError(err, afterErr)
 				} else {
 					err = afterErr
 				}
@@ -340,17 +340,12 @@ func (a *App) Command(name string) *Command {
 	return nil
 }
 
-// Categories returns a slice containing all the categories with the commands they contain
-func (a *App) Categories() *CommandCategories {
-	return a.categories
-}
-
 // VisibleCategories returns a slice of categories and commands that are
 // Hidden=false
-func (a *App) VisibleCategories() []*CommandCategory {
-	ret := []*CommandCategory{}
-	for _, category := range a.categories.Categories() {
-		if visible := func() *CommandCategory {
+func (a *App) VisibleCategories() []CommandCategory {
+	ret := []CommandCategory{}
+	for _, category := range a.Categories.Categories() {
+		if visible := func() CommandCategory {
 			if len(category.VisibleCommands()) > 0 {
 				return category
 			}
--- a/app.go
+++ b/app.go
@@ -27,7 +27,7 @@ type App struct {
 	// Version of the program
 	Version string
 	// List of commands to execute
-	Commands []Command
+	Commands []*Command
 	// List of flags to parse
 	Flags []Flag
 	// Boolean to enable bash completion commands
@@ -37,7 +37,7 @@ type App struct {
 	// Boolean to hide built-in version flag and the VERSION section of help
 	HideVersion bool
 	// Populate on app startup, only gettable through method Categories()
-	categories CommandCategories
+	categories *CommandCategories
 	// An action to execute when the bash-completion flag is set
 	BashComplete BashCompleteFunc
 	// An action to execute before any subcommands are run, but after the context is ready
@@ -55,7 +55,7 @@ type App struct {
 	// Compilation date
 	Compiled time.Time
 	// List of all authors who contributed
-	Authors []Author
+	Authors []*Author
 	// Copyright of the binary if any
 	Copyright string
 	// Writer writer to write output to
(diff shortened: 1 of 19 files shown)


category.go
```diff
@@ -1,40 +1,48 @@
 package cli
 
 // CommandCategories is a slice of *CommandCategory.
-type CommandCategories []*CommandCategory
```
jszwedko: Was there an advantage to converting this from a opaque type to a struct?

command.go
```diff
@@ -37,7 +36,7 @@ type Command struct {
 	// Execute this function if a usage error occurs.
 	OnUsageError OnUsageErrorFunc
 	// List of child commands
-	Subcommands Commands
```
jszwedko: Was there a reason you dropped this? I kind of like the opaque type in case we want to hang any other behavior on it.

command.go
```diff
@@ -231,6 +223,22 @@ func (c Command) startApp(ctx *Context) error {
...
 	return visibleFlags(c.Flags)
 }
+
+func (c *Command) appendFlag(fl Flag) {
+	if !hasFlag(c.Flags, fl) {
+		c.Flags = append(c.Flags, fl)
+	}
+}
+
+func hasCommand(commands []*Command, command *Command) bool {
```
jszwedko: This is an example of something we could hang off of `Commands` if we chose.

context.go
```diff
@@ -147,56 +147,69 @@ func (c *Context) Lineage() []*Context {
...
 }
 
 // NArg returns the number of the command line arguments.
 func (c *Context) NArg() int {
-	return len(c.Args())
+	return c.Args().Len()
+}
+
+// Args wraps a string slice with some convenience methods
+type Args struct {
```
jszwedko: Could this just be `type Args []string`?

jszwedko:
:trumpet: :100: -- just a few questions about our use (or non-use) of opaque types.

meatballhat:
I admit I'm probably irrationally against opaque types for slices, as the
public "methods" that come along with the slice can't be gatekept or
modified.  My experience with this pattern has been painful.  In this
library alone, the StringSlice and IntSlice types were un-fun to migrate.
That being said, I'll back down if it's the right thing to do :smile_cat:

jszwedko:
I forgot about the difficulty of that transition. You make a good point regarding protecting access, but does that mean we should make the internal slices private (and expose them via `.Slice()` methods as you do once here)?

meatballhat:
I'm inconsistent that way :-P  I'm happy to make the pattern: wrap with
struct, make slice public by some sensible name that's not necessarily
"Slice".

meatballhat:
OK, so I might be taking back some of that last comment.  I like what's going on with `Args.Slice()` because a copy of the internal slice is returned.  I'm uneasy about what's going on with `CommandCategories.Categories`, tbh, and I think I'd rather have a method that returns a copy as well.  Whaddya think?

meatballhat:
@jszwedko OK, so I went ahead and tried to make the slice wrapping pattern usage a wee bit more consistent.  The two big exceptions are `App` and `Command`, although given that part of what makes this library handy is struct declaration of an app, I'm hesitant to change that.  I feel like the wrapped slices that are "protected" have different expectations because they are more of what I'd consider library internals.

category.go
```diff
@@ -1,49 +1,56 @@
...
 func (c *CommandCategories) Swap(i, j int) {
-	c.Categories[i], c.Categories[j] = c.Categories[j], c.Categories[i]
+	c.slice[i], c.slice[j] = c.slice[j], c.slice[i]
 }
 
-// AddCommand adds a command to a category.
-func (c *CommandCategories) AddCommand(category string, command *Command) *CommandCategories {
-	for _, commandCategory := range c.Categories {
+// AddCommand adds a command to a category, creating a new category if necessary.
+func (c *CommandCategories) AddCommand(category string, command *Command) {
```
jszwedko: Maybe `command` should be variadic here? Would ease appending a list of commands.

category.go
```diff
@@ -1,49 +1,56 @@
 package cli
 
-// CommandCategories is a slice of *CommandCategory.
+// CommandCategories wraps a slice of *CommandCategory.
 type CommandCategories struct {
```
jszwedko: One downside is that there isn't a good way to remove a category (admittedly this is probably an edge case for normal usage).

jszwedko: Or a command from a category.

errors.go
```diff
@@ -16,18 +16,18 @@ var ErrWriter io.Writer = os.Stderr
 
 // MultiError is an error that wraps multiple errors.
 type MultiError struct {
-	Errors []error
+	errors []error
```
jszwedko: Is there a reason we don't want to expose the list of errors anymore? I feel like that is useful to the user (e.g. to be able to iterate and format some sort of output if they choose). If we choose not to expose that member, I think it would be reasonable to make the struct itself private.

category.go
jszwedko: Doh, I missed that this mimics the existing interface. Though now that we are hiding the slice, it might make sense to expose methods for removing elements :smile:

jszwedko:
:clap: for consistency, thank you for giving that thought! I think what you have here is a reasonable approach .

category.go
jszwedko: Missed that this is the existing interface :) Ignore me.

meatballhat: :heart:

meatballhat:
Alright I switched it up again.  Latest version reintroduces opaque types, but makes them all private, adding interfaces where public surface area is (I assume) still desirable.

app.go
```diff
@@ -113,11 +113,11 @@ func (a *App) Setup() {
 	}
 	a.Commands = newCmds
 
-	a.categories = NewCommandCategories()
+	a.Categories = newCommandCategories()
```
jszwedko: 👍

args.go
```diff
@@ -6,55 +6,55 @@ var (
...
-// Args wraps a string slice with some convenience methods
-type Args struct {
-	slice []string
-}
-
-// Get returns the nth argument, or else a blank string
-func (a *Args) Get(n int) string {
-	if len(a.slice) > n {
-		return a.slice[n]
+type Args interface {
```
jszwedko: Nifty 👏

args.go
```diff
@@ -6,55 +6,55 @@ var (
...
-// First returns the first argument, or else a blank string
-func (a *Args) First() string {
+func (a *args) First() string {
 	return a.Get(0)
 }
 
-// Tail returns the rest of the arguments (not the first one)
-// or else an empty string slice
-func (a *Args) Tail() []string {
+func (a *args) Tail() []string {
```
jszwedko: Maybe this should call `Slice()` and _slice_ the slice? Would eliminate a little duplication, but honestly this is fine.

category.go
```diff
@@ -1,54 +1,80 @@
...
+		&commandCategory{name: category, commands: []*Command{command}}))
+	(*c) = newVal
 }
 
-// Categories returns a copy of the category slice
-func (c *CommandCategories) Categories() []*CommandCategory {
-	ret := make([]*CommandCategory, len(c.slice))
-	copy(ret, c.slice)
+func (c *commandCategories) Categories() []CommandCategory {
+	ret := []CommandCategory{}
```
jszwedko: Maybe another `make([]CommandCategory, len(*c))` here to save a few allocations?

errors.go
```diff
@@ -15,55 +15,65 @@ var OsExiter = os.Exit
 var ErrWriter io.Writer = os.Stderr
 
 // MultiError is an error that wraps multiple errors.
-type MultiError struct {
-	errors []error
+type MultiError interface {
+	error
+	// Errors returns a copy of the errors slice
+	Errors() []error
```
jszwedko: 👏

jszwedko:
Interfaces, genius.

category.go
meatballhat: I couldn't convince `copy` to just be cool about the type/interface
mismatch.

meatballhat: or I could just use `make` with a for loop :facepalm:

category.go
```diff
@@ -39,9 +39,9 @@ func (c *commandCategories) AddCommand(category string, command *Command) {
 }
 
 func (c *commandCategories) Categories() []CommandCategory {
-	ret := []CommandCategory{}
-	for _, cat := range *c {
-		ret = append(ret, cat)
+	ret := make([]CommandCategory, len(*c))
```
jszwedko: :clap:

jszwedko:
This looks good!

PR #431, opened 2016-05-24
Ensure flag "Name" field values are un-stringly-fied

### LINKED ISSUE #426: Unexpected "Incorrect usage" message
I have been trying to make cli work as intended, but all I get when I set my custom flags is "Incorrect usage" without any explanation as to what went wrong.

Here's the code:

```
	app.Flags = []cli.Flag {
		cli.StringFlag{
			Name: "db, d",
			Usage: "Name of the database",
		},
		cli.StringFlag{
			Name: "user, u",
			Value: "postgres",
			Usage: "Username for the database",
		},
		cli.StringFlag{
			Name: "password, p",
			Value: "",
			Usage: "Password for the database",
		},
	}
```

I have tried to set them both using the shorthand, -p, -u etc. and using the longer --password, --user etc. What am I doing wrong?

EDIT: I am also confused as to why the dash is missing from my shorthands in the help output:
```
GLOBAL OPTIONS:
   --db, d value        Name of the database
   --user, u value      Username for the database (default: "postgres")
   --password, p value  Password for the database
   --help, -h           show help
   --version, -v        print the version
```

I'm using the v2 branch.

meatballhat:
to (hopefully) help with bug triage & pinpointing usage issues since ripping out
stringly typed "Name".

- Also removed now-unnecessary whitespace trimming of env var names.

Closes #426

### DIFF
--- a/flag.go
+++ b/flag.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"regexp"
 	"runtime"
 	"strconv"
 	"strings"
@@ -14,7 +15,11 @@ import (
 
 const defaultPlaceholder = "value"
 
-var slPfx = fmt.Sprintf("sl:::%d:::", time.Now().UTC().UnixNano())
+var (
+	slPfx = fmt.Sprintf("sl:::%d:::", time.Now().UTC().UnixNano())
+
+	commaWhitespace = regexp.MustCompile("[, ]+.*")
+)
 
 // BashCompletionFlag enables bash-completion for all commands and subcommands
 var BashCompletionFlag = BoolFlag{
@@ -95,7 +100,6 @@ func (f GenericFlag) Apply(set *flag.FlagSet) {
 	val := f.Value
 	if f.EnvVars != nil {
 		for _, envVar := range f.EnvVars {
-			envVar = strings.TrimSpace(envVar)
 			if envVal := os.Getenv(envVar); envVal != "" {
 				val.Set(envVal)
 				break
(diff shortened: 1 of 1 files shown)


flag.go
```diff
@@ -610,6 +607,19 @@ func withEnvHint(envVars []string, str string) string {
 	return str + envText
 }
 
+func flagNames(f Flag) []string {
```
jszwedko: :+1:

jszwedko:
I think this sort of thing would also be addressed by a migration guide so I feel less strongly about including this code if we were to add that instead. What do you think @meatballhat ?

jszwedko:
Though I do agree that the failure mode is a bit weird in that it still _seems_ to be functioning properly in the help output so this is probably a good idea.

flag.go
```diff
@@ -610,6 +607,19 @@ func withEnvHint(envVars []string, str string) string {
...
+	ret := []string{}
+
+	name := flagStringField(f, "Name")
+	aliases := flagStringSliceField(f, "Aliases")
+
+	for _, part := range append([]string{name}, aliases...) {
```
jszwedko: What do you think about including a comment here so future spelunkers aren't confused as to what is happening here?

meatballhat:
@jszwedko Your latter comment is exactly my thinking behind this change.  I don't think this is the kind of troubleshooting casual users of this library should be expected to do:

``` bash
$ go run .t/426.go -d foo
Incorrect Usage
# ...
$ go run .t/426.go --"db, d" foo
# runs OK
```

PR #439, opened 2016-05-30
Migrator touchup and testing

meatballhat:
The intent here is to ensure that any breaking changes to the v1 API will require a migrator to be added in order for CI to pass.

- [x] add more examples to either the v1 README or a separate file to cover the v1 API (pending #440 merge)
- [x] add migrators for remaining v1 ➡️ v2 changes (pending #444)
- [x] <del>reimplement script in go with proper gofix-like AST awareness</del> (save for later)


jszwedko:
This is great, I was just thinking about how to best ease the migration path and this is exactly it.

jszwedko:
:raised_hands: :confetti_ball:

meatballhat:
:tada: :bomb:

PR #440, opened 2016-05-31
Enable more examples as runnable in README

meatballhat:
so that they can be verified by the migrator script over in the v2 branch.

In support of #439


jszwedko:
:clap:

PR #444, opened 2016-06-01
Add more v1 API examples

meatballhat:


jszwedko:
:+1: I feel like we may need a TOC at some point.

meatballhat:
Agreed re:TOC.  I like what some repos have done with a bulleted list of
features, too.  And there's also linking to alternatives that have
different strengths.  All TODO.

PR #446, opened 2016-06-07
Fix ActionFunc signature of ShowAppHelp action

jamescun:
Often in helper utilities with many subcommands, I use the `ShowAppHelp` action which does not match the new ActionFunc signature introduced in 1.15.0, generating the deprecation notice.


meatballhat:
Thank you! :heart_eyes_cat:

PR #447, opened 2016-06-08
fix panic getting visible flags

joshuarubin:
When using `altsrc` for flags, (e.g. `altsrc.NewStringFlag(cli.StringFlag{...})`) the `cli.Flag` returned is a pointer to an `altsrc` flag type.

When `app.VisibleFlags` is called (often as part of getting help/usage information), the app panics:

```
panic: reflect: call of reflect.Value.FieldByName on ptr Value [recovered]
        panic: reflect: call of reflect.Value.FieldByName on ptr Value
```

This is a simple fix to restore the intended behavior.

### DIFF
--- a/flag.go
+++ b/flag.go
@@ -512,7 +512,7 @@ func (f Float64Flag) GetName() string {
 func visibleFlags(fl []Flag) []Flag {
 	visible := []Flag{}
 	for _, flag := range fl {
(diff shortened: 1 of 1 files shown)


meatballhat:
Thank you!  I'm rolling this into #448 by backporting the `flagValue` func present in the v2 branch.

PR #448, opened 2016-06-08
Merging joshuarubin master

meatballhat:


flag.go
```diff
@@ -578,8 +578,16 @@ func withEnvHint(envVar, str string) string {
 	return str + envText
 }
 
-func stringifyFlag(f Flag) string {
+func flagValue(f Flag) reflect.Value {
 	fv := reflect.ValueOf(f)
+	for fv.Kind() == reflect.Ptr {
+		fv = reflect.Indirect(fv)
+	}
+	return fv
```
joshuarubin: no need to test for the pointer, `Indirect` returns the `Value` if it is not a pointer. You can just do:

```go
return reflect.Indirect(reflect.ValueOf(f))
```

meatballhat: Ack!  Pointers to pointers aren't covered, AFAICT, although that's getting pretty edgy.  Maybe compare to the zero value instead?  (ref https://golang.org/src/reflect/value.go#L2075)

joshuarubin: ha, didn't see that it was a `for`, presumed it was an `if`.

meatballhat: Whaddya think, then?  Leave as-is?

joshuarubin: Yeah, couldn't hurt.

PR #452, opened 2016-06-09
Focus on catching fewer panics in HandleAction

### LINKED ISSUE #451: Panic being swallowed
Within a CLI Action function, I have some code that called a library method and it had a panic: "runtime error: invalid memory address or nil pointer dereference". The Action function never reported the issue and just swallowed the panic and exited.

I finally traced it down by putting a panic call in the action function directly and got:
"ERROR unknown Action error: See https://github.com/urfave/cli/blob/master/CHANGELOG.md#deprecated-cli-app-action-signature", which finally led me to discover that CLI is just eating panics when they don't happen directly in the Action function. Eating panics is just not an acceptable way to handle errors.

I was able to work around the issue by putting the following code at the top of all my actions methods:

```
defer func() {
	if r := recover(); r != nil {
		fmt.Println("ERROR: ", r, string(debug.Stack()))
	}
}()
```

I'm not sure where of why the CLI library is doing its bad behaving `recover` yet, but I wanted to submit the bug report to help anyone else who can't trace down their error. Hope this behavior is easily corrected.

meatballhat:
so that "unknown" panics can still bubble up.

Closes #451

### DIFF
--- a/app.go
+++ b/app.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"sort"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


meatballhat:
@mattmc3 Is this a reasonable solution?

mattmc3:
I can confirm - after a quick before/after test - this fixed the issue for me. The panic makes it up the call stack without being improperly swallowed. Thank you!

PR #454, opened 2016-06-11
Adding Int64Flag type and related tests

bfreis:


meatballhat:
Thanks for the contribution!  :heart_eyes_cat: I hate to ask this after you've already done the work, but are you able to do `int64(ctx.Int("myintflag"))` instead?  I won't claim that  *nobody* needs to pass 64-bit ints via command line, but I think it's safe to assume that smaller ints cover over 80% of use cases.

bfreis:
Hi!

I definitely agree with you -- probably way over 80% of use cases would be able to use smaller ints.

I'm writing a (personal) CLI tool to interact with a 3rd party API. For certain operations, that API expects an ID which is defined as a 64-bit integer. Right now, I believe all IDs I've manually interacted with on that API all would fit in 32-bit integers, but I can't be sure that it is / always will be the case.

While in my current laptop int == int64, that might not be the case in every platform. That's why I wanted to make sure that I wouldn't have an overflow issue in the future, so I included this code to be able to parse a "strongly-typed" int64 from the command line.

Although Int64Flag is what I truly need now, I might also need to include an Int64SliceFlag later. If that's the case, and if you think this PR makes sense, I can definitely send a PR for the Int64SliceFlag if I ever write it.

By the way - this was my first contact with this CLI lib, it's great! As I have almost 0 experience with this lib, I don't really know the idioms or conventions. I noticed there's a v1 and a v2 branch... I was using master, which I think is v1, so that's where I wrote the code... hope this is good enough.

Cheers!
Bruno

bfreis:
And there you go... I thought it would take longer before I would need an Int64Slice.

flag.go
```diff
@@ -245,6 +245,77 @@ func (f IntSliceFlag) GetName() string {
...
+	if err != nil {
+		return err
+	}
+	*f = append(*f, tmp)
+	return nil
+}
+
+// String returns a readable representation of this value (for usage defaults)
+func (f *Int64Slice) String() string {
+	return fmt.Sprintf("%d", *f)
```
meatballhat: Shouldn't this be `"%v"` or maybe `"%#v"`?

context_test.go
```diff
@@ -9,17 +9,21 @@ import (
 func TestNewContext(t *testing.T) {
 	set := flag.NewFlagSet("test", 0)
 	set.Int("myflag", 12, "doc")
+	set.Int("myflagInt64", 12, "doc")
 	set.Float64("myflag64", float64(17), "doc")
 	globalSet := flag.NewFlagSet("test", 0)
 	globalSet.Int("myflag", 42, "doc")
+	globalSet.Int("myflagInt64", 42, "doc")
```
meatballhat: Can you use the `.Int64(` setter for these so that the relationship with wrapping stdlib `flag` is clearer?

context_test.go
```diff
@@ -39,6 +50,14 @@ func TestContext_GlobalInt(t *testing.T) {
 	expect(t, c.GlobalInt("nope"), 0)
 }
 
+func TestContext_GlobalInt64(t *testing.T) {
+	set := flag.NewFlagSet("test", 0)
+	set.Int64("myflagInt64", 12, "doc")
```
meatballhat: Yes, just like this :smiley:

flag.go
```diff
@@ -420,6 +491,51 @@ func (f IntFlag) GetName() string {
 	return f.Name
 }
 
+// Int64Flag is a flag that takes a 64-bit integer
+// Errors if the value provided cannot be parsed
```
meatballhat: TRIVIAL: I don't think the second line of this comment is truthful, so it should probably be removed from the other flag type comments as well (for later).

meatballhat:
After taking another look at stdlib `flag`, I'm taking back my earlier comment.  I think that as long as we're wrapping stdlib `flag`, we should *at the very least* provide access to the same flag types that it defines.  This means that we're also missing `UintFlag` and `Uint64Flag` (for later).

I dropped a few inline comments, but otherwise I think this is great.  Can you also add something to the `Added` subsection of `Unreleased` in the change log?  Thank you again!  :smile_cat:

PR #466, opened 2016-06-23
Generate code for flag types and context accessors

meatballhat:


jszwedko:
:clap: :trumpet: :100:

jszwedko:
It'd be cool to fit `Apply` in here too, but I can see how that could be complicated.

This is really cool though, I think it'll reduce errors resulting from code duplication. What do you think about documenting the format `flag-types.json` (either in `generate-flag-types`, README, or ???)?.

jszwedko:
Documentation looks great!

PR #468, opened 2016-06-29
adding support for Float64SliceFlag

bfreis:
Hi,

I just added support for Float64Slice on the "v2" branch, which is the one I'm using right now. I did "./runtests" and it seems to pass - although after it runs the tests it complains it couldn't find a "gopkg.in/yaml.v2" package... I believe this will pass the CI tests though.

Let's see.

Cheers!
Bruno

### DIFF
--- a/context.go
+++ b/context.go
@@ -84,6 +84,15 @@ func (c *Context) Int64Slice(name string) []int64 {
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


jszwedko:
Looks good, thank you @bfreis !

PR #469, opened 2016-06-29
Allow passing metadata between Before/Action/After

radeksimko:
We have a list of commands and many of them have some kind of common bootstrap logic (initiating connection to APIs etc).

We could obviously create a separate standalone function and call that function inside each `Action`, but `Action` usually contains more complex logic and hence is typically maintained in a separate file/function, whereas `Before` or `After` are (at least in most cases I've seen in the wild) defined as inline functions.

This patch would allow us to do something like this:

```go
func main() {
	app := NewApp()
	app.Commands = []Command{
		{
			Name: "foo",
			Before: func(c *Context) error {
				c.App.SetMetadata("aws", initAWS())
				return nil
			},
			Action: FooCommand,
		},
		{
			Name: "bar",
			Before: func(c *Context) error {
				c.App.SetMetadata("aws", initAWS())
				return nil
			},
			Action: BarCommand,
		},
	}
}

func FooCommand(c *Context) error {
	conn, err := c.App.GetMetadata("aws")
	if err != nil {
		return err
	}

	conn.DoSomethingFancy()

	return nil
}

func BarCommand(c *Context) error {
	conn, err := c.App.GetMetadata("aws")
	if err != nil {
		return err
	}

	conn.DoSomethingElse()

	return nil
}
```

I'm open to alternative/cleaner solutions for the problem described above.

### DIFF
--- a/app.go
+++ b/app.go
@@ -160,6 +160,10 @@ func (a *App) Setup() {
 		a.categories = a.categories.AddCommand(command.Category, command)
 	}
 	sort.Sort(a.categories)
+
+	if a.Metadata == nil {
+		a.Metadata = make(map[string]interface{})
+	}
 }
 
 // Run is the entry point to the cli app. Parses the arguments slice and routes
@@ -438,6 +442,18 @@ func (a *App) appendFlag(flag Flag) {
 	if !a.hasFlag(flag) {
 		a.Flags = append(a.Flags, flag)
 	}
+}
+
+func (a *App) SetMetadata(key string, value interface{}) {
+	a.Metadata[key] = value
+}
+
+func (a *App) GetMetadata(key string) (interface{}, error) {
+	v, ok := a.Metadata[key]
+	if !ok {
+		return nil, fmt.Errorf("Metadata key %q not found", key)
+	}
+	return v, nil
 }
 
 // Author represents someone who has contributed to a cli project.
--- a/command_test.go
+++ b/command_test.go
@@ -73,6 +73,48 @@ func TestCommand_Run_DoesNotOverwriteErrorFromBefore(t *testing.T) {
 	}
 }
 
+func TestCommand_Run_BeforeSavesMetadata(t *testing.T) {
+	var receivedMsgFromAction string
+	var receivedMsgFromAfter string
+
+	app := NewApp()
+	app.Commands = []Command{
+		{
+			Name: "bar",
+			Before: func(c *Context) error {
+				c.App.SetMetadata("msg", "hello world")
+				return nil
+			},
+			Action: func(c *Context) error {
+				msg, err := c.App.GetMetadata("msg")
+				receivedMsgFromAction = msg.(string)
+				return err
+			},
+			After: func(c *Context) error {
+				msg, err := c.App.GetMetadata("msg")
+				receivedMsgFromAfter = msg.(string)
+				return err
+			},
+		},
+	}
+
+	err := app.Run([]string{"foo", "bar"})
+	if err != nil {
+		t.Fatalf("expected no error from Run, got %s", err)
+	}
+
+	expectedMsg := "hello world"
+
+	if receivedMsgFromAction != expectedMsg {
+		t.Fatalf("expected msg from Action to match. Given: %q\nExpected: %q",
+			receivedMsgFromAction, expectedMsg)
+	}
+	if receivedMsgFromAfter != expectedMsg {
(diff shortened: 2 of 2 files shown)


jszwedko:
Hey @radeksimko,

There was a recent addition of a `Metadata` field to `App` (https://github.com/urfave/cli/pull/339) which could conceivably be used for this (you can access the `App` via a `Context`) -- would this cover your use cases?

radeksimko:
Hi @jszwedko 
Thanks for showing me the `App.Metadata`, I did not notice that field before. 😃 

It does look like something we're looking for _in theory_.
In practice it would help if there were the helper functions so we don't have to deal with the `nil` map case inside each `Before`/`Action` nor allocate the map as part of initializing `App`.

I modified the PR to include these helpers while leveraging this field. Are you happy to accept this change? 😉

jszwedko:
Hi @radeksimko,

Thank you for revising! I agree with you regarding the behavior of the `nil` map, but I'm wondering if we started initializing the map in the [`setup` func](https://github.com/urfave/cli/blob/master/app.go#L123) if that would suffice for your use cases? The primary difference being that it would push your error handling up for the "key not found" case.

Apologies for the back and forth :smile:

radeksimko:
> I'm wondering if we started initializing the map in the setup func if that would suffice for your use cases?

From practical point of view it would suffice. I just thought that the "lazy" approach would prevent unnecessary memory allocations in cases where the user doesn't need to set any metadata.

I'm happy to move it to `setup` though - can you just confirm you're ok with the above caveat?

jszwedko:
Good point, it would be another allocation, but I think it'd be more in line with the current implementation.

The whole library could probably use a look with an eye for performance. Though it'd surprise me if anyone was using it in a performance critical path.

Thank you!

radeksimko:
PR updated. I also realised there is another caveat:

This:
```go
app := NewApp()
app.SetMetadata("tada", "yada")
app.Run()
```
would most likely crash because `Run()` is the method which triggers `Setup()`. Should we add a check and error out, recommending the user to set Metadata directly via `app.Metadata = make(map[string]interface{})`?

meatballhat:
@radeksimko Since we're exposing `app.SetMetadata`, I'd be OK with having that method call `Setup()` since calling multiple times [should be safe](https://github.com/urfave/cli/blob/1efa31f08b9333f1bd4882d61f9d668a70cd902e/app.go#L123-L126).  That way we could support both `app.Metadata = ... ; app.Run()` and `app.SetMetadata(...) ; app.Run()`.  Am I getting that right? 😸

radeksimko:
@meatballhat theoretically yes, practically no 😢 

https://github.com/TimeIncOSS/cli/blob/eeec72b46ed613e7adfc0310a348bc96ec0f1d33/app.go#L130

```go
app := NewApp()
app.SetMetadata("tada", "yada")
app.Author = "random"
app.Email = "random@random.tld"
app.Run()
```
^ that would have most likely unintended consequences, unless I'm missing something obvious.

meatballhat:
@radeksimko ah, gotcha.  Yep, I think this is the price we pay in `v1` for having kinda-sorta declarative support.  Hopefully it'll be a wee bit better experience in `v2`, although there's nothing stopping someone from setting a struct member after `.Setup()` gets called there either.  Given that situation, is it acceptable to get rid of the `.(Get|Set)Metadata` methods so there's (hopefully) less confusion about who's responsibility it is to init `.Metadata`?  The following interactions feel fairly equivalent to me, but that's just me :smile_cat:

``` go
// ...
    if value, ok := app.Metadata["key"]; ok {
        doStuff(value)
    }
```

``` go
// ...
    if value, err := app.GetMetadata("key"); err != nil {
        doStuff(value)
    }
```

radeksimko:
Hm, I agree. Sometimes I wish Go supported real constructors... but then I realise the simplicity & clarity is something I love so much about it that I can forgive Go the lack of these things. 😄

app.go
```diff
@@ -160,6 +160,10 @@ func (a *App) Setup() {
 		a.categories = a.categories.AddCommand(command.Category, command)
 	}
 	sort.Sort(a.categories)
+
+	if a.Metadata == nil {
+		a.Metadata = make(map[string]interface{})
+	}
```
meatballhat: I want this! 😻

radeksimko: ah, just that? sorry, I misunderstood your comment. 😉 I cannot reopen the PR, but I can open a new one.

PR #470, opened 2016-07-02
Add JSON InputSource to altsrc package

johnweldon:
- Implement NewJSONSource* functions for returning an InputSource from
   various JSON data sources.
 - Copy and modify YAML tests for the JSON InputSource

### DIFF
--- a/README.md
+++ b/README.md
@@ -547,9 +547,9 @@ the yaml input source for any flags that are defined on that command.  As a note
 the "load" flag used would also have to be defined on the command flags in order
 for this code snipped to work.
 
-Currently only YAML files are supported but developers can add support for other
-input sources by implementing the altsrc.InputSourceContext for their given
-sources.
+Currently only YAML and JSON files are supported but developers can add support
+for other input sources by implementing the altsrc.InputSourceContext for their
+given sources.
 
 Here is a more complete sample of a command using YAML support:
 
--- a/altsrc/json_command_test.go
+++ b/altsrc/json_command_test.go
@@ -1,0 +1,324 @@
+package altsrc
+
+import (
+	"flag"
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/urfave/cli"
+)
+
+const (
+	fileName   = "current.json"
+	simpleJSON = `{"test": 15}`
+	nestedJSON = `{"top": {"test": 15}}`
+)
+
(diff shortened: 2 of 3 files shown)


johnweldon:
Is travis-ci not completing the build because the pull request is against the v2 branch?

jszwedko:
Hmm, I don't see why Travis didn't get or respond to the webhook for this PR. I pulled and pushed your changes just now so that we can follow https://travis-ci.org/urfave/cli/builds/141945098 for build status.

Thank you for adding JSON support! Reviewing now.

jszwedko:
This looks good! Did you consider just unmarhaling into a `map` to reuse the `MapInputSource` code (similar to what the YAML `altsrc` does)? Just wondering if there were technical limitations as it would DRY this up a bit.

Build is green: https://travis-ci.org/urfave/cli/builds/141945098

Thanks again!

johnweldon:
@jszwedko thanks!  Yes, I actually started with reusing the `MapInputSource`, but the `json.Unmarshal` doesn't work with `map[interface{}]interface{}`, only with `map[string]interface{}`

When I manually updated that top level to `map[interface{}]interface{}` after unmarshalling, I ran into type differences when using `MapInputSource`, I don't recall the specific errors, but for example, the json unmarshal treats all numbers as `float64`, which the `MapInputSource` couldn't handle when trying to get an `int`.

jszwedko:
Ah, interesting, I do see that behavior regarding parsing all numbers as `float64` which alone would make it difficult to use `MapImportSource`. Thank you for clarifying!

Will merge when the builds finish.

anarcat:
i'm a little confused - this *looks* like it was merged, but the v2 README doesn't mention JSON support at all... where did that patch go?

jszwedko:
Hey @anarcat, this was merged into the `v2` branch, which is intended to be the precursor to a breaking `2.0` release, but I see that there really isn't anything breaking here, so I'll cherry pick to master as a separate release.

You are correct that it still needs to be added to the README though :smile:

jszwedko:
Opened https://github.com/urfave/cli/pull/598 to backport.

anarcat:
well, the reason i'm confused is that the change to the README *is* in this PR (see 91563b8)

jszwedko:
@anarcat indeed, I realized that when I went to port the feature to master. If you get a chance, would appreciate feedback on the PR (specifically if it works for your use cases!).

PR #474, opened 2016-07-07
Print completion code 

nlewo:
A proposal to print completion code from the cli.

- User has to manualy supply the program path (as it is already done with shell variable PROG). I tryied to discover the program name/path, but it seems to be hard to handle all corner cases. I then prefer to let the user know what he is doing.

- I didn't test zsh generated code (I don't use zsh). Somenone could please try it? Otherwise, I could remove the commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -787,10 +787,30 @@ func main() {
 
 #### Enabling
 
-Source the `autocomplete/bash_autocomplete` file in your `.bashrc` file while
-setting the `PROG` variable to the name of your program:
+You can enable bash or zsh completion by using the flag `--bash`,
+respectively `--zsh`. These flags take the name of your program as
+parameter.
 
-`PROG=myprogram source /.../cli/autocomplete/bash_autocomplete`
+To setup for bash (if your program is in your PATH):
+
+```
+eval "`myprogram --bash myprogram`"
+```
+
+If your program is in the current directory:
+
+```
+eval "`./myprogram --bash ./myprogram`"
+```
+
+Alternatively, you can put the completion code in your `.bashrc` file:
+```
+myprogram --bash myprogram >> ~/.bashrc
+```
+
+Note: you have to specify the path of your program because you can define
+your own completion functions. The shell has then to know where your
+program is in order to generate possible completions.
 
 #### Distribution
 
--- a/app.go
+++ b/app.go
@@ -150,6 +150,7 @@ func (a *App) Setup() {
 	if a.EnableBashCompletion {
 		a.appendFlag(GenerateCompletionFlag)
 		a.appendFlag(BashCompletionFlag)
+		a.appendFlag(ZshCompletionFlag)
 	}
 
 	if !a.HideVersion {
@@ -185,6 +186,10 @@ func (a *App) Run(arguments []string) (err error) {
 	}
 
 	if checkBashCompletion(context) {
+		return nil
+	}
+
+	if checkZshCompletion(context) {
 		return nil
 	}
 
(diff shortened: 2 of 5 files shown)


nlewo:
@meatballhat could you help me on the README.md CI problem?

meatballhat:
@nlewo I know there's a lot of output in there, but I'm seeing this error near the end of the logs indicating an issue with one of the README.md examples:

```
time="2016-07-07T07:03:00Z" level=debug msg="captured output" source=README.md stderr="# command-line-arguments\n.\\example.go:21: cannot use cli.BoolFlag literal (type cli.BoolFlag) as type cli.StringFlag in assignment\n" stdout=
```

I assume this is due to switching the type of the bash completion flag 😸

meatballhat:
@nlewo Please forgive my naïveté, but what are the corner cases you ran into for determining progname?  I would have done the following and assumed all is well 😸 https://play.golang.org/p/lxCxRU6ttT

* I think that switching the type of `cli.BashCompletionFlag` is going to be more trouble than it's worth.  If we can get away with supporting `filepath.Base(os.Args[0])` on Linux, OSX, and Windows, that's good enough for me, and it would mean that we can stick with a boolean `cli.BashCompletionFlag`.

* I'm uncomfortable with the flag names `--bash` and `--zsh`, as they don't describe their actions (to me), and also feel too general, thereby being more likely to overlap with existing flags used by consumers of `cli`.  I would prefer `--init-completion [shell-name]` where `shell-name` defaults to `bash`.  I feel that this keeps the naming a bit more in line with the existing `--(enable|generate)-bash-completion` flags (even though I would like to see the latter changed to `--(enable|generate)-completion [shell-name]` given we support more than `bash`).

* In the example code in the README, I would prefer they use `$()` subshells instead of backticks.

* To your point about the zsh completion being untested, it'd be cool to get some integration tests in place for completion in general, but that definitely doesn't need to be a part of your PR 😸 🎺

nlewo:
@meatballhat 

- A corner case is if the prog is not in the `PATH`. To launch it, we then have to do `$ myprogram`. In this case the progname is not sufficient because the completion function will not be able to call the program. 
But if we use the whole `os.Args[0]` string, it should do the job.

- Just to be sure. We would then have two flags regarding completion. One is `generate-bash-completion` which is hidden. The other one would be `init-completion value` where `value` could be `bash` or `zsh` with `bash` as default value.

meatballhat:
@nlewo I'm fine with not covering the corner case where the executable isn't in `PATH`, as it can be managed (fairly painlessly?) by anyone who really needs that particular combination.  The inconvenience imposed on that set is far outweighed by the convenience of the complementary set not having to include the executable basename as a positional arg. &lt;/opinions&gt;

To your second point: Yep, we'd have two flags if we keep `generate-bash-completion`.  I think the design I personally want is:

- `--init-completion {shell}` - to generate completion code
- `--generate-completion {shell}` - to be invoked from within completion code

In order to keep `v1` backward compatible, we'll have to leave the `generate-bash-completion` flag in place and handle both that and `--generate-completion {shell}`, but then `v2` can drop `--generate-bash-completion` along with any other bash-specific flags, fields, or funcs.

What do you think?  Am I making any sense? 😸

nlewo:
@meatballhat I completely agree with your `v2` design. 

Personnaly, I'd prefer to only implement this feature in the `v2` branch in order to avoid backward compatibility. Are you fine with this?

Concerning the executable name problem, I think we can cover all use cases if we consider `os.Args[0]` instead of `filepath.Base(os.Args[0])`.

meatballhat:
@nlewo I'm in favor of implementing this on the `v2` branch, and sticking with `os.Args[0]` sounds good to me, too :smile_cat:

nlewo:
New PR #490 has been created for the v2 branch. This one can be closed.

PR #479, opened 2016-07-10
Drop testing of go1.1.2

meatballhat:
since it's already a fairly diverged set of tests, and the fact that it passes appears to be causing issues with overall CI build status, which is probably some odd issue that surfaces because we're using `matrix.include`


jszwedko:
:heart:

Issue #235, opened 2015-06-09
Support for TOML in altsrc


aacanakin:
What I currently need is to be able to use this library by using a defined configuration. I have a revel.Config or TOML configuration. I need to be able to inject this config structure into context or app struct. Is this possible currently ? If not, is there any plan to integrate a configurations module with this package ?

wrouesnel:
+1 to this. Environment variable support is nice, but it'd be perfect if there was an official way to set default configuration from a file ala iniflags.

jszwedko:
I agree that this would be useful.

freddy33:
+1 repeating flags on each call is better solved with conf file in user home than with env vars.

zpatrick:
+1

jszwedko:
This is supported by https://github.com/codegangsta/cli/pull/306 for YAML -- changing the title to reflect `TOML` support, but it should be possible just by implementing the `altsrc.InputSourceContext` interface.

jszwedko:
Support added in #491

PR #491, opened 2016-07-23
Add TOML config file loader

ykanda:
I tried to make the loader of TOML files in issue #235
It uses the https://github.com/BurntSushi/toml to implement
The same test as the YAML loader has path

### DIFF
--- a/altsrc/toml_command_test.go
+++ b/altsrc/toml_command_test.go
@@ -1,0 +1,310 @@
+// Disabling building of tom support in cases where golang is 1.0 or 1.1
+// as the encoding library is not implemented or supported.
+
+// +build go1.2
+
+package altsrc
+
+import (
+	"flag"
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/urfave/cli"
+)
+
+func TestCommandTomFileTest(t *testing.T) {
+	app := cli.NewApp()
+	set := flag.NewFlagSet("test", 0)
+	ioutil.WriteFile("current.toml", []byte("test = 15"), 0666)
+	defer os.Remove("current.toml")
+	test := []string{"test-cmd", "--load", "current.toml"}
+	set.Parse(test)
+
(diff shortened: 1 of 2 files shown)


altsrc/toml_command_test.go
```diff
@@ -0,0 +1,310 @@
+// Disabling building of tom support in cases where golang is 1.0 or 1.1
```
meatballhat: This should be "toml support", yes?

meatballhat:
LGTM :+1: and I'd like someone more familiar with the guts of ./altsrc to review, too :smile_cat:

meatballhat:
@roboll @ChrisPRobinson :wave: 😺

altsrc/toml_file_loader.go
```diff
@@ -0,0 +1,113 @@
...
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/BurntSushi/toml"
+	"github.com/urfave/cli"
+)
+
+type TomlMap struct {
```
ChrisPRobinson: Is it required to have this type be public? It seems that its just used internally and its not needed to be public, but not sure if the underlying deserializer requires a public container type

ChrisPRobinson:
I would update Consider updating front page markup to indicate that now it supports TOML too

ChrisPRobinson:
👍 Looks good.

altsrc/toml_command_test.go
ykanda: Yes, I miss it.
I Fix the typo.

altsrc/toml_file_loader.go
ykanda: It's exactly as you say.
Fixed the name of the type to tomlMap.

jszwedko:
Added notes to README and CHANGELOG in https://github.com/urfave/cli/commit/b616f6088660d2eaa33739718f0583f8d467a178

Thanks @ykanda and @ChrisPRobinson !

PR #495, opened 2016-07-24
Ensure context lookups traverse lineage

### LINKED ISSUE #487: Parent context flag lookup broken
Hi there, 

I might have noticed a regression: If you declare a flag to the application level, it is no longer accessible on a command level. Here is a minimal example:

```go
package main

import (
	"fmt"
	"os"

	"gopkg.in/urfave/cli.v2"
)

func main() {

	app := &cli.App{}

	app.Flags = []cli.Flag{
		&cli.StringFlag{
			Name:    "f",
			Usage:   "This flag is transmitted to the default Action but not to subcommands",
		},
	}

	app.Action = Print
	
		{
			Name: "fail",
			Action: Print,
		},
	}

	app.Run(os.Args)

}

func Print(ctx *cli.Context) error {
	fmt.Println(ctx.String("f"))
	return nil
}
```

The following command prints the flags value, as the app.Action function is called.
```
$ go run main.go -f ok
$ ok
```

The following command prints nothing, as a app.Command function is called.
```
$ go run main.go -f ok fail
$
```

In the past, there was a ctx.GlobalString() method that was very useful to get global flags. The function has been removed but the ctx.String() method could access local and global flags. It can no longer.

I believe a workaround is to declare the global flags as global vars, and put them in the command. However, it's not a pretty workaround.

meatballhat:
Closes #487

### DIFF
--- a/context_test.go
+++ b/context_test.go
@@ -35,57 +35,89 @@ func TestNewContext(t *testing.T) {
 func TestContext_Int(t *testing.T) {
 	set := flag.NewFlagSet("test", 0)
 	set.Int("myflag", 12, "doc")
-	c := NewContext(nil, set, nil)
+	parentSet := flag.NewFlagSet("test", 0)
+	parentSet.Int("top-flag", 13, "doc")
+	parentCtx := NewContext(nil, parentSet, nil)
+	c := NewContext(nil, set, parentCtx)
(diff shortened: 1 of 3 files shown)


meatballhat:
@moxar see any problems with this? 😸

jszwedko:
:trumpet:

PR #496, opened 2016-07-24
Write err to stderr, exit 1 if err != ""

### LINKED ISSUE #475: Simplify cli.NewExitError
Hi,

I need to update a CLI tool because we have some deprecation warnings in the output.
While I thought I could simple return errors from commands, I end up with things like:

```go
 f, err := os.Create("config.yml")
 if err != nil {
     return cli.NewExitError(err.Error(), 1)
 }
 defer f.Close()
 err:= project.Configure(project.Slug, os.Stdin, f)
 if err != nil {
     return cli.NewExitError(err.Error(), 1)
 }
 [...]
 return nil
```

It would be so much easier to simply return `err` in case of error, and cli would display the error, and exit with code 1. Sounds like a very reasonable default, doesn't it?

meatballhat:
Closes #475

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -13,6 +13,19 @@ import (
 	"testing"
 )
 
+var (
+	lastExitCode = 0
+	fakeOsExiter = func(rc int) {
+		lastExitCode = rc
+	}
+	fakeErrWriter = &bytes.Buffer{}
+)
+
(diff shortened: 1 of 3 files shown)


jszwedko:
:heart:

PR #489, opened 2016-07-20
Improve zsh autocompletions

jereksel:
For now it replaces bash autocompletion (I'm waiting for #474 to be merged). Preview:
Before: https://asciinema.org/a/3wuz5g3qkf6o63cpmw1tt4gnw
After: https://asciinema.org/a/2xcdywm9qmbkg1ek2kubym0w9

PR #490, opened 2016-07-22
Generate completion code

nlewo:

PR #498, opened 2016-07-24
Merging #489 (plus hack)

meatballhat:
On top of the changes proposed in #489, I've added a hacky hack for zsh completion that I don't especially like.  I'm happy to alter this patch given a better alternative.  At the very least, I want to ensure such hackery isn't necessary in the `v2` branch.


jszwedko:
:+1: I see what you mean -- it looks like we should be able to drop the environment variable hack in #490

PR #499, opened 2016-07-25
Use non-TLS gocov badges

meatballhat:


jszwedko:
It looks like the issue is that gocover.io is serving an invalid certificate so we could also fix them by just using `http://` -- unless there was another reason you wanted to remove the image?

meatballhat:
I originally switched to `http://` but that wasn't enough to appease the github camo caching servers 😐

meatballhat:
I also tried some cache busting query params, but no luck there either.

jszwedko:
Hrm, looking at https://github.com/github/markup/issues/224 it looks like it should respect cache headers returned by the request. In the current case (`https`) it should, and appears to, fail to fetch the resource.

I'd advocate for reverting to the `http` and, if it doesn't clear up in a day or two, reach out to their support.

meatballhat:
le ugh 😛

jszwedko:
:cry: 

Lemme reach out to their support.

jszwedko:
Also opened an issue in gocover.io regarding the cert being served: https://github.com/vieux/gocover.io/issues/22

meatballhat:
Whoops, no longer needed!

jszwedko:
:tada:

Issue #294, opened 2015-11-12
Context.IsSet not working with values set from environment variables


varun06:
I used IsSet() to check for some values. It works the string flag is passed from command line but not when the flag values are set with environment values.

jszwedko:
Indeed, this appears to be true, thanks for raising the issue @varun06 !

AaronO:
@jszwedko I just encountered this bug too. Any update on this ?

jszwedko:
@AaronO unfortunately I haven't had the time to address this yet, but PRs are welcome!

bweston92:
Also just come across this, I'm a Go newb so my PR wouldn't look good.

meatballhat:
tl;dr - this is not straightforward, and a proper solution may require reimplementing a lot stuff

My first adventures with this have me hitting some unpleasantness.  A travelogue, all prefixed with a strong *AFAICT* :

`cli.Context.IsSet` accepts a string, which is currently used to look up a matching stdlib `*flag.Flag` via `flag.FlagSet.Visit`, which only calls the given func on flags that have been set via stdlib flag parsing (not env vars).  There is not currently a mapping of stdlib `*flag.Flag` to `cli.Flag` available to a given context, so getting at the right env vars for a given flag when all we have is *one of* the names it uses isn't straightforward.  The guts of `altsrc/flag.go` include a method for checking via both `cli.Context.IsSet` *and* directly looking at the env via `os.Getenv`, but that's because at that level it's accessing a full `cli.Flag`.

I started down the path of building a name ➡️  `cli.Flag` mapping inside `*cli.Context`, but there's issues there as well.  For example, the `cli.Context.Command` is not set during `cli.NewContext`, but assigned during `cli.Command.Run`, which means that we'd have to update the index at assignment time, namespacing the keys of whatever index to distinguish between flags from `cli.Context.Command` and `cli.Context.App`.  Then there's the issues of avoiding index staleness, handling potential `Command` or `App` reassignment (they're public, after all,) and any bugs lurking in the fact that throughout the library is a mix of struct pointers and struct copies.

varun06:
Thanks for looking into it.

jszwedko:
This is fixed by #502 (some additional, related, work is being done in #530 to address #517).

henrikhodne:
This still/again appears to be broken, `c.IsSet("foo")` returns false, but `c.String("foo")` returns a value with the following flag:

```
&cli.StringFlag{
    Name: "foo",
[...]

PR #502, opened 2016-07-31
Fix context.(Global)IsSet to respect environment variables

jszwedko:
This appeared to be the least messy approach to hack in support for
IsSet also checking environment variables to see if a particular
cli.Flag was set without making backwards incompatible changes to the
interface.

I intend to fix this more properly in v2, probably by adding another
method to the cli.Flag interface to push the responsibility down as it
occurred to me that it was really the `Flag`s themselves that offer
support for configuration via the environment as opposed to the
`context` or other supporting structures. This opens the door for the
anything implementing the `Flag` interface to have additional sources of
input while still supporting `context.IsSet`.

### DIFF
--- a/context.go
+++ b/context.go
@@ -3,6 +3,8 @@ package cli
 import (
 	"errors"
 	"flag"
+	"os"
+	"reflect"
 	"strings"
 )
 
@@ -11,12 +13,11 @@ import (
 // can be used to retrieve context-specific Args and
 // parsed command-line options.
 type Context struct {
-	App            *App
-	Command        Command
-	flagSet        *flag.FlagSet
-	setFlags       map[string]bool
-	globalSetFlags map[string]bool
-	parentContext  *Context
+	App           *App
+	Command       Command
+	flagSet       *flag.FlagSet
+	setFlags      map[string]bool
+	parentContext *Context
 }
 
 // NewContext creates a new context. For use in when invoking an App or Command action.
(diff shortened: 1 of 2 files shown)


jszwedko:
Fixes #294

context.go
```diff
@@ -43,28 +44,70 @@ func (c *Context) GlobalSet(name, value string) error {
...
+				flags = c.App.Flags
+			}
+		}
+		for _, f := range flags {
+			eachName(f.GetName(), func(name string) {
+				if isSet, ok := c.setFlags[name]; isSet || !ok {
+					return
+				}
+
+				envVars := reflect.ValueOf(f).FieldByName("EnvVar").String()
```
meatballhat: Given that this is a context level operation in v1, I think the lookup of the EnvVar field should be more defensive since it's not a part of the Flag interface.

meatballhat:
LGTM other than that one lil note :shipit:

context.go
jszwedko: During my testing, I discovered that you just end up with `""` in the case that `EnvVar` doesn't exist on the struct.

meatballhat: Hot diggety!  Ignore me then :smile_cat:

jszwedko: Actually, scratch that, you end up with `<invalid Value>`. I must have made a mistake before, will update.

jszwedko:
Haha, too quick for me, will update on master for the `EnvVar` missing case.

jszwedko:
Fixed in 168c954

PR #508, opened 2016-08-21
remove the possiblity of end-user's seeing deprecation warnings

### LINKED ISSUE #507: API change breaks github.com/codegangsta/cli
DEPRECATED Action signature.  Must be `cli.ActionFunc`.  This is an error in the application.

Is it really an application error or did someone make a breaking change to the API without changing the URL? Is this considered an OK thing to do now? 

I am using the import 'github.com/codegangsta/cli'

jszwedko:
Instead use deprecation pattern described in
https://blog.golang.org/godoc-documenting-go-code.

This will reduce the chances that the application developer will notice that
the deprecations, but I can see the argument that @alienscience is making in
 #507 that the current approach is actually an API breakage rather than
a deprecation since the warnings are printed at runtume. More practically, this
has a chance to create a poor user experience for users of applications that
use `cli` if the application developer doesn't notice the warnings.

Fixes #507

### DIFF
--- a/app.go
+++ b/app.go
@@ -62,10 +62,11 @@ type App struct {
 	// An action to execute after any subcommands are run, but after the subcommand has finished
 	// It is run even if Action() panics
 	After AfterFunc
+
 	// The action to execute when no subcommands are specified
+	// Expects a `cli.ActionFunc` but will accept the *deprecated* signature of `func(*cli.Context) {}`
+	// *Note*: support for the deprecated `Action` signature will be removed in a future version
 	Action interface{}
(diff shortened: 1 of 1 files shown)


jszwedko:
If we like this, I can backport it the released versions since 1.15.0.

alienscience:
Thanks, that works much better for me now.

I had problems updating with both Gin (the web server) and CLI last weekend. However, despite two grumpy bug reports both projects have responded politely and fast. My apologies for the tone of my bug report and many thanks for providing CLI.

meatballhat:
🎷 🐻

jszwedko:
No worries, thank you too, @alienscience

Issue #473, opened 2016-07-06
Import path in altsrc


nobu-k:
Because altsrc package imports urfave/cli as "github.com/urfave/cli", it doesn't work with cli imported from gopkg.in like "gopkg.in/urfave/cli.v1". For example,

```
// cli.go
package main

import (
	"gopkg.in/urfave/cli.v1"
	"gopkg.in/urfave/cli.v1/altsrc"
)

func main() {
	altsrc.ApplyInputSourceValues(&cli.Context{}, &altsrc.MapInputSource{}, []cli.Flag{})
}

$ go run cli.go
./cli.go:9: cannot use "gopkg.in/urfave/cli.v1".Context literal (type *"gopkg.in/urfave/cli.v1".Context) as type *"github.com/urfave/cli".Context in argument to altsrc.ApplyInputSourceValues
./cli.go:9: cannot use []"gopkg.in/urfave/cli.v1".Flag literal (type []"gopkg.in/urfave/cli.v1".Flag) as type []"github.com/urfave/cli".Flag in argument to altsrc.ApplyInputSourceValues
```

compiling this code fails due to type mismatch. The reason is that cli is imported from "gopkg.in/urfave/cli.v1" but altsrc expects to receive types in "github.com/urfave/cli". I think altsrc should depend on gopkg.in version of urfave/cli. It seems impractical to support both "github.com/urfave/cli" and "gopkg.in/urfave/cli.vx".

meatballhat:
I agree that supporting both is impractical.  I would like to see the `altsrc` imports move to using the `gopkg.in` import for the respective branch/release.

meatballhat:
Alternatively, could we use `import cli ".."` (untested)?

nobu-k:
No, relative import paths don't work for general Go programs and libraries: https://golang.org/cmd/go/#hdr-Relative_import_paths

meatballhat:
@nobu-k :heart_eyes_cat:

PR #511, opened 2016-08-27
Fix import paths in altsrc

grubernaut:
Uses gopkg.in as the import path for the `altsrc` package.

Fixes: #473


jszwedko:
This seems reasonable to me, what do you think @meatballhat ?

jszwedko:
Just saw your note on #512 :smile: 

Merging both

Issue #504, opened 2016-08-05
Set default value text


Alexendoo:
If I want to have a port flag that is random if not specified, it would be helpful to indicate that by having the default value display as `(default: random)` rather than 0 for example. Usage would be along the lines of

```go
cli.IntFlag{
    Name:        "port, p",
    DefaultText: "random",
},
```

An option to not display the default text would also work

jszwedko:
Hi @Alexendoo,

This seems like a reasonable addition.

You could work around this deficiency for now by using a `cli.GenericFlag` where the `Value` used for has a `String()` method that returns `"random"`. Though you'd then also need to handle the int parsing yourself.

Thanks for the suggestion!

jszwedko:
Fixed by #514 -- thanks again for suggesting this @Alexendoo

PR #514, opened 2016-08-28
Add DefaultValue text for flags

grubernaut:
Allows a user to override the default value of a flag in the displayed help output.

Ex:
```
cli.IntFlag{
	Name: "foo, f",
	DefaultText: "random foo",
	Value: "bar",
}
```

Running `(app.name) -h` will now yield:

```
--foo value     (default: "random foo")
```

Fixes: #504

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ applications in an expressive way.
     + [Alternate Names](#alternate-names)
     + [Values from the Environment](#values-from-the-environment)
     + [Values from alternate input sources (YAML, TOML, and others)](#values-from-alternate-input-sources-yaml-toml-and-others)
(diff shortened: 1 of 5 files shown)


jszwedko:
This looks good to me, thank you for the contribution @grubernaut! Do you mind adding some notes to the README regarding this new feature's usage?

@Alexendoo can you verify that this implementation satisfies what you were looking for?

Alexendoo:
Well it's _very_ minor but dropping the automatic quotes would be good `(default: random)`, but yeah that looks good to me. Thanks @grubernaut

grubernaut:
@jszwedko sure thing! I'll finalize this up by EOD. 

@Alexendoo Ah that would actually be preferred then, if a default text value was not a string. Will fix.

Issue #427, opened 2016-05-23
Allow global flags after command


fgblomqvist:
This is just a suggestion, but to me it seems logical to allow global flags after the command as well, not only before it.

For example, if you have a verbose flag, -v:
`myapp -v server`
vs.
`myapp server -v`

To me the second one does visually make more sense. Are there cases when this is a bad idea (and therefore the reason to why this is not how it is currently implemented)?

jszwedko:
Hi @fgblomqvist, I agree with you completely. Unfortunately this is baggage that we carry from using the Go stdlib `flag` library (which stops parsing the moment it hits a non-flag) so we cannot support it unless we were to rip that out and do our own flag parsing (which we have discussed a bit, but is a non-trivial effort). You can see some more conversation about this in #355.

I'll leave this open as a future enhancement though!

fgblomqvist:
Okay, I see.

noffle:
The core `flag` package is BSD licensed, so perhaps the easiest "fix" would be to include a copy of said package, modified to support intermixed arguments/flags. This has the benefit of keeping the API identical, making the integration simpler than a rewrite or using a different 3rd-party package. Does this sound reasonable?

jszwedko:
@noffle mmm, that is not a bad idea given the tradeoffs. I'll try to give that a shot in the near future.

Thank you for the suggestion!

0xdeafcafe:
Also interested in this.

nikkolasg:
Bump up for this one too ! 
But in the meantime, @jszwedko : Thank you so far for the good job you've done ;)

marwan-at-work:
wondering if this issue is still up for a fix :)

lynncyrin:
Given that this is from last year, I think I'm comfortable closing it 🙂 feel free to re-open / open a new issue / comment in support if there's still interest here!

lynncyrin:
Oh!!! I just noticed there's 22 👍s. Re-opening now 📈

lynncyrin:
This may have been fixed by v2? I'd appreciate if someone could investigate that 🙏 

🔗 => https://github.com/urfave/cli/releases/tag/v2.0.0

tanghaibao:
I loved cli library before but this issue of not able to mix arguments and flags is ridiculous. 

[...]

PR #515, opened 2016-08-30
Treat all flags which followd behind the last arguments as normal arg…

keloyang:
If SkipFlagParsing is not considered,cli treat all words which start by '-' as flag, so when the subcommand has a argument which start by '-', it also will treated as cli' flag, not a argument for subcommand.
e.g.
```
root@ubuntu:~/# runc exec test ls -la    
Incorrect Usage.
```
I think cli should treat all flags which followd behind the last arguments as normal arguments,just like this:
```
root@ubuntu:~/# runc exec test ls -la            
total 44
drwxr-xr-x   12 root     root          4096 Aug 30 06:37 .
drwxr-xr-x   12 root     root          4096 Aug 30 06:37 ..
-rwxr-xr-x    1 root     root             0 Aug 30 02:23 .dockerenv
-rwxr-xr-x    1 root     root             0 Aug 30 02:23 .dockerinit
...
```
https://github.com/opencontainers/runc/issues/752 for details.
Signed-off-by: Shukui Yang <yangshukui@huawei.com>

### DIFF
--- a/command.go
+++ b/command.go
@@ -46,6 +46,8 @@ type Command struct {
 	Flags []Flag
 	// Treat all flags as normal arguments if true
 	SkipFlagParsing bool
+	// Treat the flags as normal arguments if true
+	SkipLastFlagParsing bool
 	// Boolean to hide built-in help command
 	HideHelp bool
 	// Boolean to hide this command from help or completion
@@ -90,6 +92,7 @@ func (c Command) Run(ctx *Context) (err error) {
 	set.SetOutput(ioutil.Discard)
 
 	if !c.SkipFlagParsing {
+		firstFlagAdjust := -1
 		firstFlagIndex := -1
 		terminatorIndex := -1
 		for index, arg := range ctx.Args() {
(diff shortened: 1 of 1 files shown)


keloyang:
ping @jszwedko

gravis:
LGTM
If no one else is taking this, I'll merge the PR in 10 days.

jszwedko:
The usual "unix-y" way to handle this (which this library supports thanks to stdlib `flag` support) is to use `--` to specify that all following arguments should be parsed as arguments rather than flags.

I can see the argument you are making for `... exec`-like commands though. Would the support for `SkipFlagParsing` support your use case? Or would the `exec` subcommand also have flags of its own in addition to the freeform command?

keloyang:
SkipFlagParsing don't work for me,
e.g.
```
runc exec --cwd /  test ls -la 
```
it will skip -la, but it also skip --cwd

jszwedko:
@keloyang gotcha, so the subcommand also has flags.

The behavior you describe is actually the default behavior of the underlying `flag` library that we use, but we do some argument rearranging (that I removed in the unreleased `v2` branch in https://github.com/urfave/cli/pull/398 because it only worked under very specific circumstances and supporting complete reordering would require much more effort).

I can imagine a future where `cli` supports flags and arguments being intermixed (a need expressed in https://github.com/urfave/cli/issues/427), but because this is not currently supported and wouldn't be in this particular version (nor probably in `v2`), I'm OK with including this. Given it will be the default behavior in `v2`, I'm almost inclined to call the option `NoArgumentReorder`, document that it is backported behavior, and make it closer to the behavior in `v2`.

Let me know what you think about this and I can backport it.

Thanks for bringing this up! (and apologies for stepping on your toes @gravis :smile: -- this issue just reminded me of the work I'd done in #427 around removing reordering).

keloyang:
@jszwedko thank for your reply, It's nice to  backport  https://github.com/urfave/cli/pull/398 and it can resolve the problem above.

meatballhat:
@gravis you've still got this? :smile_cat:

jszwedko:
I was planning on working on the backport of #398 today, but let me know if anyone is actively working on this!

PR #528, opened 2016-09-11
Backport no argument reordering as an option

jszwedko:
In the unreleased version 2, the argument reordering has been removed
(in f585ec7cb8b0d9189ac9016fe83f911f3a360f2e) since it only worked if
all of the arguments appeared before all of the flags, but not if they
were intermixed which was of limited utility and caused some confusion.
This commit allows enabling of this future behavior via SkipArgReorder.

Ideally we'd support complete intermingling of flags and arguments, but
this is unlikely to happen until we switch flag parsers.

Fixes #515

### DIFF
--- a/command.go
+++ b/command.go
@@ -46,6 +46,11 @@ type Command struct {
 	Flags []Flag
 	// Treat all flags as normal arguments if true
 	SkipFlagParsing bool
+	// Skip argument reordering which attempts to move flags before arguments,
+	// but only works if all flags appear after all arguments. This behavior was
+	// removed n version 2 since it only works under specific conditions so we
+	// backport here by exposing it as an option for compatibility.
+	SkipArgReorder bool
 	// Boolean to hide built-in help command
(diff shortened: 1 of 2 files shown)


jszwedko:
Let me know what you think of this @meatballhat @keloyang @gravis

meatballhat:
Lovely! 🎷 🐻 

:shipit:

keloyang:
Nice , thank you!

command.go
```diff
@@ -46,6 +46,11 @@ type Command struct {
 	Flags []Flag
 	// Treat all flags as normal arguments if true
 	SkipFlagParsing bool
+	// Skip argument reordering which attempts to move flags before arguments,
+	// but only works if all flags appear after all arguments. This behavior was
+	// removed n version 2 since it only works under specific conditions so we
```
gravis: there's a typo here :(
`%s/removed n version 2/removed in version 2/`

PR #530, opened 2016-09-12
Consider empty environment variables as set

### LINKED ISSUE #517: c.IsSet returns true for empty env vars
I want to be able to differentiate between empty env vars and unset env vars, but they are treated the same right now:

```
package main

import (
    "fmt"
    "os"

    "github.com/urfave/cli"
)

func main() {
    app := cli.NewApp()
    app.Name = "boom"
    app.Usage = "make an explosive entrance"
    app.Flags = []cli.Flag{
        cli.StringFlag{
            Name:   "thing",
            EnvVar: "THING",
        },
    }
        fmt.Printf("isset: %t\n", c.IsSet("thing"))
        return nil
    }
    os.Setenv("THING", "")
    //os.Setenv("THING", "set")

    app.Run(os.Args)
}
```

This prints

```
isset: false
```

If you uncomment the commented out line it prints:

```
isset: true
```

I need both cases to be treated as set. Right now I have some ugly workarounds to do this.

jszwedko:
When assigning values to flags (also when interogatting via
`context.(Global)IsSet`.

Using `syscall.Getenv` rather than `os.LookupEnv` in order to support older Golang versions.

For boolean flags, consider empty as `false`.

One important consideration here is that `context.(Global)IsSet()` will return
`true` for flags whose specified environment variable has been set even when
the value of that variable fails to be parsed as the type that the flag
requires; however, I think this behavior is better than the previous behavior
where `IsSet` would return false even if the variable truly was set by the
environment.

In order to properly support fix considering environment variables with invalid
values (which are currently silently ignored) is to augment the `Flag`
interface's `Apply()` method to be able to return an error (which I'd suggest
we do in `v2` since we cannot do this here without breaking compatibilty).

Fixes #517

### DIFF
--- a/altsrc/flag.go
+++ b/altsrc/flag.go
@@ -2,9 +2,9 @@ package altsrc
 
 import (
 	"fmt"
-	"os"
 	"strconv"
 	"strings"
+	"syscall"
 
 	"gopkg.in/urfave/cli.v1"
 )
@@ -237,13 +237,11 @@ func (f *Float64Flag) ApplyInputSourceValue(context *cli.Context, isc InputSourc
 func isEnvVarSet(envVars string) bool {
 	for _, envVar := range strings.Split(envVars, ",") {
 		envVar = strings.TrimSpace(envVar)
-		if envVal := os.Getenv(envVar); envVal != "" {
+		if _, ok := syscall.Getenv(envVar); ok {
 			// TODO: Can't use this for bools as
 			// set means that it was true or false based on
 			// Bool flag type, should work for other types
-			if len(envVal) > 0 {
-				return true
-			}
+			return true
 		}
 	}
 
--- a/app.go
+++ b/app.go
@@ -174,7 +174,11 @@ func (a *App) Run(arguments []string) (err error) {
 	a.Setup()
 
 	// parse flags
-	set := flagSet(a.Name, a.Flags)
+	set, err := flagSet(a.Name, a.Flags)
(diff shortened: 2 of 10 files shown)


jszwedko:
I have a couple of ideas for poor hacks to make `IsSet` consistent with the `Apply` implementations wherein it returns `false` if the env var was set, but was not parsable (while maintaining backwards compatibility).

I'd be interested in feedback on the current changes though (particularly around `bool` handling).

/cc @meatballhat @vikstrous

meatballhat:
LGTM :metal:

vikstrous:
This sounds reasonable. I agree that values that failed to parse shouldn't be considered set. If any variables fail to parse the program shouldn't even get to execute IMO. The behaviour you describe is indeed an improvement and works for my use case.

jszwedko:
After discussing briefly with @meatballhat, we decided that it is closer to the original intent of parsing environment variables to error if they are set to unparsable values (rather than silently ignoring). I pushed a second commit, https://github.com/urfave/cli/pull/530/commits/b816af8abccd27a018d816aeb7964af7c2fa33a7, which does this and, by consequence, `(Global)IsSet` will not return `true` if the environment variables failed to be parsed.

@meatballhat do you mind reviewing the implementation here? It's a bit hacky right now to maintain compatibility, but I think we can slim it down in `v2` by modifying the existing `Flag` interface rather than having to have both and reflect.

PR #531, opened 2016-09-12
bug fix when SkipFlagParsing is false and SkipArgReorder is true

keloyang:
Signed-off-by: Shukui Yang <yangshukui@huawei.com>


keloyang:
I have this pr to fix the if case, could you give a review ? thank you!
 ping @jszwedko

jszwedko:
Oops, thank you for fixing this!

PR #535, opened 2016-09-17
Hide version if set to empty string

### LINKED ISSUE #524: Version behavior is inconsistent
In https://github.com/urfave/cli/issues/248 there was a change made so if a user empties the app.Version then the version is not printed as part of the help output.

This has left the -v, --version flags though which then creates the somewhat strange output of "app version" with no version number.

I would really prefer that the Version string is not initialized at all in cli.NewApp and then if the app.Version string is empty set `HideVersion = true`

nthnca:
When the version string is empty, automatically set the HideVersion setting to
true. This fixes #524.

### DIFF
--- a/app.go
+++ b/app.go
@@ -152,6 +152,10 @@ func (a *App) Setup() {
 		a.appendFlag(BashCompletionFlag)
 	}
 
+	if a.Version == "" {
+		a.HideVersion = true
+	}
+
 	if !a.HideVersion {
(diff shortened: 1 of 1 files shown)


jszwedko:
@nthnca awesome, thank you! Do you mind adding a test? This is good otherwise.

meatballhat:
@nthnca Ping! Test, please?

PR #536, opened 2016-09-20
Pass environment around in the context

zimbatm:
Hi,

what would you think of injecting the environment variables as part of the run? This work is not complete but I wanted some feedback before I put the finishing touches.

The main downside is that it loses compatibility with the flag.FlagSet.Parse() interface. I don't think it's a big problem since cli implements all the of the flag-provided types already. The might also be a slight performance hit because of the conversion but it's probably negligible.

On the plus side this will makes it easier to do dependency injection during testing or to do some advanced environment-based logic. I think os.Environ() was available quite early in the go releases so this also solves the problem of not being able to use `os.LookupEnv()` to check the existence of an environment variable (which I've seen mentioned in another issue).


jszwedko:
Hi @zimbatm,

Apologies for the delay!

I think this is pretty interesting and opens up some advanced use cases for the library. I imagine this feature would also be lightly used, but as long as it isn't too invasive -- which I believe your approach to not be -- then I'm OK with adding it.

If we are going to break the compatibility with `flag.Flag`s `Apply()` method, it might be better, from a compatibility standpoint, to inject a `cli.FlagApp/yContext` or similar instead of `cli.Env`. This way we can add additional fields to the context in the future without having to modify the signature. I'm curious what your thoughts are on that.

Either way, I'm happy to see this added if you want to continue iterating!

lynncyrin:

PR #540, opened 2016-09-30
Call HandleExitCoder for all members of MultiError.Errors

hirose31:
[The document of HandleExitCoder](https://godoc.org/github.com/urfave/cli#HandleExitCoder) say:

> If the given error is a MultiError, then this func is called on all members of the Errors slice.

but HandleExitCoder is called only once for the first element of Errors slice.

Because HandleExitCoder which is called in HandleExitCoder will `os.Exit`.

``` go
    if multiErr, ok := err.(MultiError); ok {
        for _, merr := range multiErr.Errors {
            HandleExitCoder(merr) // will exit
        }
        return
    }
```

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -74,7 +74,7 @@ func (ee *ExitError) ExitCode() int {
 // HandleExitCoder checks if the error fulfills the ExitCoder interface, and if
 // so prints the error to stderr (if it is non-empty) and calls OsExiter with the
 // given exit code.  If the given error is a MultiError, then this func is
-// called on all members of the Errors slice.
+// called on all members of the Errors slice and calls OsExiter with the last exit code.
 func HandleExitCoder(err error) {
 	if err == nil {
 		return
--- a/errors.go
+++ b/errors.go
@@ -86,8 +86,9 @@ func HandleExitCoder(err error) {
 
 	if multiErr, ok := err.(MultiError); ok {
 		for _, merr := range multiErr.Errors {
-			HandleExitCoder(merr)
+			fmt.Fprintln(ErrWriter, merr)
 		}
+		OsExiter(1)
 		return
 	}
 
--- a/errors.go
+++ b/errors.go
@@ -85,10 +85,8 @@ func HandleExitCoder(err error) {
 	}
 
 	if multiErr, ok := err.(MultiError); ok {
-		for _, merr := range multiErr.Errors {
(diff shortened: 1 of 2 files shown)


jszwedko:
Hi @hirose31,

I believe I see your intention, but this implementation wouldn't handle the case where the `MultiError` itself consists of `ExitCoder`s with differing status codes besides `1`. I can see the case for printing all of the error messages before exiting, but I think we'll want need to make a decision about which status code to exit with (I'm inclined to exit with the code of the first matching `ExitCoder` and `1` if none in the list are `MultiError`s).

What do you think?

hirose31:
> I'm inclined to exit with the code of the first matching ExitCoder and 1 if none in the list are MultiErrors

I agree 👍

errors.go
```diff
@@ -96,3 +95,18 @@ func HandleExitCoder(err error) {
...
+
+func handleMultiError(multiErr MultiError) int {
+	code := 1
+	for _, merr := range multiErr.Errors {
+		if multiErr2, ok := merr.(MultiError); ok {
+			code = handleMultiError(multiErr2)
+		} else {
+			fmt.Fprintln(ErrWriter, merr)
+			if exitErr, ok := merr.(ExitCoder); ok {
+				code = exitErr.ExitCode()
```
jszwedko: I believe this would result in the last exit code being used. I think this is fine as long as we document the behavior.

hirose31:
@jszwedko 
I've just added document on exit code and test cases. Could you please review commits?

errors_test.go
```diff
@@ -46,8 +50,10 @@ func TestHandleExitCoder_MultiErrorWithExitCoder(t *testing.T) {
 	called := false
 
 	OsExiter = func(rc int) {
-		exitCode = rc
-		called = true
+		if !called {
```
jszwedko: Could you add another `ExitCoder` to this test case? I want to assert that it uses the last one (to avoid inadvertently changing that behavior).

jszwedko:
@hirose31 thank you! Just one more note, but otherwise this is looking good.

errors_test.go
hirose31: Added adec15acf537c7b696c299223a16c60cb3eeeed8

jszwedko:
Perfect, thank you @hirose31

PR #542, opened 2016-10-07
Display error instead of just say command is incorrect

mhmxs:


jszwedko:
:+1: thanks @mhmxs! This seems reasonable to me. Do you mind making the analogous change in `app.go` and adding a couple of tests that assert this behavior?

app.go
```diff
@@ -194,7 +194,7 @@ func (a *App) Run(arguments []string) (err error) {
 			HandleExitCoder(err)
 			return err
 		}
-		fmt.Fprintf(a.Writer, "%s\n\n", "Incorrect Usage.")
+		fmt.Fprintf(a.Writer, "%s %s\n\n", "Incorrect Usage.", err.Error())
```
mhmxs: Covered by test

app.go
```diff
@@ -315,7 +315,7 @@ func (a *App) RunAsSubcommand(ctx *Context) (err error) {
...
```
mhmxs: New test added

command.go
```diff
@@ -138,7 +138,7 @@ func (c Command) Run(ctx *Context) (err error) {
 			HandleExitCoder(err)
 			return err
 		}
-		fmt.Fprintln(ctx.App.Writer, "Incorrect Usage.")
+		fmt.Fprintln(ctx.App.Writer, "Incorrect Usage:", err.Error())
```
mhmxs: Covered by test

app.go
mhmxs: New test added

mhmxs:
@jszwedko plz review

jszwedko:
Looks great, thank you for adding the tests!

PR #543, opened 2016-10-12
app: Add App.Description

wking:
So you can describe what the application is for without requiring users to drill down into a particular command.

If this idea sounds usefull, I can add a test-case to this PR.

### DIFF
--- a/app.go
+++ b/app.go
@@ -42,6 +42,8 @@ type App struct {
 	ArgsUsage string
 	// Version of the program
 	Version string
+	// Description of the program
+	Description string
 	// List of commands to execute
 	Commands []Command
 	// List of flags to parse
--- a/app.go
+++ b/app.go
@@ -458,10 +458,10 @@ type Author struct {
 func (a Author) String() string {
 	e := ""
 	if a.Email != "" {
-		e = "<" + a.Email + "> "
+		e = " <" + a.Email + ">"
 	}
 
-	return fmt.Sprintf("%v %v", a.Name, e)
+	return fmt.Sprintf("%v%v", a.Name, e)
 }
 
 // HandleAction uses ✧✧✧reflection✧✧✧ to figure out if the given Action is an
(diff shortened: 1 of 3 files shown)


jszwedko:
Hi @wking,

This was the intention of the `Usage` field -- do you have an example use case that might better show the distinction you are making here?

Thank you!

wking:
On Tue, Oct 18, 2016 at 08:28:11PM -0700, Jesse Szwedko wrote:
> This was the intention of the `Usage` field -- do you have an
> example use case that might better show the distinction you are
> making here?

CommandHelpTemplate (vs. AppHelpTemplate) has separate USAGE and
DESCRIPTION sections.  I'm just proposing App allow the same
distinction.  And man(1) has [1]:

  Conventional section names include NAME, SYNOPSIS, CONFIGURATION,
  DESCRIPTION, OPTIONS, EXIT STATUS, RETURN VALUE, ERRORS,
  ENVIRONMENT, FILES, VERSIONS, CONFORMING TO, NOTES, BUGS, EXAMPLE,
  AUTHORS, and SEE ALSO.

Help generated by this package uses “USAGE” instead of “SYNOPSIS”, but
clearly distinguishing between invocation templates and plain text
discriptions is a fairly standard thing.

One benefit of the distinction is that you will be able to put plain
text in App.Description while still relying on the auto-generated
USAGE entries which would have been clobbered had you set
App.UsageText.

[1]: http://man7.org/linux/man-pages/man1/man.1.html#DESCRIPTION

jszwedko:
@wking aha, I see! I missed that `Command` already allows for this. Do you mind adding a test for this? Otherwise I think it looks great.

wking:
Test added (along with some whitespace fixups in `AppHelpTemplate` and `Author.String`) with c4a46a7 → 3c2bce5.

jszwedko:
This looks great, thank you for adding the tests @wking

PR #544, opened 2016-10-13
flags: sort flags

runcom:
Just went ahead and implemented this and it's working over https://github.com/kubernetes-incubator/cri-o - not sure there are other code paths I missed. I'm open to have a boolean to decide if flags needs to be sorted but I assumed everyone wants flags sorted.

Sorry if I made any mistake, first time contributing to this repository.

Signed-off-by: Antonio Murdaca <runcom@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ applications in an expressive way.
   * [Flags](#flags)
     + [Placeholder Values](#placeholder-values)
     + [Alternate Names](#alternate-names)
+    + [Ordering](#ordering)
     + [Values from the Environment](#values-from-the-environment)
     + [Values from alternate input sources (YAML, TOML, and others)](#values-from-alternate-input-sources-yaml-toml-and-others)
   * [Subcommands](#subcommands)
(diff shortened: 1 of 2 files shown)


jszwedko:
Hi @runcom,

Thank you for starting this discussion!

I think users often declare the flags in the order they'd like them to be displayed to the user (often they'll list them from most commonly used to least commonly used), but I can definitely see an argument for sorting flags by name so I would open to adding what you have [here](https://github.com/runcom/ocid-1/blob/cf677a20f66c9ae06a4957948d3b03df58b676fc/cmd/server/main.go#L62-L72) to the library and documenting how you can sort them from outside of the library, reducing the code the developer needs to write to something like what you have:

```
sort.Sort(cli.FlagsByName(app.Flags))
sort.Sort(cli.FlagsByName(configCommand.Flags))
```

This would also allow us to keep compatibility.

What are your opinions?

runcom:
> sort.Sort(cli.FlagsByName(app.Flags))
> sort.Sort(cli.FlagsByName(configCommand.Flags))
> This would also allow us to keep compatibility.
> 
> What are your opinions?

I'll be fine with that as well :)

runcom:
@jszwedko I've updated this PR, PTAL

jszwedko:
This is great, thank you for the contribution!

PR #559, opened 2016-11-04
make shell autocomplete more robust

joshuarubin:
Lets say you have an app with a flag "a" of type `cli.StringFlag`.

Then this is the command typed:

```bash
$ app -a <tab>
```

The completion system calls this command: `app -a --generate-bash-completion`

Since the "a" flag expects an argument, the string "--generate-bash-completion" is used as its value and the flagset, when parsed, shows that the completion flag was not enabled.

This patch changes the handling of the `BashCompletionFlag`.

`BashCompletionFlag.Name` should only ever exist at the end of the `arguments` list. With this knowledge, we can check for its existence outside the flagset, store its value in `Context` and strip it from the `arguments` seen by the rest of the app.

Some additional reordering of the logic in `Command.Run` (to match that in `App.Run`) was required to prevent invalid usage errors from preempting the completion of commands.

### DIFF
--- a/app.go
+++ b/app.go
@@ -145,10 +145,6 @@ func (a *App) Setup() {
 		}
 	}
 
-	if a.EnableBashCompletion {
-		a.appendFlag(BashCompletionFlag)
-	}
-
 	if !a.HideVersion {
 		a.appendFlag(VersionFlag)
 	}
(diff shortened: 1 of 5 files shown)


joshuarubin:
Anything?

jszwedko:
Hi @joshuarubin,

Thank for raising this and for the implementation -- you've definitely identified what I would consider a bug with the current completion code.

Overall the implementation seems sound, but what do you think about:

* Using `shellComplete` rather than `complete` in the various places (to make it clearer what that field / those variables indicate)
* Adding a couple of tests for this behavior

Apologies for the delay in review!

joshuarubin:
OK, I've updated the naming to be more clear. I've also added a test for this functionality.

joshuarubin:
Any chance of getting this merged soon?

drekar:
Good comments @jszwedko now let's merge this baby.

jszwedko:
:+1: apologies for the delay, I'll have more time this weekend to dig into this, but at a cursory glance this looks good! Do you mind fixing the merge conflict and failing tests?

Thanks again!

joshuarubin:
Should be ready to merge now.

jszwedko:
:+1: :+1: Thank you for addressing the conflict!

PR #560, opened 2016-11-05
Close #558: detect FormattedError and print their stack trace

### LINKED ISSUE #558: better intergration with pkg/errors
Hi,

would it be possible to improve a little bit https://github.com/urfave/cli/blob/master/errors.go to have better integration of pkg/errors, thus be able to display a stack trace of the error ?

In my understanding it looks like it needs a type to feed an error rather a string
 `func NewExitError(message string, exitCode int) *ExitError {`

And the `HandleExitCoder` needs a small update to detect formatted errors (https://github.com/pkg/errors/blob/master/errors.go#L126), to apply a `fmt.Printf("%+v", err)` rather than `fmt.Fprintln(ErrWriter, err)`

Or i m doing something wrong when i do `return cli.NewExitError(err.Error(), 1)` ?

mh-cbon:
I took the path of the least effort. Let me know!

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -34,6 +34,10 @@ func (m MultiError) Error() string {
 	return strings.Join(errs, "\n")
 }
 
+type ErrorFormatter interface {
+	Format(s fmt.State, verb rune)
+}
+
 // ExitCoder is the interface checked by `App` and `Command` for a custom exit
 // code
 type ExitCoder interface {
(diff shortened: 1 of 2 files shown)


errors.go
```diff
@@ -77,7 +81,9 @@ func HandleExitCoder(err error) {
 	}
 
 	if exitErr, ok := err.(ExitCoder); ok {
-		if err.Error() != "" {
+		if _, ok := exitErr.(ErrorFormatter); ok {
```
jszwedko: I think you would also want to do this below at line 94/100, no?

mh-cbon: Im not so clear why there is this test on empty err message. Does people sends empty errors ? weird :x

jszwedko: That is a bit weird, I'm not sure why that check is there. Let me see if I can track down when it was added.

jszwedko: Ah, I see, it's just to avoid printing an empty line if there is no error message. Probably an edge case, but I could see someone setting an empty `ExitCoder` just to get the exit behavior without outputting a message.

jszwedko:
Changes, look great, thank you!

PR #564, opened 2016-11-14
Make ApplyWithError a public method on errorableFlag

jszwedko:
Add to altsrc flags. Otherwise, flagSet() was bypassing altsrc's attempt
at shadowing.


jszwedko:
Build will fail until there is a new release of `cli` with these changes given that `altsrc` imports the latest published version -- verified that the tests passed locally. Releasing momentarily.

PR #566, opened 2016-11-17
allow to load YAML configuration files on Windows

fernandezvara:
The function `loadDataFrom` does not take care of Windows users since any of the conditions met and it returns an error.

The change looks for the runtime where it's running and then if the filePath contains a `\`.

### DIFF
--- a/altsrc/yaml_file_loader.go
+++ b/altsrc/yaml_file_loader.go
@@ -11,6 +11,8 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"runtime"
+	"strings"
 
 	"gopkg.in/urfave/cli.v1"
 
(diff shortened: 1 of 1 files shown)


altsrc/yaml_file_loader.go
```diff
@@ -78,6 +80,11 @@ func loadDataFrom(filePath string) ([]byte, error) {
 			return nil, fmt.Errorf("Cannot read from file: '%s' because it does not exist.", filePath)
 		}
 		return ioutil.ReadFile(filePath)
+	} else if runtime.GOOS == "windows" && strings.Contains(u.String(), "\\") {
```
jszwedko: Hmm, don't you want this above the previous clause? Otherwise it'll fall into the `u.Path !+ ""` condition.

fernandezvara: At least on the machines I could try, `u.Path` is empty on Windows, so the previous if does not apply and falls onto the error on the last condition. It gets detected if you prefix with file:// but that seems not friendly.

jszwedko: Aha, interesting, do you mind leaving a comment to that effect? That isn't immediately obvious looking at this code.

fernandezvara: For sure. That behaviour is strange, and of course, that `if` without a comment does not help understanding.

jszwedko:
Thanks again!

PR #568, opened 2016-11-21
Fix for #556 broke the api for users who were using the ActionFunc

kklipsch:


jszwedko:
Oh jeeze, thank you for fixing that and providing a test case. That was a pretty glaring miss on my part (assuming that the assertion would match the signature).

PR #575, opened 2016-12-18
command: enable ordering commands by name

MaximilianMeister:


MaximilianMeister:
> AFAICT this implementation is incomplete. 

Might be that I missed something, but the code seems to be enough to do:

```go
sort.Sort(cli.CommandsByName(app.Commands))
```

and the commands in the output are sorted. what else would need to be implemented here?

> Are you planning to add any more code for integrating CommansdByName with App or Command?

You mean like always sorting by default, or exposing some option flag to sort the commands? here I just wanted to enable the usage with `sort` like in https://github.com/urfave/cli/commit/c516bce8f12bf57771f9be57baa7c895e2039477

jszwedko:
:+1: my only request might be to add an example to the README similar to https://github.com/urfave/cli#ordering (maybe just augmenting that example)

@meatballhat for more context, there was some discussion around this when we added similar functionality for flags (https://github.com/urfave/cli/pull/544). My thought was to make it easier for people who want sorted flags, but not require, or even default, to it.

MaximilianMeister:
@jszwedko thanks. added some example to the README

MaximilianMeister:
@jszwedko any idea how i can fix the AppVeyor test? it complains about the README change somehow

jszwedko:
Interesting, what immediately jumps out is the possible need to add the commands to the help output of the example (here: https://github.com/urfave/cli/pull/575/files#diff-04c6e90faac2675aa89e2176d2eec7d8R464), but I would have expected the Linux builds to fail as well if that were the case. I should have time this weekend to spin up a Windows VM to poke.

Thanks for making the changes @MaximilianMeister !

MaximilianMeister:
@jszwedko @meatballhat btw `appveyor` test is green now...

jszwedko:
Very weird, but I'll take it. Thanks again @MaximilianMeister !

MaximilianMeister:
@jszwedko thank you too for the pointer to the inline comment! ;) that indeed needed to be updated.

btw is it possible to backport this commit to `v1`? i could open a separate PR with a cherry-pick to the `v1` branch, or do you have another backporting strategy?

jszwedko:
For the `v1` branch, we typically merge down on an adhoc basis, but could definitely be better about that. Will do so now.

PR #583, opened 2017-01-11
Invalidate context.setFlags cache on modification.

karalabe:
The `Context.IsSet` method caches the set of flags "set" to speed up subsequent queries. Unfortunately if a program calls `IsSet`, but afterwards modifies the flagset via `Set` or `GlobalSet`, the cache will not be invalidated, causing subsequent `IsSet` and `GlobalIsSet` methods to return `false`, even though they have been just set.

I've updated both `TestContext_Set` and `TestContext_GlobalSet` to also test for the "set"-edness of the flag before and after setting it, both of which will produce test failures without the cache invalidation introduced in this PR.

### DIFF
--- a/context.go
+++ b/context.go
@@ -39,11 +39,13 @@ func (c *Context) NumFlags() int {
 
 // Set sets a context flag to a value.
 func (c *Context) Set(name, value string) error {
+	c.setFlags = nil
 	return c.flagSet.Set(name, value)
 }
 
 // GlobalSet sets a context flag to a value on the global flagset
 func (c *Context) GlobalSet(name, value string) error {
+	globalContext(c).setFlags = nil
 	return globalContext(c).flagSet.Set(name, value)
 }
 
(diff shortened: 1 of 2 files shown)


jszwedko:
Hi @karalabe,

Definitely a bug. Thank you for catching this and for submitting this patch! What do you think about explicitly setting `c.setFlags[name] = true` rather than invalidating the full cache?

karalabe:
That would need a further complexity of generating the cache in `Set` too if it was not yet generated, since until the first `IsSet`, the cache is nil if I understand the code correctly.

karalabe:
I can separate out the cache generation into a function and call in both `Set` and `IsSet` if you prefer that approach though :) Just ping we which version you'd like.

jszwedko:
@karalabe ah yes, you are correct. I'm not overly concerned about performance in this library, but if it isn't too much added complexity, I think this would be an easy place to avoid the less performant implementation.

PR #587, opened 2017-01-19
Pass the ErrWriter on the root app to subcommands

guywithnose:


jszwedko:
Doh, yep, thanks for fixing this and adding a test!

Issue #594, opened 2017-01-31
App shouldn't exit if error doesn't implement ExitCoder


awsmsrc:
## Current behavior:

Even if the error returned doesn't implement ExitCoder the app exits with an exit status of 1

- introduced here: https://github.com/urfave/cli/commit/e9688813
- further info here: https://github.com/urfave/cli/issues/565

## Expected behavior 

the app will return the error but not exit so it can be introspected as follows

``` 
err := app.Run()
if err != nil {
  if err.IsInterestingToMe() {
    reportInterestingData()
  } else {
    exit(1)
  }
}
```

## Usecase
if the users token has expired, prompt for a email/password, generate a new token and re run the command

## Alternative option 
Allow for a custom HandleExitCoder to be specified

awsmsrc:
Happy to help with whatever implementation is decided here, if it can help get things moving :)

jszwedko:
@awsmsrc :+1: thanks for filing this -- it reminded me to take a deeper dive into #565 and I came to the same conclusion you did. #595 adjusts this behavior to only exit for `ExitCoder`s.

kklipsch:
So #595 has the effect of ignoring errors coming out of the action function unless they meet the ExitCoder interface? That seems weird and wrong.  Why not change the action function to explicitly require an ExitCoder?

jszwedko:
@kklipsch it doesn't ignore the errors, rather it propagates them up to the `App.Run()` method so that the caller can decide what to do. You can use `App.RunAndExitOnError` as a convenience to exit with a code of `1` if desired.

Closing this since #595 has been merged in.

PR #595, opened 2017-02-04
Do not exit if commands return non-ExitCoder errors

### LINKED ISSUE #565: How to Run without exiting on error?
Currently it appears that if a command returns an error, `Run` will exit the program, since it calls HandleExitCoder, which always exits if the error is not nil. Is it possible to receive that error without exiting? For instance, I've implemented an interactive console that uses this library - the console should not exit just because a command returned an error

jszwedko:
Doing so limits the ability for users to have only some of their errors
cause the application to terminate while allowing others to bubble up.

This was originally an adjustment to #361 in #496 to fix #495, but, in
hindsight, I believe that the better approach is to recommend the use of
`RunAndExitOnError` for this use case (which is undeprecated here).

Fixes #565 #594

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,2 @@
+[flake8]
+max-line-length = 120


jszwedko:
@gravis this is a revert of the behavior added in #475 (see the reasoning above) so, after this is merged and you update, you'll want to switch to `RunAndExitOnError` to preserve the behavior. Let me know if you have any thoughts though!

gravis:
I think this will make the behaviour of _many_ programs uncertain. I'd recommend to add a new method instead of undeprecating an old one + change the existing `Run`. This will confuse so many users...

awsmsrc:
Will this be reaching master soon? Anything I can do to help it a long?

jszwedko:
@gravis I concede that changing the behavior back will confuse some users, but I think there may be a number of users that were (and will be when they upgrade to the latest tag of `cli`) confused by the introduction of the behavior in #475 (which, again in hindsight, I would have qualified as a breaking change in the behavior of the library). There isn't a great path forward, but I _think_ putting the behavior back will result in the least overall confusion to users of this library.

jszwedko:
Pinging @meatballhat to get his thoughts

meatballhat:
This is a tough one alright.  I think that we're in a bit of a bind.  Maybe a silly question, but can we leave master as is and target v2 with the reverted/new behavior?

awsmsrc:
@jszwedko @meatballhat the change in #475 was breaking for us in production, the advantage to reverting is that the current state is a removal of functionality which we require where as reverting the change (though a breaking change again) may cause some users to update there applications but at least both paths are available to the developer

vincentbernat:
I was also quite confused by the change. The current behaviour is a bug as this is not working as documented. This makes error handling quite inconsistent currently.

jszwedko:
@vincentbernat thank you for the nudge, after further reflection on the matter, I agree that it was a regression in behavior that will cause more confusion by being left in than by being reverted. I was trying to reconcile with master and appeared to have closed this PR, but will get it into master now.

Issue #549, opened 2016-10-20
Slice Flags & EnvVars attributes


jrnt30:
First, thanks for a great project.  Newer to go and really like the well thought out library here and the time that went into it!  It certainly is helpful to get going with. 

---

I see there is a lot of discussion in other threads about the precedence of Default Values, Command Line flags and Environment Variables when it comes to the *SliceFlag attributes. I was hoping to get clarification around my expectations and will take a stab at modifying the behavior if there is consensus around how it should work or will look into updating the documentation to make it explicit how these differ from other types of attributes. 
## Attribute Precedence

**Expected**
- Env Variables would override Default Values 
- Command Line flags would override both Default Values & Command Line Flags

**[Reality](https://github.com/urfave/cli/compare/v2...jrnt30:env-attribute-precedence?expand=1)**
- Env Variables and Command line flags are _appended_ together (defaults are properly ignored but Tests do fail)
- This differs from the relationship of other non-slice flags such as String (test added as well for validation)

| Defaults Provided | Env Variable Configured | Env Variable Present | Command Line Flags | Expected | Actual |
| --- | --- | --- | --- | --- | --- |
| Yes | Yes | Yes | No | Env Variable | Env Variable (Split) |
| Yes | No | No | Yes | Command Line Flags | Command Line Flags |
| No | Yes | Yes | Yes | Command Line Flags | **Command Line Flags AND Env Variable** |
| Yes | Yes | Yes | Yes | Command Line Flags | **Command Line Flags AND Env Variable** |
## Attribute Parsing

**Expected**
- Env Variable & Command Line flags would be parsed the same 

**Reality** 
- Env var files are split on comma
- Command line flags are treated as a single value 

NOTE: This slicing on `,` for env vars vs. command line flags may be intentional, just slightly different than what I expected when looking at the unit tests

jszwedko:
Hi @jrnt30,

Thank you for this well written issue!

The behavior with environment variables splitting on `,` is to facilitate setting multiple values as an environment variable. E.g. `LANGS=english,french,spanish` will get converted to `[]string{"english", "french", "spanish"}`. However, I agree that that behavior differing from what happens when `--langs english,french,spanish` is provided is confusing. To that end, I'm wondering if we should either:

* split the argument values on ","s too (seems like moderately surprising behavior)
[...]

PR #605, opened 2017-03-07
Handle slice flag values like slice env values

vincentbernat:
Command-line flags and environment variables are parsed differently for
slices. Command-line flags need one value per flag (`--serve 10 --serve 20`)
while environment variables need to use a comma-separated
value (`X_SERVER=10,20`).

This commit makes command-line flags work like environment
variables (`--serve 10,20`). Previous behavior is still
possible (`--serve 10 --serve 20`).

This change can be surprising for users expecting to be able to use
comma for a value.

Related to #549.

### DIFF
--- a/flag.go
+++ b/flag.go
@@ -136,7 +136,10 @@ type StringSlice []string
 
 // Set appends the string value to the list of values
 func (f *StringSlice) Set(value string) error {
-	*f = append(*f, value)
+	for _, s := range strings.Split(value, ",") {
+		s = strings.TrimSpace(s)
+		*f = append(*f, s)
+	}
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


flag.go
```diff
@@ -136,7 +136,10 @@ type StringSlice []string
 
 // Set appends the string value to the list of values
 func (f *StringSlice) Set(value string) error {
-	*f = append(*f, value)
+	for _, s := range strings.Split(value, ",") {
```
jszwedko: What do you think about making this delimiter optional and specifiable on the `StringSlice` itself? This would allow us to maintain backwards compatibility.

jszwedko:
Thanks @vincentbernat ! I really like this idea. Just had one note about whether we could make this behavior optional (and being able to control the delimiter).

flag.go
vincentbernat: No strong opinion. If the delimiter was configurable, this wouldn't solve the fact that environment variables and flags are treated differently.

jszwedko: Good point, the choice of `,` here is consistent with how environment variables are treated. I'm just concerned with unexpectedly breaking compatibility with use cases that depend on allowing `,`s to be specified in their values.

In hindsight, it feels like the delimiter for environment variables could also be configurable.

jgillich:
What's the status of this one? I was surprised that comma-separating values was not supported, definitely would like to see this merged.

If we don't want to break existing programs, maybe only apply to v2? Since that branch is considered unstable as per readme.

dgrisham:
Want to throw out a +1 for this, came across this issue hoping for the same feature.

henrywallace:
What are the unresolved questions/changes for getting this merged?

lynncyrin:

@henrywallace if you're interested in this, can you add a comment to https://github.com/urfave/cli/issues/549? I like for the issue (rather than the PR) to be the canonical reference for "people are interested in this"

Issue #613, opened 2017-03-31
Read flag value from file (Kubernetes, Docker Swarm)


bradrydzewski:
I would like to propose new functionality (which I'm willing to implement) to load individual arguments from file:

```diff
type StringFlag struct {
	Name        string
	Usage       string
	EnvVar      string
+	FilePath    string
	Hidden      bool
	Value       string
	Destination *string
}
```

When declaring flags one could specify an optional path for loading the value:

```diff
cli.StringFlag{
	Name: "mysql-password",
	EnvVar: "MYSQL_PASSWORD",
+	File: "/run/secrets/mysql_password",
}
```

This is functionally similar to loading a flag value from environment variable. If you follow the 12-factor pattern loading values from environment variable can be very helpful. If you are using systems like Docker Swarm and Kubernetes you may desire to load parameters from the contents of mounted files [1][2]

An example use case would be starting my server with a mysql password. When using Kubernetes or Docker swarm I might choose to load this parameter value from the filesystem using the built-in secret stores.

So for local development I might provide the mysql password using a flag:

```
$ ./app --mysql-password=<value>
```

For a 12 factor deployment I might provide the mysql password using an environment variable:

```
$ MYSQL_PASSWORD=<value> ./app
```

And for a Kubernetes or Swarm deployment I might source the mysql password using the built-in secret capabilities of the platform, from the file system.

```
$ docker secret create mysql_password <value>
$ docker run --secret=mysql_password ...
```

The implementation would be pretty straightforward:

```
if envVal, ok := flagFromFileEnv(f.FilePath, f.EnvVar); ok {
	envValDuration, err := time.ParseDuration(envVal)
	if err != nil {
		return fmt.Errorf("could not parse %s as duration for flag %s: %s", envVal, f.Name, err)
	}

	f.Value = envValDuration
}
```

[1] https://docs.docker.com/engine/swarm/secrets/#simple-example-get-started-with-secrets
[2] https://kubernetes.io/docs/user-guide/secrets/#using-secrets-as-files-from-a-pod

jmalloc:
As a user who is interested in this functionality for the same reasons (Docker specifically), I'd like to suggest allowing the filename to be specified via a second environment variable, so as to avoid hard-coding the details of Docker's `/run/secrets` filesystem layout into the binary.

```diff
cli.StringFlag{
	Name: "mysql-password",
	EnvVar: "MYSQL_PASSWORD",
[...]

PR #614, opened 2017-03-31
Ability to load variable from file

bradrydzewski:
This demonstrates a potential solution for https://github.com/urfave/cli/issues/613. If there is interest in possibly adding this feature we can flush out the implementation

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ applications in an expressive way.
     + [Alternate Names](#alternate-names)
     + [Ordering](#ordering)
     + [Values from the Environment](#values-from-the-environment)
+    + [Values from files](#values-from-files)
     + [Values from alternate input sources (YAML, TOML, and others)](#values-from-alternate-input-sources-yaml-toml-and-others)
   * [Subcommands](#subcommands)
   * [Subcommands categories](#subcommands-categories)
@@ -579,6 +580,38 @@ func main() {
       Value: "english",
       Usage: "language for the greeting",
       EnvVar: "LEGACY_COMPAT_LANG,APP_LANG,LANG",
+    },
+  }
+
+  app.Run(os.Args)
+}
+```
+
+#### Values from Files
+
+You can also have the default value set from file via `FilePath`.  e.g.
+
+<!-- {
+  "args": ["&#45;&#45;help"],
+  "output": "language for the greeting.*APP_LANG"
+} -->
+``` go
+package main
+
+import (
+  "os"
+
+  "github.com/urfave/cli"
+)
+
+func main() {
+  app := cli.NewApp()
+
+  app.Flags = []cli.Flag {
+    cli.StringFlag{
+      Name: "password, p",
+      Usage: "password for the mysql database",
+      FilePath: "/etc/mysql/password",
     },
   }
 
--- a/README.md
+++ b/README.md
@@ -587,13 +587,13 @@ func main() {
 }
 ```
 
-#### Values from Files
+#### Values from files
 
 You can also have the default value set from file via `FilePath`.  e.g.
 
 <!-- {
   "args": ["&#45;&#45;help"],
-  "output": "language for the greeting.*APP_LANG"
+  "output": "password for the mysql database"
 } -->
 ``` go
 package main
@@ -618,6 +618,9 @@ func main() {
   app.Run(os.Args)
 }
 ```
+
+Note that default values set from file (e.g. `FilePath`) take precedence over
+default values set from the enviornment (e.g. `EnvVar`).
 
 #### Values from alternate input sources (YAML, TOML, and others)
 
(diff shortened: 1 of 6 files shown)


bradrydzewski:
This demonstrates a potential solution for https://github.com/urfave/cli/issues/613. If there is interest in this feature we can flush out the implementation

bradrydzewski:
The failed unit test is because I updated the documentation in the readme without realizing there were validations. I wasn't exactly sure how to fix, but if someone can point me in the right direction I can update the pr

tboerger:
Awesome, that looks great

jszwedko:
@bradrydzewski 👍 this is looking good, thanks for the contribution!

My only thought is that we may want to more explicitly document in the README what the order of precedence of the sources is (I was surprised we didn't already, but it feels prudent to do so now given the addition of the new source).

README.md
```diff
@@ -32,6 +32,7 @@ applications in an expressive way.
     + [Alternate Names](#alternate-names)
     + [Ordering](#ordering)
     + [Values from the Environment](#values-from-the-environment)
+    + [Values from files](#values-from-files)
```
jszwedko: Looks like the linter wants this to match the heading (capital "Files")

README.md
```diff
@@ -586,6 +587,38 @@ func main() {
 }
 ```
 
+#### Values from Files
+
+You can also have the default value set from file via `FilePath`.  e.g.
+
+<!-- {
+  "args": ["&#45;&#45;help"],
+  "output": "language for the greeting.*APP_LANG"
```
jszwedko: `gfmrun` expects the output specified here to match the program output when ran (this helps make the examples are not invalidated by code changes). In this particular case, I might suggest writing a temporary file in the example and then having a default action on `app` that prints out the value, updating this expected `output` to match.

You can test locally via:

```
go get github.com/urfave/gfmrun/...
./runtests gfmrun
```

jszwedko: Also removing the `--help` from the args.

jszwedko:
I am wondering if we should try to integrate this into the help output for flags with the `FilePath` defined.

README.md
jszwedko: You can find more details about `gfmrun` here: https://github.com/urfave/gfmrun#go

bradrydzewski:
Thanks for the review! I'll plan on updating the pr accordingly this coming week

> My only thought is that we may want to more explicitly document in the README what the order of precedence of the sources is (I was surprised we didn't already, but it feels prudent to do so now given the addition of the new source).

I can certainly add this to the documentation.

I also meant to solicit feedback on precedence. Should it be Flag > Env > File? Or Flag > File > Env? Given the primary use case for loading parameters from file (secrets) it made me wonder whether or not it should take precedence over environment, although I'll admit I'm having difficulty articulating why one would be more important than the other. Any thoughts?

bradrydzewski:
> My only thought is that we may want to more explicitly document in the README what the order of precedence of the sources is

I added a section to document precedence for the file path. Let me know if you have any thoughts or feedback on the wording (writing good documentation is harder than writing code, IMO 😄 )

> I am wondering if we should try to integrate this into the help output for flags with the FilePath defined.

I also added the filepath to help. Below is an example output. Please let me know if you have any feedback regarding the formatting.

```
GLOBAL OPTIONS:
   --password value  password for the mysql database [/var/secret/password]
```

bradrydzewski:
ok, updated the PR based on feedback. Please note that a few of the Travis builders are failing for reasons that I can't explain. I noticed these same builders (Go 1.5 and Go 1.6) are also failing for #616 as well.

bradrydzewski:
friendly ping @jszwedko ... I'm sure you are busy so I want to do as much as I can on my end to help push this through. I think the PR addresses most (all?) of your comments, but it looks like the pipeline is acting up. For the life of me I can't figure out what is going on. Any thoughts?

Thanks so much!!

jszwedko:
Hi @bradrydzewski,

Apologies for the delay in reviewing this! This looks good to me.

Regarding the build failures, this appears to be due to some incompatibilities in the golang.org/x/tools tree for some Go versions. I'm working on fixing this over in #624. I'll drop a note here when that is merged.

jszwedko:
#624 has been merged, updating this branch

jszwedko:
And we're green, thanks again @bradrydzewski !

jszwedko:
Actually, I have one more thought :smile: 

The other fields that are like this, e.g. `EnvVar`, support multiple values. What do you think about having `FilePaths` be a `[]string` so that users could specify multiple files (e.g. `[]string{"~/.mysql_password", "/etc/mysql/password"}`) where it would try them until it found one to read from.

bradrydzewski:
yes, I'll update the pull request next week with that capability. Thanks!

bradrydzewski:
closing now that https://github.com/urfave/cli/pull/675 was merged (thanks @jmccann )

PR #615, opened 2017-04-04
Remove the error return signature

uudashr:
I didn't see any error returned... so better remove the signature


jszwedko:
Thanks @uudashr !

PR #616, opened 2017-04-06
Fix autocomplete issue when multiple apps run autocomplete from etc/bash_completion.d directory

tbaud0n:
If there is more than 1 app that uses bash_autocomplete from /etc/bash_completion.d, the first one initialize PROG to it's name.
Then, for the others apps, PROG is already set so it's doesn't complete the right cli app.

I've just unset PROG variable to fix this issue.


jszwedko:
Looks great, thank you @tbaud0n !

PR #618, opened 2017-04-14
Return raw slices in Getter impls

sttts:
The Getter interface should return the value, not the wrapped value, i.e. `[]string`, not `StringSlice`.

### DIFF
--- a/flag.go
+++ b/flag.go
@@ -152,7 +152,7 @@ func (f *StringSlice) Value() []string {
 
 // Get returns the slice of strings set by this flag
 func (f *StringSlice) Get() interface{} {
-	return *f
+	return []string(*f)
 }
 
 // Apply populates the flag given the flag set and environment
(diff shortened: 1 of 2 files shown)


jszwedko:
Hi @sttts,

Thanks for putting this together! I agree that this feels like a better interface to expose, but hesitate to make this change on `master` as it would technically break backwards compatibility for users of the library. I know it isn't ideal, but you can work around it via something like `[]string(s.Get().(cli.StringSlice))`.

I'm curious what your thoughts are though, there may be some use cases I'm missing that are more difficult to work around which could motivate breaking compatibility.

meatballhat:
@sttts Ditto to what @jszwedko said, and I'll add that this should be trivial in `v2` as the slice types are defined as structs with inner raw slice members.  What do you think about targeting `v2` instead?

lynncyrin:

Issue #546, opened 2016-10-19
flag for -h does not work after set HelpFlag to something else


Robin11:
Hello! I need the "-h" flag for my command ,so I set the HelpFlag to "help",without "h".But it turned out that "-h" still show for help which is not I want!
First I set helpFLag to "help"

```
cli.HelpFlag = cli.BoolFlag{
        Name:"help",
        Usage:"To show help for the tool",
    }
```

Then in my command "ls", I have a flag "h",but it still shows for help.Hope someone would help!

jszwedko:
Hi @Robin11,

With

```golang
package main

import (
        "os"

        "github.com/urfave/cli"
)

func main() {
        cli.HelpFlag = cli.BoolFlag{
                Name:  "help",
                Usage: "To show help for the tool",
        }

        cli.NewApp().Run(os.Args)
}
```

You should see:

```
➜  cli git:(master) go run /tmp/tmp.go -h 
Incorrect Usage.

NAME:
   tmp - A new cli application

USAGE:
   tmp [global options] command [command options] [arguments...]
   
VERSION:
   0.0.0
   
COMMANDS:
     help, h  Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --help         To show help for the tool
   --version, -v  print the version
```

As you can see, the help message does print, but that is because it considers the passed flag, `-h`, invalid. Is this the behavior you are describing?

You can override this behavior by setting `OnUsageError`, like the following:

```
package main

import (
        "fmt"
        "os"

        "github.com/urfave/cli"
)

func main() {
        cli.HelpFlag = cli.BoolFlag{
                Name:  "help",
                Usage: "To show help for the tool",
        }

        app := cli.NewApp()

        app.OnUsageError = func(context *cli.Context, err error, isSubcommand bool) error {
                fmt.Println("invalid usage")

                return nil
        }

        app.Run(os.Args)
}
```

Which will show:

```
➜  cli git:(master) go run /tmp/tmp.go -h
invalid usage
```

Let me know if this helps!

Robin11:
thanks for your help!I have already see the situation you post,but this is not what i want.Below is the sample code I use.In the comand ls,I have a "-h " flag,but it does not work!
````
package main

import (
	"github.com/urfave/cli"
	"fmt"
	"os"
)

var Commands = []cli.Command{
	CommandDu,
	CommandMkdir,
	CommandRm,
	CommandCp,
	CommandList,
}
var CommandList = cli.Command{
[...]

PR #619, opened 2017-04-16
Fix can not set flag '-h' for custom

chensongz:
Users may want to set flag '-h' for other usage rather than "help".

But function `checkCommandHelp` and `checkSubcommandHelp` check every command for flag '-h' and '--help'. User-defined flag '-h' will always show help message.

Only keep flag '--help' more reasonable.

Relate to #546

### DIFF
--- a/help.go
+++ b/help.go
@@ -235,7 +235,7 @@ func checkHelp(c *Context) bool {
 }
 
 func checkCommandHelp(c *Context, name string) bool {
-	if c.Bool("h") || c.Bool("help") {
+	if c.Bool("help") {
 		ShowCommandHelp(c, name)
 		return true
 	}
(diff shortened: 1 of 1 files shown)


help.go
```diff
@@ -235,7 +235,7 @@ func checkHelp(c *Context) bool {
 }
 
 func checkCommandHelp(c *Context, name string) bool {
-	if c.Bool("h") || c.Bool("help") {
```
jszwedko: I think it'd be more accurate to call the `checkHelp` method defined above.

lynncyrin:
It's been ~2 years since the changes were requested, so I'm comfortable closing this. Please feel free to re-open / create a new PR though!

Issue #617, opened 2017-04-13
Does it support POSIX-style short flag combining?


chensongz:
I want to use flag `-lh` for `-l` `-h`.
How to support this?

SCKelemen:
I don't think this can be compatible. The aliases are just a collection of strings. There is no requirement that they be one character in length. Therefore, it seems like there would be considerable overhead and searching arguments, and parsing them into valid flags.

jszwedko:
This was added as optional behavior in #684 -- thanks for raising the issue @chensongz !

PR #621, opened 2017-04-17
Support POSIX-style short flag combining

chensongz:
Parse flag `-ab` to `-a` `-b`

Relate to #617

### DIFF
--- a/command.go
+++ b/command.go
@@ -137,7 +137,19 @@ func (c Command) Run(ctx *Context) (err error) {
 				flagArgs = args[firstFlagIndex:]
 			}
 
-			err = set.Parse(append(flagArgs, regularArgs...))
+			// separate combined flags
+			var flagArgsSeparated []string
+			for _, flagArg := range flagArgs {
+				if strings.HasPrefix(flagArg, "-") && strings.HasPrefix(flagArg, "--") == false && len(flagArg) >2 {
+					for _, flagChar := range flagArg[1:] {
+						flagArgsSeparated = append(flagArgsSeparated, "-" + string(flagChar))
+					}
+				} else {
+					flagArgsSeparated = append(flagArgsSeparated, flagArg)
+				}
+			}
+
+			err = set.Parse(append(flagArgsSeparated, regularArgs...))
 		} else {
 			err = set.Parse(ctx.Args().Tail())
 		}


jszwedko:
Hi @chensongz,

Thank you for putting this together! Do you mind:

* Adding tests for this logic
* Adding a note to the README about this support

command.go
```diff
@@ -137,7 +137,19 @@ func (c Command) Run(ctx *Context) (err error) {
 				flagArgs = args[firstFlagIndex:]
 			}
 
-			err = set.Parse(append(flagArgs, regularArgs...))
+			// separate combined flags
+			var flagArgsSeparated []string
+			for _, flagArg := range flagArgs {
+				if strings.HasPrefix(flagArg, "-") && strings.HasPrefix(flagArg, "--") == false && len(flagArg) >2 {
```
jszwedko: Why `len(flagArg) > 2`?

rhatdan:
Really this this fix.  Did @chensongz disappear.

command.go
rhatdan: Probably should be just one I guess.

rhatdan: Actually I think this is correct.  Since the flagArg starts with - followed by at least one char, then he needs to make sure their are two or more chars after the "-"

-t is valid and should be just handled

-ti 
Should be treated as -t and -i
-tip should be treated as -t -i -p

rhatdan:
Just pulled this into my project and it does exactly what I need.

rhatdan:
@baude WDYT

command.go
chensongz: Yeah, that's right.

chensongz:
Hello, I'm here.  : )
I will add tests for this feature and notes to the README recently.

baude:
I looked this over again today.  A couple of issues I see.  The current suggestion breaks the tests.  Also, in looking deeper at the tests, there are several tests that use options like "-option".  That is going to be problem for what I think is considered POSIX (-o, --option).  Anyways, we might need a flag like ForcePosix or UsePosix to accomplish what we want.  What do people think, specifically @jszwedko ?

rhatdan:
@jszwedko We are using your library for https://github.com/projectatomic/libpod and we want to mimic as much of the DOCKER CLI as possible.  A very common use case is to type

kpod run -ti fedora sh
docker run -ti fedora sh

Without this fixed, we could be forced to use a different CLI Library, which we really do not want to.  Since we like your library, and have it heavily embedded in our projects including CRI-O and Buildah.

command.go
jszwedko: 👍

jszwedko:
Aha, good point @baude; the current CLI does allow `-option` options implicitly (since `flag` does).

I agree, to maintain backwards compatibility, that we should introduce a switch to enable parsing of combined short flags; however, I hesitate to call it anything `Posix` unless we also ensure that it meets the other `POSIX` recommendations (http://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html).

I'm still open to reviewing additional PRs around this.

Thanks all!

rhatdan:
How about TraditionalShortOptionHandling=True  Or something like that?

jszwedko:
👍 something like that would be fine

chensongz:
Can someone help to complete this MR?

baude:
I willk pick this PR up, rebase, add the flag, and a test or two.

jszwedko:
Closing this up in lieu of #684

PR #627, opened 2017-04-24
Rely on Command context in Run()

### LINKED ISSUE #609: cli.Context.Command not available to cli.Command.OnUsageError
If `OnUsageError` is triggered for a command, `cli.Command` is not yet available to the function via `cli.Context.Command`. To reproduce:
```
func main() {
	app := cli.NewApp()
	app.Name = "myapp"

	app.Commands = []cli.Command{
		cli.Command{
			Name: "mycmnd",
			OnUsageError: func(c *cli.Context, err error, isSubcommand bool) error {
				fmt.Printf("command name: %#v\n", c.Command.FullName())
				return nil
			},
		},
	}

	app.Run(os.Args)
}
```

Running `go run main.go mycmnd -a` will now produce:
```
command name: ""
```
I might be going this the wrong way or missing something but what I want to achieve is being able to print an error message that is aware of its calling command similar to CLIs like `docker` & `vault` e.g.
```
myapp mycmnd: flag provided but not defined: -a
Try 'myapp mycmnd -help' for more information
```
Looking at the current `cli/command.go` I can see why this is show in `func (c Command) Run(ctx *Context) (err error)` where `context.Command = c` is set after `c.OnUsageError, c.Before, c.After` are handled. What I am referring to here would have been solved if `context.Command = c` was set much earlier but there might be something guiding this decision that I am missing here as well. Any feedback would be gladly be welcomed.

jszwedko:
Was previously relying on the parent context which caused things like
`.Command` to not be available to OnUsageError().

Fixes #609

### DIFF
--- a/command.go
+++ b/command.go
@@ -154,19 +154,20 @@ func (c Command) Run(ctx *Context) (err error) {
 	}
 
 	context := NewContext(ctx.App, set, ctx)
+	context.Command = c
 	if checkCommandCompletions(context, c.Name) {
 		return nil
 	}
 
 	if err != nil {
 		if c.OnUsageError != nil {
-			err := c.OnUsageError(ctx, err, false)
+			err := c.OnUsageError(context, err, false)
 			HandleExitCoder(err)
 			return err
 		}
-		fmt.Fprintln(ctx.App.Writer, "Incorrect Usage:", err.Error())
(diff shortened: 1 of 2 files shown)


josephlr:
This PR fixes my problem and seems to work.

jszwedko:
Thanks for confirming @josephlr !

Issue #625, opened 2017-04-24
Do not show stack trace when using pkg/error


jszwedko:
This functionality was added in #560, but https://github.com/urfave/cli/issues/558#issuecomment-281531299 points out that this shows the stack trace to the end-user by default which is generally not desirable.

mh-cbon:
hi,

what if instead the package provides some error handlers in order to let the user decide what to do with the error.

Before that change about `pkg/errors`, there was just no such data available.
Now, there s too much data.
I think, it is only lacking of handling.

Where an error handler could detect that this error is stacktraced, thus appropriately handle it.
An handler which would be bale to read some env to display full stack en dev, partial stack/nostack on production.
To go further a parsing error might return file/line/pos, and an error handler detecting that might print a more appropriate error to the end user, in a generic way.

But, this does not really help current users in the sense that they ll need to appropriately  return typed error to trigger the appropriate display behavior.

That being said, a default error handler might be defined to reproduce old behavior.
Other users would then subscribe the handlers they are interested into to replace that default.

mh-cbon:
To add some example,

3 kind of error output,

> $ myprogram
> wrong arguments
> // I don t want more except maybe help section of the man.

> $ myprogram
> panic(...)
> // I expect to get a trace


> $ myprogram
> wrong syntax at xx ll:pp
> // here you might want to show the source and point the place.

lynncyrin:
Given that this is from last year, I think I'm comfortable closing it 🙂 feel free to re-open / open a new issue / comment in support if there's still interest here!

Issue #578, opened 2016-12-24
Make %+v formatting of ErrorFormatter optional


cyphar:
Not all users want, by default, to have their `pkg/error` errors' stack traces printed. This should be handled by a `--debug` flag or `Debug` option in the App, so that you only get a stack trace if the user actually asked for one.

jszwedko:
:+1: I can definitely see that argument, apologies for not conceiving of it when reviewing #560.

I'll mark this as a bug.

lynncyrin:
Given that this is from last year, I think I'm comfortable closing it 🙂 feel free to re-open / open a new issue / comment in support if there's still interest here!

PR #628, opened 2017-04-25
Allow custom ExitError handler function

phinnaeus:
Forgive me if this is already possible, I ready through the code and documentation extensively but couldn't find a way to do what I wanted.

There are a couple of open issues related to not showing the stack traces of errors. Since this is a tricky situation to solve and please all users, I think it makes sense to allow users to just define their own error handler function.

https://github.com/urfave/cli/issues/625
https://github.com/urfave/cli/issues/578

In my case, I like the stack traces, but I want them printed to a file instead of stderr. With this change it's easy for me to do something like this:

```
func main() {
  app := cli.NewApp()
  app.Name = "boom"
  app.Usage = "make an explosive entrance"
    fmt.Println("boom! I say!")
    return nil
  }

  // of course this is contrived, I would want to print something out to stderr at the same time :)
  app.ExitErrHandler = func(err error) {
    if file, err := os.Create("error-log.txt"); err == nil {
      defer file.Close()
      file.WriteString(fmt.Sprintf("%+v\n", err))
    }
  }

  app.Run(os.Args)
}
```

### DIFF
--- a/app.go
+++ b/app.go
@@ -468,7 +468,7 @@ func (a *App) appendFlag(flag Flag) {
 	}
 }
 
-func (a *App) handleExitCoder(context *Context, error) {
+func (a *App) handleExitCoder(context *Context, err error) {
 	if a.ExitErrHandler != nil {
 		a.ExitErrHandler(context, err)
 	} else {
--- a/app.go
+++ b/app.go
@@ -121,7 +121,6 @@ func NewApp() *App {
 		Action:       helpCommand.Action,
 		Compiled:     compileTime(),
 		Writer:       os.Stdout,
-		ExitErrHandler: HandleExitCoder,
 	}
 }
 
@@ -211,7 +210,7 @@ func (a *App) Run(arguments []string) (err error) {
 	if err != nil {
 		if a.OnUsageError != nil {
 			err := a.OnUsageError(context, err, false)
-			a.ExitErrHandler(err)
+			a.handleExitCoder(err)
 			return err
 		}
 		fmt.Fprintf(a.Writer, "%s %s\n\n", "Incorrect Usage.", err.Error())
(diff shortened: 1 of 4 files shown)


phinnaeus:
If this change is something that you might want to include, I can add some info and an example to the README under the [Exit Code](https://github.com/urfave/cli#exit-code) heading.

jszwedko:
:+1: this seems like a reasonable addition to me. Do you mind resolving the conflicts and perhaps adding a test for the defaulting behavior?

Apologies for the delay in review.

phinnaeus:
Will add the test and fix the conflict. One thing I was wondering about after I submitted this is if the ExitErrHandlerFunc function type should have the context passed in as well?

Since writing this I ran into a use-case where I wanted to log some metrics when an fatal error occurred (using a custom exit error handler), and seemed like it would be nice to access things in the context to include with the metrics.

jszwedko:
👍  I could see having the `context` being useful for that.

phinnaeus:
Uh, sorry, closed by accident. Will re-open shortly with updated code.

phinnaeus:
(Working on adding a test)

phinnaeus:
This should be ready to merge now.

phinnaeus:
I agree it's not ideal. I'm open to suggestions to improving it now or in the future. But since this is marked approved can it be pulled?

Thanks!

phinnaeus:
Rebased in the hopes that this can be pulled soon, or if there are more suggestions for improvements I will do my best to take care of them.

phinnaeus:
@jszwedko @meatballhat Rebased and fixed conflicts -- can this be pulled? Thanks!

jszwedko:
Agreed, I think this is a step in the right direction, thanks @phinnaeus !

PR #629, opened 2017-05-05
Subcommand OnUsageError should be passed to app

josephlr:
Similar to #589, not all of the `Command` components were being passed into the `App` created in `startApp()`. This PR fixes the issue by directly passing the command's `OnUsageError`. I came across this issue when trying to use a custom `OnUsageError` in my application. This PR and #627 fix the problems for me.

The only thing to really consider here is if the `App`-level `OnUsageError` should be forwarded if one isn't set in the `Command`. Something like:
```go
if c.OnUsageError != nil {
	app.OnUsageError = c.OnUsageError
} else {
	app.OnUsageError = ctx.App.OnUsageError
}
```
I opted not to do this; nothing like it is done in the `startApp()` command.

### DIFF
--- a/command.go
+++ b/command.go
@@ -291,6 +291,7 @@ func (c Command) startApp(ctx *Context) error {
 	} else {
 		app.Action = helpSubcommand.Action
 	}
+	app.OnUsageError = c.OnUsageError
 
 	for index, cc := range app.Commands {
 		app.Commands[index].commandNamePath = []string{c.Name, cc.Name}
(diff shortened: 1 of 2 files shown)


josephlr:
Any progress on this or #630?

jszwedko:
:+1: this looks good, but do you mind adding a test asserting this behavior to avoid regressions?

Apologies for the delay, just getting back from vacation.

josephlr:
Thanks @jszwedko test has been added, and I confirmed that it would catch a regression in app.OnUsageError.

Sorry for the delay.

jszwedko:
Thanks @josephlr !

PR #630, opened 2017-05-06
Add ability to use custom Flag types

josephlr:
Users can now use custom flags types (conforming to the `Flag` interface) in their applications. It is mentioned [here](https://github.com/urfave/cli/blob/master/flag.go#L55-L63) that users should be able to define their own `Flag` types. Before this was not possible due to some unsafe reflection in `visableFlags` (the other uses of reflection are safe as they either check for validity or use string-based reflection).

To fully support the use of custom flags, the global flags `HelpFlag`, `VersionFlag`, and `BashCompletionFlag` are made to be of type `Flag`. The package only uses their names, so this change shouldn't break anything. 

Tests were also added.

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -1520,6 +1520,63 @@ func TestApp_OnUsageError_WithWrongFlagValue_ForSubcommand(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


jszwedko:
This looks great, thank you!

Issue #523, opened 2016-09-05
Don't want any commands in the CLI interface (including help)


nthnca:
Is it possible to use cli for tools without subcommands?

Currently the help always prints out something like this in the usage:
scratch [global options] command [command options] [arguments...]

Is there anyway to get rid of the help command and have it print something more like:
scratch [global options] [arguments...]

It appears if I set "app.HideHelp = true" it gets close to what I want but I would still like to keep the global options of -h, --help.

nthnca:
So I have figured out how to get close to what I want:

        app.HideHelp = true                                                     
        app.Flags = []cli.Flag{                                                 
                cli.BoolFlag{Name: "help, h", Usage: "show help"},              
        }                                                                       
                                                                                
        app.Action = func(c *cli.Context) error {                               
                if c.Bool("help") {                                             
                        cli.ShowAppHelp(c)                                      
                        return nil                                              
                }                                                               
                fmt.Println("Hello!")                                           
                return nil                                                      
        } 

There are just two problems:
1. The help output still uses the wording "Global Options".
2. It feels like this is a lot of code for what is for many simple tools should be the default behavior.

Solutions (I am happy to write the code if these are wanted behaviors)
1. The help output should only use the word "Options" if there are no subcommands defined.
2. I would suggest that the "help" command should not be added unless there are user defined commands.

npcode:
My workaround is calling `app.Setup()` and empty `app.Commands` before calling `app.Run()`. e.g.
```
app.Setup()
app.Commands = []cli.String{}
if err := app.Run(os.Args); err != nil {
    fmt.Fprintf(os.Stderr, "%+v\n", err)
    os.Exit(1)
}
```

jszwedko:
@nthnca I agree (apologies for the delay in triaging this!).

I'm definitely happy to review PRs augmenting this behavior. For the current version, I think we could probably introduce another property for controlling this to keep compatibility, but, in the `v2` branch, I could see us doing something more drastic like not automatically adding the `help` subcommand or `-h` flag, but instead just provide them as package level variables (they are currently private) and let the user add them when initializing the `App` struct.

sorsasampo:
Any update on this? Not having any commands at all seems like a really popular usecase.

twlz0ne:
Just set [CustomAppHelpTemplate](https://github.com/urfave/cli/blob/8e01ec4cd3e2d84ab2fe90d8210528ffbb06d8ff/help.go#L139):
```go
[...]

Issue #222, opened 2015-05-12
Proposal: option to hide help subcommand and keep global option


ahmetalpbalkan:
I have a use case where I want to hide the `help` subcommand that gets added to the app by default; however, I want to keep `-h, --help` because that's generally what people run when they need help.

Unfortunately, currently `app.HideHelp=true` hides both. 

```
$ mycmd
NAME:
   mycmd

USAGE:
   mycmd [global options] command [command options] [arguments...]

[...]
COMMANDS:
   foo		does foo
   bar		does bar
   help, h	Shows a list of commands or help for one command  <------- hide this 👎

GLOBAL OPTIONS:
   --help, -h		show help  <------- keep this 👍
   --version, -v	print the version

```

We can have another option `app.HideHelpSubcommand` to hide the subcommand and `HideHelp` would still hide both globab option and the subcommand to keep it backwards compatible.

How do folks feel about such a feature?

wking:
I was missing this too, but it looks like there's something similar currently in flight with #636.

PR #636, opened 2017-06-12
Add HelpFlag as part of VisibleFlags.

harshavardhana:
Additionally also allow help command to be hidden

### DIFF
--- a/app.go
+++ b/app.go
@@ -45,8 +45,10 @@ type App struct {
 	Flags []Flag
 	// Boolean to enable bash completion commands
 	EnableBashCompletion bool
-	// Boolean to hide built-in help command
+	// Boolean to hide built-in help flag
 	HideHelp bool
+	// Boolean to hide built-in help command
+	HideHelpCommand bool
 	// Boolean to hide built-in version flag and the VERSION section of help
 	HideVersion bool
 	// Populate on app startup, only gettable through method Categories()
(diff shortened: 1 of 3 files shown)


harshavardhana:
Any feedback on this @jszwedko ?

jszwedko:
Hi @harshavardhana,

Apologies for the delay in review. Do you mind adding some additional details about the what this change is enabling?

It also appears to alter the existing interface by changing the behavior of the `HideHelp` field. If that was intentional, we may need to retarget this to the `v2` branch to avoid breaking compatibility.

wking:
Part of what this PR is doing is splitting `HideHelp` into separate booleans for the command and option.  That would let us address #222 and avoid things like listing:

```
USAGE:
    [global options] command [command options] [arguments...]
```

[for an app whose only command is `help`][1] where the meat of the app is in the app-level `Run`.  With this PR, setting `HideHelpCommand` but not `HideHelp` will give you:

```
USAGE:
    [global options] [arguments...]
```

I agree that there are backwards-compat concerns, but think we can address them with a slight tweak.  Instead of re-purposing `HideHelp`, I'd deprecate it (to be removed in v2) and add both `HideHelpCommand` and `HideHelpFlag`.  Then you could hide if the specific property was set or the deprecated generic property was set.  That covers all cases except:

1. Alice's new code sets `HideHelpFlag` `true`, and hands the `App` off to Bob's code.
2. Bob's old code sets `HideHelp` false, because it wants to ensure that the `--help` flag is visible.

Bob's usecase would work now (because the only way Alice could have disabled `--help` is via `HideHelp`), but Bob's usecase would break with my proposed approach.  Still, I think that passing `App` and `Command` instances around with different parties having different opinions on whether `--help` should be displayed is unlikely.  In most cases, these are only used locally within a single repository.

[1]: https://github.com/urfave/cli/blob/v1.20.0/app_test.go#L189-L193

jszwedko:
@wking aha, I see, thank you for providing additional context. I like your idea of splitting out the flags and deprecating `HideHelp`.

devmattrick:
Sorry for the ping, but are still plans to implement this?

AkihiroSuda:
Any chance to get this reopened?

lynncyrin:
@AkihiroSuda @devmattrick either of you are free to open a new PR! Please create an associated issue first though.

AkihiroSuda:
Will try, https://github.com/urfave/cli/issues/523 seems the existing issue ticket that covers this.

AkihiroSuda:
PR: https://github.com/urfave/cli/pull/1083

Issue #599, opened 2017-02-15
TOML/YAML Parsing int64, uint64, uint, time.Duration 


timjchin:
Love the library, thank you all for working on it. 

Ran into an issue when using toml/yaml imports, `int64`, `uint64`, `uint` and `time.Duration` are not parsed. 

For example:

test.toml
```
test-int=5
test-int-64=9223372036854775807
test-uint-64=100
test-uint=50
# test-duration="4m49s"
```

test.yaml
```
test-int: 5
test-int-64: 9223372036854775807
test-uint-64: 100
test-uint: 50
# test-duration: "4m49s"
```

main.go:
```
package main

import (
	"fmt"
	"os"
	"time"

	"gopkg.in/urfave/cli.v1"
	"gopkg.in/urfave/cli.v1/altsrc"
)

func main() {
	app := cli.NewApp()

	var (
		testInt64    int64
		testUint64   uint64
		testUint     uint
		testInt      int
		testDuration time.Duration
	)

	app.Flags = []cli.Flag{
		altsrc.NewInt64Flag(cli.Int64Flag{
			Name:        "test-int-64",
			Destination: &testInt64,
		}),
		altsrc.NewUint64Flag(cli.Uint64Flag{
			Name:        "test-uint-64",
			Destination: &testUint64,
		}),
		altsrc.NewUintFlag(cli.UintFlag{
			Name:        "test-uint",
			Destination: &testUint,
		}),
		altsrc.NewIntFlag(cli.IntFlag{
			Name:        "test-int",
			Destination: &testInt,
		}),
		altsrc.NewDurationFlag(cli.DurationFlag{
			Name:        "test-duration",
			Destination: &testDuration,
		}),
		cli.StringFlag{
			Name: "config, c",
		},
	}

	app.Action = func(c *cli.Context) error {
		fmt.Println(testInt)
		fmt.Println(testInt64)
		fmt.Println(testUint64)
		fmt.Println(testUint)
		fmt.Println(testDuration)
		return nil
	}

	// TOML:
	app.Before = altsrc.InitInputSourceWithContext(app.Flags, altsrc.NewTomlSourceFromFlagFunc("config"))

	// YAML:
	// app.Before = altsrc.InitInputSourceWithContext(app.Flags, altsrc.NewYamlSourceFromFlagFunc("config"))
	app.Run(os.Args)
}
```

Running this with YAML or TOML results in the same outputs. 

Results from running  with `test-duration` comented out:

[...]

PR #641, opened 2017-07-06
altsrc: Parse durations from strings

cbranch:
Fixes part of #599.
I don't have a good idea (or the requirement) for fixing the other datatypes in that issue; presumably it should show the same leniency with exact types when parsing from weakly-typed data sources.

### DIFF
--- a/altsrc/map_input_source.go
+++ b/altsrc/map_input_source.go
@@ -71,22 +71,26 @@ func (fsm *MapInputSource) Int(name string) (int, error) {
 func (fsm *MapInputSource) Duration(name string) (time.Duration, error) {
 	otherGenericValue, exists := fsm.valueMap[name]
 	if exists {
-		otherValue, isType := otherGenericValue.(time.Duration)
-		if !isType {
-			return 0, incorrectTypeForFlagError(name, "duration", otherGenericValue)
-		}
-		return otherValue, nil
+		return castDuration(name, otherGenericValue)
 	}
 	nestedGenericValue, exists := nestedVal(name, fsm.valueMap)
 	if exists {
-		otherValue, isType := nestedGenericValue.(time.Duration)
-		if !isType {
-			return 0, incorrectTypeForFlagError(name, "duration", nestedGenericValue)
-		}
-		return otherValue, nil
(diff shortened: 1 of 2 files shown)


jszwedko:
:+1: This seems reasonable to me, but do you mind adding a couple of tests for this behavior?

cc/ @ChrisPRobinson since I think you still have the most context on `altsrc`

ChrisPRobinson:
@cbranch 👍, seems good other than tests. 

Reviewing my code I think what I should have done re #599 was just to add those types to the system.
Uint, UInt64, Int64? Then have similar impls for those. But that would be a larger change. I think I mapped the types that existed at the time, or I just missed this. Not a fan of the boiler plate that is required but I don't see another way. Sorry for the bugs/design flaws...

cbranch:
Finally got back to this one!
I added a simple test and moved the code into its own function to reduce the code reuse.
Regarding Uint/Uint64/Int64, these are larger changes because it requires an implementation of ApplyInputSourceValue for each of the flags, plus adding the new types on InputSourceContext interface/MapInputSource implementation. Since the workaround is easy (just use int) I'll punt for now.

altsrc/map_input_source.go
```diff
@@ -71,24 +71,29 @@ func (fsm *MapInputSource) Int(name string) (int, error) {
...
-		return otherValue, nil
+		return castDuration(name, nestedGenericValue)
 	}
 
 	return 0, nil
 }
 
+func castDuration(name string, value interface{}) (time.Duration, error) {
+	otherValue, isType := value.(time.Duration)
+	if !isType {
```
jszwedko: Nitpick, but can we do an early return like

```go
if isType {
    return otherValue, nil
}

...
```

Just to remove some of the nesting.

cbranch: Agreed, done.

jszwedko:
Just waiting for build to finish (the failure looked like it might have been a CI flake).

jszwedko:
Fixed the build failure in https://github.com/urfave/cli/pull/703. Apologies for the delay!

PR #644, opened 2017-07-07
Fix for gitlab.com/ayufan/golang-cli-helpers

firelizzard18:
The library `gitlab.com/ayufan/golang-cli-helpers` creates cli flags from struct fields. Flags with multiple names all map to the same struct field value. `copyFlag` on `context.go:239` calls `*flag.FlagSet.Set(name, value string) error`. This calls `clihelpers.StructFieldValue.Set(val string) error` (which matches the `flag.Value` API). For slices, this appends the new value to the slice. The end result is `normalizeFlags -> copyFlag -> FlagSet.Set -> StructFieldValue.Set` results in something like `["a", "b", "c", "[a, b, c]"]`, which is clearly undesired. To combat this, I added `flagsAreIdentical`, which checks to see if all of the (string) values of the named flags are identical; and I modified `normalizeFlags` to skip a flag if all of its aliases already have identical values. There's no reason to normalize values if they're all already the same.

### DIFF
--- a/context.go
+++ b/context.go
@@ -244,6 +244,31 @@ func copyFlag(name string, ff *flag.Flag, set *flag.FlagSet) {
 	}
 }
 
+func flagsAreIdentical(names []string, set *flag.FlagSet) bool {
+	// fix for situations such as gitlab.com/ayufan/golang-cli-helpers; in that
+	// case, the different names point to the same struct field value so
+	// copyFlag breaks things
+
+	if len(names) == 0 {
+		return false
+	}
+
+	f := set.Lookup(strings.Trim(names[0], " "))
+	if f == nil {
+		return false
+	}
+
+	for i := 1; i < len(names); i++ {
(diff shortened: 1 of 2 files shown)


jszwedko:
Hi @firelizzard18,

Thank you for this proposal. I'm having a bit of a hard time following the issue as-is; do you mind adding a few test cases that might make it more clear what this guards against (and also to avoid breaking it in the future)?

firelizzard18:
I added a test-case. The basics are, without my fix, `m.Test` will be `[1, 2, [1, 2]]` (`fmt`), and with my fix, it will be the expected `[1, 2]` (given the arguments my test uses: `cmd hi --test 1 --test 2`)

dolmen:
This test adds a dependency on `gitlab.com/ayufan/golang-cli-helpers`. Could you rewrite the test to show the issue with that dependency?

lynncyrin:

Issue #557, opened 2016-11-03
Confusing "Help" Results


kenji-isuntv:
This issue may be related to #163.

Here is the source code (`bug.go`, also available at [<kbd>The Go Playground</kbd>](https://play.golang.org/p/H80jY3BWX4)) for the bug report:

```go
package main

import (
	"os"

	"gopkg.in/urfave/cli.v1"
)

func dummyAction(_ *cli.Context) error {
	return nil
}

var (
	command1 = cli.Command{
		Name:        "command1",
		Usage:       "command1.Usage",
		Description: "command1.Description",
		Subcommands: []cli.Command{
			{
				Action:      dummyAction,
				Name:        "action1",
				Usage:       "command1action1.Usage",
				ArgsUsage:   "<arg1>",
				Description: "command1action1.Description",
			},
			{
				Action:      dummyAction,
				Name:        "action2",
				Usage:       "command1action2.Usage",
				ArgsUsage:   "<arg1> [<arg2>]",
				Description: "command1action2.Description",
			},
			{
				Action:      dummyAction,
				Name:        "action3",
				Usage:       "command1action3.Usage",
				ArgsUsage:   "<arg1> [<arg2>]...",
				Description: "command1action3.Description",
			},
		},
	}
	command2 = cli.Command{
		Action:      dummyAction,
		Name:        "command2",
		Usage:       "command2.Usage",
		Description: "command2.Description",
		Subcommands: []cli.Command{
			{
				Action:      dummyAction,
				Name:        "action1",
				Usage:       "command2action1.Usage",
				ArgsUsage:   "<arg1>",
				Description: "command2action1.Description",
			},
			{
				Action:      dummyAction,
				Name:        "action2",
				Usage:       "command2action2.Usage",
				ArgsUsage:   "<arg1> [<arg2>]",
				Description: "command2action2.Description",
			},
			{
				Action:      dummyAction,
				Name:        "action3",
				Usage:       "command2action3.Usage",
				ArgsUsage:   "<arg1> [<arg2>]...",
				Description: "command2action3.Description",
			},
		},
	}
	command3 = cli.Command{
		Action:      dummyAction,
		Name:        "command3",
		Usage:       "command3.Usage",
		ArgsUsage:   "<arg1> [<arg2>]...",
		Description: "command3.Description",
	}
)

func main() {
	app := cli.NewApp()
	app.Name = "app.Name"
	app.Usage = "app.Usage"
	app.Version = "app.Version"
[...]

PR #647, opened 2017-07-26
WIP: Fix error where app name would be used instead of command help name

Strift:
This should fix the issue #557 

Sorry this is WIP. I will update when I'm done.

### DIFF
--- a/command.go
+++ b/command.go
@@ -241,7 +241,7 @@ func (c Command) startApp(ctx *Context) error {
 	app.Metadata = ctx.App.Metadata
 	// set the name and usage
 	app.Name = fmt.Sprintf("%s %s", ctx.App.Name, c.Name)
-	if c.HelpName == "" {
+	if c.HelpName != "" {
 		app.HelpName = c.HelpName
 	} else {
 		app.HelpName = app.Name


Strift:
I'm afraid I do not understand what is the actual problem.

Failed assertions happens at:
- app_test.go:1097
- app_test.go:1100
- app_test.go:1159
- app_test.go:1162
- app_test.go:1190
- app_test.go:1193

But all the outputs look good to me... :/

Strift:
I know there are failed assertions when running this fork. But I have no idea why are the tests failing.

@jszwedko If you could tell me what I should do to make the tests pass / refactor the tests, I would gladly investigate a bit more.

jszwedko:
Hi @Strift,

Thanks for taking a crack at this! Apologies for the delay, just getting back from traveling.

I believe this change does break behavior that was introduced in https://github.com/urfave/cli/pull/267 (see that thread for more discussion) which is why you see the failing tests here.

I think there are more nuances described in #557 than what you are describing here though. I think to address that issue a more drastic overhaul of the help system will be required to handle all of the cases described there. I think it can be simplified along the way though.

lynncyrin:

PR #651, opened 2017-08-03
Alter runtests for py2 + py3 compat

meatballhat:
which I noticed when running `make` on an Arch system where `python` is `python3`.

:memo: Before this can pass, the `v2` branch will need to be up to date with `master`, as the examples in the `v1` README use `v1`-specific bits that have no corresponding bits in `v2`.


meatballhat:
:sparkles: nothing changed anymore :rofl:

PR #652, opened 2017-08-03
Update travis config

meatballhat:
to only test on 1.8.x, and ensure we route to Ubuntu Trusty and a recent macOS image.


jszwedko:
:+1: :+1:

PR #655, opened 2017-08-04
Merging from v1

meatballhat:
I still have some work to do with the tests, but I wanted to get this PR up for review in case any of my conflict resolution choices look silly.

**Update**: This is :white_check_mark: and ready for scrutiny :tada:

### DIFF
--- a/altsrc/flag_generated.go
+++ b/altsrc/flag_generated.go
@@ -27,6 +27,13 @@ func (f *BoolFlag) Apply(set *flag.FlagSet) {
 	f.BoolFlag.Apply(set)
 }
 
+// ApplyWithError saves the flagSet for later usage calls, then calls the
+// wrapped BoolFlag.ApplyWithError
+func (f *BoolFlag) ApplyWithError(set *flag.FlagSet) error {
+	f.set = set
+	return f.BoolFlag.ApplyWithError(set)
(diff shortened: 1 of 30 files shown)


altsrc/flag_generated.go
```diff
@@ -141,6 +183,13 @@ func (f *IntSliceFlag) Apply(set *flag.FlagSet) {
 	f.IntSliceFlag.Apply(set)
 }
 
+// ApplyWithError saves the flagSet for later usage calls, then calls the
+// wrapped IntSliceFlag.ApplyWithError
+func (f *IntSliceFlag) ApplyWithError(set *flag.FlagSet) error {
```
jszwedko: I was thinking in `v2` we could just switch the signature of `Apply` to return `error`, but I'm OK with doing that as a second pass since this is just an additive change as-is.

cli.go
```diff
@@ -12,6 +12,7 @@
 //			 Usage: "say a greeting",
 //			 Action: func(c *cli.Context) error {
 //				 println("Greetings")
+//         return nil
```
jszwedko: :nitpick: indentation

I also find it amusing that this example uses `println`

meatballhat: Ack!

PR #656, opened 2017-08-07
Allow Ignoring Usage Errors

grubernaut:
In some rare cases, it's handy to allow the application to handle
use-case errors in other ways, without requiring an error to be
returned. This allows the application written to handle the usage
error in other ways. Or, allow the application to pass the flags
on to subsequent child processes regardless of the host application's
flag parsing.

```
$ go test -run="TestApp_IgnoreOnUsageError"
PASS
ok      github.com/grubernaut/cli       0.001s
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1306,6 +1306,7 @@ func main() {
       fmt.Fprintf(c.App.Writer, "WRONG: %#v\n", err)
       return nil
     },
+    IgnoreUsageError: true,
     Action: func(c *cli.Context) error {
       cli.DefaultAppComplete(c)
       cli.HandleExitCoder(errors.New("not an exit coder, though"))
--- a/app.go
+++ b/app.go
@@ -51,6 +51,8 @@ type App struct {
 	Action ActionFunc
 	// Execute this function if the proper command cannot be found
 	CommandNotFound CommandNotFoundFunc
+	// Ignore usage errors to be handled later
+	IgnoreUsageError bool
 	// Execute this function if an usage error occurs
 	OnUsageError OnUsageErrorFunc
 	// Compilation date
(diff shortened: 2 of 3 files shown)


meatballhat:
@grubernaut IIRC this is part of the original intent behind the `OnUsageError` func, so that folks can choose to define it with a `return nil` to do whatever handling is desired.  I'm generally OK with having this kind of usage be more explicit via a struct field like you've implemented, but I'm curious how the current feature(s) aren't meeting your needs :smiley:

grubernaut:
@meatballhat ah yea, I noticed that function. However, that returns the execution loop from the CLI app, without allowing for further interpolation from flags. It allows subcommands to pass through, but not incorrectly passed-in flags. 
For example: 
```
#fooApp/main.go
package main

import (
	"fmt"
	"os"

	cli "gopkg.in/urfave/cli.v2"
)

func main() {
	app := cli.App{}
	app.Name = "foo"
	app.Usage = "dont"
	app.OnUsageError = func(c *cli.Context, err error, isSubcommand bool) error {
		return nil
	}
		fmt.Printf("It's working!!!")
		return nil
	}
	app.Run(os.Args)
}
```
```
$ ./fooApp
It's working!!!%
$ ./fooApp command
It's working!!!%
$ ./fooApp -f 
$
```
The `-f` flag threw the `OnUsageError` function, which ended application execution, and returned a `nil` error. 

Albeit, this is a small use-case, but I'm currently working on a wrapper utility to run a conditional set of child programs via `os.Exec()`, and need to pass any argument flags to the wrapper binary that can be passed to any children. 

This could have been changed by modifying how the error is returned here:https://github.com/urfave/cli/blob/v2/app.go#L193-L194, but I didn't want to cause any changing behavior for any consumers that were returning `nil` from the `OnUsageError` function, who actually wanted execution to stop on a usage error. 

OT: but not quite sure how I broke the Travis build :scream:

lynncyrin:

PR #661, opened 2017-08-26
Allow customization of prefixes and environment variable hints in flag help strings

rliebz:
This will allow users to customize the prefix section or env hint section of the flag entries in the help menu without having to reimplement the rest of the logic required in defining FlagStringer.

For example, to keep all other defaults, but change placeholder display from:
```
-f value, --flag value	flag description here
```

To only include it once as such:
```
-f, --flag value	flag description here
```

One would currently have to redefine the following unexported functions:
- stringifyFlag (the only current entry point)
- flagValue
- withEnvHint
- stringifyIntSliceFlag
- stringifyInt64SliceFlag
- stringifyStringSliceFlag
- stringifySliceFlag
- unquoteUsage
- prefixedNames

This change should both be backward compatible and allow changing the prefix or environment hint independently, so either of those can be redefined by simple pure functions that don't actually depend on the rest of the `stringifyFlag` logic.

I kept the original function signatures here, but if it makes more sense, for example, for `withEnvHint` to be replaced by something like `func envHint(envVar string) string` to decouple it from the rest of the message at the cost of being able to manipulate that string in other ways, I would be happy to do so.

### DIFF
--- a/flag.go
+++ b/flag.go
@@ -37,6 +37,14 @@ var HelpFlag Flag = BoolFlag{
 // to display a flag.
 var FlagStringer FlagStringFunc = stringifyFlag
 
+// FlagNamePrefixer converts a full flag name and its placeholder into the help
+// message flag prefix. This is used by the default FlagStringer.
+var FlagNamePrefixer FlagNamePrefixFunc = prefixedNames
+
+// FlagEnvHinter annotates flag help message with the environment variable
+// details. This is used by the default FlagStringer.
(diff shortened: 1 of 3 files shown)


jszwedko:
:+1: I like the flexibility created by this.

Just for sanity, do you mind adding a test or two?

rliebz:
Just did a quick rebase and added a few test cases for each of the exported functions. Let me know if you have any other feedback.

jszwedko:
This looks great, thank you for adding in the tests!

PR #662, opened 2017-08-29
lint + DefaultCommand

### LINKED ISSUE #657: default 'app.commands' if one is not is supplied
I am having a hard time figuring out how to make one of my app.commands default when a command is not supplied at run time.
for example
```
...
app.Commands = []cli.command{
                       {
                        Name: "command that I want to be default"
			Usage:  "Does something",
			Action: runthedefaultcommand,
			Flags: []cli.Flag{
				cli.StringFlag{
					Name:   "something.something",
					Usage:  "stuff",
					EnvVar: "ENV_VAR",
				},
                          },
                      },
                      {
			Name:   "second optional command that overwrites default",
			Usage:  "Do something else",
			Action: runoptcommand,
			Flags: []cli.Flag{
				cli.StringFlag{
					Name:  "something.somethingelse",
					Usage: "stuff",
				},
                          },
                      },
...
```

OneOfOne:
This adds a global function that can be passed to app.Action to execute a default command.

Fixes #657

### DIFF
--- a/app.go
+++ b/app.go
@@ -212,7 +212,7 @@ func (a *App) Run(arguments []string) (err error) {
 
 	if err != nil {
 		if a.OnUsageError != nil {
-			err := a.OnUsageError(context, err, false)
+			err = a.OnUsageError(context, err, false)
 			HandleExitCoder(err)
 			return err
 		}
@@ -479,4 +479,12 @@ func (a *Author) String() string {
 	}
 
 	return fmt.Sprintf("%v%v", a.Name, e)
+}
+
+// DefaultAppComplete returns an ActionFunc to run a default command if non were passed.
+// Usage: `app.Action = DefaultCommand("command")`
+func DefaultCommand(name string) ActionFunc {
+	return func(ctx *Context) error {
+		return ctx.App.Command(name).Run(ctx)
+	}
 }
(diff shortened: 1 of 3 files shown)


jszwedko:
I missed this somehow, will review soon.

app.go
```diff
@@ -480,3 +480,11 @@ func (a *Author) String() string {
 
 	return fmt.Sprintf("%v%v", a.Name, e)
 }
+
+// DefaultAppComplete returns an ActionFunc to run a default command if non were passed.
+// Usage: `app.Action = DefaultCommand("command")`
```
jszwedko: Because the target of this comment is likely to be end-users, could we use `cli.DefaultCommand("command")`?

flag.go
```diff
@@ -778,7 +778,7 @@ func visibleFlags(fl []Flag) []Flag {
 func prefixFor(name string) (prefix string) {
 	if len(name) == 1 {
 		prefix = "-"
-	} else {
+	} else if name[0] != '-' {
```
jszwedko: Did you mean to include this change? It seems unrelated.

OneOfOne: Oops, yeah, I don't even remember why I did that, but I needed it for something, removed it tho.

omerxx:
Hey @OneOfOne and @jszwedko, great addition. I think an addition to the documentation would be great. I got here by mistake after searching through issues to find this default option. Found it only after implementing it on my own...

Issue #666, opened 2017-09-16
File paths and altsrc


VMitov:
When having a flag representing file path and using alternative sources if becomes hard to figure out what the file path is relative to. 
- if you're running from the same folder the config file is - cmd current dir is the same as the config file dir, so it doesn't matter
- if you're running from different folder and taking the option from the file - path should be treated as relative to the config file 
- if you're running from different folder and taking the option from the cmd - path should be treated as relative to the folder you're running the cmd from

I'm using v2 btw.

jszwedko:
:+1: I agree with these three assertions, can you qualify which are not currently met?

VMitov:
It depends on the implementation. In my case the last one is failing. If there is a way to find out where is the source of a flag - I would easily fix it.

VMitov:
I solved it with custom PathFlag. Do you fancy PR? The only caveat is - how I know the directory where the config file is in the PathFlag ApplyInputSourceValue?

jszwedko:
You should be able to get the current working directory via `os.Getwd`. PRs definitely welcome :smile:

VMitov:
What I mean is that in ApplyInputSourceValue don't have the flag name for the configuration file.

VMitov:
And that's what I mean :) https://github.com/urfave/cli/pull/670/files#diff-b409136834e66f5a239c3d6ca4df876dR255 .

lynncyrin:
Hiya! There's a known issue with the current implementation of the CLI, where altsrc is generally clunky and poorly documented. My current idea is that I'm going to move all of the `altsrc` functionality into the main package in some way, and that's how all of the _"altsrc is weird and different???"_ issues people are having will get resolved. I'm tracking that work here => https://github.com/urfave/cli/issues/833

PR #670, opened 2017-09-26
Add PathFlag

VMitov:


VMitov:
This need additional work that is discussed in https://github.com/urfave/cli/issues/666.

flag.go
```diff
@@ -490,6 +490,33 @@ func (f *StringFlag) ApplyWithError(set *flag.FlagSet) error {
 
 // Apply populates the flag given the flag set and environment
 // Ignores errors
+func (f *PathFlag) Apply(set *flag.FlagSet) {
+	f.ApplyWithError(set)
+}
+
+// ApplyWithError populates the flag given the flag set and environment
+func (f *PathFlag) ApplyWithError(set *flag.FlagSet) error {
```
jszwedko: I take it we don't need to resolve the path in here since it'll be relative to the working directory.

altsrc/flag.go
```diff
@@ -231,3 +255,19 @@ func isEnvVarSet(envVars []string) bool {
 func float64ToString(f float64) string {
 	return fmt.Sprintf("%v", f)
 }
+
+func stringToPath(context *cli.Context, path string) (string, error) {
+	inputSourcePath := context.String("load")
```
jszwedko: I see, yes, it'll be difficult to access the path of the loaded file from here.

One approach could be to add another method to `InputSourceContext` that returns the filepath of the source; this value would then be passed in by the YAML and TOML loaders to the constructed MapInputSource.

VMitov: Yeah, I was thinking in the same direction. I'll do that. Cheers!

VMitov: Done.

VMitov:
FYI I'll rebase and squash the commits when ready.

altsrc/input_source_context.go
```diff
@@ -9,6 +9,8 @@ import (
 // InputSourceContext is an interface used to allow
 // other input sources to be implemented as needed.
 type InputSourceContext interface {
+	Source() string
```
jszwedko: Can we add a comment here to make it more clear for implementers what this should return?

jszwedko:
This looks great @VMitov ! Thank you for adding the comprehensive tests. I just left one note about commenting the interface, but I think this is good to go after that.

jszwedko:
:+1: looks great! Let me know if you still want to squash.

VMitov:
Rebased.

PR #675, opened 2017-10-26
Ability to load variable from file - UPDATED

jmccann:
This is a continuation of #614 

It includes the requested updates @ https://github.com/urfave/cli/pull/614#issuecomment-296794519 and is rebased from master.

### DIFF
--- a/flag.go
+++ b/flag.go
@@ -125,9 +125,9 @@ func (f GenericFlag) Apply(set *flag.FlagSet) {
 // provided by the user for parsing by the flag
 func (f GenericFlag) ApplyWithError(set *flag.FlagSet) error {
 	val := f.Value
-	if envVal, ok := flagFromFileEnv(f.FilePath, f.EnvVar); ok {
-		if err := val.Set(envVal); err != nil {
-			return fmt.Errorf("could not parse %s as value for flag %s: %s", envVal, f.Name, err)
+	if fileEnvVal, ok := flagFromFileEnv(f.FilePath, f.EnvVar); ok {
+		if err := val.Set(fileEnvVal); err != nil {
+			return fmt.Errorf("could not parse %s as value for flag %s: %s", fileEnvVal, f.Name, err)
 		}
 	}
 
@@ -658,8 +658,6 @@ func stringifyFlag(f Flag) string {
 			),
 		)
 	case Int64SliceFlag:
-		// return FlagEnvHinter(fv.FieldByName("EnvVar").String(),
-		// 	stringifyInt64SliceFlag(f.(Int64SliceFlag)))
 		return FlagFileHinter(
 			fv.FieldByName("FilePath").String(),
 			FlagEnvHinter(
@@ -668,8 +666,6 @@ func stringifyFlag(f Flag) string {
 			),
 		)
 	case StringSliceFlag:
-		// return FlagEnvHinter(fv.FieldByName("EnvVar").String(),
-		// 	stringifyStringSliceFlag(f.(StringSliceFlag)))
 		return FlagFileHinter(
 			fv.FieldByName("FilePath").String(),
 			FlagEnvHinter(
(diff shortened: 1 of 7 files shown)


flag.go
```diff
@@ -718,14 +650,33 @@ func stringifyFlag(f Flag) string {
...
+			fv.FieldByName("FilePath").String(),
+			FlagEnvHinter(
+				fv.FieldByName("EnvVar").String(),
+				stringifyIntSliceFlag(f.(IntSliceFlag)),
+			),
+		)
 	case Int64SliceFlag:
-		return FlagEnvHinter(fv.FieldByName("EnvVar").String(),
-			stringifyInt64SliceFlag(f.(Int64SliceFlag)))
+		// return FlagEnvHinter(fv.FieldByName("EnvVar").String(),
```
jszwedko: Can we drop this comment?

flag.go
```diff
@@ -718,14 +650,33 @@ func stringifyFlag(f Flag) string {
...
+			fv.FieldByName("FilePath").String(),
+			FlagEnvHinter(
+				fv.FieldByName("EnvVar").String(),
+				stringifyInt64SliceFlag(f.(Int64SliceFlag)),
+			),
+		)
 	case StringSliceFlag:
-		return FlagEnvHinter(fv.FieldByName("EnvVar").String(),
-			stringifyStringSliceFlag(f.(StringSliceFlag)))
+		// return FlagEnvHinter(fv.FieldByName("EnvVar").String(),
```
jszwedko: Here too

flag.go
```diff
@@ -120,15 +125,9 @@ func (f GenericFlag) Apply(set *flag.FlagSet) {
...
-	if f.EnvVar != "" {
-		for _, envVar := range strings.Split(f.EnvVar, ",") {
-			envVar = strings.TrimSpace(envVar)
-			if envVal, ok := syscall.Getenv(envVar); ok {
-				if err := val.Set(envVal); err != nil {
-					return fmt.Errorf("could not parse %s as value for flag %s: %s", envVal, f.Name, err)
-				}
-				break
-			}
+	if envVal, ok := flagFromFileEnv(f.FilePath, f.EnvVar); ok {
```
jszwedko: It might be more accurate to rename all of these to `val` now given the logic changes.

flag.go
```diff
@@ -805,3 +761,18 @@ func stringifySliceFlag(usage, name string, defaultVals []string) string {
...
+			return string(data), true
+		}
+	}
+	for _, envVar := range strings.Split(envName, ",") {
+		envVar = strings.TrimSpace(envVar)
+		if envVal, ok := syscall.Getenv(envVar); ok {
+			return envVal, true
+		}
+	}
+	return
```
jszwedko: I prefer to eschew naked returns as they are less readable.

flag.go
```diff
@@ -805,3 +761,18 @@ func stringifySliceFlag(usage, name string, defaultVals []string) string {
 	usageWithDefault := strings.TrimSpace(fmt.Sprintf("%s%s", usage, defaultVal))
 	return fmt.Sprintf("%s\t%s", FlagNamePrefixer(name, placeholder), usageWithDefault)
 }
+
+func flagFromFileEnv(filePath, envName string) (val string, ok bool) {
```
jszwedko: This puts the precedence as:

* File
* Env Var

Especially given that the altsrc (reading from configuration files) takes a lower precedence than env vars, what do you think of swapping these two?

flag.go
jmccann: I renamed `envVal` to `fileEnvVal`.  I did not use `val` since it's already defined on the previous line 127.  I hope this is OK otherwise let me know what you'd like me to do.

jmccann:
@jszwedko There are already tests to check for loading precedence @ https://github.com/urfave/cli/pull/675/files#diff-f959b18b5d42c386fb387a60b0d27376R1296

I did update the test to reflect the preferred way to load, from EnvVar then File.

Let me know if there's anything else!

jszwedko:
👍 thanks @jmccann !

PR #676, opened 2017-10-28
Consider case when sorting strings

rliebz:
This makes sorting flags and other sections consistent with how most
command line tools function, by placing both flags `-A` and `-a` before
a flag `-B`.

Whereas previously you'd get generated documentation like this:
```
   -A           Big option A
   -B           Big option B
   -a           Little option a
```

Now it is more consistent with GNU-style and other command-line tools:
```
   -A           Big option A
   -a           Little option a
   -B           Big option B
```

### DIFF
--- a/category.go
+++ b/category.go
@@ -10,7 +10,7 @@ type CommandCategory struct {
 }
 
 func (c CommandCategories) Less(i, j int) bool {
-	return c[i].Name < c[j].Name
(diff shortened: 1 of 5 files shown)


jszwedko:
👍 good thought @rliebz 

This does change the sort order, but I believe this ordering to be the original intention for flag sorting.

PR #684, opened 2017-11-13
Combine bool short names

baude:
Adds the ability to allow the combination of bool
short-name options.  For example,
    
cmd foobar -ov
    
This is done through a bool "UseShortOptionHandler" set in
the command struct.

Built upon PR #621

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ applications in an expressive way.
   * [Version Flag](#version-flag)
     + [Customization](#customization-2)
     + [Full API Example](#full-api-example)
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM

rhatdan:
https://github.com/jszwedko PTAL

jszwedko:
Awesome, this looks good to me. Thanks for adding the tests and README documentation @baude!

rhatdan:
BTW We have found issues with this and the use of SkipArgReorder, basically they don't work together.  We have been fooling around with patches to fix this.  But have not completed the work yet.

PR #685, opened 2017-11-17
Export Value function in context

tsarpaul:
Hi, I need this function to retrieve a flag's value :)
My constraint is that I only have a context object and a type of interface Flag, and I couldn't find a convenient way.


tsarpaul:
Nevermind, I found out I could use GlobalGeneric so I'm not sure why that internal function exists if you can just use that...

PR #686, opened 2017-11-20
Handle ShortOptions and SkipArgReorder

baude:
There was a bug in parsing when both ShortOptions
and SkipArgReorder were being used together.

Signed-off-by: baude <bbaude@redhat.com>

### DIFF
--- a/command.go
+++ b/command.go
@@ -115,57 +115,29 @@ func (c Command) Run(ctx *Context) (err error) {
 		return err
 	}
 	set.SetOutput(ioutil.Discard)
-
+	firstFlagIndex, terminatorIndex := getIndexes(ctx)
+	flagArgs, regularArgs := getAllArgs(ctx.Args(), firstFlagIndex, terminatorIndex)
+	if c.UseShortOptionHandling {
+		flagArgs = translateShortOptions(flagArgs)
+	}
 	if c.SkipFlagParsing {
 		err = set.Parse(append([]string{"--"}, ctx.Args().Tail()...))
 	} else if !c.SkipArgReorder {
-		firstFlagIndex := -1
-		terminatorIndex := -1
-		for index, arg := range ctx.Args() {
-			if arg == "--" {
-				terminatorIndex = index
-				break
-			} else if arg == "-" {
-				// Do nothing. A dash alone is not really a flag.
-				continue
-			} else if strings.HasPrefix(arg, "-") && firstFlagIndex == -1 {
-				firstFlagIndex = index
-			}
-		}
-
 		if firstFlagIndex > -1 {
-			args := ctx.Args()
-			regularArgs := make([]string, len(args[1:firstFlagIndex]))
-			copy(regularArgs, args[1:firstFlagIndex])
-
-			var flagArgs []string
-			if terminatorIndex > -1 {
-				flagArgs = args[firstFlagIndex:terminatorIndex]
-				regularArgs = append(regularArgs, args[terminatorIndex:]...)
-			} else {
-				flagArgs = args[firstFlagIndex:]
-			}
-			// separate combined flags
-			if c.UseShortOptionHandling {
-				var flagArgsSeparated []string
-				for _, flagArg := range flagArgs {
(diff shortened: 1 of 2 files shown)


command.go
```diff
@@ -233,6 +206,77 @@ func (c Command) Run(ctx *Context) (err error) {
...
+	}
+	if len(ctx.Args()) > 0 && !strings.HasPrefix(ctx.Args()[0], "-") && firstFlagIndex == -1{
+		return -1, -1
+	}
+
+	return firstFlagIndex, terminatorIndex
+
+}
+
+func copySlice(slice []string, start, end int) []string{
```
jszwedko: This should be equivalent to `foo = slice[start:end]`. I'd recommend dropping this method and using the slice "slicing".

command.go
```diff
@@ -233,6 +206,77 @@ func (c Command) Run(ctx *Context) (err error) {
...
+	var newSlice []string
```
jszwedko: I think something like

```
newSlice := make([]string, end - start)
copy(newSlice, slice[start:end)
return newSlice
```

would be a bit more performant. Maybe also call this `copyStringSlice`?

command.go
```diff
@@ -233,6 +206,77 @@ func (c Command) Run(ctx *Context) (err error) {
...
+	for i := start; i < end; i++ {
+		newSlice = append(newSlice, slice[i])
+	}
+	return newSlice
+}
+
+func getAllArgs(ctx *Context, firstFlagIndex, terminatorIndex int) (Args, []string) {
```
jszwedko: It seems like this method only needs the args rather than the full `Context`. I think it could also benefit from some commenting, naming, or both of the return values. I think the return values should either both be `[]string` or `Args`.

command.go
```diff
@@ -115,57 +115,30 @@ func (c Command) Run(ctx *Context) (err error) {
 		return err
 	}
 	set.SetOutput(ioutil.Discard)
-
+	firstFlagIndex, terminatorIndex := getIndexes(ctx)
```
jszwedko: I liked that you started pulling this logic into functions. What do you think about also pulling this whole flag handling logic into a separate function? It'd allow us to do things like return early to reduce nesting as well as compartmentalize this handling.

command.go
baude: Ok fixed

command.go
baude: Fixed, let me know if not verbose enough

command.go
baude: I would have done more, but I couldn't logically see a path through this.  I was thinking in V2 it might be easier ?

baude:
@jszwedko I did some minor reworks addressing your comments.  Take a peek please.  I'll look for you on IRC as well.

command.go
jszwedko: 👍 we can always do it as a second pass.

command.go
```diff
@@ -115,57 +115,30 @@ func (c Command) Run(ctx *Context) (err error) {
...
-			} else {
-				err = set.Parse(append(flagArgs, regularArgs...))
-			}
+			err = set.Parse(append(flagArgs, regularArgs...))
 		} else {
 			err = set.Parse(ctx.Args().Tail())
 		}
 	} else {
-		err = set.Parse(ctx.Args().Tail())
+		if c.UseShortOptionHandling {
```
jszwedko: What do you think about pulling this up as an `else if`? Would eliminate some nesting.

I'm not super clear on what this conditional logic is doing here either. Could you add a comment or two that describes the behavior?

jszwedko:
Looking better, thanks for the changes @baude ! Left a couple of comments and this still needs to be `gofmt`ed, but I think it's getting close.

command.go
baude: done

baude:
@jszwedko gofmt'd and took the nice suggestion to bump up the conditional resulting in less nesting.  Nice catch and touch!

rhatdan:
@jszwedko PTAL

jszwedko:
@rhatdan looks great, thank you so much for addressing this regression!

PR #687, opened 2017-11-21
Don't clobber slices with EnvVar

joshuarubin:
For the "slice" flag types, if there is an `EnvVar` set, the new slice pointer clobbers any slice that may have already existed.

This change simply copies the slice by value if there is already one there.


jszwedko:
👍 good catch, thanks @joshuarubin !

PR #691, opened 2017-12-03
Refactor flag handling logic

jszwedko:
Refactor logic introduced by #686 

Was able to simplify up a little by extracting to a separate function (to allow returning early) and rewriting the flag reordering logic to be more straightforward.

Would love your thoughts if you get a moment @rhatdan

### DIFF
--- a/command.go
+++ b/command.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"flag"
 	"fmt"
 	"io/ioutil"
 	"sort"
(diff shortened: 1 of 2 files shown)


jszwedko:
This diff is a bit easier to review side-by-side than inline.

command.go
```diff
@@ -205,60 +170,65 @@ func (c Command) Run(ctx *Context) (err error) {
...
+		if readFlagValue {
+			readFlagValue = false
+			flags = append(flags, arg)
+			continue
+		}
+
+		if arg != "-" && strings.HasPrefix(arg, "-") {
+			flags = append(flags, arg)
+
+			readFlagValue = !strings.Contains(arg, "=")
```
rhatdan: Will this handle commands like

command --foo=bar argument ...

Won't argument end up being marked as a flag?

rhatdan:
@baude PTAL

command.go
jszwedko: It does, but you made me realize there wasn't an explicit test for that logic. Added in https://github.com/urfave/cli/pull/691/commits/0671b166dcacb3dc1215ba65bf986dab194581dc

baude:
@jszwedko I reviewed this today and it also fixes a bug in the logic that I missed and now have exposed in our project. And also checked and couldn't find an instance of any regression. Please merge this.

jszwedko:
Awesome, thanks for taking a look @baude and @rhatdan.

Just updating the branch, but I'll merge shortly.

PR #704, opened 2018-01-29
Fix unnecessary uses of Sprintf

dolmen:
- use strconv directly
- use concatenation for "%s%s"

### DIFF
--- a/flag.go
+++ b/flag.go
@@ -636,7 +636,7 @@ func withEnvHint(envVar, str string) string {
 			suffix = "%"
 			sep = "%, %"
 		}
-		envText = fmt.Sprintf(" [%s%s%s]", prefix, strings.Join(strings.Split(envVar, ","), sep), suffix)
+		envText = " [" + prefix + strings.Join(strings.Split(envVar, ","), sep) + suffix + "]"
 	}
 	return str + envText
 }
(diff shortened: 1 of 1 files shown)


flag.go
```diff
@@ -636,7 +636,7 @@ func withEnvHint(envVar, str string) string {
 			suffix = "%"
 			sep = "%, %"
 		}
-		envText = fmt.Sprintf(" [%s%s%s]", prefix, strings.Join(strings.Split(envVar, ","), sep), suffix)
```
jszwedko: I find the original a bit easier to read here, but I'm guessing that you feel the opposite?

flag.go
```diff
@@ -766,8 +766,8 @@ func stringifySliceFlag(usage, name string, defaultVals []string) string {
 		defaultVal = fmt.Sprintf(" (default: %s)", strings.Join(defaultVals, ", "))
 	}
 
-	usageWithDefault := strings.TrimSpace(fmt.Sprintf("%s%s", usage, defaultVal))
-	return fmt.Sprintf("%s\t%s", FlagNamePrefixer(name, placeholder), usageWithDefault)
```
jszwedko: Same here.

jszwedko:
👍 dropping type checking is one I hadn't considered.

Thanks again for the submission @dolmen !

PR #705, opened 2018-01-29
Simplify slice types

dolmen:
- Pass by value when we just forward to another function.
- Convert to bare `[]int64` when calling `Sprintf` to simplify `Sprintf` work.

### DIFF
--- a/flag.go
+++ b/flag.go
@@ -270,18 +270,18 @@ func (f *Int64Slice) Set(value string) error {
 }
 
 // String returns a readable representation of this value (for usage defaults)
-func (f *Int64Slice) String() string {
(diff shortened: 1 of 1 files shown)


jszwedko:
@dolmen thanks for this PR! I'm curious what impact you see this having on the user of the library. Or do you just believe this to be cleaner internally?

This does have some slight performance implications with the additional memory allocations, though I'm not overly concerned with that due to the fact that this is a CLI library.

dolmen:
There is no increase of memory allocation: this is just zero-cost type casting, not conversions that require allocation. Proof: https://play.golang.org/p/PNXeTENui0b

lynncyrin:

PR #712, opened 2018-02-10
Fix args reordering when bool flags are present

windler:
# Bug description
Args reordering possibly leads to wrong args when `BoolFlags` are present.

# How to reproduce
Define some Command "`cmd`":
```go
Flags: []Flag{
     StringFlag{Name: "option", Value: "", Usage: "some option"},
     BoolFlag{Name: "boolflag", Usage: "some bool"},
}
```
Call the command like the following:
`cmd my-arg --boolflag --option my-option`

# Cause
Args reordering expects the next arg to be a value when no "`=`" is present. Providing a `BoolFlag`, the next args is not a value.

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -329,6 +329,39 @@ func TestApp_CommandWithArgBeforeFlags(t *testing.T) {
 	expect(t, firstArg, "my-arg")
 }
 
+func TestApp_CommandWithArgBeforeBoolFlags(t *testing.T) {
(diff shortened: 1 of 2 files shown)


jszwedko:
Good catch, thank you for addressing and adding a test case!

petemoore:
Unfortunately, `git bisect` tells me that this commit is what broke [our tests](https://github.com/taskcluster/taskcluster-lib-artifact-go/pull/5/files).

PR #717, opened 2018-02-22
Ensure multiple line description strings are indented correctly.

jservice-rvbd:
Previously, if you set a description with multiple lines only the first line after `DESCRIPTION` within the help text would be indented, which looks messy.

This PR solves this problem by splitting the description variable on new lines and ranging over them within the help text templates.


lynncyrin:

PR #718, opened 2018-02-22
Ensure App provides Reader, Writer and ErrWriter fields.

jservice-rvbd:
This allows users to easily use Reader/Writer/ErrWriter in preference to os.Stdin/os.Stdout/os.Stderr, and is helpful in testing applications which may read from stdin.


lynncyrin:

PR #720, opened 2018-02-24
Expose a constructor for Map Input Source

sargun:
This makes it so people can more easily write their own sources
on top of altsrc. In addition, it changes the struct from map[interface{}]
to map[string] as all keys must be strings.


lynncyrin:

PR #735, opened 2018-04-14
Add app-wide support for combining short flags

rliebz:
This change allows an entire application to support short-flag combining, rather than just individual commands. Setting the following will enable on the application's root action, as well as all commands and sub commands:

```go
app := cli.NewApp()
app.UseShortOptionHandling = true
```

I also added a bit of context to the documentation, and cleaned up a couple of related functions.


command_test.go
```diff
@@ -315,5 +315,6 @@ func TestCommandSkipFlagParsing(t *testing.T) {
 		err := app.Run(c.testArgs)
 		expect(t, err, c.expectedErr)
 		expect(t, args, c.expectedArgs)
+		expect(t, value, "")
```
rliebz: Looking at some of the other PRs that have started failing recently, this line is needed to fix a `go vet` failure that was likely introduced with [an update to `vet` in Go 1.10](https://golang.org/doc/go1.10#vet).

rliebz:
@jszwedko @meatballhat — Let me know if there's anything I can do to help get this PR through.

asahasrabuddhe:
Thanks for this. I think this is going to be a great addition to this library

README.md
```diff
@@ -921,6 +921,76 @@ func main() {
...
+If your program has any number of bool flags such as `serve` and `option`, and
+optionally one non-bool flag `message`, with the short options of `-s`, `-o`,
+and `-m` respectively, setting `UseShortOptionHandling` will also support the
+following syntax:
+
+```
+$ cmd -som "Some message"
+```
+
+If you enable the `UseShortOptionHandling`, then you must not use any flags that
```
asahasrabuddhe:
```suggestion
If you enable `UseShortOptionHandling`, then you must not use any flags that
```

parse.go
```diff
@@ -0,0 +1,80 @@
...
+			return nil, err
+		}
+
+		err = set.Parse(args)
+		if !ip.useShortOptionHandling() || err == nil {
+			return set, err
+		}
+
+		errStr := err.Error()
+		trimmed := strings.TrimPrefix(errStr, "flag provided but not defined: ")
```
AudriusButkevicius: This is super hacky. Can't we return a:
```
type FlagNotFoundError error {
    error
    flag string
}
```
and then do a type assertion here, having access to the field member that we need?

AudriusButkevicius:
Do we know if this clashes with: https://github.com/urfave/cli/pull/808/files#diff-7d1d20bb38d1a5a83da81f4df47018c9R186  that got merged recently?

parse.go
rliebz: That error is coming straight from the standard library `flag` package, which does not offer anything in the way of sentinel errors or custom types: https://golang.org/pkg/flag/#FlagSet.Parse

If you dig through the source, you'll eventually find this is what we're dealing with in the underlying `parseOne` function (which I can't figure out how to link to directly): 
```go
f.failf("flag provided but not defined: -%s", name)
```

Since that's just wrapping `fmt.Errorf`, I don't think there's anything we can do better here.

rliebz:
@AudriusButkevicius - From what I can tell, that PR doesn't support combined short options, which I'm not sure is necessarily common to begin with. 

Since this is extending the existing short flag support, and the new use case of top-level short flag combinations should be consistent with the experience of commands, I don't think there should be any issue merging this in as well.

rliebz:
Thanks for the reviews!

Let me know if there's anything else I need to do to get this merged

lynncyrin:
Feel free to merge it whenever 👍

PR #738, opened 2018-04-20
Allow zsh completion to be distributed like bash

zjs:
The readme describes a distribution pattern for bash_autocomplete. In
this pattern, the file is renamed to match the name of the program for
which the  autocompletion support is being added. When following this
pattern, it is not necessary to set the PROG environment variable.

Update zsh_autocomplete so that it can be distributed the same way.


gnowxilef:
+1 bump

lynncyrin:
I'm going to close this specific PR since it's a year old - but it looks like it's solving the same problem as you describe in https://github.com/urfave/cli/issues/810 @gnowxilef?

PR #739, opened 2018-04-27
Add flag error handling to App

crosbymichael:
This allows the user to set how flag parsing should work.  By default,
the common flag const are not useful in the context of cli but if you
want to clear it by setting something like `flag.ContinueOnError -1`
then you don't get the invalid flag error.

This is useful for multiple projects implementing the same CLI interface
but do not define all the flags because they don't apply.

Signed-off-by: Michael Crosby <crosbymichael@gmail.com>

### DIFF
--- a/app.go
+++ b/app.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"flag"
 	"fmt"
 	"io"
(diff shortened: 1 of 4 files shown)


lynncyrin:

PR #741, opened 2018-05-02
Exposed the `value` accessor in `Context`

corruptmemory:
I wrote a small utility to populate a `struct` reflectively from command-line-arguments, but to do this I needed to get at the values stored via `value`.

### DIFF
--- a/context.go
+++ b/context.go
@@ -114,7 +114,7 @@ func (c *Context) Lineage() []*Context {
 	return lineage
 }
 
-// value returns the value of the flag corresponding to `name`
+// Value returns the value of the flag corresponding to `name`
 func (c *Context) Value(name string) interface{} {
 	return c.flagSet.Lookup(name).Value.(flag.Getter).Get()
 }
--- a/context.go
+++ b/context.go
@@ -115,7 +115,7 @@ func (c *Context) Lineage() []*Context {
 }
 
 // value returns the value of the flag corresponding to `name`
-func (c *Context) value(name string) interface{} {
+func (c *Context) Value(name string) interface{} {
 	return c.flagSet.Lookup(name).Value.(flag.Getter).Get()
 }
 
--- a/context_test.go
+++ b/context_test.go
@@ -3,8 +3,8 @@ package cli
 import (
 	"context"
 	"flag"
-	"sort"
 	"os"
+	"sort"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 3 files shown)


AudriusButkevicius:
Not sure why? I've written something simillar and not sure I needed that.

command_test.go
```diff
@@ -305,7 +304,6 @@ func TestCommandSkipFlagParsing(t *testing.T) {
 					},
 					Action: func(c *Context) {
 						fmt.Printf("%+v\n", c.String("flag"))
-						value = c.String("flag")
```
AudriusButkevicius: This does exercise the getter, why are we removing this?

AudriusButkevicius:
actually, it's not a unreasonable change.

I am using reflection on the destination, and then based on that I pick the "getter", so it's reasonable to want to do the inverse.

corruptmemory:
Nice to see some life coming to the PR.  I'll rebase and push sometime this weekend.

lynncyrin:
Oh no, this PR has been assailed by merge conflicts! @corruptmemory can ya take care of that?

lynncyrin:
👋 heya, I think I'll close this within the next _~week_ if you don't have time for it @corruptmemory! You are very much encouraged to re-open it later, though 🙏

corruptmemory:
Sorry for the late response.  I'll clean this up this week.

corruptmemory:
I see that there is now a collision between calling `(*context.Context).Value(interface{})` and `(*cli.Context).Value(string)` and the second one is winning causing the NPE.

This change [in line 331 in `context_test.go`](https://github.com/corruptmemory/cli/blob/expose-value/context_test.go#L331) seems to address the issue:

```go
val := ctx.Context.Value("key")
```

Napas:
@lynncyrin This PR introduces breaking changes.

`Value(string) interface{}` overrides `Value(interface{}) interface{}` from the `context.Context` and `cli.Context` stops implementing `context.Context` interface.

Please revert it ASAP.

asahasrabuddhe:
@Napas Are you using a dependency management system? I think not. I would suggest that pin to a release prior to `v2.1.0` instead of getting the latest code directly from the `master`.

Napas:
> @Napas Are you using a dependency management system? I think not. I would suggest that pin to a release prior to `v2.1.0` instead of getting the latest code directly from the `master`.

I am doing that until it will be reverted, but minor version release should not have breaking changes.

On top of that, it makes testing much harder as `cli.Context` can't have mocked and used as `context.Context` downstream.

lynncyrin:
I'll revert this now, but just so it's 100% clear - there's no specific reason why reverting this should be blocked on me. Anyone can submit a reversion PR.

lynncyrin:
@Napas @asahasrabuddhe => https://github.com/urfave/cli/pull/1039

lynncyrin:
> I am doing that until it will be reverted, but minor version release should not have breaking changes

Just in case anyone isn't aware - by no means did anyone involved in this PR merge a breaking change on purpose.

PR #745, opened 2018-05-19
2018

marfi:


AudriusButkevicius:
There is no need to update copyright by law, the code is copyright from the year it was produced.

PR #761, opened 2018-07-25
fix ExitErrHandler not used with subcommands

### LINKED ISSUE #736: Errors returned by subcommands are ignored by ExitErrHandler
When a subcommand receiver outputs an error, app.ExitErrHandler is not called.

```go
func main() {
	app := cli.NewApp()
	app.Commands = []cli.Command{
		{
			Name:        "firstlevel",
			Action:      firstlevel,
		},
		{
			Name:        "parent",
			Subcommands: []cli.Command{
				{
					Name:        "sub1",
					Action:      sub1,
				}, {
					Name:        "sub2",
					Aliases: []string{"d"},
					Action:      sub2,
				},
			},
		},
	}

	app.ExitErrHandler = func(context *cli.Context, err error) {
		fmt.Println(err)
	}

	app.Run(os.Args)
}

// [...] all receivers return an error for this example
```

When an error is returned on firstlevel `firstlevel()`, it gets printed. However, if either sub1 or sub2 return it it is not. Not sure if this is a bug or I'm trying to do some weird things, but some insights would be appreciated anyway :)

travisperson:
Errors returned by subcommands no longer bypass the toplevel ExitErrHandler

Fixes #736


jkanywhere:
AppVeyer build is failing because `go vet` 
```
python runtests vet
45# github.com/urfave/cli
46.\command_test.go:328:3: value declared but not used
47vet: typecheck failures
48runtests: go vet ./...
...
```
which is odd because `value` defined on [line 328](
https://github.com/urfave/cli/pull/761/files#diff-2fc2009ba1969a36b69136d7fb7b2072R328) is used on [line 340](/urfave/cli/pull/761/files#diff-2fc2009ba1969a36b69136d7fb7b2072R340), and this diff doesn't alter that part of the file.

phinnaeus:
@jkanywhere It's failing on master, without these changes at all. I created https://github.com/urfave/cli/pull/765 to fix that.

HenrikPoulsen:
This is still an issue. What is preventing us from merging this in at this point?

Issue #340, opened 2016-03-05
Is it possible to generate static command documentation?


bmatsuo:
I would like to have static documentation for my cli command. More specifically, I would like to have a link to HTML documentation for the command.  But I think there are some problems to using godoc comments (and godoc.org) for this.

Does the cli package recommend some way other than godoc.org to achieve static docs?

These are the problems that I see with using godoc.org for command documentation:
- In the typical case, it will only document the latest (in-development) version of the application. People can avoid this by making sure their "master" branch is stable production code. But I don't think that is the norm.
- For cli commands, go doc comments are often either vague or redundant. The command help text and flags already document the command. And manually duplicating that information into the godoc comments is a pain -- which leads to out-of-sync comments.

The first problem is solved by generating static documentation -- command documentation can be embedded in the release archive (tarball), or hosted using something like github.io. The second problem could potentially be solved by having a "godoc" output option for the doc generator that spits out a `doc.go` file with command documentation.

(edit: I'm willing to help implement such a solution. But it sounds like a fairly impressive undertaking so I wanted to run it by the maintainers first)

jszwedko:
I think this could also be addressed by adding the ability to generate `man` page documentation -- would that fit with what you are thinking?

bmatsuo:
@jszwedko generating `man` pages would certainly be a start if that is possible. But I think I am looking for ideally would be a more extensible solution. IMO `man` is a little archaic. And I don't know that it would be ideal on Windows (though I'm sure it probably does/can work).

I imagine that a set of `man` pages can be transformed into HTML using some existing tools. But I doubt the results look spectacular. And I am looking for something that produces reasonably modern looking HTML. Please point out anything you are aware of to do that @jszwedko :smile: 

I imagine that the primary documentation reference for my project will be online. I intend on hosting versioned copies of generated documentation (so v1 docs at github.io/bmatsuo/foo/v1, v2 docs at github.io/bmatuso/foo/v2, etc). And I would like to style these docs to integrate with the project website as a whole.

I think something like sphynx doc generation may be more what I'm looking for. But from what I know I don't really like the look of sphynx generated sites.

I was thinking something along the lines of a Go interface that could could be implemented to generate `man` page documents, sphynx style docs, markdown, or raw HTML.

yepher:
I could see a MarkDown output that then could be fed into any number of tools to generate PDF, DOC or any number of other outputs.

jszwedko:
[...]

PR #769, opened 2018-09-18
Move templates to their own files and load via vfsgen

jribe:
As I mentioned in https://github.com/urfave/cli/issues/340, writing markdown template strings in go source is a huge pain because there's no way to escape backticks in a multiline string. This change replaces the existing template strings with file that are read into a virtual filesystem by vfsgen. I split this into its own change since it's a little bit orthogonal to markdown support.

~~This has WIP status right now because I can't get Travis to pass on my fork because the repo path is different.~~ Travis passed on the PR. I'm reasonably confident the existing tests cover these changes.


lynncyrin:
This is great!!!

I'm very interested in getting this change merged 👍 Although the diff is a bit large and I assume you probably aren't as noticed anymore @jribe? So that may be a bit hard. I want to leave this open, and see what we can do here 🔍

help.go
```diff
@@ -7,77 +7,35 @@ import (
...
-   {{.Category}}{{end}}{{if .Description}}
-
-DESCRIPTION:
-   {{.Description}}{{end}}{{if .VisibleFlags}}
-
-OPTIONS:
-   {{range .VisibleFlags}}{{.}}
-   {{end}}{{end}}
-`
+var CommandHelpTemplate = getTemplate("/commandHelp.gotxt")
```
AudriusButkevicius: The comments are sort of obsolete now

templates_generate.go
```diff
@@ -0,0 +1,53 @@
...
+	"os"
+	"time"
+
+	"github.com/shurcooL/vfsgen"
+)
+
+// zeroModTimeFileSystem is an http.FileSystem wrapper.
+// It exposes a filesystem exactly like Source, except
+// all file modification times are changed to zero.
+// See https://github.com/shurcooL/vfsgen/pull/40#issuecomment-355416103
```
AudriusButkevicius: Why do we not want mtime?

asahasrabuddhe: We do not need modified time because if the modified time changes but there is no content change, it would cause vfsgen to output a new file once more. This can lead to additional files in commits etc. Mod time isn't even relevant to our use case so I think this is the correct approach.

AudriusButkevicius: But the new file would be of the same content hence where would not be a diff?

asahasrabuddhe: this would probably save us from generating the same stuff again.

AudriusButkevicius: If the mtime is the same why would it generate it again? If the mtime is different its likely different, if its not, there will be no diff hence nothing to commit.

asahasrabuddhe: agreed. let's remove this

lynncyrin:
Ping @jribe, let us know if you're still interested in this PR? If not (and that's ok!) I'll close it

lynncyrin:
To expound - I'm still interested in this change! I would be closing the PR so that anyone else interested can open their own PR.

lynncyrin:
Closing pending user feedback!

PR #776, opened 2018-10-19
Bring Go version current

gliptak:


gliptak:
@meatballhat please review

AudriusButkevicius:
What material impact does this have, why is it important to do this? I am not disagreeing that it's the right thing to do, but I wonder why people care about this.

gliptak:
IMHO keeping up with (patch and/or compatible) compiler/runtime releases helps with bugs and improves performance. For a library, compilation with the last three releases ```Go``` might be a good goal per https://golang.org/doc/devel/release.html

PR #777, opened 2018-10-19
Bump Travis versions

HaraldNordgren:


HaraldNordgren:
@AudriusButkevicius @jszwedko Ping

AudriusButkevicius:
Why does this matter?

HaraldNordgren:
@AudriusButkevicius It serves to prove that the project builds on newer Go versions and not just Go 1.8.

AudriusButkevicius:
Go is backwards compatible. So I am not sure why do we need to prove this to anyone?

I don't mind the change, yet I don't see the material impact of this or the importance of this.

AudriusButkevicius:
Done by a different pr

PR #779, opened 2018-10-29
Modify Context struct to add context.Context

sjug:
By adding a `context.Context` to the main cli.Context struct we can add compatibility for the [OpenTracing API](https://opentracing.io/guides/golang/). This gives people a good starting point to conveniently implement OpenTracing with their CLI applications that use this library.

### DIFF
--- a/context.go
+++ b/context.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"context"
 	"errors"
 	"flag"
 	"os"
@@ -20,6 +21,7 @@ type Context struct {
 	flagSet       *flag.FlagSet
 	setFlags      map[string]bool
 	parentContext *Context
+	Ctx           context.Context
 }
 
 // NewContext creates a new context. For use in when invoking an App or Command action.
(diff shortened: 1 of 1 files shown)


txgruppi:
My concern is that adding `context.Context` will give an idea that timeout and cancellation is supported, which is not true at the moment.

What do you think?

AudriusButkevicius:
The handlers implementes could just capture the context from the outer scope? Why does it need to be passed down via the library?

sjug:
@txgruppi I hadn't considered that people would think that, maybe we can add a comment? Considering the field nor any goroutines are used anywhere in the library I don't know that there would be much expectation for it to do more.

@AudriusButkevicius  I'm not entirely clear on what you're suggesting. If the context is initialized in main, there is no way to pass it through to the library to the various cli.Commands.

AudriusButkevicius:
In only makes sense to have a single context in a cli, in which case it might as well be global, or captured by action closures when being defined in main?

txgruppi:
I agree with Audrius.

So far there is no need for a way to manage `context.Context` in a cli app.

Examples:

1. you could create a package with `DefaultContext` similar to how `net/http` has a `DefaultClient` and use as the context for you app.

2. You can create each command as a struct that implements `Run() error` and have public fields that will get populated right before calling `Run() error`, one of these fields can be the `context.Context`.

sjug:
Thanks for your time and feedback @AudriusButkevicius and @txgruppi, closing.

PR #782, opened 2018-11-07
skip before and after actions if shellComplete flag is set

jackzzheng:
I think we should skip before and after actions if we are doing tab complete.


lynncyrin:
ping @jackzzheng, can you explain this PR a bit more? If not (and that's ok!) I'll close it.

jackzzheng:
hi @lynncyrin 
sorry for the late reply, the reason for this change was we had before and after hooks defined in our cli, which made calls to external services via an api call. It seemed unnecessary if the cli has to make an api call for tab complete.
i didn't make it return early because that felt like it might skip part of the core code that i had very little knowledge in, so i wanted to minimize the footprint of my change.

AudriusButkevicius:
I think this change is actually ok, just needs to be brought up to date.

AudriusButkevicius:
Actually, I don't know how shell completion gets here, perhaps it shouldn't and that should be the fix.

lynncyrin:
> it seems like shell completion should early return rather than skipping particular lines?
> 
> Actually, I don't know how shell completion gets here, perhaps it shouldn't and that should be the fix.

These are both describing the same idea, yeah? If so, I can write up an issue about it! @AudriusButkevicius

lynncyrin:
👋 hello @jackzzheng! It's been about a month since the last update for this PR, do you still plan on working on it?

lynncyrin:
Closing due to age 🕙 feel free to re-open this if you're still interested!

Issue #711, opened 2018-02-09
Alternate input sources YAML doesn't work when flag includes alternate name


wcwynn:
Alternate names are described here: https://github.com/urfave/cli#alternate-names
Support for YAML is described here: https://github.com/urfave/cli#values-from-alternate-input-sources-yaml-toml-and-others

When any alternate source flag includes an alternate name, the alternate source input doesn't function as it should (i.e. it doesn't use the alternate input source data).

Code to reproduce:
```golang
package main

import (
    "fmt"
    "os"

    "github.com/urfave/cli/altsrc"
    "gopkg.in/urfave/cli.v1"
)

func main() {
    app := cli.NewApp()

    flags := []cli.Flag{
        /*  
            // OK
            altsrc.NewIntFlag(cli.IntFlag{Name: "test", Value: -1}),
            altsrc.NewStringFlag(cli.StringFlag{Name: "flag", Value: "default"}),
        */

        // NG #1
        altsrc.NewIntFlag(cli.IntFlag{Name: "test, t", Value: -1}),
        altsrc.NewStringFlag(cli.StringFlag{Name: "flag", Value: "default"}),

        /*  
            // NG #2
            altsrc.NewIntFlag(cli.IntFlag{Name: "test", Value: -1}),
            altsrc.NewStringFlag(cli.StringFlag{Name: "flag, f", Value: "default"}),

            // NG #3
            altsrc.NewIntFlag(cli.IntFlag{Name: "test, t", Value: -1}),
            altsrc.NewStringFlag(cli.StringFlag{Name: "flag, f", Value: "default"}),
        */

        cli.StringFlag{Name: "load, l"},
    }   

    app.Action = func(c *cli.Context) error {
        fmt.Println("yaml ist rad")
        fmt.Printf("load: %s\n", c.String("load"))
        fmt.Printf("test: %d\n", c.Int("test"))
        fmt.Printf("flag: %s\n", c.String("flag"))
        return nil 
    }   

    app.Before = altsrc.InitInputSourceWithContext(flags, altsrc.NewYamlSourceFromFlagFunc("load"))
    app.Flags = flags

    app.Run(os.Args)
}
```
Test file contents:
```
iMac:buggy chris$ cat buggy.txt
test: 123
flag: excellent!
iMac:buggy chris$
```

Output:
```
iMac:buggy chris$ go run buggy.go --load buggy.txt
yaml ist rad
load: buggy.txt
test: -1
flag: excellent!
iMac:buggy chris$
[...]

PR #791, opened 2019-01-08
support for getting alternate name flag values from yaml/toml/json

pysqz:
fix issue #711. please check~


mdbraber:
It seems that this fix doesn't work yet with `NewStringSliceFlag`. Any ideas what could be changed?

pysqz:
Repush a patch #792. It can fix this bug with StringSliceFlag and IntSliceFlag. Please review again~

PR #792, opened 2019-01-15
support for getting alternate name flag values from other file input …

pysqz:
…sources(yaml/toml/json)


pysqz:
refix issue #711. please check again~

lynncyrin:
Hiya! There's a known issue with the current implementation of the CLI, where altsrc is generally clunky and poorly documented. My current idea is that I'm going to move all of the `altsrc` functionality into the main package in some way, and that's how all of the _"altsrc is weird and different???"_ issues people are having will get resolved. I'm tracking that work here => https://github.com/urfave/cli/issues/833

Issue #800, opened 2019-02-03
Since one possibly hijacks their "Before" by using altsrc we should have another statement run after the context is loaded but after Before


johnwyles:
As you can see in a few examples floating around the web the `Before` method is sort of _hijacked_ by the use of `altsrc` in loading YAML or TOML:

- https://github.com/urfave/cli/issues/793
- https://github.com/urfave/cli/issues/570

This leaves other initiation you may want to perform that perhaps is some preparation for your entire command and subsequent subcommands (for example setting up generic logging or setting some values all subcommands will need).  If you are not loading configuration using `altsrc` then `Before` is just fine for your use case.  However in the event you _do_ want to load configuration from YAML or TOML you are left with your `Before` held hostage with no option to perform any other `Before` actions.  Thus I am proposing a post-`Flags` and post-`Before` directived I've dubbed `Prepare`.

The following code from how you would do this currently is pretty standard in your searches around the documentation and the web for examples of loading configuration. Note that the `Before` method is "hostage" (e.g. it can no longer do anything else) to this process of loading configuration leaving you little option to perform any other `Before` actions you may want to take:

With the YAML file `file.yml`:
```yaml
test: 10
```

In previous `test.go`:
```go
package main

import (
        "fmt"
        "os"

        "gopkg.in/urfave/cli.v1"
        "gopkg.in/urfave/cli.v1/altsrc"
)

func main() {
        app := cli.NewApp()

        flags := []cli.Flag{
                altsrc.NewIntFlag(cli.IntFlag{Name: "test"}),
                cli.StringFlag{Name: "load"},
        }

        app.Action = func(c *cli.Context) error {
                fmt.Println("action")
                fmt.Println(c.Int("test"))
                return nil
        }

        app.Before = altsrc.InitInputSourceWithContext(flags, altsrc.NewYamlSourceFromFlagFunc("load"))

        app.Flags = flags

        app.Run(os.Args)
}
```

Output:
```shell
$ go run test.go --load file.yaml
action
10
```

Instead I have proposed the use of a `Prepare` directive that will run after `Before` so you can still load configuration and then do something with everything loaded.  Take the above example modified with this new `Prepare` directive:

The file `test.go`:
```go
package main

import (
        "fmt"
        "os"

        "gopkg.in/urfave/cli.v1"
        "gopkg.in/urfave/cli.v1/altsrc"
)

func main() {
        app := cli.NewApp()

        flags := []cli.Flag{
[...]

Issue #793, opened 2019-01-18
V2: Making `altsrc` flag optional in `Before:`


johnwyles:
Is there a way we can get an `altsrc` wrapper function to accomplish the following? Given that all of these features (env variables, flags, and config files) all override each other you do not want to mandate a configuration file be specified and instead make it OPTIONAL that it is specified:

In `test.go`:
```go
package main

import (
	"fmt"
	"os"

	cli "gopkg.in/urfave/cli.v2"
	"gopkg.in/urfave/cli.v2/altsrc"
)

func main() {
	flags := []cli.Flag{
		altsrc.NewStringFlag(
			&cli.StringFlag{
				Name:  "foo",
				Value: "BAR",
			},
		),
		&cli.StringFlag{
			Name: "load",
		},
	}

	app := &cli.App{

		Action: func(c *cli.Context) error {
			fmt.Printf("load: %s\n", c.String("load"))
			fmt.Printf("foo: %s\n", c.String("foo"))
			return nil
		},

		// This is seen all over and works but REQUIRES you provide "--load" and a TOML file
		// Before: altsrc.InitInputSourceWithContext(flags, altsrc.NewTomlSourceFromFlagFunc("load")),

		// Instead something like this would be desired:
		Before: altsrc.InitInputSourceWithContext(
			flags,
			func(context *cli.Context) (altsrc.InputSourceContext, error) {
				if context.IsSet("load") {
					return altsrc.NewTomlSourceFromFile(context.String("load"))
				}

				return &altsrc.MapInputSource{}, nil
			},
		),

		Flags: flags,
	}

	if err := app.Run(os.Args); err != nil {
		fmt.Errorf("An error occurred: %#v", err)
		os.Exit(1)
	}
}
```

In `test.toml`:
```toml
foo = "FOO"
```

And the above `test.go` will work as expected:

```shell
# Works as expected:
$ go run test.go
load:
foo: BAR

# Works as expected:
$ go run --foo OOF
load:
foo: OOF

# Works as expected:
$ go run test.go --load test.toml --foo OOF
load: test.toml
foo: OOF

# Works as expected:
$ go run test.go --load test.toml
load: test.toml
foo: FOO
```

johnwyles:
I've actually run *further* into problems in that I would like to do more initialization in `Before:` after the configuration is loaded, but still before additional commands are loaded, so this is something I think that could be useful.

johnwyles:
[...]

PR #802, opened 2019-02-04
V1: Issue #800: Adding Prepare directive which runs after Before but before Action

johnwyles:
This is for https://github.com/urfave/cli/issues/800

### DIFF
--- a/README.md
+++ b/README.md
@@ -1334,6 +1334,10 @@ func main() {
         fmt.Fprintf(c.App.Writer, "brace for impact\n")
         return nil
       },
+      Prepare: func(c *cli.Context) error {
+        fmt.Fprintf(c.App.Writer, "we've almost made it!\n")
+        return nil
+      },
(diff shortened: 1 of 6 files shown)


johnwyles:
@meatballhat Are you able to review and merge this? Associated issues are: https://github.com/urfave/cli/issues/793 https://github.com/urfave/cli/issues/800

lynncyrin:
Hiya! I see that you put a bunch of work into this, and I appreciate that you were so inspired 🙏 I want to mention though, that none of the maintainers acknowledged that we want this feature - so I don't think it's appropriate for this PR to be on our plates. Also relevant here (and I mention this in https://github.com/urfave/cli/issues/800) is the fact that I don't totally understand the feature request, and without that understanding I definitely can't review a PR.

lynncyrin:
I'm going to close this, actually! We can talk about this change request in this issue https://github.com/urfave/cli/issues/800

PR #808, opened 2019-03-21
Add support for flags bash completion

### LINKED ISSUE #188: Add completion for flags
I am trying to work on the auto-completion part using this framework, and it does not work with the flags option. Here is my piece of code:

package main

import (
    "fmt"
    "os"

```
"github.com/codegangsta/cli"
```

)

func main() {
    app := cli.NewApp()
    app.Name = "greet"
    app.Usage = "sample command-line app by greet"
    app.Author = "abc"
    app.Email = "xyz@aaa.com"
    app.Commands = []cli.Command{
        {
            Name:      "read",
            ShortName: "r",
            Usage:     "read something",
            Subcommands: []cli.Command{
                {
                    Name:   "articles",
                    Usage:  "read articles",
                    Action: readArticles,
                },

```
            {
                    Name:   "tweets",
                    Usage:  "read Tweets",
                    Flags: []cli.Flag{
                            cli.StringFlag{
```

Name: "account",
                                        Value: "SomeThing",
                                        Usage: "name of Twitter account",
                                },
                        },
                        Action: readTwitter,
                },
            },
        },
    }
    app.Run(os.Args)
}

func readArticles(ctx *cli.Context) {
    fmt.Println("Go to http://www.google.com to read articles!")
}

func readTwitter(ctx *cli.Context) {
        fmt.Printf("Go to https://twitter.com/%s to read tweets!", ctx.String("account"))
}

When i try to do an autocomplete on the flags, it does not work. Could anybody let me know if this is supported by the framework?

Example:

./greet read tweets --a [TAB][TAB] does not work.

yogeshlonkar:
Fixed #188 

Short and long flags will be generated with `--generate-bash-completion` when last argument is hyphen (`-`) or double hyphen (`--`).

If flag is already provided in cli args it will not be generated. Also for double hyphens single character flags will not be generated.

Adds default BashCompletion for Commands and SubCommands.

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -221,6 +221,60 @@ func ExampleApp_Run_subcommandNoAction() {
 
 }
 
+func ExampleApp_Run_bashComplete_withShortFlag() {
+	os.Args = []string{"greet", "-", "--generate-bash-completion"}
+
+	app := NewApp()
+	app.Name = "greet"
+	app.EnableBashCompletion = true
+	app.Flags = []Flag{
+		IntFlag{
+			Name: "other,o",
+		},
+		StringFlag{
+			Name: "xyz,x",
+		},
+	}
+
+	app.Run(os.Args)
+	// Output:
+	// --other
+	// -o
+	// --xyz
+	// -x
+	// --help
+	// -h
+	// --version
+	// -v
+}
+
+func ExampleApp_Run_bashComplete_withLongFlag() {
+	os.Args = []string{"greet", "--s", "--generate-bash-completion"}
+
+	app := NewApp()
+	app.Name = "greet"
+	app.EnableBashCompletion = true
+	app.Flags = []Flag{
+		IntFlag{
+			Name: "other,o",
+		},
+		StringFlag{
+			Name: "xyz,x",
+		},
+		StringFlag{
+			Name: "some-flag,s",
+		},
+		StringFlag{
+			Name: "similar-flag",
+		},
+	}
+
+	app.Run(os.Args)
+	// Output:
+	// --some-flag
+	// --similar-flag
+}
+
 func ExampleApp_Run_bashComplete() {
 	// set args for examples sake
 	os.Args = []string{"greet", "--generate-bash-completion"}
--- a/app_test.go
+++ b/app_test.go
@@ -274,6 +274,35 @@ func ExampleApp_Run_bashComplete_withLongFlag() {
 	// --some-flag
 	// --similar-flag
 }
+func ExampleApp_Run_bashComplete_withMultipleLongFlag() {
+	os.Args = []string{"greet", "--st", "--generate-bash-completion"}
+
+	app := NewApp()
+	app.Name = "greet"
+	app.EnableBashCompletion = true
+	app.Flags = []Flag{
+		IntFlag{
+			Name: "int-flag,i",
+		},
+		StringFlag{
+			Name: "string,s",
+		},
+		StringFlag{
+			Name: "string-flag-2",
+		},
+		StringFlag{
+			Name: "similar-flag",
+		},
+		StringFlag{
+			Name: "some-flag",
+		},
+	}
+
+	app.Run(os.Args)
+	// Output:
+	// --string
+	// --string-flag-2
+}
 
 func ExampleApp_Run_bashComplete() {
 	// set args for examples sake
--- a/autocomplete/bash_autocomplete
+++ b/autocomplete/bash_autocomplete
@@ -3,6 +3,7 @@
 : ${PROG:=$(basename ${BASH_SOURCE})}
 
 _cli_bash_autocomplete() {
+  if [[ "${COMP_WORDS[@]:0:$COMP_CWORD}" != "source" ]]; then
     local cur opts base
     COMPREPLY=()
     cur="${COMP_WORDS[COMP_CWORD]}"
@@ -13,8 +14,8 @@ _cli_bash_autocomplete() {
     fi
     COMPREPLY=( $(compgen -W "${opts}" -- ${cur}) )
     return 0
+  fi
 }
 
-complete -F _cli_bash_autocomplete $PROG
-
+complete -o bashdefault -o default -o nospace -F _cli_bash_autocomplete $PROG
 unset PROG
--- a/autocomplete/bash_autocomplete
+++ b/autocomplete/bash_autocomplete
@@ -7,7 +7,7 @@ _cli_bash_autocomplete() {
     local cur opts base
     COMPREPLY=()
     cur="${COMP_WORDS[COMP_CWORD]}"
-    if [[ $cur == -* ]]; then
+    if [[ "$cur" == "-"* ]]; then
       opts=$( ${COMP_WORDS[@]:0:$COMP_CWORD} ${cur} --generate-bash-completion )
     else
       opts=$( ${COMP_WORDS[@]:0:$COMP_CWORD} --generate-bash-completion )
--- a/autocomplete/bash_autocomplete
+++ b/autocomplete/bash_autocomplete
@@ -11,9 +11,6 @@ _cli_bash_autocomplete() {
     else
       opts=$( ${COMP_WORDS[@]:0:$COMP_CWORD} --generate-bash-completion )
     fi
-    if [[ "$opts1" == "$cur1" ]]; then
-      return 0
-    fi
     COMPREPLY=( $(compgen -W "${opts}" -- ${cur}) )
     return 0
 }
--- a/autocomplete/bash_autocomplete
+++ b/autocomplete/bash_autocomplete
@@ -3,7 +3,7 @@
 : ${PROG:=$(basename ${BASH_SOURCE})}
 
 _cli_bash_autocomplete() {
-  if [[ "${COMP_WORDS[@]:0:$COMP_CWORD}" != "source" ]]; then
+  if [[ "${COMP_WORDS[0]}" != "source" ]]; then
     local cur opts base
     COMPREPLY=()
     cur="${COMP_WORDS[COMP_CWORD]}"
--- a/autocomplete/bash_autocomplete
+++ b/autocomplete/bash_autocomplete
@@ -6,7 +6,14 @@ _cli_bash_autocomplete() {
     local cur opts base
     COMPREPLY=()
     cur="${COMP_WORDS[COMP_CWORD]}"
-    opts=$( ${COMP_WORDS[@]:0:$COMP_CWORD} --generate-bash-completion )
+    if [[ $cur == -* ]]; then
+      opts=$( ${COMP_WORDS[@]:0:$COMP_CWORD} ${cur} --generate-bash-completion )
+    else
+      opts=$( ${COMP_WORDS[@]:0:$COMP_CWORD} --generate-bash-completion )
+    fi
+    if [[ "$opts1" == "$cur1" ]]; then
+      return 0
+    fi
     COMPREPLY=( $(compgen -W "${opts}" -- ${cur}) )
     return 0
 }
--- a/help.go
+++ b/help.go
@@ -162,6 +162,22 @@ func DefaultAppComplete(c *Context) {
 }
 
 func DefaultAppCompleteWithFlags(cmd *Command) func(c *Context) {
+	cliArgContains := func(flagName string) bool {
+		for _, name := range strings.Split(flagName, ",") {
+			name = strings.Trim(name, " ")
+			count := utf8.RuneCountInString(name)
+			if count > 2 {
+				count = 2
+			}
+			flag := fmt.Sprintf("%s%s", strings.Repeat("-", count), name)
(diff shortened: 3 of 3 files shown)


yogeshlonkar:
@jszwedko Hey, can you review and merge this?

jszwedko:
Hi @yogeshlonkar. Thanks for the PR! I'm no longer an active maintainer of this project, but tagging @meatballhat who is (per https://github.com/urfave/cli/blob/master/MAINTAINERS.md).

yogeshlonkar:
@meatballhat could please review and merge?

luis-silva:
@meatballhat Anything we can do to expedite review of this PR?

asahasrabuddhe:
@lynncyrin mind taking a :eyes: here?

app_test.go
```diff
@@ -221,6 +221,60 @@ func ExampleApp_Run_subcommandNoAction() {
...
+		StringFlag{
+			Name: "similar-flag",
+		},
+	}
+
+	app.Run(os.Args)
+	// Output:
+	// --some-flag
+	// --similar-flag
+}
```
lynncyrin: This is an interesting pattern! 👍

autocomplete/bash_autocomplete
```diff
@@ -3,14 +3,19 @@
...
+      opts=$( ${COMP_WORDS[@]:0:$COMP_CWORD} --generate-bash-completion )
+    fi
     COMPREPLY=( $(compgen -W "${opts}" -- ${cur}) )
     return 0
+  fi
 }
 
-complete -F _cli_bash_autocomplete $PROG
-
+complete -o bashdefault -o default -o nospace -F _cli_bash_autocomplete $PROG
```
lynncyrin: I don't 100% follow everything that's going on here - but I think we can definitely merge and iterate on bug fixes 🔧

help.go
```diff
@@ -152,24 +154,91 @@ func ShowAppHelp(c *Context) (err error) {
 	return nil
 }
 
+var shortFlagRegex = regexp.MustCompile(`^-`)
+
 // DefaultAppComplete prints the list of subcommands as the default app completion method
 func DefaultAppComplete(c *Context) {
-	for _, command := range c.App.Commands {
+	DefaultCompleteWithFlags(nil)(c)
```
lynncyrin: `(nil)(c)` is a bit of a confusing invocation, but I don't have enough context to suggest any alternatives here 🤔

help.go
```diff
@@ -152,24 +154,91 @@ func ShowAppHelp(c *Context) (err error) {
...
+			}
+			flagCompletion := fmt.Sprintf("%s%s", strings.Repeat("-", count), name)
+			if strings.HasPrefix(name, cur) && cur != name && !cliArgContains(flag.GetName()) {
+				fmt.Fprintln(writer, flagCompletion)
+			}
+		}
+	}
+}
+
+func DefaultCompleteWithFlags(cmd *Command) func(c *Context) {
```
lynncyrin: I can't tell offhand if these functions have any test coverage? I'll investigate that a bit! But I don't want to block this PR on it ^^

autocomplete/bash_autocomplete
yogeshlonkar: Sourcing this file causes bash builtin `source` command to throw `-bash: source: --: invalid option` error on <kbd>TAB</kbd> . I was not able detect why in the first place `source` builtin is getting linked to this function. So excluded it and fallback to defaults

help.go
yogeshlonkar: Yah kinda confusing. This method generates sub commands or flags using current/last command in argv. Problem is app it self is command so it doesn’t have any thing to pass

yogeshlonkar:
@urfave/cli Can someone test this on their local and merge?

asahasrabuddhe:
@yogeshlonkar I will do this today or tomorrow in the evening depending on how much time I get. Sorry for not being able to get to this quickly as I was working on some other fixes and features.

autocomplete/bash_autocomplete
```diff
@@ -3,14 +3,19 @@
 : ${PROG:=$(basename ${BASH_SOURCE})}
 
 _cli_bash_autocomplete() {
+  if [[ "${COMP_WORDS[@]:0:$COMP_CWORD}" != "source" ]]; then
```
saschagrunert: Some shellcheck hint about this: 
`SC2199: Arrays implicitly concatenate in [[ ]]. Use a loop (or explicit * instead of @).`
See https://github.com/koalaman/shellcheck/wiki/SC2199

autocomplete/bash_autocomplete
```diff
@@ -3,14 +3,19 @@
...
     local cur opts base
     COMPREPLY=()
     cur="${COMP_WORDS[COMP_CWORD]}"
-    opts=$( ${COMP_WORDS[@]:0:$COMP_CWORD} --generate-bash-completion )
+    if [[ $cur == -* ]]; then
+      opts=$( ${COMP_WORDS[@]:0:$COMP_CWORD} ${cur} --generate-bash-completion )
```
saschagrunert: Some shellcheck hint about this:
`SC2068: Double quote array expansions to avoid re-splitting elements.`
See https://github.com/koalaman/shellcheck/wiki/SC2068

autocomplete/bash_autocomplete
```diff
@@ -3,14 +3,19 @@
...
+    else
+      opts=$( ${COMP_WORDS[@]:0:$COMP_CWORD} --generate-bash-completion )
```
saschagrunert: The same for this one :point_up:

help.go
```diff
@@ -155,24 +157,91 @@ func ShowAppHelp(c *Context) (err error) {
 	return nil
 }
 
+var shortFlagRegex = regexp.MustCompile(`^-`)
```
saschagrunert: I would recommend to put this regex close to the source where it is used and not making it public.

help.go
```diff
@@ -155,24 +157,91 @@ func ShowAppHelp(c *Context) (err error) {
...
-				fmt.Fprintf(c.App.Writer, "%s\n", name)
+				fmt.Fprintf(writer, "%s\n", name)
 			}
 		}
 	}
 }
 
+func cliArgContains(flagName string) bool {
+	for _, name := range strings.Split(flagName, ",") {
+		name = strings.Trim(name, " ")
```
saschagrunert: I'd suggest using `TrimSpace` here.

yogeshlonkar: done

autocomplete/bash_autocomplete
yogeshlonkar: done

autocomplete/bash_autocomplete
yogeshlonkar: replace with exact element match in array

help.go
yogeshlonkar: this is not making the variable public. keeping it in function will make regex compile on every invocation of method.

autocomplete/bash_autocomplete
yogeshlonkar: this seems false positive. `${COMP_WORDS[@]:0:$COMP_CWORD}` in not part for any `[[ ]]` and it is just using shell parameter expansion to get rid of what ever characters there are after current position of cursor to. This is already part current code and It completes current word/flag the cursor at instead of last word/flag in cli

saschagrunert: Ah I think shellcheck was complaining about double quoting to avoid word splitting. :)

help.go
saschagrunert: Yeah I agree. Then we leave it as it currently is.

autocomplete/bash_autocomplete
yogeshlonkar: that wont be an issue here I guess due to `$( ... )`

AudriusButkevicius: You still want to quote it I think, as if the word contains quotes this will blow up?

app_test.go
```diff
@@ -221,6 +221,60 @@ func ExampleApp_Run_subcommandNoAction() {
...
+	// Output:
+	// --other
+	// -o
+	// --xyz
+	// -x
+	// --help
+	// -h
+	// --version
+	// -v
+}
```
AudriusButkevicius: These tests do not assert anything. Why can't we set a custom writer and assert the output?

help.go
```diff
@@ -155,24 +157,91 @@ func ShowAppHelp(c *Context) (err error) {
...
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func printFlagSuggestions(lastArg string, flags []Flag, writer io.Writer) {
+	cur := shortFlagRegex.ReplaceAllString(lastArg, "")
+	cur = shortFlagRegex.ReplaceAllString(cur, "")
```
AudriusButkevicius: Why not just strip prefix? Firing up a regex engine for stripping leading dashes is a bit insane.

help.go
```diff
@@ -155,24 +157,91 @@ func ShowAppHelp(c *Context) (err error) {
...
+	for _, flag := range flags {
+		if bflag, ok := flag.(BoolFlag); ok && bflag.Hidden {
+			continue
+		}
+		for _, name := range strings.Split(flag.GetName(), ",") {
+			name = strings.Trim(name, " ")
```
AudriusButkevicius: Trim space?

help.go
```diff
@@ -155,24 +157,91 @@ func ShowAppHelp(c *Context) (err error) {
...
+			count := utf8.RuneCountInString(name)
+			if count > 2 {
+				count = 2
+			}
+			if strings.HasPrefix(lastArg, "--") && count == 1 {
+				continue
+			}
```
AudriusButkevicius: Whats the purpose of this if? What does it try to do? Should have a comment as it looks magical.

app_test.go
yogeshlonkar: // Output:
is test right? it is expecting output to match

help.go
```diff
@@ -155,24 +157,91 @@ func ShowAppHelp(c *Context) (err error) {
...
+	for _, flag := range flags {
+		if bflag, ok := flag.(BoolFlag); ok && bflag.Hidden {
+			continue
+		}
+		for _, name := range strings.Split(flag.GetName(), ",") {
+			name = strings.Trim(name, " ")
+			count := utf8.RuneCountInString(name)
+			if count > 2 {
+				count = 2
+			}
```
AudriusButkevicius: This kind of implies that you cannot have a single character long flag. Is that actually a limitation by the library or are we committing to not supporting that just here?

help.go
yogeshlonkar: I had compiled it on go 1.0.0 that does not have TrimPrefix. But yah make sense here

help.go
```diff
@@ -155,24 +157,91 @@ func ShowAppHelp(c *Context) (err error) {
...
+			if strings.HasPrefix(lastArg, "--") && count == 1 {
+				continue
+			}
+			flagCompletion := fmt.Sprintf("%s%s", strings.Repeat("-", count), name)
+			if strings.HasPrefix(name, cur) && cur != name && !cliArgContains(flag.GetName()) {
+				fmt.Fprintln(writer, flagCompletion)
+			}
```
AudriusButkevicius: Flag completion only used in the if block so should be moved there.

help.go
```diff
@@ -155,24 +157,91 @@ func ShowAppHelp(c *Context) (err error) {
...
+		}
+	}
```
AudriusButkevicius: This whole flag parseroo is a big ugly and should be responsibility of the flag I guess, but that's a separate battle I guessm

autocomplete/bash_autocomplete
yogeshlonkar: I don't think so that's the purpose of that command to invoke it

help.go
AudriusButkevicius: Even if it doesn't, a for loop is less work than a regexp

app_test.go
AudriusButkevicius: First time I am hearing that. Is that a go thing or our magical test runner? (I am new to this library)

AudriusButkevicius: :astonished: been writing go since 2013, did not know this.

help.go
yogeshlonkar: done

help.go
yogeshlonkar: done

help.go
yogeshlonkar: done

AudriusButkevicius:
The regex is now unused?

help.go
yogeshlonkar: I don't think so, parsing flag is generic use function and each flag should not be responsible to implementation. Besides this is not parsing the flag just pattern matching it for auto completion

help.go
yogeshlonkar: No it doesn't it. Added comments for better understanding

help.go
yogeshlonkar: done

PR #817, opened 2019-06-28
fix sliceValue not work with altsrc

julionlee:


lynncyrin:
Hiya! In general I think it's not totally clear how we handle issues with `altsrc`, so apologies if this PR takes a bit of extra work for me to understand.

altsrc/flag.go
```diff
@@ -116,7 +116,7 @@ func (f *IntSliceFlag) ApplyInputSourceValue(context *cli.Context, isc InputSour
 				eachName(f.Name, func(name string) {
 					underlyingFlag := f.set.Lookup(f.Name)
 					if underlyingFlag != nil {
-						underlyingFlag.Value = &sliceValue
```
lynncyrin: I'm not sure what the issue is here, can you add a test case?

lynncyrin:
ping @julionlee ~ can you explain this PR a bit more?

altsrc/flag.go
```diff
@@ -94,7 +94,7 @@ func (f *StringSliceFlag) ApplyInputSourceValue(context *cli.Context, isc InputS
...
+						*(underlyingFlag.Value.(*cli.StringSlice)) = sliceValue
```
AudriusButkevicius: Can we just do a copy instead?

lynncyrin:
Heya @julionlee, I'm trying to keep the open PRs list for maintainers updated and relevant - can you respond to the feedback above in the next week? If not (and that's fine!) i'll close this.

julionlee:
Sorry,i have no attention for this recently.

```
package main

import (
	"fmt"
	"github.com/urfave/cli/altsrc"
	"gopkg.in/urfave/cli.v1"
	"log"
	"os"
)

type SSTest struct {
	arga cli.StringSlice
	argb cli.StringSlice
}

var jsonobj = []byte(`{"arga":["aa","bb","cc"]}`)

func main() {
	var ss SSTest
	app := cli.NewApp()
	flags := []cli.Flag{
		altsrc.NewStringSliceFlag(cli.StringSliceFlag{
			Name:  "arga",
			Value: &(ss.arga),
			Usage: "string slice arga",
		}),
		altsrc.NewStringSliceFlag(cli.StringSliceFlag{
			Name:  "argb",
			Value: &(ss.argb),
			Usage: "string slice argb",
		})}
	app.Before = altsrc.InitInputSource(flags,
		func() (context altsrc.InputSourceContext, e error) {
			return altsrc.NewJSONSource(jsonobj)
		})
	app.Flags = flags
	app.Action = func(c *cli.Context) error { return nil }
	err := app.Run(os.Args)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println("arga:", ss.arga, len(ss.arga))
	fmt.Println("argb:", ss.argb, len(ss.argb))
}
```
```
# go run sstest.go --argb 11 --argb=22  
arga: [] 0
argb: [11 22] 2
```

The problem is that, the object address is a copy, and pointer assignment does not affect the original value. For example:

1. origin object is ss.argb ,set StringSliceFlag `Value: &(ss.arga)` (address:0x1000)
2. from json object , sliceValue = \["aa","bb","cc"\](address:0x2000)
3. typeof(sliceValue) is cli.StringSlice, typeof(underlyingFlag.Value) is *cli.StringSlice
4. underlyingFlag.Value = &sliceValue (underlyingFlag.Value=0x2000)，but the address of ss.arga is still 0x1000.
5. so the value of ss.arga not take effect.

```
@@ -94,7 +94,7 @@ func (f *StringSliceFlag) ApplyInputSourceValue(context *cli.Context, isc InputS
				eachName(f.Name, func(name string) {
					underlyingFlag := f.set.Lookup(f.Name)
					if underlyingFlag != nil {
						underlyingFlag.Value = &sliceValue 
                         // so i made this change 
                        // *(underlyingFlag.Value.(*cli.StringSlice)) = sliceValue
```

The same problem occurs in IntSliceFlag.

AudriusButkevicius:
My comment about magic copying still applies.

julionlee:
> My comment about magic copying still applies.

I don't think there is anything wrong with object copying.
Of course, I also look forward to a better way.

AudriusButkevicius:
I am ok with copying, but it's a slice underneath, let's just use the copy keyword?

julionlee:
> I am ok with copying, but it's a slice underneath, let's just use the copy keyword?

I think it's only an array pointer underneath.
Can you show the code?

lynncyrin:
👋  hello @julionlee! It's been about a month since the last update for this PR, do you still plan on working on it?

lynncyrin:
Closing due to age 🕙 feel free to re-open this if you're still interested!

AudriusButkevicius:
My only complain about this PR was the magical pointer deref based copying. I have not tried but I suspect that can be done with the copy keyword.

PR #823, opened 2019-08-02
Make the exit code example more clear

### LINKED ISSUE #820: Exit code
package main

import (
  "log"
  "os"

  "github.com/urfave/cli"
)

func main() {
  app := cli.NewApp()
  app.Flags = []cli.Flag{
    cli.BoolTFlag{
      Name:  "ginger-crouton",
      Usage: "is it in the soup?",
    },
  }
    if !ctx.Bool("ginger-crouton") {
      return cli.NewExitError("it is not in the soup", 86)
    }
    return nil
  }

  err := app.Run(os.Args)
  if err != nil {
    log.Fatal(err)
  }
}

the line of 13 should be 		cli.BoolFlag{   not 		cli.BoolTFlag{

xordspar0:
The purpose of this example is to show that you can exit with an error
code if a flag is unspecified, but with the code as it is, the only way
to cause a non-zero exit is by adding the flag `--ginger-crouton=false`,
which is not explained in the example.

In this new version of the example, running the command with no flag
will exit with an error, and running it with the flag will exit
normally.

This resolves #820.

### DIFF
--- a/README.md
+++ b/README.md
@@ -901,14 +901,14 @@ import (
 func main() {
   app := cli.NewApp()
   app.Flags = []cli.Flag{
-    cli.BoolTFlag{
+    cli.BoolFlag{
       Name:  "ginger-crouton",
-      Usage: "is it in the soup?",
+      Usage: "Add ginger croutons to the soup",
     },
   }
   app.Action = func(ctx *cli.Context) error {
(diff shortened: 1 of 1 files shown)


lynncyrin:
Thank you!

PR #830, opened 2019-08-03
Add markdown and man page docs generation methods

saschagrunert:
This adds two new methods to the `App` struct:

- `ToMarkdown`: creates a markdown documentation string
- `ToMan`: creates a man page string

Example markdown output (had to modify the triple backticks to not render here):
```
% Harrison
% August 2019

# NAME

greet - Some app

# SYNOPSIS

greet

``
[--another-flag|-b]
[--flag|--fl|-f]=[value]
[--socket|-s]=[value]
``

# DESCRIPTION

app [first_arg] [second_arg]

**Usage**:

``
greet [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
``

# GLOBAL OPTIONS

**--another-flag, -b**: another usage text
**--flag, --fl, -f**="": _no description available_
**--socket, -s**="": some usage text (default: value)

# COMMANDS

## config, c

another usage test

**--another-flag, -b**: another usage text
**--flag, --fl, -f**="": _no description available_

### sub-config, s, ss

another usage test

**--sub-command-flag, -s**: some usage text
**--sub-flag, --sub-fl, -s**="": _no description available_

## info, i, in

retrieve generic information

## some-command

_no description available_
```

I'm now totally open to discussions about the look and feel of the
markdown / man page. But please consider that the generated markdown
has to be consumed by md2man, so we're maybe not as free as we want to be.

What do you think? Any suggestions? :)

Closes https://github.com/urfave/cli/issues/340


docs.go
```diff
@@ -0,0 +1,289 @@
...
+
+		args = append(args, modifiedArg)
+
+	}
+	sort.Strings(args)
+	return args
+}
+
+// flagTakesValue returns true if the flag takes a value, otherwise false
+func flagTakesValue(flag Flag) bool {
```
lynncyrin: ❓ _(blocking question)_

Is there some other attribute on our `Flag` types that could be used to obtain this information?

If not, can you push this attribute up into the base `Flag` types? eg. `BoolFlag.takesValue`, GenericFlag.takesValue`.

docs.go
```diff
@@ -0,0 +1,289 @@
...
+		}
+	}
+	if f, ok := flag.(Uint64Flag); ok {
+		description = f.Usage
+		value = fmt.Sprintf("%d", f.Value)
+	}
+	if f, ok := flag.(UintFlag); ok {
+		description = f.Usage
+		value = fmt.Sprintf("%d", f.Value)
+	}
```
lynncyrin: ⭐⭐⭐ _(must change)_

I'm _pretty sure_ these various conditions for flags can be represented as a single if statement. So like

```golang
		description = f.Usage
		if f.Value != nil {
		    value = fmt.Sprintf("%f", f.Value)
		}
```

So that we can avoid the cognitive overhead of the type checks.

docs.go
```diff
@@ -0,0 +1,289 @@
...
+
+type CliTemplate struct {
+	App          *App
+	Date         string
+	Commands     []string
+	GlobalArgs   []string
+	SynopsisArgs []string
+}
+
+func (a *App) write(w io.Writer) error {
```
lynncyrin: ⭐⭐⭐ _(must change)_

This needs a more specific name, like `writeDocTemplate`, since it's on the widely shared `App` type.

docs.go
```diff
@@ -0,0 +1,289 @@
...
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"sort"
+	"strings"
+	"text/template"
+	"time"
+
+	"github.com/cpuguy83/go-md2man/md2man"
```
lynncyrin: ❓ _(blocking question)_

I think that since you added modules support, this goes into the `go.mod` file now?

docs.go
```diff
@@ -0,0 +1,289 @@
...
+{{ .App.Name }} [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
+` + "```" + `
+{{ if .GlobalArgs }}
+# GLOBAL OPTIONS
+{{ range $v := .GlobalArgs }}
+{{ $v }}{{ end }}
+{{ end }}{{ if .Commands }}
+# COMMANDS
+{{ range $v := .Commands }}
+{{ $v }}{{ end }}{{ end }}`
```
lynncyrin: ⭐⭐ _(seriously consider)_

So this is a big ask, but I wanted to ask it anyways! Can you consider possibly [generating this template from / sharing it with] the main help output template? When markdown generation is its own stand-alone template, people might not realize that it exists.

docs.go
saschagrunert: Yes, I added a `TakesValue() bool` method to the `Flag` interface and removed the function completely. Does this match your intention?

docs.go
saschagrunert: As above , I added a `GetUsage() string` and `GetValue() string` method to the `Flag` interface and removed the type assertions completely.

docs.go
saschagrunert: Renamed it as suggested. 👍

docs.go
saschagrunert: Yes, rebased on top of the latest master and added it to `go.mod`.

docs.go
saschagrunert: Hm, I moved the templates to `template.go`, made them private and const (breaking change). I could also let them var and public but I saw no real use case for having them public.

I added the markdown template there as well, but I'm not sure if and how we should make them dependent of each other... What do you think?

app_test.go
```diff
@@ -975,17 +975,17 @@ func TestRequiredFlagAppRunBehavior(t *testing.T) {
...
 				Flags: []Flag{StringFlag{Name: "requiredFlag", Required: true}},
 			}},
 		},
 		{
 			testCase:    "valid_case_required_flag_input_on_subcommand",
 			appRunInput: []string{"myCLI", "myCommand", "mySubCommand", "--requiredFlag", "cats"},
-			appCommands: []Command{Command{
+			appCommands: []Command{{
 				Name: "myCommand",
-				Subcommands: []Command{Command{
```
saschagrunert: gofmt removed these, which should be fine.

docs.go
```diff
@@ -0,0 +1,153 @@
...
+type CliTemplate struct {
+	App          *App
+	Date         string
+	Commands     []string
+	GlobalArgs   []string
+	SynopsisArgs []string
+}
+
+func (a *App) writeDocTemplate(w io.Writer) error {
+	now := time.Now()
```
AudriusButkevicius: For reproducibility, there should be a way to modify this.

docs.go
```diff
@@ -0,0 +1,153 @@
...
+func (a *App) ToMan() (string, error) {
+	var w bytes.Buffer
+	if err := a.writeDocTemplate(&w); err != nil {
+		return "", err
+	}
+	man := md2man.Render(w.Bytes())
+	return string(man), nil
+}
+
+type CliTemplate struct {
```
AudriusButkevicius: Does this really need to be exported?

docs.go
```diff
@@ -0,0 +1,153 @@
...
+	for i := range commands {
+		command := &commands[i]
+		prepared := strings.Repeat("#", level+2) + " " +
+			strings.Join(command.Names(), ", ") + nl
+
+		usage := noDescription
+		if command.Usage != "" {
+			usage = command.Usage
+		}
+		prepared += nl + usage + nl
```
AudriusButkevicius: Use a string buildee for good practice, this is a ton of copying for no reason.

docs.go
```diff
@@ -0,0 +1,153 @@
...
+
+		flags := prepareArgsWithValues(command.Flags)
+		if len(flags) > 0 {
+			prepared += nl
+		}
+		prepared += strings.Join(flags, nl)
+		if len(flags) > 0 {
+			prepared += nl
+		}
```
AudriusButkevicius: Seems this could be a single nl and an if, or just an if with nl, flags,nl ? I

docs.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	sort.Strings(args)
+	return args
+}
+
+// flagDetails returns a string containing the flags metadata
+func flagDetails(flag Flag) string {
+	description := flag.GetUsage()
+	if flag.GetUsage() == "" {
+		description = noDescription
```
AudriusButkevicius: Does this actually get printed? Is it not better to show nothing?

template.go
```diff
@@ -0,0 +1,108 @@
+package cli
+
+// AppHelpTemplate is the text template for the Default help topic.
+// cli.go uses text/template to render templates. You can
+// render custom help text by setting this variable.
+var AppHelpTemplate = `NAME:
```
AudriusButkevicius: Const or do we want it modifyable?

docs_test.go
```diff
@@ -0,0 +1,82 @@
...
+func TestToMan(t *testing.T) {
+	// Given
+	app := testApp()
+
+	// When
+	_, err := app.ToMan()
+
+	// Then
+	// TODO: extend test case
+	expect(t, err, nil)
```
AudriusButkevicius: You could check in expected output, or dump it in here. Atleast the md stuff would be viewable in github.

docs.go
saschagrunert: I will remove the date for now, I don't think that this is neccesary.

docs.go
saschagrunert: No, I made it internal

docs.go
saschagrunert: Okay, used `fmt.Sprintf` and combined some statements together. :+1:

docs.go
saschagrunert: I think this can be completely removed if I adapt the template, let's check.

docs.go
saschagrunert: Yes, I will go for that.

template.go
saschagrunert: This provides the user the possibility to modify the template for customization purposes. We do the same for the other templates so I thought this would be reasonable, WDYT?

docs_test.go
saschagrunert: Yes, I add this as well.

template.go
AudriusButkevicius: Sure as long as it's a conscious decision. Perhaps dependency injection would be better, but life is hard and then we all die.

saschagrunert: > Sure as long as it's a conscious decision. Perhaps dependency injection would be better, but life is hard and then we all die.

I keep that in mind for a later improvement...the dependency injection, not the fact that we all will die.

docs.go
```diff
@@ -0,0 +1,145 @@
...
+			strings.Repeat("#", level+2),
+			strings.Join(command.Names(), ", "),
+			usage,
+		)
+
+		flags := prepareArgsWithValues(command.Flags)
+		if len(flags) > 0 {
+			prepared += nl
+		}
+		prepared += strings.Join(flags, nl)
```
AudriusButkevicius: Is this not a noop if flags is empty? In which case it could be in the if.

This could still use a string builder for good practice.

saschagrunert: Yes, I fixed that one and a couple of more too.

AudriusButkevicius:
Sprintf only solves part of the problem. Any reason why you are not using https://www.calhoun.io/concatenating-and-building-strings-in-go/ etc?

I guess this is nitpicking at this point and I don't care as it looks ok, but just genuinely curious.

go.sum
```diff
@@ -1,5 +1,9 @@
 github.com/BurntSushi/toml v0.3.1 h1:WXkYYl6Yr3qBf1K79EBnL4mak0OimBfB0XUf9Vl28OQ=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
+github.com/cpuguy83/go-md2man v1.0.10 h1:BSKMNlYxDvnunlTymqtgONjNnaRV1sTpcovwwjF22jk=
+github.com/cpuguy83/go-md2man v1.0.10/go.mod h1:SmD6nW6nTyfqj6ABTjUi3V3JVMnlJmwcJI5acqYI6dE=
+github.com/russross/blackfriday v1.5.2 h1:HyvC0ARfnZBqnXwABFeSZHpKvJHJJfPz81GNueLj0oo=
+github.com/russross/blackfriday v1.5.2/go.mod h1:JO/DiYxRf+HjHt06OyowR9PTA263kcR/rfWxYHBV53g=
```
AudriusButkevicius: Where is this coming from? Is it actually a needed dep?

app_test.go
lynncyrin: I didn't realize you could do this!!! Nice

help.go
```diff
@@ -183,7 +110,7 @@ func ShowCommandHelpAndExit(c *Context, command string, code int) {
 func ShowCommandHelp(ctx *Context, command string) error {
 	// show the subcommand help for a command with subcommands
 	if command == "" {
-		HelpPrinter(ctx.App.Writer, SubcommandHelpTemplate, ctx.App)
```
lynncyrin: ❓ _(blocking question)_

Can you confirm that this isn't a breaking change? I'm not sure if the CLI supports people importing this variable and changing the template that way.

go.sum
lynncyrin: its indirect https://github.com/cpuguy83/go-md2man/blob/f3b8c87600b8883c04e9834382df4af7f4b11bf7/md2man/roff.go#L9

flag_generated.go
```diff
@@ -30,6 +31,22 @@ func (f BoolFlag) GetName() string {
...
+// GetUsage returns the usage string for the flag
+func (f BoolFlag) GetUsage() string {
+	return f.Usage
+}
+
+// GetValue returns the flags value as string representation and an empty
+// string if the flag takes no value at all.
+func (f BoolFlag) GetValue() string {
+	return ""
+}
```
lynncyrin: ⭐⭐⭐ _(must change)_

It looks like you edited this directly? What you'll want to do here is change the `generate-flag-types` file.

And that file is going to be removed in this PR => https://github.com/urfave/cli/pull/836 which means that this PR now depends on that one ^^

go.sum
saschagrunert: Yes :)

saschagrunert:
> Sprintf only solves part of the problem. Any reason why you are not using https://www.calhoun.io/concatenating-and-building-strings-in-go/ etc?
> 
> I guess this is nitpicking at this point and I don't care as it looks ok, but just genuinely curious.

Ah sorry then I got you wrong. For sure we could use `strings.Builder`

flag_generated.go
lynncyrin: https://github.com/urfave/cli/pull/836 has been merged! ✔️

flag.go
```diff
@@ -73,6 +73,16 @@ type Flag interface {
...
+	// TakesValue returns true of the flag takes a value, otherwise false
+	TakesValue() bool
+
+	// GetUsage returns the usage string for the flag
+	GetUsage() string
+
+	// GetValue returns the flags value as string representation and an empty
+	// string if the flag takes no value at all.
+	GetValue() string
 }
```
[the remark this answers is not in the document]
saschagrunert: Alright, I will it change that way.

flag_generated.go
saschagrunert: Alright I will change it

help.go
saschagrunert: Hm no I change it back.

saschagrunert:
So I hope I got every review. :)

PR #831, opened 2019-08-03
Add go module support

saschagrunert:
This adds a go.{mod,sum} file to official support go modules.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,8 @@
+module github.com/urfave/cli
+
+go 1.12
+
+require (
+	github.com/BurntSushi/toml v0.3.1
+	gopkg.in/yaml.v2 v2.2.2
+)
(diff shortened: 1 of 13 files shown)


go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/urfave/cli
+
+go 1.12
+
+require (
+	github.com/BurntSushi/toml v0.3.1
+	gopkg.in/urfave/cli.v1 v1.20.0
```
AudriusButkevicius: Require self? A bit strange no?

saschagrunert: The package `github.com/urfave/cli/altsrc`seems to rely on it :)

AudriusButkevicius: I think the imports should be rewritten to use github, which I think would address this confusion?

I think gopkg is a bad idea going forward given go.mod landed?
Or am I misunderstanding stuff.

saschagrunert: Sure, I rewrote the imports and now the dependency is gone 👍

PR #832, opened 2019-08-03
Go mod support

asahasrabuddhe:
Added a go.mod file 

Henceforth, we can support go modules through our version tags. Better vendoring :)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,9 @@
+module github.com/urfave/cli
+
+go 1.12
+
+require (
+	github.com/BurntSushi/toml v0.3.1
+	gopkg.in/urfave/cli.v1 v1.20.0
+	gopkg.in/yaml.v2 v2.2.2
+)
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/urfave/cli
+
+go 1.12
+
+require (
+	github.com/BurntSushi/toml v0.3.1
+	gopkg.in/urfave/cli.v1 v1.20.0
```
AudriusButkevicius: This is back

asahasrabuddhe: I think that is there due to the package being referenced in the flag type in the `altsrc` package. We would need to find a way to eventually merge it into the main package and get rid of this too in the process.

AudriusButkevicius:
See #831

asahasrabuddhe:
thanks i will use that instead

Issue #834, opened 2019-08-03
Remove python dependency in the build process


lynncyrin:
via https://github.com/urfave/cli/pull/784#issuecomment-517913044

lynncyrin:
I recently saw this tweet about self-hosting compilers https://twitter.com/TheLexTimes/status/1157076403252109314. Related to that, I realized that `urfave/cli` could use itself during the build process!

asahasrabuddhe:
@lynncyrin you read my mind. I have implemented a CLI app using `urfave/cli` to help generate the necessary code files. I am about to generate a PR and start a discussion to finalize the same :)

asahasrabuddhe:
I reviewed the runtests file today. It is a python application to bootstrap the build / test process. Do you guys think I should port it to bash or should I write a go app to do those same things and then use it?

lynncyrin:
> I reviewed the runtests file today. It is a python application to bootstrap the build / test process. Do you guys think I should port it to bash or should I write a go app to do those same things and then use it?

I would be in favor of either of those options! Writing it in bash sounds more "conventional", whereas writing in as a go app sounds more "fun". I'd write it as a go app first, and see how that feels 🔍

asahasrabuddhe:
this is done

PR #784, opened 2018-11-26
generate-flag-type: Use standard tag for generated code

dolmen:
... to prevent manual edit.

See official spec at https://golang.org/s/generatedcode.

PR #836, opened 2019-08-03
Flag Generation in the CLI

asahasrabuddhe:
Hello Team,

As suggested in #784 and in continuation of #834 here is the first implementation of my idea:

The idea is to remove the existing python script used to generate the flag types and have the library generate the necessary files using itself. This removes the dependency on Python and makes it easier for the maintainers to update this feature in the future as it is now written in Go.

The first thing we need to change now is how Travis runs. We would now need to first run

`go get https://github.com/urfave/cli/fg`

to build the binary and place it in the go path.  Once this is done,

`go generate ./...` 

should take care of generating the necessary files. This would be followed by tests. Please share your ideas :)


asahasrabuddhe:
What I really need an opinion is should the generation be a part of build process ? The generated files are already a part of the version control. In this case, it would be more practical for the developer to manually re-generate the files in case they change the `flag-types.json`

They would have to do it anyways in order to be able to test the changes they have done. What do you guys think?

asahasrabuddhe:
> Commenting to acknowledge that I saw this, and that I agree with your plan! I don't specifically have the bandwidth to craft the path forward here, but totally support you doing so

Yes but I would still appreciate you going through the code and sharing any pointers to make it better :)

altsrc/altsrc.go
```diff
@@ -1,3 +1,3 @@
 package altsrc
 
-//go:generate python ../generate-flag-types altsrc -i ../flag-types.json -o flag_generated.go
```
lynncyrin: ⭐ _(nice to have)_

It'd be nice to have `generate-flags` instead of `fg` here

flag-types.json
```diff
@@ -80,12 +96,16 @@
...
+    "value": true,
+    "dest": true,
     "context_default": "0",
     "parser": "strconv.ParseUint(f.Value.String(), 0, 64)"
   },
   {
     "name": "Uint",
     "type": "uint",
+    "value": true,
+    "dest": true,
```
lynncyrin: wow I have no idea what this file is even for 🤔

altsrc/altsrc.go
lynncyrin: Or `flag-generator`, either one!

flag_generated.go
```diff
@@ -685,7 +685,7 @@ func (f UintFlag) GetName() string {
 	return f.Name
 }
 
-// IsRequired returns the whether or not the flag is required
```
lynncyrin: 🙏  nice!

fg/main.go
```diff
@@ -0,0 +1,202 @@
...
+	Value          bool   `json:"value"`
+	Destination    bool   `json:"dest"`
+	Doctail        string `json:"doctail"`
+	ContextDefault string `json:"context_default"`
+	ContextType    string `json:"context_type"`
+	Parser         string `json:"parser"`
+	ParserCast     string `json:"parser_cast"`
+}
+
+var flagTemplate = `// Code generated by fg; DO NOT EDIT.
```
lynncyrin: ⭐⭐ _(seriously consider)_

This template is fairly long, so I think it'd be a good idea to move it into a file? https://github.com/urfave/cli/pull/769 implements something like that

fg/main.go
```diff
@@ -0,0 +1,202 @@
...
+			log.Fatal(err)
+		}
+
+		return nil
+	}
+
+	err := app.Run(os.Args)
+	if err != nil {
+		log.Fatal(err)
+	}
```
lynncyrin: 💯  I like how easy this reads!

fg/main.go
```diff
@@ -0,0 +1,202 @@
...
+`
+
+func main() {
+	var packageName, inputPath, outputPath string
+
+	app := cli.NewApp()
+
+	app.Name = "fg"
+	app.Usage = "Generate flag type code!"
+	app.Version = "v0.1.0"
```
lynncyrin: ⭐⭐ _(seriously consider)_

I think it may be confusing to have a version here?

fg/main.go
```diff
@@ -0,0 +1,202 @@
...
+
+	app.Flags = []cli.Flag{
+		cli.StringFlag{
+			Name:        "package, p",
+			Value:       "cli",
+			Usage:       "`PACKAGE` for which the flag types will be generated",
+			Destination: &packageName,
```
lynncyrin: ⭐⭐ _(seriously consider)_

We only have the two packages yeah? Can we make this a string slice, and have it default to both `cli` and `altsrc`? (or whatever the two names are)

fg/main.go
```diff
@@ -0,0 +1,202 @@
...
+		cli.StringFlag{
+			Name:        "input, i",
+			Usage:       "path to the `INPUT JSON FILE` which defines each type to be generated",
+			Destination: &inputPath,
+		},
+		cli.StringFlag{
+			Name:        "output, o",
+			Usage:       "path to the `OUTPUT GO FILE` which will contain the flag types",
+			Destination: &outputPath,
+		},
```
lynncyrin: ⭐⭐ _(seriously consider)_

I would strongly recommend either using defaults for both of these, or possibly removing them entirely? I don't think most people will need to be passing in specific options here.

fg/main.go
```diff
@@ -0,0 +1,202 @@
+package main
```
lynncyrin: ⭐ _(nice to have)_

meta: can we add a `Makefile` or similar so that there's a `make build` command that runs this file?

lynncyrin:
> What I really need an opinion is should the generation be a part of build process ?

100% yes 👍 I think there should be a `make build` script that does two things:

- runs this flag generation code
- then `go build ...`

_(I made another comment about adding `make build` in my code reviews)_

flag-types.json
asahasrabuddhe: This file is the source that is parsed to generate the flag types. The `FlagType` struct corresponds to this.

fg/main.go
asahasrabuddhe: I thought that the generator should have versions to track bugs etc. To make things easier, I'll have the generator binary share the version with the library.

altsrc/altsrc.go
asahasrabuddhe: `flag-generator` it is

fg/main.go
asahasrabuddhe: I used this approach as this started as a port of the existing python script. I think having defaults makes more sense as:

- The use of this tool is highly local to `urfave/cli`
- we would be able to remove the current two //go generate from the codebase to only a single one
- would clean up the eventual build process by requiring only a single command to generate the necessary bits

fg/main.go
asahasrabuddhe: agreed

fg/main.go
asahasrabuddhe: refer above conversation

fg/main.go
lynncyrin: 👍👍👍👍

fg/main.go
asahasrabuddhe: This is now done. Kindly review :)

fg/main.go
asahasrabuddhe: Done. All switches to accept input are now gone. We generate code for cli and altsrc for default

asahasrabuddhe:
> > What I really need an opinion is should the generation be a part of build process ?
> 
> 100% yes  I think there should be a `make build` script that does two things:
> 
> * runs this flag generation code
> * then `go build ...`
> 
> _(I made another comment about adding `make build` in my code reviews)_

But wouldn't it generate the same files over and over again? Unless the flag-types.json is changed, the regeneration process won't have any effect. Suppose you, for example, made a change in the flag-types.json file, you would explicitly have to run the code generation in order for you to be able to test the new functionality. Your next commit would have the latest files generated by default. We would not achieve anything new by running the generation logic once again.

Just a thought.

asahasrabuddhe:
There still exists one more python dependency `runtests` that I am looking at now. I think once we replace this with a `Makefile` most of our problems would be automatically taken care of

Makefile
```diff
@@ -0,0 +1,30 @@
...
+
+.PHONY: generate
+generate:
+	go generate ./...
+	git diff --exit-code
+
+.PHONY: test
+test:
+	go test -v --coverprofile=cli.coverprofile github.com/urfave/cli/
+	go test -v --coverprofile=altsrc.coverprofile github.com/urfave/cli/altsrc
```
AudriusButkevicius: This now requires make installed, which is not available on Windows. Everyone has go installed when compiling go, why can't we have a build.go script that you go run

AudriusButkevicius:
Please don't do forced pushes to branches, this destroys any sensible way to review them, as everything unconditionally becomes outdated.

flag-gen/assets_generate.go
```diff
@@ -0,0 +1,53 @@
...
+func (f file) Stat() (os.FileInfo, error) {
+	fi, err := f.File.Stat()
+	return fileInfo{fi}, err
+}
+
+type fileInfo struct {
+	os.FileInfo
+}
+
+func (fi fileInfo) ModTime() time.Time { return time.Time{} }
```
AudriusButkevicius: I still disagree with this and I have not heard an good argument why we need to bodge normal filesystem operations.

asahasrabuddhe: I hear you. Will remove this in the next commit

Makefile
asahasrabuddhe: the forced push was me reverting the previous commit including make. Now we are back to what we had previously. I am now working on a build.go that will take care of these things.

flag-gen/assets_generate.go
asahasrabuddhe: Hey @AudriusButkevicius 

I found a valid argument for this point. Please check `assets_vsfdata.go` file. The issue here is that on regeneration during builds, the mod time of the file will change. This will reflect inside `assets_vsfdata.go`. In each build, we check if there is any diff in between newly generated files and previous files. This change in modtime will cause a git diff everytime a build is run causing our build to fail.

asahasrabuddhe: mtime

Check this out

asahasrabuddhe:
All done, including code coverage :sunglasses: 

https://codecov.io/gh/urfave/cli

flag_generated.go
```diff
@@ -6,17 +8,15 @@ import (
...
-	Required    bool
-	Hidden      bool
-	Destination *bool
+	Name		string
+	Usage		string
+	EnvVar		string
+	FilePath	string
+	Required	bool
+	Hidden		bool
+	Destination	*bool
```
lynncyrin: Nice 💯 was this spacing done automatically?

asahasrabuddhe: Oh yes :)

lynncyrin: 👏👏👏

flag-gen/assets_generate.go
asahasrabuddhe: @AudriusButkevicius ping

appveyor.yml
```diff
@@ -8,19 +8,19 @@ clone_folder: c:\gopath\src\github.com\urfave\cli
...
+  - go get -v -t ./...
 
 build_script:
-- python runtests vet
-- python runtests test
-- python runtests gfmrun
+  - go run build/build.go generate
+  - go run build/build.go vet
+  - go run build/build.go test
+  - go run build/build.go gfmrun
```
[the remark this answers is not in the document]
asahasrabuddhe: go run builds and immediately runs it. The reason I am not using a built CLI is because right now there is no way to go get that package because it does not yet exist on master. This works well as I tested it on my machine

build/build.go
```diff
@@ -0,0 +1,179 @@
...
+			Name:   "toc",
+			Action: TocActionFunc,
+		},
+		cli.Command{
+			Name:   "generate",
+			Action: GenActionFunc,
+		},
+	}
+
+	packages = []string{"cli", "altsrc"}
```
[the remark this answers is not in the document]
asahasrabuddhe: go cannot have slice constants. I can have it as a constant array but this serves the purpose very well.

build/build.go
```diff
@@ -0,0 +1,179 @@
...
+
+func TestActionFunc(c *cli.Context) error {
+	for _, pkg := range packages {
+		var packageName string
+
+		if pkg == "cli" {
+			packageName = fmt.Sprintf("github.com/urfave/cli")
+		} else {
+			packageName = fmt.Sprintf("github.com/urfave/cli/%s", pkg)
+		}
```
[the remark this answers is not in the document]
asahasrabuddhe: there could be but then this condition would move below into the cleanup function. Either way, how does this improve on the existing implementation?

flag-gen/assets_generate.go
asahasrabuddhe: the thing is when we do a git clone, the mod time changes. this prompts the vfsgen plugin to generate new assets file as it perceives the files as changed. hence the need for the zero mod time fs work around.

.travis.yml
```diff
@@ -5,23 +5,28 @@ osx_image: xcode8.3
...
-- go get golang.org/x/tools/cmd/goimports
-- if [ ! -f node_modules/.bin/markdown-toc ] ; then
+  - go get github.com/shurcooL/vfsgen || true
+  - go get github.com/shurcooL/httpfs/union || true
+  - go get github.com/urfave/gfmrun/... || true
+  - go get golang.org/x/tools/cmd/goimports
+  - if [ ! -f node_modules/.bin/markdown-toc ] ; then
     npm install markdown-toc ;
-  fi
+    fi
```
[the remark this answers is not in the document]
asahasrabuddhe: done

asahasrabuddhe:
Once this is done, I will create a similar PR for v2

build/build.go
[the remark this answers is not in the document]
AudriusButkevicius: Sure, but a set of non-changing values, assigned to a global, somewhere in some function, is completely magical. A reader sees packages, finds the definition, and it's empty...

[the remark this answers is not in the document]
asahasrabuddhe: let's convert this to an const array then

appveyor.yml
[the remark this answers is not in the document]
AudriusButkevicius: I guess my question is why don't we run `go build ./...` as part of CI?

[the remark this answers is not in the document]
asahasrabuddhe: I did not try this. Let me try and see how it goes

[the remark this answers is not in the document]
asahasrabuddhe: We can try building a binary but it will cause us issues when writing scripts as windows will need .exe and mac/linux won't. There will be a conflict in go generate directive.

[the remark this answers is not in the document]
AudriusButkevicius: I think we are misunderstanding each other. I think you are talking about using the library itself for the build script. I am not talking about that. I am saying we generate files, run the tests as part of CI, yet don't actually try to build the code of the library. P

build/build.go
```diff
@@ -0,0 +1,177 @@
+package main
```
AudriusButkevicius: I think this should live at the root with a go build that that skipa building it. It will still be runnable I think.

build/build.go
```diff
@@ -0,0 +1,177 @@
...
+func VetActionFunc(_ *cli.Context) error {
+	return exec.Command("go", "vet").Run()
+}
+
+func TestActionFunc(c *cli.Context) error {
+	for _, pkg := range packages {
+		var packageName string
+
+		if pkg == "cli" {
+			packageName = fmt.Sprintf("github.com/urfave/cli")
```
AudriusButkevicius: Could just be a string without the sprintf

flag-gen/assets_generate.go
AudriusButkevicius: Fine

build/build.go
asahasrabuddhe: if we move to root, we won't be able to have `package main`. and without `package main`, the main method won't run

build/build.go
asahasrabuddhe: agree. will change this

appveyor.yml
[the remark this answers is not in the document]
asahasrabuddhe: That's what I am doing now. I am confused as to what you want me to do. Go build will output a binary however go run will run it without generating a build.

build/build.go
AudriusButkevicius: Is that true if the file is excluded from building with a nobuild directive?

appveyor.yml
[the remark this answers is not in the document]
AudriusButkevicius: I don't see go build anywhere in this file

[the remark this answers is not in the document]
asahasrabuddhe: how will go build make the solution any better than what we have currently?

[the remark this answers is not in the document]
AudriusButkevicius: Well currently we're not even building the code? So why would building the code make it worse?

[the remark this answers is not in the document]
asahasrabuddhe: we are building it. go run first builds the code and then immediately runs it without outputting the binary.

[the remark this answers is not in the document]
AudriusButkevicius: I don't see `go build` in either of the CI files, hence why I am confused. How does this get built if there is no go build anywhere? Some CI magic?

Sorry, I know it feels like we are going in circles but just want to understand it.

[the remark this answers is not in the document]
asahasrabuddhe: The `go run` builds it! The `go run` command is roughly equivalent of the following:

`go build X.go -o /tmp/random-tmp-folder/exe && /tmp/random-tmp-folder/exe`

so it will build and immediately run.

another way to look at this would be if you could point me to the line of code where we were using something analogous to go build in the old python scripts.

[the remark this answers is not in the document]
AudriusButkevicius: But we don't go rum the cli itself, as it's a library. I understand go run builds the flag generator, but I am talking about building the library after the flags are generated to see if it still builds.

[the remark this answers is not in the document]
asahasrabuddhe: oh! I agree this is not done but then it wasn't even done before. If we examine the code closely, we will see that if the flags are not generated right, the test cases would fail.

build/build.go
asahasrabuddhe: this is done

build/build.go
asahasrabuddhe: fair point.I have moved the file

appveyor.yml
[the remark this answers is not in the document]
AudriusButkevicius: It's possible that running tests will not build a certain subpackage.
For example, if I have
`cli/` and  `cli/argparse`, and `cli/argparse` has no `_test.go` files, that package would not be built part of the tests (no tests found), yet might be incompatible with what the flags generated and break when building the library (oppose to when running the library)

build.go
```diff
@@ -0,0 +1,179 @@
...
+			Name:   "toc",
+			Action: TocActionFunc,
+		},
+		cli.Command{
+			Name:   "generate",
+			Action: GenActionFunc,
+		},
+	}
+
+	err := app.Run(os.Args)
```
AudriusButkevicius: Perhaps the default behaviour for this (no arguments), would be to run all of these steps, or just run a build?

asahasrabuddhe: that is up for discussion. this is just a port of the old python script. we can change it to what we feel would be the best

appveyor.yml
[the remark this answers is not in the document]
asahasrabuddhe: yes but in our scenario, there is only one different package, altsrc which does have its tests.

build.go
AudriusButkevicius: Agree, feel free to ignore me.

asahasrabuddhe:
@AudriusButkevicius I am facing a new problem. One of the examples in README.md returns error code 86. Due to this, gfmrun reports an error causing tests to fail. Does anyone here or from the old team know about this?

AudriusButkevicius:
No idea, perhaps try binary bisecting the file to see where it blows up?

lynncyrin:
> I am facing a new problem. One of the examples in README.md returns error code 86. Due to this, gfmrun reports an error causing tests to fail. Does anyone here or from the old team know about this?

🤷‍♀

appveyor.yml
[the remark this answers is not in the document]
AudriusButkevicius: That doesn't mean it will always be the case. 

Regardless, I agree that it's not a change for this PR, but doing a simple go build would definately up my confidence.

asahasrabuddhe:
> No idea, perhaps try binary bisecting the file to see where it blows up?

on it

asahasrabuddhe:
After a lot of digging, here is what I found:

#823 caused the code snippet to return an error. This was not being handled by `gfmrun` leading to failed tests.

I have created a PR https://github.com/urfave/gfmrun/pull/22 to achieve this. Once this is merged, we can proceed ahead.

@AudriusButkevicius please do the honors or extend write access to that repo for our team.

Regards,
Ajitem

asahasrabuddhe:
> Lgtm, there are other changes I'd do in the future, but as @asahasrabuddhe correctly pointed out, these are somewhat out of scope for this PR.

please create issues about the changes you envision and assign them to me. I will take a look and implement them step by step :)

asahasrabuddhe:
@lynncyrin Please do the needful here :)

asahasrabuddhe:
this is a huge relief! I will work on this same thing for v2 now. :)

PR #837, opened 2019-08-03
Use codeowners instead of maintainers

lynncyrin:
Inspired by cpython https://github.com/python/cpython/blob/master/.github/CODEOWNERS, I added github's codeowners file https://help.github.com/en/articles/about-code-owners. 

Using this file unlocks some nice automation (like automatic review pings) and removes the issue of "oh, we forgot to update the maintainers file"!


asahasrabuddhe:
this is brilliant!

Issue #839, opened 2019-08-05
implement context.Context


marwan-at-work:
*cli.Context should implement context.Context and tie the context cancellation (Done()) to a user hitting ctrl+C

lynncyrin:
✔️ https://github.com/urfave/cli/pull/840

PR #840, opened 2019-08-05
add context.Context to cli.Context

marwan-at-work:
This is a naive implementation of attaching context.Context to the *cli.Context

The goroutine that waits for a kill signal hangs forever. I imagine that's okay since running a cli command does not hang forever. Otherwise, I'd be happy to add a way to end the goroutine. 

Fixes https://github.com/urfave/cli/issues/839

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -503,6 +503,7 @@ func TestApp_Float64Flag(t *testing.T) {
 }
 
 func TestApp_ParseSliceFlags(t *testing.T) {
+	var parsedOption, firstArg string
 	var parsedIntSlice []int
 	var parsedStringSlice []string
 
@@ -517,11 +518,15 @@ func TestApp_ParseSliceFlags(t *testing.T) {
 				Action: func(c *Context) error {
 					parsedIntSlice = c.IntSlice("p")
 					parsedStringSlice = c.StringSlice("ip")
+					parsedOption = c.String("option")
+					firstArg = c.Args().First()
 					return nil
 				},
 			},
 		},
 	}
+	var _ = parsedOption
+	var _ = firstArg
 
 	app.Run([]string{"", "cmd", "-p", "22", "-p", "80", "-ip", "8.8.8.8", "-ip", "8.8.4.4", "my-arg"})
 
--- a/app_test.go
+++ b/app_test.go
@@ -236,7 +236,7 @@ func ExampleApp_Run_shellComplete() {
 	os.Args = []string{"greet", fmt.Sprintf("--%s", genCompName())}
 
 	app := &App{
-		Name: "greet",
+		Name:                  "greet",
 		EnableShellCompletion: true,
 		Commands: []*Command{
 			{
@@ -503,7 +503,6 @@ func TestApp_Float64Flag(t *testing.T) {
 }
 
 func TestApp_ParseSliceFlags(t *testing.T) {
-	var parsedOption, firstArg string
 	var parsedIntSlice []int
 	var parsedStringSlice []string
 
@@ -518,8 +517,6 @@ func TestApp_ParseSliceFlags(t *testing.T) {
 				Action: func(c *Context) error {
 					parsedIntSlice = c.IntSlice("p")
 					parsedStringSlice = c.StringSlice("ip")
-					parsedOption = c.String("option")
-					firstArg = c.Args().First()
 					return nil
 				},
 			},
--- a/context.go
+++ b/context.go
@@ -1,11 +1,14 @@
 package cli
 
 import (
+	"context"
 	"errors"
 	"flag"
 	"os"
+	"os/signal"
 	"reflect"
 	"strings"
+	"syscall"
 )
 
 // Context is a type that is passed through to
@@ -13,6 +16,7 @@ import (
 // can be used to retrieve context-specific args and
 // parsed command-line options.
 type Context struct {
+	context.Context
 	App           *App
 	Command       *Command
 	shellComplete bool
@@ -24,6 +28,14 @@ type Context struct {
 // NewContext creates a new context. For use in when invoking an App or Command action.
 func NewContext(app *App, set *flag.FlagSet, parentCtx *Context) *Context {
 	c := &Context{App: app, flagSet: set, parentContext: parentCtx}
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+		defer cancel()
+		sigs := make(chan os.Signal, 1)
+		signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+		<-sigs
+	}()
+	c.Context = ctx
 
 	if parentCtx != nil {
 		c.shellComplete = parentCtx.shellComplete
--- a/context.go
+++ b/context.go
@@ -29,12 +29,10 @@ type Context struct {
 func NewContext(app *App, set *flag.FlagSet, parentCtx *Context) *Context {
 	c := &Context{App: app, flagSet: set, parentContext: parentCtx}
 	if parentCtx != nil {
-		if parentCtx.Context == nil {
-			parentCtx.Context = context.Background()
-		}
 		c.Context = parentCtx.Context
 		c.shellComplete = parentCtx.shellComplete
-	} else {
+	}
+	if c.Context == nil {
 		ctx, cancel := context.WithCancel(context.Background())
 		go func() {
 			defer cancel()
--- a/context.go
+++ b/context.go
@@ -28,17 +28,21 @@ type Context struct {
 // NewContext creates a new context. For use in when invoking an App or Command action.
 func NewContext(app *App, set *flag.FlagSet, parentCtx *Context) *Context {
 	c := &Context{App: app, flagSet: set, parentContext: parentCtx}
-	ctx, cancel := context.WithCancel(context.Background())
-	go func() {
-		defer cancel()
-		sigs := make(chan os.Signal, 1)
-		signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
-		<-sigs
-	}()
-	c.Context = ctx
-
 	if parentCtx != nil {
+		if parentCtx.Context != nil {
+			parentCtx.Context = context.Background()
+		}
+		c.Context = parentCtx.Context
 		c.shellComplete = parentCtx.shellComplete
+	} else {
+		ctx, cancel := context.WithCancel(context.Background())
+		go func() {
+			defer cancel()
+			sigs := make(chan os.Signal, 1)
+			signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+			<-sigs
+		}()
+		c.Context = ctx
 	}
 
 	return c
--- a/context.go
+++ b/context.go
@@ -29,7 +29,7 @@ type Context struct {
 func NewContext(app *App, set *flag.FlagSet, parentCtx *Context) *Context {
 	c := &Context{App: app, flagSet: set, parentContext: parentCtx}
 	if parentCtx != nil {
-		if parentCtx.Context != nil {
+		if parentCtx.Context == nil {
 			parentCtx.Context = context.Background()
 		}
 		c.Context = parentCtx.Context
--- a/context_test.go
+++ b/context_test.go
@@ -264,17 +264,20 @@ func TestContext_lookupFlagSet(t *testing.T) {
 	}
 }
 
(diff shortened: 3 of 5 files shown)


go.mod
```diff
@@ -0,0 +1,3 @@
+module gopkg.in/urfave/fli.v2
```
AudriusButkevicius: Fli? Also, why gopkg.in?

context.go
```diff
@@ -24,6 +28,14 @@ type Context struct {
...
 func NewContext(app *App, set *flag.FlagSet, parentCtx *Context) *Context {
 	c := &Context{App: app, flagSet: set, parentContext: parentCtx}
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+		defer cancel()
+		sigs := make(chan os.Signal, 1)
+		signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+		<-sigs
+	}()
+	c.Context = ctx
```
AudriusButkevicius: If parent context is not nil, should we not inherit? Also a bit grim the client can't inject a context from higher up, so he could use it to cancel his own commands.

context.go
```diff
@@ -24,6 +28,14 @@ type Context struct {
 // NewContext creates a new context. For use in when invoking an App or Command action.
 func NewContext(app *App, set *flag.FlagSet, parentCtx *Context) *Context {
 	c := &Context{App: app, flagSet: set, parentContext: parentCtx}
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+		defer cancel()
+		sigs := make(chan os.Signal, 1)
+		signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+		<-sigs
```
AudriusButkevicius: I don't think a random library should yank signal handling of the whole process because it feels like it.
It should be up to the user to set this up.

go.mod
marwan-at-work: @AudriusButkevicius ops, will fix that. What import path should it be? The README for v2 says use gopkg.in which is why I added that

context.go
marwan-at-work: @AudriusButkevicius the client can wrap the context if they wish to. So that should be possible. Note that net/http does the same thing with the `*http.Request.Context()` in that it automatically includes cancellation if the client cancels the request.

context.go
marwan-at-work: @AudriusButkevicius similar to the comment above, this follows the net/http request Context pattern. If the user wants to use their own context, they can simple create one and ignore the given one, or they can wrap the given one. 

Furthermore, the Context is exposed, so the user can replace it. 

However, if you'd like me to change that to just be context.Background() and have each user implement the boiler plate above on their side, that's okay with me too :+1:

go.mod
AudriusButkevicius: I guess this needs a wider discussion on how we deal with this going forward. With go.mod support I think gopkg.in is obsolete, and people just `go get github.com/urfave/cli@2.0.0`

context.go
AudriusButkevicius: Then we should leave it empty by default, or context.Background()?

go.mod
marwan-at-work: @AudriusButkevicius `github.com/urfave/cli@2.0.0` would not work. Per "Semantic Import Versioning" it would have to be `github.com/urfave/cli/v2` and similarly `go get github.com/urfave/cli/v2@v2.1.9` and so on. 

Note that Go Modules has a special handling of Gopkg.in so it should work okay. That said, I'm happy to remove the go.mod file from this PR and open a new issue for what the import path should be/

context.go
AudriusButkevicius: This was not about that, but about hijacking the processes signal handlers, and net/http not doing that by default because it's being used.

It seems go abstracts signal handling, so calling notify multiple times does not override the process signal handler repeatedly, in which case I am ok this this.

context.go
AudriusButkevicius: My question is whether we should create a new context for every subcommand or just pass the same context down.

go.mod
AudriusButkevicius: Why would @2.0.0 not work, I was under the assumption that just resolves the release tag. Not sure why it needs to move to a separate subpackage to work?

context.go
marwan-at-work: @AudriusButkevicius I see. Is there ever a situation where the CLI runs multiple subcommands at the same time? Or does urfave/cli only run one sub command per binary invocation? If the latter, then we're never gonna create multiple contexts. 

That said, if you can point to any changes, I'm happy to make those :+1:

go.mod
marwan-at-work: @AudriusButkevicius "Semantic Import Versioning" is a Module rule where it forces the major tag (2 and above) to be included in the import path. See here: https://github.com/golang/go/wiki/Modules#semantic-import-versioning

And this image is a good summary: 

image

context.go
AudriusButkevicius: We run NewContext for every subcommand, as we traverse down the command tree:
https://github.com/urfave/cli/blob/7675649a174ac724b1d7fa9d5932eff3dd1582a2/app.go#L323

So for `command subcomand-a subcommand-a-b subcommand-a-b-c` we'd init 4 contexts, all of which would set up their own signal handlers and have nothing in common with the parent context, hence cancellation at the root would do nothing I think.

marwan-at-work: @AudriusButkevicius sounds good. Thanks for the clarification. I will verify and let you know :+1:

go.mod
AudriusButkevicius: Ok, I am not sure how we'll handle this. 
Ideally I'd like to get away from gopkg.in, as we'll always live with master being v1, and go get github.com/urface/cli fetching v1.

marwan-at-work: @AudriusButkevicius `go get github.com/urfave/cli` will *always* fetch v1, and `go get github.com/urfave/cli/v2` will always fetch v2 and so on.

context.go
marwan-at-work: @AudriusButkevicius ok i just updated the PR to only run one signal handler, and all sub-commands will inherit that context. 

Interestingly, `command sub-a sub-a-b sub-a-b-c` will call NewContext on each one as you pointed out, but the `Action` function that will run is only `sub-a-b-c` which makes me think that maybe urfave/cli should not create multiple contexts? It should probably just create one context for the Action it will run. 

If I'm not missing something, that's a topic for another PR. For now let me know if this is good to merge.

Thanks!

go.mod
AudriusButkevicius: Use, unless we merge v2 to master and roll with that. I don't think two versions of the library should exist in the same repo basically.

marwan-at-work: @AudriusButkevicius Go Modules does not care about what's in master, when you have semver tags. It will basically look at git tags and determine what version to get you.

marwan-at-work: @AudriusButkevicius for this PR however, should I just remove the go.mod file? Or should I change it to be `github.com/urfave/cli/v2`?

context.go
```diff
@@ -24,9 +28,21 @@ type Context struct {
 // NewContext creates a new context. For use in when invoking an App or Command action.
 func NewContext(app *App, set *flag.FlagSet, parentCtx *Context) *Context {
 	c := &Context{App: app, flagSet: set, parentContext: parentCtx}
-
 	if parentCtx != nil {
+		if parentCtx.Context != nil {
+			parentCtx.Context = context.Background()
+		}
```
AudriusButkevicius: This has a side effect on the parentCtx passed in, which I as a user would find suprising.

AudriusButkevicius: Also, I think the intended condition in the if was == nil it seems?

I think this needs tests, as we can't rely on reviewers to prove this works

go.mod
AudriusButkevicius: No idea, thats why I want someone else from @urfave/cli to comment and drive the direction of the migration.

context.go
marwan-at-work: @AudriusButkevicius this is just to make sure the context is never nil, so that we never panic. Should I remove it and leave the risk up to the user?

AudriusButkevicius: It's not true, the if statement is wrong and it always overrides it if its *NOT* nil, and leaves it nil otherwise. I think we should not cause side effects on objects given to us, so if its nil, create a new one locally, and then validate this behaviour with a few tests.

marwan-at-work: @AudriusButkevicius great catch, that was a mistake on my part. I will change != nil to == nil

marwan-at-work: @AudriusButkevicius i also added some propagation tests to ensure the parent context does not get overridden.

app_test.go
```diff
@@ -503,7 +503,6 @@ func TestApp_Float64Flag(t *testing.T) {
 }
 
 func TestApp_ParseSliceFlags(t *testing.T) {
-	var parsedOption, firstArg string
```
marwan-at-work: the Go command complained that these were declared and not used

context.go
marwan-at-work: I'd like to add sigterm tests, but i'm not sure how I can send a sigterm signal to the process, without killing the tests themselves.

app_test.go
AudriusButkevicius: These are not used but they do have side effects/assert evaluation logic so should probably be called?

marwan-at-work: @AudriusButkevicius the code didn't even compile though, so im not sure how that's supposed to work? In any case, I can just add `var _ = parsedOption` below

context.go
AudriusButkevicius: I still think we should not mutate the parent context.
```
c.Context = parentCtx.Context
```

Then outside the if (oppose to else):
```
If c.Context == nil {
  // Create cancel context, spin up signal handler ctx.
}
```

context_test.go
```diff
@@ -262,3 +263,24 @@ func TestContext_lookupFlagSet(t *testing.T) {
...
 }
+
+// TestContextPropagation tests that
+// *cli.Context always has a valid
+// context.Context
+func TestContextPropagation(t *testing.T) {
+	ctx := NewContext(nil, nil, nil)
+	if ctx.Context == nil {
+		t.Fatal("expected a non nil context when no parent is present")
+	}
```
AudriusButkevicius: Should be a separate test

context_test.go
```diff
@@ -262,3 +263,24 @@ func TestContext_lookupFlagSet(t *testing.T) {
...
+	parent := NewContext(nil, nil, nil)
+	parent.Context = context.WithValue(context.Background(), "key", "val")
+	ctx = NewContext(nil, nil, parent)
```
AudriusButkevicius: What about a test where parent.Context is nil, but parent not nil.

go.mod
```diff
@@ -0,0 +1,3 @@
+module gopkg.in/urfave/cli.v2
+
+go 1.12
```
AudriusButkevicius: Lets leave this out. The team needs to decide whats happening with the different versions.

marwan-at-work: @AudriusButkevicius alright, I will remove the go.mod file for now.

app_test.go
AudriusButkevicius: Its not the variable declarations, its the calls that assign to the unused variables is what matters.

context_test.go
marwan-at-work: @AudriusButkevicius I made this into a separate test

context_test.go
marwan-at-work: @AudriusButkevicius i added that scenario to the bottom of this test.

context.go
marwan-at-work: @AudriusButkevicius done

app_test.go
marwan-at-work: @AudriusButkevicius again, the code doesn't even compile and tests don't run :)

marwan-at-work: This is because we're just re-assigning those variables, but not really doing anything with them. It could be that go1.8 (what travis is using) didn't care, but Go 1.13 is smarter.

AudriusButkevicius:
lgtm

lynncyrin:
Oh rather, LGTM pending the tests passing 🙂

marwan-at-work:
@lynncyrin I'm not sure why the build is failing. It has to do with the python script which im not familiar with. Can you please point me in the right direction to fix this? Thanks

AudriusButkevicius:
I think the test is a red herring, we need to bump travis version.

PR #843, opened 2019-08-07
Support also GoLang 1.11 compiler

lafriks:
Currently when using `github.com/urfave/cli` in project it requires GoLang compiler version 1.12

Error:
```
go build github.com/urfave/cli: module requires Go 1.12
```


lafriks:
@asahasrabuddhe done

PR #848, opened 2019-08-08
Add fish shell completion support

saschagrunert:
This commit adds a new method `ToFishCompletion` to the `*App` which can
be used to generate a fish completion string for the application.

Relates to: #351, #833

PR #851, opened 2019-08-09
Add `TakesFile` indicator to flag

saschagrunert:
This new member of `Flag` indicates if the flag expects a file as input.
This is especially useful for documentation and shell completion purposes.

If we add this, then the fish shell completion as well as the documentation
could indicate that a file is expected, which would make especially the completion
more usable.

Relates to https://github.com/urfave/cli/pull/848


saschagrunert:
Thanks!

AudriusButkevicius:
It's super grim that our boolean flag now takes file. This whole autogeneration stuff is mad, we have like 10 types, hand rolling the code would get rid of tons of this insanity.

AudriusButkevicius:
I think more consideration should go into PRs that extend the user facing facade, as I think we've gone full YOLO mode. 

You can always add flags, but removing them is hard, and then suddenly trying to remove TakesFiles from a IntSlice becomes a API breaking change requiring a V99 of the library.

asahasrabuddhe:
There is no need to reverse this to avoid the boolean flag to take a file. We can simply make a few patches to the json generator source and the generation template.

It is a little surprising to see you not having a problem with code generation earlier and having a problem now suddenly. As far as I can tell, this library had code generation since a long time!

Please let's discuss problems to come up with sensible solutions instead of rashly going about doing things.

saschagrunert:
I would also extend the code generator to fix this up for the flags which cannot take a file at all.

AudriusButkevicius:
I never participated in development of this library, nor do I really want to now, but I just want to make sure that it doesn't turn into a mess where every PR is merged without considering if its the right thing to do.

I am not going to fix this, but I think that it's better to not have this at all than to have it in it's current form, hence why I opened the revert (as it's just easy button clicking). If there is an alternative PR that fixes this, great, we can close my revert, but if that does not happen, we should revert before this goes to the wild.

I think the flag generation is completely magical and perhaps it should not exist, as it's like 10 flags that can be hand maintained, yet that is my opinion and I guess it matters less, if you guys are ok with this.

saschagrunert:
I'm going to fix this. Please review https://github.com/urfave/cli/pull/860

PR #854, opened 2019-08-11
[WIP] Feature: Flags validation support

asahasrabuddhe:
This is the first draft of the support for flags validation. I have managed to implement it for Int, Int64, Float, Float64 and String flags. Slice flags are pending. Also support for altsrc is pending. Hence the draft and WIP.

I have however run into an error with a failed test for `TestToMan` function. @saschagrunert could you take a look as to why that might be? Meanwhile, I will continue development and pushing commits.


app.go
```diff
@@ -245,6 +245,14 @@ func (a *App) Run(arguments []string) (err error) {
 		return cerr
 	}
 
+	// Validation Logic
+	verr := validateFlags(a.Flags, context)
```
lynncyrin: I believe you'll want to add this to `Command.Run` also https://github.com/urfave/cli/blob/7e49cc210a231eec218c2fba82df106af06d05b5/command.go#L100

Unrelated to this PR: I've been thinking about how to DRY `App.Run` / `App.RunAsSubCommand` / `Command.Run` 🤔

flag.go
```diff
@@ -75,6 +75,15 @@ type Flag interface {
 	GetName() string
 }
 
+// Flags represents a collection of Flags
+type Flags []Flag
+
+type Validator interface {
```
lynncyrin: I'd possible call this `ValidatedFlag`?

lynncyrin:
👏  looking great so far!

saschagrunert:
> I have however run into an error with a failed test for `TestToMan` function. @saschagrunert could you take a look as to why that might be? Meanwhile, I will continue development and pushing commits.

This seems to be solved, right? I don't see any further influences on this test case.

asahasrabuddhe:
> > I have however run into an error with a failed test for `TestToMan` function. @saschagrunert could you take a look as to why that might be? Meanwhile, I will continue development and pushing commits.
> 
> This seems to be solved, right? I don't see any further influences on this test case.

I don't think so. I was also very surprised to see this. I'd really appreciate if you could take a pull of my branch and try running the tests.

saschagrunert:
> > > I have however run into an error with a failed test for `TestToMan` function. @saschagrunert could you take a look as to why that might be? Meanwhile, I will continue development and pushing commits.
> > 
> > 
> > This seems to be solved, right? I don't see any further influences on this test case.
> 
> I don't think so. I was also very surprised to see this. I'd really appreciate if you could take a pull of my branch and try running the tests.

I did, and running the unit tests works on my local machine. :)

AudriusButkevicius:
I think this is sort of a anti feature. 

You could do validation of flags yourself in the command, as this is essentially what you are doing anyway, just at the flag definition time oppose to command invocation time. 

Also, flags might be contextual to the command. Imagine cases kubectl, where it supports --ouput json flag for most commands other than describe. This simply would not work.

Also, the whole min max check but for anything else you need a validator is somewhat pointless. Why don't we have IsPowerOfN validation? 

If we decide this is a good idea, this should probably do `Validators: []Validator{Min(0), Max(10), Mod(2)}`, but in general I think this is a "and a kitchen sink" type of feature.

asahasrabuddhe:
what do we do about this guys? do I continue working with this or abandon ship and move on?

lynncyrin:
I personally would prefer we not work any new features for a while, there's a largish backlog of bugs / tech debt / strange behavior that can be addressed instead

AudriusButkevicius:
I guess it's more of a ideological question, do we believe the library should support validation, and if yes, is this the right implementation (and implementation that cannot do different validation of global flags based on the command in question)

lynncyrin:
I believe the library would be improved if it supported validation. But I personally would rather avoid being pulled into more heated discussions about what a library should / shouldn't be doing - those kinds of discussions burn me out.

AudriusButkevicius:
Well, you are the maintainers, you have to make the shots, or are we going to merge everything unconditionally, without thinking about the direction of where this library is going?

lynncyrin:
I'm going to defer all new feature PRs to everyone else. Either defer or state that I won't be engaging with them until more bugs or fixed.

lynncyrin:
...which includes this PR 🙂

So my formal opinion on "flags validation" is that I personally won't be approving it, and defer any decisions about this feature to the other maintainers.

AudriusButkevicius:
@saschagrunert do you have an opinion?

saschagrunert:
> @saschagrunert do you have an opinion?

Generally I think that the feature would be a valuable addition, independently of the implementation details. So I propose to go forward and bring this one in since a lots of the main work has been already done.

I see your concerns @AudriusButkevicius and I guess having some sort of prioritization for the next couple of weeks / releases, like @lynncyrin did, is already something other projects don’t have. I’m convinced that we're all smart enough to add features without messing up everything :)

lynncyrin:
I consider this PR blocked pending more discussion

PR #857, opened 2019-08-12
Add `TakesFile` to fish shell completion

saschagrunert:
The new `TakesFile` flag will be now consumed by the fish shell
completion generator.


saschagrunert:
We need https://github.com/urfave/cli/pull/860 before continuing here.

saschagrunert:
Rebased on top of the latest master to get the latest takes-file related changes in.

fish.go
```diff
@@ -126,11 +126,15 @@ func (a *App) prepareFishFlags(
 
 		var completion strings.Builder
 		completion.WriteString(fmt.Sprintf(
-			"complete -c %s -f -n '%s'",
+			"complete -c %s -n '%s'",
 			a.Name,
 			a.fishSubcommandHelper(previousCommands),
 		))
 
+		if !flag.GetTakesFile() {
```
AudriusButkevicius: Sorry, why not just flag.TakesFile? What's the point of this getter?

flag.go
```diff
@@ -96,6 +96,10 @@ type DocGenerationFlag interface {
 	// GetValue returns the flags value as string representation and an empty
 	// string if the flag takes no value at all.
 	GetValue() string
+
+	// GetTakesFile indicates that the command takes a file as input, which
+	// will be used for documentation and completion generation purposes.
+	GetTakesFile() bool
```
AudriusButkevicius: Given we have TakesValue() this should be TakesFile()?

flag-gen/templates/cli_flags_generated.gotpl
```diff
@@ -60,6 +60,12 @@ func (f {{ $flag.Name }}Flag) GetValue() string {
 	{{ $flag.ValueString }}
 }
 
+// GetValue returns the flags value as string representation and an empty
+// string if the flag takes no value at all.
+func (f {{ $flag.Name }}Flag) GetTakesFile() bool {
+	{{ if eq $flag.TakesFile true }}return f.TakesFile{{ else }}return false{{ end }}
+}
```
AudriusButkevicius: Just don't generate these for things that do not take files, and do an interface assertion in the generation code.

flag.go
saschagrunert: Golang will complain if we have a method and a value with the same name and it seems that we follow a `Get...` pattern for accessing the values from the interface. WDYT?

flag-gen/templates/cli_flags_generated.gotpl
saschagrunert: Hm, the issue is that we have this method as part of the `DocGenerationFlag`, which we already type assert in fish.go. I we remove the function from the other flags we would not be able any more to generate docs/completions for them.

AudriusButkevicius: Why? You can assert something more than once.

flag.go
AudriusButkevicius: Then why do we a method at all?

saschagrunert: I see, I removed it completely.

flag-gen/templates/cli_flags_generated.gotpl
saschagrunert: Okay, I added a helper function to check if we can access the `TakesFile` variable.

saschagrunert:
PTAL, It's now way simpler and the tests are passing as before :)

fish.go
```diff
@@ -158,6 +160,25 @@ func (a *App) prepareFishFlags(
...
+	addFileExclusionFlag := true
+	if f, ok := flag.(GenericFlag); ok && f.TakesFile {
+		addFileExclusionFlag = false
+	}
+	if f, ok := flag.(StringFlag); ok && f.TakesFile {
+		addFileExclusionFlag = false
+	}
+	if f, ok := flag.(StringSliceFlag); ok && f.TakesFile {
+		addFileExclusionFlag = false
+	}
```
AudriusButkevicius:
```
switch v := flag.(type) {
   case StringFlag, GenericFlag, StringSliceFlag:
        if !v.TakesFile {
            completion.WriteString(" -f")
        }
}
```

saschagrunert: The comma separated type switch does not seem to work, fallthrough does also not work so I had to add dedicated statements for each type.

AudriusButkevicius:
Why does fallthrough not work?  You just need to use fallthrough keyword. Or are you saying it does not work for tyoe switches?

saschagrunert:
> Why does fallthrough not work? You just need to use fallthrough keyword. Or are you saying it does not work for tyoe switches?

The compiler says that fallthrough is not allowed in type switches :(

fish.go
```diff
@@ -158,6 +160,27 @@ func (a *App) prepareFishFlags(
 	return completions
 }
 
+func fishAddFileFlag(
+	flag Flag,
+	completion *strings.Builder,
+) {
```
AudriusButkevicius: Is gofmt happy with this?

saschagrunert: Yes :)

AudriusButkevicius: Looks fairly non-standard. There aren't that many arguments that it would require this ackward breaking. Is this standard in this project or are you bringing your own twist to this?

Reason I am asking is that format consistency is important in a project with multiple collaborators

saschagrunert: It is standard in this file, but since I'm the main author I would say it is just my own twist. I will align with the project standard (single line) and change the other functions, too.

PR #858, opened 2019-08-12
Revert "Add `TakesFile` indicator to flag"

AudriusButkevicius:
Reverts urfave/cli#851

Suggest we rethink this before this spreads too far.


lynncyrin:
@AudriusButkevicius can you give a more thorough explanation of why you're opening this PR? I don't have enough context to make a decision here.

lynncyrin:
If we want to revert this, I think the reversion needs to come with a thorough explanation in the PR description. I don't understand why would should be reverting this, which means I am very likely to unknowingly approve similar changes like this in the future. If there's some API stability north star we are working towards, then it would help me out if that north star was described. My current state is that I have no idea why someone would / would not add a certain feature, and I need to be educated on this if the goal is for me to make decisions and moving the API to a good place.

asahasrabuddhe:
Let's first finalize the discussion on #851 and then come to a conclusion about whether or not this PR is even necessary

AudriusButkevicius:
The rationale is on the original PR

lynncyrin:
I consider this PR blocked pending more discussion, similarly to https://github.com/urfave/cli/pull/860

AudriusButkevicius:
We can probably close this and merge #860 before we release anything.

lynncyrin:
> We can probably close this and merge #860 before we release anything.

👍

PR #860, opened 2019-08-13
Update `TakesFile` flag to apply only to supported flags

saschagrunert:
This changes the code generation to apply the `TakesFile` flag only at types where it makes sense.

Closes https://github.com/urfave/cli/pull/858


AudriusButkevicius:
Not related to this PR, but just to demonstrate my point, we are adding 1 line per flag to a json file, 3 lines to a template, to add 2 lines to a go file.

Everything that would have to be hand written in the go file is somewhat already written in the JSON file.

lynncyrin:
> Not related to this PR, but just to demonstrate my point, we are adding 1 line per flag to a json file, 3 lines to a template, to add 2 lines to a go file.
>
> Everything that would have to be hand written in the go file is somewhat already written in the JSON file.

Is there idea here that we should consider ditching the json file + template based code generation? Because if so, then I totally agree @AudriusButkevicius!

lynncyrin:
> Is there idea here that we should consider ditching the json file + template based code generation? 

Followup => https://github.com/urfave/cli/issues/870

saschagrunert:
Okay, so I think this is on hold now.

saschagrunert:
@lynncyrin I put it into status blocked but want to be sure if this is common practice here :innocent:

lynncyrin:
👍 blocked totally makes sense here 👍

AudriusButkevicius:
I don't think we have to be blocked on killing flag generation, as I assume we'll just take the last generated output and maintain it by hand, hence this can go ahead?

lynncyrin:
I was moreso thinking that this PR is blocked b/c we need to discuss the intention behind the `TakesFile` property - and determine if we want to support it.

lynncyrin:
I was looking at the various `TakesFile` PRs, and this one is approved already. AudriusButkevicius mentioned closing https://github.com/urfave/cli/pull/858, so I just did that. So as far as I can tell (given the new information I now how) this PR can move forward.

AudriusButkevicius:
bring it up to date and lets merge it.

PR #862, opened 2019-08-14
Rewrite the "pinning to v1" section.

russoj88:
@lynncyrin here's the PR for the README.  Let me know if you'd like to change anything!

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,13 @@ import (
 
 ### Using `v1` releases
 
-Use the github URL to use the v1 branch.  This is illustrated in the Getting Started section below.
+$ go get github.com/urfave/cli
+
+package main
+
+import (
+  "github.com/urfave/cli"
+)
 
 Do not use the gopkg.in URL as it does not work for versions >= 1.21.0.
 
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ applications in an expressive way.
 - [Installation](#installation)
   * [Supported platforms](#supported-platforms)
   * [Using the `v2` branch](#using-the-v2-branch)
-  * [Pinning to the `v1` releases](#pinning-to-the-v1-releases)
+  * [Using `v1` releases](#using-v1-releases)
 - [Getting Started](#getting-started)
 - [Examples](#examples)
   * [Arguments](#arguments)
--- a/README.md
+++ b/README.md
@@ -110,14 +110,17 @@ import (
 
 ### Using `v1` releases
 
+```
 $ go get github.com/urfave/cli
+```
 
-package main
-
+```go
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -108,25 +108,12 @@ import (
 ...
 ```
 
-### Pinning to the `v1` releases
+### Using `v1` releases
```
lynncyrin: The build is failing because this line needs to be changed in the table of contents too 🙂

README.md
```diff
@@ -108,25 +108,12 @@ import (
...
 
-Similarly to the section above describing use of the `v2` branch, if one wants
-to avoid any unexpected compatibility pains once `v2` becomes `master`, then
-pinning to `v1` is an acceptable option, e.g.:
+Use the github URL to use the v1 branch.  This is illustrated in the Getting Started section below.
```
lynncyrin:
```suggestion
$ go get github.com/urfave/cli

package main

import (
  "github.com/urfave/cli"
)
```

I know its _right below_, but for some things (like install directions) its fine to copy paste IMO.

README.md
```diff
@@ -108,25 +108,12 @@ import (
...
-```
-
-``` go
-...
-import (
-  "gopkg.in/urfave/cli.v1" // imports as package "cli"
-)
-...
-```
+Do not use the gopkg.in URL as it does not work for versions >= 1.21.0.
```
lynncyrin:
```suggestion
```

I would remove this line entirely. It might be enough to say _"do this correct thing"_ (eg. use the github url) and avoid saying _"don't do this wrong thing"_ (eg. don't use the gopkg.in url). If it turns out people are still mentioning gopkg.in a bunch, then we can remove it.

README.md
russoj88: Fixed

russoj88:
@lynncyrin looks like gfmrun isn't building.  Not sure what that is.  Any thoughts on why the build is  breaking?

asahasrabuddhe:
> @lynncyrin looks like gfmrun isn't building. Not sure what that is. Any thoughts on why the build is breaking?

gfmrun - github flavoured markdown runner is a tool that extracts code snippets from within the README.md and executes them as go code. If that code refuses to build and execute, it fails. I think an edit in the README.md is breaking the build. A quick look shows me that gopkg has been renamed to github.com when importing the package. I don't know what immediate effect this has but it can be prod in the right direction for you to start debugging.

Let me know what you come up with otherwise, I'd be taking a closer look at this tomorrow

lynncyrin:
I have no idea how gfmrun works, so anyone else's guess is as good as mine 🙂

Issue #850, opened 2019-08-08
Release 1.21.0 regression when using pass-through commands


lynncyrin:
In `1.20.0` this used to work

```
myCLI myCommand --someArg someInput docker run --rm ubuntu bash
```

In `1.21.0` it's broken, with this error message

```
Incorrect Usage: flag provided but not defined: -rm
```

The issue here is that the CLI is reading `--rm` as input to `myCommand`, rather than input to `docker run`.

lynncyrin:
Pass through commands in general are (...) a bit difficult to support. The hotfix here is to invoke docker like this

```diff
- myCLI myCommand --someArg someInput docker run --rm ubuntu bash
+ myCLI myCommand --someArg someInput -- docker run --rm ubuntu bash
```

lynncyrin:
Tangentially, when you hit this bug it exits with code `0`, which can very easily "gotcha!" you if you're expecting a given command to failed on errors.

lynncyrin:
Ah, the lack of an error response was because my specific CLI wasn't checking for an error from `app.Run`. I think a previous version of the readme suggested not checking for `app.Run` errors.

lynncyrin:
I'm going to start looking into this next week 📆

lynncyrin:
If anyone drops into this issue and **definitely** knows how to fix the problem, feel free to take this off my hands ^_^ You'll have to race me for the PR though 🏃‍♀

lynncyrin:
Just ran a bisect, and it looks like this is the commit that caused the bug https://github.com/urfave/cli/commit/df562bf1a8626f2d16f91fcbf7230a5bdca3d592

lynncyrin:
Which was merged in this PR https://github.com/urfave/cli/pull/691

Issue #826, opened 2019-08-03
maint: release 2.X


lynncyrin:
If anyone is interested in the work of merging the `master` branch into v2, add let us know here with a _"🤚I'm interested in working on this"_ comment! 

---

The comments below describe a random assortment of release `2.X` concerns 📝

lynncyrin:
I'm thinking that I'll keep the current 2.0 branch as release 2.0 forever, and the very next release I make will be 2.1.

lynncyrin:
I'm also going to try to merge 2.0 into master ASAP. I may keep maintaining a 1.0 branch with bugfixes? If people are interested in that.

AudriusButkevicius:
Isn't that a breaking change in terms of API? I am not sure its cool everyone that don't do vendoring. Also, current api and implementation is battle tested, v2 is not.

lynncyrin:
> Isn't that a breaking change in terms of API? I am not sure its cool everyone that don't do vendoring.

My understanding of the go community is that the vast majority of people install from a tag, eg. `1.2.4` - and don't install from the default branch, eg `master`. So there's no API compatibility guarantee around the `master` branch.

That said, I just checked the `yaml` package and apparently they changed their default branch to be `v2` https://github.com/go-yaml/yaml. So that may be the way to go here.

> Also, current api and implementation is battle tested, v2 is not.

That's true! Are you suggesting we have a beta release period, or similar?

lynncyrin:
Got a comment via Stephen Houston in the gophers general channel on slack

> Traditionally I would say you keep master based off your current stable with bugfixes added in to prepare for point releases ... I.e. if your current stable is 1.0 ... master is on 1.0 but with bugfixes added in to be able to tag 1.1 etc... you have a develop branch that is for features/future releases... when you get ready to do a new major release, i.e. v1 -> v2... master becomes develop and you tag... and then master follows 2.0 and gets bugfixes for 2.0 in order to do point releases 2.1 etc... and develop moves on to new features/future releases.

lynncyrin:
to echo @tsingson's comment here 

> this gopkg.in/urfave/cli.v2 is working fine for me in product about half year.

https://github.com/urfave/cli/issues/824#issuecomment-517824045

marwan-at-work:
Since v2 was never officially released, it might be okay to make minimal breaking changes to the v2 branch before merging and releasing it? 

I'm happy either way, but if that's more convenient then it should be okay since people know that v2 has not been released yet

[...]

PR #872, opened 2019-08-23
Fix 1.21.0 pass through regression

lynncyrin:
## Changes

- adds an explicit "regression test" in its own file, distinct from the other tests
- changes the `reorderArgs` logic to ignore args that aren't defined for the command you are running
- refactors `reorderArgs` for understandability
- changes a `TestCommandFlagParsing` test case that was expecting an error, to no longer expect an error

## Motivation

Fixes https://github.com/urfave/cli/issues/850

https://github.com/urfave/cli/issues/850 caused a _stop everything and fix this now_ P0 bug that since then taken a few dozen hours of work to fully root-out and stabilize _(eg. at my dayjob)_. Because of that, I've been working to get a change into this package to avoid that occurring in the future.

## What I'm looking for here

- I feel like it's valuable to keep the new regression test I added distinct from the other tests, but do let me know if that doesn't make sense. I'm very ok with moving that test into `command_test.go`
- I put a lot of effort into making the logic in `reorderArgs` as clear as possible, please let me know if I can add more context / docs / update variable names / etc.
- The change to `TestCommandFlagParsing` is conceptually dangerous, but in a practical sense I think it's actually safe. Please let me know if you agree or disagree? I think it's safe in a practical sense because user's shouldn't be strictly relying on this package to raise _"flag provided but not defined"_ errors in all kinds of nuanced places.

## Future Work

The whole idea of argument re-ordering feels dangerous to me. Immediately after this PR I plan on investigating how I could potentially remove that functionality, without impacting the API.


lynncyrin:
https://github.com/urfave/cli/pull/691 <= I looks like this PR is what caused the bug, but it doesn't revert cleanly. Which is to be expected, since it's 2 years old 😅 I'll figure this out

lynncyrin:
I believe what happened was that the reorder args functionality here https://github.com/urfave/cli/pull/691/files#diff-9c42b524cb14ca001c61267826cbefb1R206 started accidentally pulling in args meant for pass-through commands. So like the `--no-trunc` in my regression test.

lynncyrin:
There's a line in the flag re-ordering function that checks for `--`, https://github.com/urfave/cli/blob/6cc7e987c4fa553caa5014c7dbc1e7acaea9f0f1/command.go#L223. This hints to me that we already kind of knew that flag re-ordering could cause issues with pass-through commands.

command.go
```diff
@@ -214,34 +214,60 @@ func (c *Command) useShortOptionHandling() bool {
...
 	}
 
-	return append(flags, nonflags...)
+	return append(reorderedArgs, remainingArgs...)
+}
+
+// argIsFlag checks if an arg is one of our command flags
+func argIsFlag(commandFlags []Flag, arg string) bool {
+	// this line turns `--flag` into `flag`
+	arg = strings.Replace(arg, "-", "", -1)
```
saschagrunert: This turns `--my-flag` into `myflag`, which is not intended, right?

command.go
```diff
@@ -214,34 +214,60 @@ func (c *Command) useShortOptionHandling() bool {
...
+	// this line turns `flag=value` into `flag`
+	arg = strings.Split(arg, "=")[0]
+	// look through all the flags, to see if the `arg` is one of our flags
+	for _, flag := range commandFlags {
+		for _, key := range strings.Split(flag.GetName(), ",") {
+			if key == arg {
```
saschagrunert: I think we have to trim the `key` from white space.

command.go
```diff
@@ -214,34 +214,60 @@ func (c *Command) useShortOptionHandling() bool {
 	return c.UseShortOptionHandling
 }
 
-// reorderArgs moves all flags before arguments as this is what flag expects
-func reorderArgs(args []string) []string {
-	var nonflags, flags []string
+// reorderArgs moves all flags (via reorderedArgs) before the rest of
+// the arguments (remainingArgs) as this is what flag expects.
+func reorderArgs(commandFlags []Flag, args []string) []string {
+	var remainingArgs []string
```
saschagrunert:
```suggestion
	var remainingArgs, reorderedArgs []string
```

command.go
lynncyrin: nice catch 👍

lynncyrin:
I'm putting this `on hold` until the v2 released plan is figured out => https://github.com/urfave/cli/issues/826

lynncyrin:
👋  I have returned to this PR

command.go
lynncyrin: ✔️ 	// this line turns `--flag` into `flag`

lynncyrin: ✔️  https://github.com/urfave/cli/pull/872/commits/51259808fff1f7384b556f6e127a090fe0362305

command.go
lynncyrin: ✔️  https://github.com/urfave/cli/pull/872/commits/64d3555482218336121cd318c27454041cc479a1

command.go
```diff
@@ -219,34 +219,73 @@ func (c *Command) useShortOptionHandling() bool {
...
+		return false
+	}
+	// this line turns `--flag` into `flag`
+	if strings.HasPrefix(arg, "--") {
+		arg = strings.Replace(arg, "-", "", 2)
+	}
+	// this line turns `-flag` into `flag`
+	if strings.HasPrefix(arg, "-") {
+		arg = strings.Replace(arg, "-", "", 1)
+	}
```
saschagrunert: We could deduce the logic to this one (non blocking depending on which one you prefer):
```suggestion
arg = strings.TrimLeft(arg, "-")
```

lynncyrin: I'm try grab this one in followup I think, I wanna see if I can ship this since it's a regression case.

saschagrunert: Sure, sounds good.

lynncyrin:
review ping @AudriusButkevicius / @asahasrabuddhe

command.go
```diff
@@ -219,34 +219,73 @@ func (c *Command) useShortOptionHandling() bool {
...
-// reorderArgs moves all flags before arguments as this is what flag expects
-func reorderArgs(args []string) []string {
-	var nonflags, flags []string
+// reorderArgs moves all flags (via reorderedArgs) before the rest of
+// the arguments (remainingArgs) as this is what flag expects.
+func reorderArgs(commandFlags []Flag, args []string) []string {
+	var remainingArgs, reorderedArgs []string
 
-	readFlagValue := false
+	nextIndexMayContainValue := false
```
asahasrabuddhe:
```suggestion
	var nextIndexMayContainValue bool
```

This is also a possibility as it would be false by default!

asahasrabuddhe:
This looks good :+1: I have added a very optional change. You may or may not choose to do it :smile:

command.go
lynncyrin: I'm gonna opt against adding this into the PR - but only because I wanna merge this ASAP so that we can fix the regression 🔧

PR #873, opened 2019-08-23
build: show failures when running tests

lynncyrin:
## Before

When you have failing tests, you get output like this

```
go run build.go generate
go run build.go vet
go run build.go test
exit status 1
exit status 1
Command exited with code 1
```

The above output is the literal terminal output, it doesn't tell the user what the source of the test failure is.

Adding the verbose flag like so

```
go run -v build.go test
```

Does not fix the issue.

## After

When you have failing tests, you get output like this

```
go run -v build.go test
...
--- FAIL: TestToMan (0.00s)
...
FAIL
coverage: 84.0% of statements
FAIL	github.com/urfave/cli	0.089s
exit status 1
exit status 1
```

This gives the user enough information to resolve the issue.

## Changes

I hooked `Stdin`, `Stdout`, `Stderr` into the ox exec calls.

## Impact

This change effects the local build system, and does not relate to the functionality of the cli.

## Context

This change is followup from https://github.com/urfave/cli/pull/836

### DIFF
--- a/build.go
+++ b/build.go
@@ -52,8 +52,8 @@ func main() {
 	}
 }
 
-func runCmd(args ...string) error {
-	cmd := exec.Command(args[0], args[1:]...)
+func runCmd(arg string, args ...string) error {
+	cmd := exec.Command(arg, args...)
 
 	cmd.Stdin = os.Stdin
 	cmd.Stdout = os.Stdout
--- a/build.go
+++ b/build.go
@@ -53,7 +53,12 @@ func main() {
 }
 
 func runCmd(args ...string) error {
-	cmd := exec.Command(args[0], args[1:]...)
+	var cmd *exec.Cmd
+	if len(args) > 1 {
+		cmd = exec.Command(args[0], args[1:]...)
+	} else {
+		cmd = exec.Command(args[0])
+	}
 
 	cmd.Stdin = os.Stdin
 	cmd.Stdout = os.Stdout
--- a/build.go
+++ b/build.go
@@ -53,12 +53,7 @@ func main() {
 }
 
 func runCmd(args ...string) error {
-	var cmd *exec.Cmd
-	if len(args) > 1 {
-		cmd = exec.Command(args[0], args[1:]...)
-	} else {
-		cmd = exec.Command(args[0])
-	}
+	cmd := exec.Command(args[0], args[1:]...)
 
 	cmd.Stdin = os.Stdin
 	cmd.Stdout = os.Stdout
--- a/build.go
+++ b/build.go
@@ -52,14 +52,18 @@ func main() {
 	}
 }
 
-func VetActionFunc(_ *cli.Context) error {
-	cmd := exec.Command("go", "vet")
+func runCmd(args ...string) error {
+	cmd := exec.Command(args[0], args[1:]...)
 
 	cmd.Stdin = os.Stdin
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 
 	return cmd.Run()
+}
+
+func VetActionFunc(_ *cli.Context) error {
+	return runCmd("go", "vet")
 }
 
 func TestActionFunc(c *cli.Context) error {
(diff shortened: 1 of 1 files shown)


build.go
```diff
@@ -52,7 +53,13 @@ func main() {
 }
 
 func VetActionFunc(_ *cli.Context) error {
-	return exec.Command("go", "vet").Run()
+	cmd := exec.Command("go", "vet")
+
+	cmd.Stdin = os.Stdin
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
```
saschagrunert: We may write a wrapping function for this recurring pattern, but this is a non-blocking nit pick.

AudriusButkevicius: Also, I assume there should be some function exec that sets up the process this way. I find it strange that this is not the default.

lynncyrin: I also find it strange that there isn't a function in that package that does this by default. Here is the documentation => https://golang.org/pkg/os/exec/#Command

AudriusButkevicius: eek, horrible.

build.go
```diff
@@ -51,8 +52,18 @@ func main() {
 	}
 }
 
+func runCmd(args ...string) error {
+	cmd := exec.Command(args[0], args[1:]...)
```
saschagrunert: I’m not sure if this hurts because it is only the `build.go` but it can result in an index out of bounds access if `len(args) < 1`.

lynncyrin: https://github.com/urfave/cli/pull/873/commits/1095838cca9d596e55cff88bcb35b67cf83bf4e4 ✔️

build.go
lynncyrin: https://github.com/urfave/cli/pull/873/commits/29ad6ee6ad7e02dbec3334e8843bb6711c011b55 ✔️

build.go
```diff
@@ -51,8 +52,23 @@ func main() {
 	}
 }
 
+func runCmd(args ...string) error {
+	var cmd *exec.Cmd
+	if len(args) > 1 {
+		cmd = exec.Command(args[0], args[1:]...)
+	} else {
+		cmd = exec.Command(args[0])
+	}
```
AudriusButkevicius: I think args[1:] would return a nil slice, so it should be ok?

lynncyrin: I think the code would panic without the length check. Either way, if something is broken we can fix it.

AudriusButkevicius: I think not:
https://play.golang.org/p/wxxS-Es9Qe6

lynncyrin: Do you have a specific suggestion as to how this code should change?

AudriusButkevicius:
```
	if len(args) > 1 {
		cmd = exec.Command(args[0], args[1:]...)
	} else {
		cmd = exec.Command(args[0])
	}
```
becomes
````
	cmd = exec.Command(args[0], args[1:]...)
```

lynncyrin: Right, that's what I had before and it was suggested I change it. I'll change it back.

lynncyrin: https://github.com/urfave/cli/pull/873/commits/c2d1a132082e3b02a219e61eeef49da364d2c315 ✔️

AudriusButkevicius: Actually the suggestion was something else. There is no args is nil check.

lynncyrin: > Actually the suggestion was something else. There is no args is nil check.

Got it 👍  I'll add that

lynncyrin: Ah, I cant. It's not clear what you would return if there are no args.

AudriusButkevicius: If we have the same signature as exec (arg, args...), it makes it impossible to have 0 arguments.

lynncyrin: Hey so, this is a PR for cleaning up a specific problem with the build system? I'm not etching code into a stone tablet here, we can change it if it's problematic. I have more important things to do than engage in _"what if this..."_ comments about the build system.

 I'm going to stop responding to this thread now, because it's starting to tank my mental health. If there's some more changes that should be made here, someone else can push to my branch and make them.

AudriusButkevicius: So I've made the change, but I'd suggest not starting PRs if you don't have the will to finish them.

It wastes everyones time, your time on writing code that will not get merged as it's not finished, and our time for reviewing code that will not be driven to completion.

PR #879, opened 2019-08-29
Escape single quotes in fish shell completion

saschagrunert:
Single quotes can break the generated fish shell completion and should
be escaped correctly.


saschagrunert:
Rebased on top of the latest master.

PR #883, opened 2019-09-09
Remove flag generation

asahasrabuddhe:
Removes the old flag generation logic written in Go and also removes remaining legacy python scripts for running tests etc.


asahasrabuddhe:
moving the flags to their own directory, say flag, would also introduce a collision with the flag package from the standard library that we are using

asahasrabuddhe:
:sunglasses: It's all green now! Please let me know if this is Ok to be merged!

lynncyrin:
You're good to go in my opinion!

AudriusButkevicius:
I am not that concerned with flag package colliding in name, we can call it flags. I am more concerned about the repo layout and tons of files in a single directory.

asahasrabuddhe:
Let me give it a shot and see how it goes

asahasrabuddhe:
@AudriusButkevicius I took a look at the code and it is going to be much difficult than just moving a couple of files into a directory. The dependencies are spread out across multiple files and we would need to carve out a lot of different packages to avoid import loops.

If you have a better way in mind, I am open to suggestions

AudriusButkevicius:
Sure, my approval is already there, even though I am sad about the state of affairs, but I understand there is not much we can do here.

asahasrabuddhe:

Issue #884, opened 2019-09-09
Mismanaged Branches


asahasrabuddhe:
I just wanted to share something that I have experienced working with this codebase. Currently, the master branch is not aligned with the latest stable release. This is not correct. As per convention, the master branch must always contain production-ready code which in our case is tagged at v1.21.0

We could have another branch say develop where we could first merge PRs and then eventually carve out releases from that branch once we are satisfied.

The major issue here is that people not using Go Modules would get master code which is not yet stable or consistent with our idea of the latest release that we would support and maintain.

If we do make this change, we would have better management of code for ourselves as well as help fix a lot of problems for people not currently using Go Modules.

lynncyrin:
I'm 👍! I think we would just need a some instructions in `CONTRIBUTING.md` describing how this works.

lynncyrin:
It would also be nice to have an example of a repo that does branches correctly 🙏

asahasrabuddhe:
@lynncyrin Unfortunately I do not have a repo that perfectly follows the model that I have in mind. However, I can share this: https://nvie.com/posts/a-successful-git-branching-model/

This blog post is the source of inspiration behind my idea. I have already implemented this in couple of repos at work which I am not able to share. I would very much like to have this flow in place here in our projects.

What we would need to do now is follows:

1. Rename current master branch to develop.
2. Create a new branch called master from the current release commit 1.21.0 or 1.22.0
3. Update the CONTRIBUTING.md with new instructions. In short, every branch will be named to indicate the purpose. feature, hotfix, bug, patch etc. All PRs would only be directed to develop branch. And we as maintainers would then choose the PRs and craft out a release branch which would then eventually get merged with the master and we would have a new release.

I don't think we have enough rights to do #1. We would need @AudriusButkevicius 's help if he is on board. Then we can do #2 and finally I will update CONTRIBUTING.md with #3 

What do you guys think?

@saschagrunert your opinion would also be needed in this matter.

AudriusButkevicius:
I somewhat disagree. Master is for development, tags are for versioning. Sure, people without go mod might have some pain, but various different options for addressing this in go have existed for years. Go mod is our vendoring system and people should use it.

I'd like to see an example (a go repo) where this is done differently. Even go itself follows the same model.

asahasrabuddhe:
If not this then I suppose we should at least have dedicated release branches to avoid the mess that I am seeing at the moment.

AudriusButkevicius:
[...]

PR #885, opened 2019-09-09
Go modules support

asahasrabuddhe:
In line with Go's release policy, dropping support for Go 1.10 and adding support for Go 1.13

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,6 @@ go:
 os:
   - linux
   - osx
-  - windows
 
 env:
   GO111MODULE=on
@@ -23,6 +22,7 @@ before_script:
   - go get github.com/urfave/gfmrun/cmd/gfmrun
   - go get golang.org/x/tools/cmd/goimports
   - npm install markdown-toc
+  - go mod tidy
 
 script:
   - go run build.go generate
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ os:
 
 env:
   GO111MODULE=on
+  GOPROXY=https://proxy.golang.org
 
 cache:
   directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,16 +20,14 @@ cache:
     - node_modules
 
 before_script:
-  - go get github.com/shurcooL/vfsgen || true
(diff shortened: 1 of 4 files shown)


asahasrabuddhe:
At a very interesting point here:

This PR would never have all the test cases pass. There are two things going on here:

1. The build fails at the gfmrun stage where the gfmrun is not able to successfully run two of the README.md examples reference ShortOptionHandling (in master) which is not available in the latest release 1.21.0 of the cli.

2. Support I were to remove these references from the code and push, the gfmrun stage would pass but the build would again fail at the git diff stage as the README.md pushed with the PR would again not match the README.md at the master.

This brings me to the issue: #884

lynncyrin:
> two of the README.md examples reference ShortOptionHandling (in master) which is not available in the latest release 1.21.0 of the cli

I have a PR for releasing `1.22.0` with `ShortOptionHandling ` here => https://github.com/urfave/cli/pull/882

.travis.yml
```diff
@@ -1,31 +1,30 @@
...
   - 1.11.x
   - 1.12.x
+  - 1.13.x
 
 os:
   - linux
   - osx
 
+env:
+  GO111MODULE=on
```
saschagrunert: do we want to set `GOPROXY=https://proxy.golang.org` to speed things up?

asahasrabuddhe: Good Idea! I must have missed it. I have added it now.

asahasrabuddhe:
This PR is based of the original flags PR.

asahasrabuddhe:
I can remove that and separate it if you like

AudriusButkevicius:
Right, either lets make the changes completely independent, or not file them at the same time if they are, and wait for required change to land.

asahasrabuddhe:
Fair enough. I will clean this up

asahasrabuddhe:
I have rebased this and cleaned up.

asahasrabuddhe:
This build will fail until such time that 1.22.0 is merged and UseShortHandOptions is available for the tests to succeed.

asahasrabuddhe:
All green here too :)

PR #886, opened 2019-09-09
Adds ability to create custom flag errors for required flags

anberns:
App creators can provide custom error messages for required flags through the addition of the `RequiredFlagErr` field, which implements the underlying internal interface. Apps can have a combination of default and custom flag messages. Tests and documentation are included.


AudriusButkevicius:
Why do you need this? Why is the default message not sufficient?

asahasrabuddhe:
Although we are removing flag generation, looking at the current state of development, it is unwise to edit a file specifically marked as not to be edited...

lynncyrin:
> Why do you need this? Why is the default message not sufficient?

I've done user testing of this CLI package internally at Textio, and in approximately 50% of the cases the default error message isn't sufficient.

Also from an implementer point of view, having required flags is useless if I can't define custom error messages. I need to be able to define *why* the flag is required and *what* values should be input for it, inside the error message.

Also from an implementer point of view, something like 40% of the bug reports I get are essentially "the error message from urfave/cli" was confusing. If you don't define an api for error messages, people will start doing ridiculous things like matching on the literal text "required flag". Adding an api here gets ahead of that issue.

Finally, in my experience getting a "why do you need this" comment in a pull request is very anxiety inducing. I would appreciate if we did two things:

- surface any concerns like this inside the GitHub issue, rather than the pull request. I'll create a new "approval required" label to reflect this.
- used language that respects and appreciates new contributors, like for example "Can you help me understand what's driving the need for this feature?"

context.go
```diff
@@ -289,39 +289,67 @@ func normalizeFlags(flags []Flag, set *flag.FlagSet) error {
...
+}
+
+func (e *errRequiredFlags) getMissingCustomFlags() []string {
+	return e.missingCustomFlags
 }
 
 func checkRequiredFlags(flags []Flag, context *Context) requiredFlagsErr {
-	var missingFlags []string
+	var missingDefaultFlags []string
+	var missingCustomFlags []string
```
lynncyrin: ❓ _(blocking question)_

Can you help me understand the motivation behind separating `missingDefaultFlags` and `missingCustomFlags`? Thinking apart this functionality conceptually, I'm not sure that separation is strictly required.

flag.go
```diff
@@ -83,6 +83,15 @@ type RequiredFlag interface {
 	IsRequired() bool
 }
 
+// RequiredFlagErr is an interface that allows users to redefine errors on required flags
+// it allows flags with user-defined errors to be backwards compatible with the Flag interface
+type RequiredFlagErr interface {
+	Flag
+
+	IsCustom() bool
```
lynncyrin: ❓ _(blocking question)_

This is probably related to the other blocking question -- can you help me understand the motivation behind the `IsCustom` attribute here?

AudriusButkevicius:
So I am not a native English speaker, hence my expressions don't always give warm fuzzy feelings.

The fear I have with this change is the ever growing surface of the flag structs. I wonder (and this is just me wonder, it's not a suggestion of a direction), whether App.OnRequiredFlagMissing is not a better option. I can see how this could be a worse option, but I am just paranoid about growing the flag surface.

context.go
anberns: Yeah, that was just the way I came up with to allow for both default and custom flag errors in the same app. It does seem longwinded. I'd be glad to update it if you have suggestions.

flag.go
anberns: There's also probably a better way to do this and I played around with removing the field but again, helps allow both default and custom messages.

README.md
```diff
@@ -560,6 +560,85 @@ Will result in help output like:
...
+
+```go
+package main
+
+import (
+  "log"
+  "os"
+  "strings"
+
+  "../cli"
```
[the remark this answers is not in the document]
anberns: Good catch. Updated.

flag_generated.go
```diff
@@ -9,16 +9,22 @@ import (
...
+	Name            string
+	Usage           string
+	EnvVar          string
+	FilePath        string
+	Required        bool
+	RequiredFlagErr FlagErr
+	Hidden          bool
+	TakesFile       bool
+	Destination     *bool
 }
```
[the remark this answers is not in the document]
anberns: Yeah no problem. Editing a 'do not edit' file is not something I would normally do, just wanted to focus on the flag errors while flag generation was being discussed / updated. I'll remove the file from this pr.

asahasrabuddhe:
You'd want to rebase now

context.go
lynncyrin: The default and custom errors can exist within the same app if you always use the default error when the custom error is undefined.

So instead of having this

```golang
      Required: true, 
      RequiredFlagErr: cli.FlagErr{
        Custom: true, 
        Message: `There's a problem: "lang" is a required flag` ,
      },
```

You have this

```golang
      Required: true, 
      RequiredFlagErr: `There's a problem: "lang" is a required flag`,
```

Although ideally you would name the attribute based on user perception, so something like so

```golang
      Required: true, 
      RequiredErrorMessage: `There's a problem: "lang" is a required flag`,
```

So my _must change_ request here is, start with this api ^ and then work on getting that code change working with everything else.

README.md
```diff
@@ -560,6 +560,85 @@ Will result in help output like:
...
+
+  app.Flags = []cli.Flag {
+    cli.StringFlag{
+      Name: "lang",
+      Value: "english",
+      Usage: "language for the greeting",
+      Required: true, 
+      RequiredFlagErr: cli.FlagErr{
+        Custom: true, 
+        Message: `There's a problem: "lang" is a required flag` ,
```
lynncyrin: ⭐⭐ _(seriously consider)_ In my experience the most common use case for required flags is when a user must pick from a set of choices. Having the example reflect that would likely help users make good decisions, so something like

> There's a problem: "lang" is required, and must be EN, ES, or RU

lynncyrin: 👋 I've had my focus shift to other things, and don't have the capacity to provide much more guidance here. My recommendation is that PR be closed. I believe that we will return here in the future, though! ✨

anberns: Me too. Just went through a major release at work so haven't had any free time. Revisiting some time in the future sounds good. Thanks!

PR #887, opened 2019-09-10
Release 1.22.1

asahasrabuddhe:
Summary

I wanted these PRs to be a part of 1.22.0 but owing to certain dependency issues it was not possible.  With 1.22.0 in place, I would like to propose this minor release that would include the new build system and support for `Go 1.13` in our latest stable version.

Let me know if we can include anything else.

Included PRs

#883 
#885


lynncyrin:
I'm 👍 here! I'll give the actual approving review once the relevant PRs are merged.

AudriusButkevicius:
We should address the deps issue before making a new release.

asahasrabuddhe:
This is not a new release. This is merely a minor patch that was originally meant to be a part of 1.22.0 but could not be integrated then due to test failures caused by dependency issues.

What deps issue are you referring to? Could you reference it here?

AudriusButkevicius:
Sorry, on the phone, about to catch a flight. #888 is what I am referring to.

asahasrabuddhe:
Thanks! On it. Have a happy and safe journey

asahasrabuddhe:
@Audrius The fix for the dep is already included in #885 

The issue is due to mismatched version of md2man library leading to an issue in mismatched version of black friday.

If you would just approve this, I'd make the release and the issue would get fixed :)

AudriusButkevicius:
But this needs more than just approvals, it needs an actual release.

asahasrabuddhe:
Yes but with your approval, it's one less required. We could also include #889 in this release. It's a nice quick fix!

AudriusButkevicius:
Sure

AudriusButkevicius:
We should document the procedure for releases as I am not sure if you've done everything required so would be more comfortable if @lynncyrin approves this too.

asahasrabuddhe:

Narendra-Kamath:
> 🚢

@lynncyrin will this release be done today?

lynncyrin:
> will this release be done today?

Probably!

PR #890, opened 2019-09-11
Fix #878

### LINKED ISSUE #878: Short flag name don't trigger required setting
``` go
app.Flags = []cli.Flag{
    cli.StringSliceFlag{
      Name: "names, N",
      Required: true,
    },
  }
```

For example, flag setted as described, and i can write in terminal something like `cli --names qwe --names asd` - and that's fine. But also i should be able to write `cli -N qwe -N asd` and that's should work too, but i've got error.

`Required flag "names" not set`

asahasrabuddhe:
Fixes #878


asahasrabuddhe:

context.go
```diff
@@ -313,9 +313,19 @@ func checkRequiredFlags(flags []Flag, context *Context) requiredFlagsErr {
 	var missingFlags []string
 	for _, f := range flags {
 		if rf, ok := f.(RequiredFlag); ok && rf.IsRequired() {
-			key := strings.Split(f.GetName(), ",")[0]
-			if !context.IsSet(key) {
-				missingFlags = append(missingFlags, key)
+			key := strings.Split(f.GetName(), ",")
+			shortName := strings.TrimSpace(key[0])
+			if len(key) > 1 {
```
saschagrunert: Wouldn't it be better to just iterate over the split result here?

context.go
```diff
@@ -313,9 +313,19 @@ func checkRequiredFlags(flags []Flag, context *Context) requiredFlagsErr {
...
+				// has short name
+				longName := strings.TrimSpace(key[1])
+				if !context.IsSet(shortName) && !context.IsSet(longName) {
+					missingFlags = append(missingFlags, key[0])
+				}
+			} else {
+				// does not have short name
+				if !context.IsSet(shortName) {
+					missingFlags = append(missingFlags, key[0])
+				}
```
AudriusButkevicius: Is it mandated the order in which short vs non short should be defined? Also the comments in the if branches seem the other way around judging by the if condition

asahasrabuddhe: The names here are only for readability. Even if you define it the other way, the code will still work. Essentially what we are trying to do here is to check if both the defined values are there in the context and only if both of them are absent, throw the error

context.go
asahasrabuddhe: If I iterate over the values, I would get false for names and the loop would most likely exit. I need to check both of them together to ensure that the conditions are met and we are very sure of the flag being absent

context.go
asahasrabuddhe: In case of definition "n, names" shortName = n and longName = names
In case of definition "names, n" shortName = names and longName = n

Either  way the code would evaluate the same :)

AudriusButkevicius: Sure but its incredibly confusing for the reader

asahasrabuddhe: would you prefer just key[0] and key[1] which would be even more ambiguous?

asahasrabuddhe: I have changed the names of the variables now. Please check

context.go
```diff
@@ -313,9 +313,17 @@ func checkRequiredFlags(flags []Flag, context *Context) requiredFlagsErr {
 	var missingFlags []string
 	for _, f := range flags {
 		if rf, ok := f.(RequiredFlag); ok && rf.IsRequired() {
-			key := strings.Split(f.GetName(), ",")[0]
-			if !context.IsSet(key) {
-				missingFlags = append(missingFlags, key)
+			key := strings.Split(f.GetName(), ",")
+			if len(key) > 1 {
+				// has short name
+				if !context.IsSet(strings.TrimSpace(key[0])) && !context.IsSet(strings.TrimSpace(key[1])) {
```
saschagrunert: What if we have more than one additional short name, like:
```go
			Name:  "long, lo, l"
```
`l` would not covered in that case, right?

asahasrabuddhe: Is that something we support? I have not seen multiple short names for a single flag even in the documentation. Anyway, if this is supported, I guess iteration would the way to go.

asahasrabuddhe:
@AudriusButkevicius @saschagrunert I have implemented the fix using the iterative logic now. It won't matter how many alternate names we add it will work. I have also added a new test case for the same.

https://github.com/urfave/cli/blob/f8bb66ae7d679973cf9b3f6f8c3dc6933404a31a/context_test.go#L520-L533

What do you think?

saschagrunert:
> @AudriusButkevicius @saschagrunert I have implemented the fix using the iterative logic now. It won't matter how many alternate names we add it will work. I have also added a new test case for the same.
> 
> https://github.com/urfave/cli/blob/f8bb66ae7d679973cf9b3f6f8c3dc6933404a31a/context_test.go#L520-L533
> 
> What do you think?

Thanks, for me it seems easier to understand the code now :)

context.go
```diff
@@ -313,9 +313,20 @@ func checkRequiredFlags(flags []Flag, context *Context) requiredFlagsErr {
...
-			if !context.IsSet(key) {
-				missingFlags = append(missingFlags, key)
+			var flagPresent bool
+			var flagName string
+			for _, key := range strings.Split(f.GetName(), ",") {
+				if len(key) > 1 {
+					flagName = key
+				}
+
+				if context.IsSet(strings.TrimSpace(key)) {
```
AudriusButkevicius: Is it ok if key is a single character?

AudriusButkevicius: Is there any purpose to check N?

context.go
```diff
@@ -313,9 +313,20 @@ func checkRequiredFlags(flags []Flag, context *Context) requiredFlagsErr {
...
+					flagPresent = true
+				}
+			}
+
+			if !flagPresent {
+				missingFlags = append(missingFlags, flagName)
```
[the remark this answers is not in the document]
asahasrabuddhe: Fixed

context.go
asahasrabuddhe: That was the bug. For a flag defined as "names, N" we would only check "names" which would not be found as we are passing "N" as the argument.

The fix, makes the library check all the alternate names, in this case "names", "N" to be present in the flagSet before coming to a conclusion that the flag is not present.

asahasrabuddhe:
@lynncyrin could you please review this?

PR #892, opened 2019-09-13
v2 master merge

asahasrabuddhe:
Here we go!


lynncyrin:
image
 
Ah heck it's doing that git thing I can never tell how to fix, where it thinks you added a file that's already present on master 🤔

asahasrabuddhe:
Hello @urfave/cli 

I have finally managed to merge v2 branch with v1.22.1 in this branch. Along the way, I made a few changes which I found throughout the code and various discussions:

1. Removed re-ordering flags
2. `Apply` method now returns error in the `Flag` interface
3. Added `IsSet` method to the `Flag` interface that has now allowed me to remove the reflection based code from the context.

Please go through this and let me know where I can improve this. Also, let's plan a release cycle. That way we can EOL the v1 and only backport fixes and some features to v1 and continue working on v2 moving forward.

Regards,
Ajitem

asahasrabuddhe:
I'm afraid that the diff isn't getting smaller anytime soon. There are significant changes in v2 over v1 like introduction of new methods and new ways of doing things like for example IsSet is now a part of flag interface.

That was the reason I merged master into v2 to get the changes in master into v2 which were not a part of v2. This is why you see CODEOWNERS in the diff

As for the conflicts, as long as we keep adding stuff master, the conflicts will continue to exist because of the differences in v1 and v2. The API breaks with v2 already. Hence I wanted to fix a release cycle, have v2 in the master and continue development ok v2 instead of v1.

lynncyrin:
> That was the reason I merged master into v2 to get the changes in master into v2 which were not a part of v2. This is why you see CODEOWNERS in the diff

This is fair, but I am absolutely 100% certain that we can use some `git` magic to remove the git diff on the files that aren't actually changing with this merge - such as `CODEOWNERS`.

I'll work on communicating to people that v2 work is happening, so new master merges should slow down dramatically. In the meantime, its impossible to merge this PR into master until the merge conflicts are resolved - feel free to ping me for a code review after the conflicts are fixed 🙏

lynncyrin:
👀  merge conflicts are fixed!!! 🎉

asahasrabuddhe:
>  merge conflicts are fixed!!!

Yes. Even the issue with files already being a part of master being shown in the diff is also solved! The latest commit also fixes the failing build :)

Let's start with the review.

README.md
```diff
@@ -65,7 +66,7 @@ the install instructions for Go](http://golang.org/doc/install.html).
 
 To install cli, simply run:
 ```
-$ go get github.com/urfave/cli
```
lynncyrin: ⭐⭐ _(seriously consider)_ With respect to changes made to this file, I would like it if we split the docs into `v1` and `v2` docs. I've started that here => https://github.com/urfave/cli/pull/896

GNUmakefile
```diff
@@ -0,0 +1,37 @@
...
+
+toc:
+	./runtests toc
+
+test: deps
+	./runtests test
+
+all: gen vet test gfmrun v1-to-v2 migrations toc
+
+.PHONY: default gen vet test gfmrun migrations toc v1-to-v2 deps all
```
lynncyrin: ⭐⭐⭐ _(must change)_ Old makefile to be removed ☝️

altsrc/flag_generated.go
```diff
@@ -4,343 +4,271 @@ package altsrc
 
 import (
 	"flag"
-	"github.com/urfave/cli"
```
lynncyrin: ❓ _(blocking question)_ this file should be removed, yea?

go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/urfave/cli
+module github.com/urfave/cli/v2
```
lynncyrin: 🚀🚀🚀 its happening dot gif !!! 🚀🚀🚀

appveyor.yml
```diff
@@ -6,16 +6,21 @@ image: Visual Studio 2017
...
+  GOPROXY: https://proxy.golang.org
 
 install:
   - set PATH=%GOPATH%\bin;C:\go\bin;%PATH%
   - go version
   - go env
-  - go get github.com/urfave/gfmrun/...
-  - go get -v -t ./...
+  - go get github.com/urfave/gfmrun/cmd/gfmrun
+  - go mod vendor
```
lynncyrin: ❓ _(blocking question)_ Not sure I understand this line of diff?

altsrc/flag_test.go
```diff
@@ -3,12 +3,12 @@ package altsrc
 import (
 	"flag"
 	"fmt"
+	"github.com/urfave/cli/v2"
```
lynncyrin: Meta comment - I'm ignoring the diff on altsrc since 🤷‍♀ who even really knows what's going on with altsrc

lynncyrin:
I think there's gonna end up being 3ish distinct review rounds? The PR is large so it's hard to view everything at once.

lynncyrin:
Actually, ignore me, I can just look at the code again. Gonna eat a snack and then dive back into reviews 👀

altsrc/flag_generated.go
asahasrabuddhe: Yes. I think I need to look the altsrc code once again

appveyor.yml
asahasrabuddhe: The older lines work with Go Modules disabled. With Go Modules enabled, the tool needs to locate a package with a tag and complete path. Hence the change.

`go mod vendor` is the new command to download dependencies.

asahasrabuddhe:
> I think there's gonna end up being 3ish distinct review rounds? The PR is large so it's hard to view everything at once.

Please take your time and do a thorough review. This is one of the changing points of this library and we should give it due time and consideration before we merge

appveyor.yml
lynncyrin: I'm reading this https://github.com/golang/go/wiki/Modules and it looks like it says that `go mod vendor` is only needed if you're using vendoring?

So I think I may be missing something, because it looks like `go get` is what you use for downloading dependencies. If you even need to run `go get` that is, since that page above says

> Note there was no go get required in the example above

Here https://github.com/golang/go/wiki/Modules#daily-workflow

asahasrabuddhe: This is different from the daily workflow where you already have a vendor directory and you might want to download a single new dependency or update some of the existing dependencies. But in the CI environment when we clone the repo, there is no vendor folder. Hence `go mod vendor` is required to initialize the vendor folder and get all the dependencies in place.

asahasrabuddhe: Also as seen here:

> Standard commands like go build or go test will automatically add new dependencies as needed to satisfy imports (updating go.mod and downloading the new dependencies)

I don't want to risk updating any dependency. I just want to download what has been mentioned in go.mod and go.sum and proceed with testing

altsrc/flag_generated.go
asahasrabuddhe: I am going to reorganize the altsrc package for good. You should see something exciting in the next commit :)

appveyor.yml
lynncyrin: > But in the CI environment when we clone the repo, there is no vendor folder. Hence go mod vendor is required to initialize the vendor folder and get all the dependencies in place.

I'm still not following? There shouldn't be any vendor folder involved here? The go modules system completely replaces the usage of the vendor folder.

> I don't want to risk updating any dependency. I just want to download what has been mentioned in go.mod and go.sum and proceed with testing

The thing I've done here (in python and javascript) is write a small test that asserts that the dependency file hasn't changed. That's fairly standard practice, I think.

---

Also, as a meta comment, I'm not sure why there's a diff on this file at all inside of this V2 PR. I think the changes on this specific file would be a bit easier to discuss inside their own PR?

asahasrabuddhe: These changes are already a part of master when we added go modules support. This is just coming into the v2 branch now.

errors.go
```diff
@@ -45,29 +59,30 @@ type ExitCoder interface {
...
-	return &ExitError{
-		exitCode: exitCode,
+// NewExitError makes a new *exitError
+func NewExitError(message interface{}, exitCode int) ExitCoder {
+	return Exit(message, exitCode)
+}
+
+// Exit wraps a message and exit code into an ExitCoder suitable for handling by
+// HandleExitCoder
+func Exit(message interface{}, exitCode int) ExitCoder {
```
lynncyrin: ❓ _(blocking question)_ `NewExitError` and `Exit` are exactly the same function, yeah? I can't tell why `Exit` is being added in this PR?

errors.go
```diff
@@ -83,9 +98,9 @@ func HandleExitCoder(err error) {
 	if exitErr, ok := err.(ExitCoder); ok {
 		if err.Error() != "" {
 			if _, ok := exitErr.(ErrorFormatter); ok {
-				fmt.Fprintf(ErrWriter, "%+v\n", err)
+				_, _ = fmt.Fprintf(ErrWriter, "%+v\n", err)
 			} else {
-				fmt.Fprintln(ErrWriter, err)
+				_, _ = fmt.Fprintln(ErrWriter, err)
```
lynncyrin: ⭐ _(nice to have)_ Given that this PR is already quite large, I would prefer that small style changes like this one ^ be removed.

command.go
```diff
@@ -298,18 +240,17 @@ func (c Command) startApp(ctx *Context) error {
...
 	app.Compiled = ctx.App.Compiled
-	app.Author = ctx.App.Author
-	app.Email = ctx.App.Email
 	app.Writer = ctx.App.Writer
 	app.ErrWriter = ctx.App.ErrWriter
+	app.ExitErrHandler = ctx.App.ExitErrHandler
 	app.UseShortOptionHandling = ctx.App.UseShortOptionHandling
 
-	app.categories = CommandCategories{}
+	app.Categories = newCommandCategories()
```
lynncyrin: ❓ _(blocking question)_ This is another `Categories` change that I'm not sure belongs in V2?

appveyor.yml
asahasrabuddhe: Check .travis.yml The appveyor does not include GO module support on Master

flag.go
```diff
@@ -70,8 +90,9 @@ func (f FlagsByName) Swap(i, j int) {
 type Flag interface {
 	fmt.Stringer
 	// Apply Flag settings to the given flag set
-	Apply(*flag.FlagSet)
-	GetName() string
+	Apply(*flag.FlagSet) error
+	Names() []string
+	IsSet() bool
```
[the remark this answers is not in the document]
asahasrabuddhe: I think we can collect all such changes and I'd love to help write the migration guide to help people migrate

errors.go
asahasrabuddhe: Good question. This is how it was named in v2 and I have followed v2 names everywhere. At places like this, I'd be glad to revert to make it more compatible with v1 and one less point for migration.

Good catch (Y)

flag_float64_slice.go
```diff
@@ -0,0 +1,165 @@
+package cli
```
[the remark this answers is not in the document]
asahasrabuddhe: we don't want float64 slice flag?

appveyor.yml
lynncyrin: > The appveyor does not include GO module support on Master

This makes sense 👍 but I would still say my ⭐⭐⭐ _(must change)_ feedback here is that I need the diff for the V2 PR to shrink, which means moving changes like this one into their own PR. I'm going to do that myself now.

lynncyrin: => https://github.com/urfave/cli/pull/898

flag_float64_slice.go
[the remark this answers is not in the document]
lynncyrin: We shouldn't be adding new flag types as a part of this V2 PR, no.

errors.go
lynncyrin: It's frustrating because the new name (`Exit`) is definitely better, but we can easily make changes like this in v `2.1` or `2.9` or whatever. So it'll make our lives easier to remove these kinds of changes from this V2 PR, yeah 👍

altsrc/flag_generated.go
lynncyrin: If you're about to make large changes to `altsrc`, I would specifically advise against doing that inside this PR!

asahasrabuddhe: I will do a separate PR for that.

flag_float64_slice.go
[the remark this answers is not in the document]
asahasrabuddhe: Float64SliceFlag is already in v2

appveyor.yml
rliebz: `go mod vendor` only populates the `vendor/` directory on the local file system. It's not a requirement to use modules, and it's actually ignored unless explicitly opted into for each `go` command.

From [the official documentation](https://github.com/golang/go/wiki/Modules#how-do-i-use-vendoring-with-modules-is-vendoring-going-away):
> By default, go commands like `go build` ignore the vendor directory when in module mode.

There's some details in those docs on why vendoring can be valuable, but I don't think it's providing any benefit in this case. Based on the conversation above, what we're probably looking for here is running `go test` (or other commands) with the `-mod=readonly` flag.

From `go help modules`:

> If invoked with -mod=readonly, the go command is disallowed from the implicit
automatic updating of go.mod described above. Instead, it fails when any changes
to go.mod are needed. This setting is most useful to check that go.mod does
not need updates, such as in a continuous integration and testing system.
The "go get" command remains permitted to update go.mod even with -mod=readonly,
and the "go mod" commands do not take the -mod flag (or any other build flags).

To guarantee that it's used consistently, the `GOFLAGS` env var can be set in the `travis.yml`:
```yaml
environment:
  # ...
  GO111MODULE: on
  GOFLAGS: -mod=vendor
```

AudriusButkevicius:
I am travelling for another few weeks with no proper size screen, so i'll leave this to you guys, or I'll have a look at it once I am back start of Oct.

appveyor.yml
lynncyrin: Big 👍  on `readonly`, I'll stand up a PR for that change in a bit.

lynncyrin:
@asahasrabuddhe hey friend, it's been about two weeks - so I've been thinking about re-assigning work on this branch to myself this weekend. Let me know if you plan on doing some work here within the next week?

appveyor.yml
lynncyrin: => https://github.com/urfave/cli/pull/903

GNUmakefile
asahasrabuddhe: done

altsrc/flag_generated.go
asahasrabuddhe: # 907

lynncyrin:
image

Click the ♻️ to request another review from me, when you've finished working through the existing comments! I'll let ya work in peace till then 🙂

README.md
asahasrabuddhe: How do we want to take this up?

asahasrabuddhe:
@lynncyrin I think I should be done by tomorrow

lynncyrin:
ping @asahasrabuddhe! You should feel encouraged to pass this responsibility off to anyone else, if you don't have a ton of time to work on it! There's always other things that can be done.

appveyor.yml
asahasrabuddhe: Setting `GOFLAGS=-mod=readonly` is causing builds to fail on Travis. However, with regard to above discussion, I have removed the `go mod vendor` and have instead added `go mod tidy`

`go mod tidy` prunes any no-longer-needed dependencies (which are added when we try to install gfmrun and goimports for go vet) from go.mod and add any dependencies needed for other combinations of OS, architecture, and build tags as described [here](https://github.com/golang/go/wiki/Modules#daily-workflow)

CI builds are working now...

app.go
```diff
@@ -50,36 +47,28 @@ type App struct {
 	HideHelp bool
 	// Boolean to hide built-in version flag and the VERSION section of help
 	HideVersion bool
-	// Populate on app startup, only gettable through method Categories()
-	categories CommandCategories
-	// An action to execute when the bash-completion flag is set
+	// Categories contains the categorized commands and is populated on app startup
+	Categories CommandCategories
```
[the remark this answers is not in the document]
asahasrabuddhe: I agree. I see no reason for this being public...

app.go
```diff
@@ -50,36 +47,28 @@ type App struct {
...
+	Action ActionFunc
 	// Execute this function if the proper command cannot be found
 	CommandNotFound CommandNotFoundFunc
 	// Execute this function if an usage error occurs
 	OnUsageError OnUsageErrorFunc
 	// Compilation date
 	Compiled time.Time
 	// List of all authors who contributed
-	Authors []Author
+	Authors []*Author
```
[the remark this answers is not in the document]
asahasrabuddhe: There are several places where I have seen we have changed from a slice of structs to a slice of pointers to struct. I think this is mainly due to performance. Here is an [example](https://stackoverflow.com/questions/27622083/slices-of-structs-vs-slices-of-pointers-to-structs)

appveyor.yml
asahasrabuddhe: What really happens here is that when we have `GOFLAGS=-mod=readonly` it prevents the `go.mod` file from being modified. This, however, causes a problem when `gfmrun` tries to execute the examples from `README.md`. It tries to download the `urfave/cli` dependency and modify the blank mod file created with the example with the dependency. Setting mod to read-only causes this to fail.

The workaround here is to add a flag to the `gfmrun` tool that would allow us to specify dependencies for running the examples or ignore the mod readonly for the while.

command.go
asahasrabuddhe: This has been updated

category.go
```diff
@@ -1,41 +1,75 @@
...
+func (c *commandCategory) VisibleCommands() []*Command {
+	if c.commands == nil {
+		c.commands = []*Command{}
+	}
+
+	var ret []*Command
+	for _, command := range c.commands {
 		if !command.Hidden {
 			ret = append(ret, command)
 		}
```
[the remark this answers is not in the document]
asahasrabuddhe: This file is changing as we are moving from the slice of structs to the slice of pointers to struct

flag.go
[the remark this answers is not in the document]
asahasrabuddhe: Do we have a separate md for this migration guide or have this as a part of the v2 specific README.md?

errors.go
```diff
@@ -45,29 +59,30 @@ type ExitCoder interface {
 	ExitCode() int
 }
 
-// ExitError fulfills both the builtin `error` interface and `ExitCoder`
-type ExitError struct {
+type exitError struct {
```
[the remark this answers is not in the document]
asahasrabuddhe: I think that this change is made because these functions are primarily for internal use and I have currently not seen an example where this function would be used by a user of this library

errors.go
asahasrabuddhe: If we can make this in 2.1 then I wonder why not in 2.0? I think we should have a new 2.0 release with all the features

lynncyrin:
👀

errors.go
lynncyrin: That's fine 👍 but I would like the record to state that I predict users will say _"V2 is changing too much"_

flag_path.go
```diff
@@ -0,0 +1,95 @@
+package cli
+
+import "flag"
+
+type PathFlag struct {
```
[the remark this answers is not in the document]
lynncyrin: Upon further thought, I'm fine with this 👍

flag.go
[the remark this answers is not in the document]
lynncyrin: I've been waiting a few weeks for a code review on me adding the folder, inside which that guide will eventually exist https://github.com/urfave/cli/pull/896 @asahasrabuddhe

asahasrabuddhe:
> * to merge my versioned docs PR so I can write v2 migration documentation #872

@lynncyrin shouldn't this be #896

asahasrabuddhe:
@lynncyrin I would try to collect all the changes to command categories in this comment. Let me know if this is enough or you have further queries. If this is good enough,  it can directly go and be a part of the documentation.

`category.go`

1. The old `CommandCategory` struct is replaced by an interface of the same name with a private struct `commandCategory` implementing the said interface. This interface is now referred to by the rest of the library.

2. The old `CommandCategories` struct (which aliased a slice of the `CommandCategory` struct) is also replaced by an interface of the same name with a private aliased slice `commandCategories` implementing it. This interface is now referred to by the rest of the library.

I personally like this approach as it allows us more flexibility when it comes to adding new or changing this functionality while keeping the interface the same.

Please let me know what your thoughts are on this one. There is no breaking change in this for the front-facing users. This is more on the development side.

flag.go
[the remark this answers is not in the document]
asahasrabuddhe: Done

lynncyrin:
I've going to push the documentation changes to this PR now

docs/CHANGELOG.md
```diff
@@ -2,7 +2,43 @@
...
+  flags](https://github.com/urfave/cli/issues/103) and [reordering of the
+  arguments](https://github.com/urfave/cli/issues/355) when the user
+  attempted to mix flags and arguments. Given the trade-offs we removed support
+  for this reordering.
+- adapter code for deprecated `Action` func signature
+- deprecated `App.Author`, `App.Email`, and `Command.ShortName` fields
+- All `Context.Global*` methods, as the non-global versions now traverse up
+  the context lineage automatically.
+- `Context.Parent` method, as this is now available via `Context.Lineage`
+- `BoolTFlag` and related code, as this is now available via `BoolFlag.Value`
```
lynncyrin: ☝️  I believe this whole section is outdated now, we'll need to go through each line and see if it needs an update

asahasrabuddhe: Agreed. Let me check this and update it now

altsrc/helpers_test.go
```diff
@@ -1,13 +1,23 @@
...
 import (
+	"os"
 	"reflect"
+	"runtime"
+	"strings"
 	"testing"
 )
 
+var (
+	wd, _ = os.Getwd()
```
saschagrunert: I would put that into the `expect` function below and check for the error. (non blocking)

altsrc/json_source_context.go
```diff
@@ -29,6 +28,7 @@ func NewJSONSourceFromFile(f string) (InputSourceContext, error) {
 	if err != nil {
 		return nil, err
 	}
+
```
saschagrunert: Looks unnecessary to me (non blocking)

altsrc/toml_command_test.go
```diff
@@ -1,26 +1,20 @@
...
-
-	"github.com/urfave/cli"
 )
 
 func TestCommandTomFileTest(t *testing.T) {
-	app := cli.NewApp()
+	app := &cli.App{}
 	set := flag.NewFlagSet("test", 0)
-	ioutil.WriteFile("current.toml", []byte("test = 15"), 0666)
+	_ = ioutil.WriteFile("current.toml", []byte("test = 15"), 0666)
```
saschagrunert: I would check these errors and early abort on failure (maybe as a follow up?)

lynncyrin:
😎 this is looking ready to merge 😎

lynncyrin:
I'm gonnaaaaaaaa merge it

purpleidea:
Is there some background on what just happened, because all our builds just broke. Thanks!

lynncyrin:
For any future people reading this, follow-up on this PR is happening on these issues:

- https://github.com/urfave/cli/issues/926
- https://github.com/urfave/cli/issues/925
- https://github.com/urfave/cli/issues/826

PR #894, opened 2019-09-13
support go module in  go1.12 +

tsingson:


asahasrabuddhe:
In accordance with Go's release policy of supporting the previous two releases, we are supporting upwards from `Go 1.11` in this library.

I'd suggest you update this to support at least `Go 1.11`. Also, take a look at #892

tsingson:
@asahasrabuddhe  I will double check...........

Issue #877, opened 2019-08-26
UseShortOptionHandling not found in v1.21.0


bygui86:
Hi guys,

First of all great job! It's an amazing framework!

I'm relatively new to Go and most of all to Go Modules, so I'm almost sure that I'm doing something wrong.
But I found something weird in the flag App.UseShortOptionHandling.
In [my example](https://github.com/bygui86/go-cli) to understand the framework, I copied/pasted your sample about [Combining short options](https://github.com/urfave/cli#combining-short-options) in [this file](https://github.com/bygui86/go-cli/blob/master/examples/combine-short-opts.go) but for some reason the compiler complains about
```
$ go run examples/combine-short-opts.go
# command-line-arguments
examples/combine-short-opts.go:13:5: app.UseShortOptionHandling undefined (type *cli.App has no field or method UseShortOptionHandling)
```

What do I miss? What do I wrong?

Thanks for your help

lynncyrin:
The PR for that was merged 21 days ago https://github.com/urfave/cli/pull/735, whereas `1.21.0` was released 25 days ago https://github.com/urfave/cli/releases/tag/v1.21.0. So this feature will be in release `1.22.0`, which I'm tracking here https://github.com/urfave/cli/issues/867.

IME more mature projects solve this problem have having docs links that are specific to the version, like this page for example https://bundler.io/v1.15/man/bundle-config.1.html which is for bundler `1.15` - and then you can see that it's changed by version `2.0` https://bundler.io/v2.0/man/bundle-config.1.html. I'll put some thought into implementing something like that 🤔

asahasrabuddhe:
I see there is a big discrepancy here.

I checked the tree of the 1.21.0 release. There is no UseShortOptionHandling option defined although the README.md encourages us to do so.

Please check:

https://github.com/urfave/cli/blob/e6cf83ec39f6e1158ced1927d4ed14578fda8edb/README.md#combining-short-bool-options

and

https://github.com/urfave/cli/blob/e6cf83ec39f6e1158ced1927d4ed14578fda8edb/app.go

lynncyrin:
Ah, my comment may not have been clear enough! `1.21.0` does not contain `UseShortOptionHandling` because `UseShortOptionHandling` was merged after `1.21.0` was released. I'll gladly lift my hold on the `1.22.0` release so that we can release `UseShortOptionHandling`, since this is becoming confusing for >= 2 people.

lynncyrin:
[...]

PR #896, opened 2019-09-16
Setup basic versioned docs

lynncyrin:
## Motivation

My proposal for v2 release cycle requires there be versioned docs (see https://github.com/urfave/cli/issues/826). These was also a recent issue where people discussed versioned documentation (see https://github.com/urfave/cli/issues/877).

## Context and Future Work

This isn't "true" versioned documentation, not yet at least. In the future I would like to consider mkdocs, docsify, docusaurus or similar, as mentioned in this comment https://github.com/urfave/cli/issues/877#issuecomment-529708311. This specific PR is focused only on setting up a `v1` and `v2` folder so that https://github.com/urfave/cli/pull/892 won't have to deal with merge conflicts. Judge this PR by that parameter, eg, does this PR help make https://github.com/urfave/cli/pull/892 easier?

## Changes

- creates a `docs` folder
- moves non-versioned documentation (like the contributing guide) to the `docs` folder. the code of conduct has not been moved because (I think) github looks for it in the root directory (I'll explore this more in the future)
- moves the versioned documentation into `docs/v1/manual.md`
- adds an empty folder for v2, `docs/v2/`, with a `.keep` file (this pattern is commonly used with rails projects)
- updates the tests for the new paths


lynncyrin:
🕙 PR timeout!

asahasrabuddhe:
I have a suggestion. How about we use good old godoc for the library? That would be consistent and would allow for auto-generated and would allow for playground-style runnable examples.

We can always generate a README.md from the godoc.

Example link: https://go.ajitem.com/pkg/go.ajitem.com/stack/

lynncyrin:
@asahasrabuddhe can you create an issue about godoc? This PR isn't quite the place to have the conversation.

PR #906, opened 2019-10-06
fix zsh autocomplete

gnowxilef:


autocomplete/zsh_autocomplete
```diff
@@ -1,3 +1,5 @@
+#compdef $PROG
```
saschagrunert: Hey @gnowxilef, can you please explain a bit how this fixes the ZSH completion?

gnowxilef: sure, on the zsh documentation page http://zsh.sourceforge.net/Doc/Release/Completion-System.html#Initialization under the "Autoloaded files" section (20.2.2)7, it states that adding `#compdef name` will make the completion file autoloadable. the current behavior of the cli does not load anything into zsh.

saschagrunert: Thanks 🙏

PR #916, opened 2019-10-31
Add missing call to Close()

Daanikus:


build.go
```diff
@@ -137,6 +137,7 @@ func GfmrunActionFunc(c *cli.Context) error {
 	if err != nil {
 		return err
 	}
+	defer file.Close()
```
lynncyrin: 2 blocking questions here:

1. what problem does this solve?
2. all the other instances of `file.Close` in this file use `err = file.Close`, why is this line using `defer`?

Daanikus: @lynncyrin Thanks for the feedback. I have updated to conform to existing style of closing files.

PR #924, opened 2019-11-09
Update readme for the v2 merge

lynncyrin:
_(title)_

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ applications in an expressive way.
 Usage documentation exists for each major version
 
 - `v1` - [./docs/v1/manual.md](./docs/v1/manual.md)
-- `v2` - 🚧 documentation for `v2` is WIP 🚧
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,7 +18,7 @@ applications in an expressive way.
 Usage documentation exists for each major version
 
 - `v1` - [./docs/v1/manual.md](./docs/v1/manual.md)
-- `v2` - 🚧 documentation for `v2` is WIP 🚧
+- `v2` - [./docs/v2/manual.md](./docs/v2/manual.md)
```
saschagrunert: Can we add a note that this belongs to the master branch as well? (See https://github.com/urfave/cli/issues/925)

lynncyrin: > Can we add a note that this belongs to the master branch as well? (See #925)

Since we require 2 approving reviews, I'll do this in a followup PR

Issue #925, opened 2019-11-11
cli.StringFlag does not implement cli.Flag (Apply method has pointer receiver)


skillful-alex:
After the last 2 months there was an error:
`cli.StringFlag does not implement cli.Flag (Apply method has pointer receiver)`
In commit ebc25651f0fc5d51acf8182565706d670edff596 error not exists.
In corrent master error exists.

AudriusButkevicius:
I don't think this has enough details to understand the problem. 
Do you have a reproduction example, of code that worked and now doesn't?
What go version are you using,? Are you using go modules? Which version of the library you are using?

ffrank:
Stumbled upon that as well. We have code as follows
```
                runFlags := []cli.Flag{
                        cli.BoolFlag{
                                Name:  flagDownload,
                                Usage: "download any missing imports (as the get command does)",
                        },
                        cli.BoolFlag{
                                Name:  "update",
                                Usage: "update all dependencies to the latest versions",
                        },
                }
```
It produces this error message:
```
lang/gapi.go:79:16: cannot use cli.BoolFlag literal (type cli.BoolFlag) as type cli.Flag in array or slice literal:

	cli.BoolFlag does not implement cli.Flag (Apply method has pointer receiver)
```

HTH!

AudriusButkevicius:
Let me guess, you are not using go modules or any form of vendoring and pull stuff straight from master and build against that?

ffrank:
You know, I had this same question 😅 and our main author has it covered:

https://github.com/purpleidea/mgmt/blob/master/docs/faq.md#why-arent-you-using-glide-godep-or-go-mod-for-dependency-management

I have since been informed that we'll probably need to update this code to match the API change. (On it.)

AudriusButkevicius:
Well this was an intended change to a new version of the API, so I guess you just have to fix your side of thongs.

We expect potentially one more of these breaking changes next year.

skillful-alex:
ok, but change [manual.md](https://github.com/urfave/cli/blob/master/docs/v1/manual.md) , line 547:
```
package main

import (
  "log"
  "os"

  "github.com/urfave/cli"
)

func main() {
  app := cli.NewApp()

  app.Flags = []cli.Flag {
    cli.StringFlag{
      Name: "password, p",
      Usage: "password for the mysql database",
      FilePath: "/etc/mysql/password",
    },
  }
```

saschagrunert:
[...]

PR #927, opened 2019-11-13
Mention `v2` first

lynncyrin:
Surfaced via @skillful-alex here https://github.com/urfave/cli/issues/925#issuecomment-552775673, people reading the documentation are under the impression that they should be reading the `v1` documentation, when they should instead be using `v2`.

This PR changes the readme to mention the `v2` documentation first, and adds a _"if you don't know, you're probably using v2"_ mention.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ $ go get github.com/urfave/cli.v2
 ```go
 ...
 import (
-  "github.com/urfave/cli.v2" // imports as package "cli"
+  "github.com/urfave/cli/v2" // imports as package "cli"
 )
 ...
 ```
--- a/README.md
+++ b/README.md
@@ -15,10 +15,40 @@ applications in an expressive way.
 
 ## Usage Documentation
 
-Usage documentation exists for each major version
+Usage documentation exists for each major version. Don't know what version you're on? You're probably using the version from the `master` branch, which is currently `v2`.
 
-- `v1` - [./docs/v1/manual.md](./docs/v1/manual.md)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,29 +15,26 @@ applications in an expressive way.
...
 
-### Supported platforms
-
-cli is tested against multiple versions of Go on Linux, and against the latest
-released version of Go on OS X and Windows.  For full details, see
-[`./.travis.yml`](./.travis.yml) and [`./appveyor.yml`](./appveyor.yml).
+```go
+...
+import (
+  "github.com/urfave/cli.v2" // imports as package "cli"
```
saschagrunert: I think this should be 
```suggestion
  "github.com/urfave/cli/v2" // imports as package "cli"
```

README.md
```diff
@@ -15,29 +15,26 @@ applications in an expressive way.
...
+- `v1` - [./docs/v1/manual.md](./docs/v1/manual.md)
 
-## Installation
-
-Make sure you have a working Go environment.  Go version 1.10+ is supported.  [See
-the install instructions for Go](http://golang.org/doc/install.html).
+### Using `v2` releases
 
-### GOPATH
+**Warning**: `v2` is in a beta state.
```
AudriusButkevicius: I guess it's not really a beta, but rather early days, as beta implies that people should use v1 for something stable, which I don't think is what we want.

lynncyrin: I'll think on the wording here for a bit, and stand up a PR for changing this line 👍

PR #946, opened 2019-11-25
Prevent panic on flagSet access from custom BashComplete

### LINKED ISSUE #944: v2 bug: ctx.NArg() panics on Command.BashComplete
## Checklist

## Describe the bug

`ctx.flagSet` is nil during `cli.Command.BashComplete` when a malformed flag is passed, either misspelled or with an invalid value, leading to a panic when calling `ctx.NArg()`. For example:

```sh
# these panic
go run main.go test --undefined --generate-bash-completion
go run main.go test --number fourty-two --generate-bash-completion
# these don't panic
go run main.go --he --generate-bash-completion
go run main.go test -- --generate-bash-completion
go run main.go test --number 42 --generate-bash-completion
```

## To reproduce

With released `v2.0.0` on `go version go1.12.5 darwin/amd64` the following example results in `panic: runtime error: invalid memory address or nil pointer dereference [signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x11752c2]`

```go
package main

import (
	"fmt"

	"github.com/urfave/cli/v2"
)

func main() {
	app := cli.NewApp()

		&cli.Command{
			Name: "test",
			BashComplete: func(c *cli.Context) {
				fmt.Println(c.NArg())
			},
		},
	}

	app.RunAndExitOnError()
}
```

## Expected behavior

`0` output to stdout. Noticed after upgrade to `v2.0.0` from `d3ae77c26ac8`, according to my go.mod.

## Additional context

I started working my way through [parser.go@06e3d38d88](https://github.com/urfave/cli/commit/06e3d38d88#diff-090598d65549267a01d4ed4132809255), which is where the `nil` value for my command's `flagSet` seems to be last seen, but this is still more of a hunch than a fact. I also feel like there's might be some initialization that I should be doing inside the command's `BashComplete` that I'm missing?

Complete panic:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x11752c2]

goroutine 1 [running]:
flag.(*FlagSet).Args(...)
	/usr/local/Cellar/go/1.12.5/libexec/src/flag/flag.go:615
github.com/urfave/cli/v2.(*Context).Args(...)
	/Users/rob/src/go/pkg/mod/github.com/urfave/cli/v2@v2.0.0/context.go:124
github.com/urfave/cli/v2.(*Context).NArg(...)
	/Users/rob/src/go/pkg/mod/github.com/urfave/cli/v2@v2.0.0/context.go:130
main.main.func1(0xc0000a4500)
	/Users/rob/Work/cli-panic/main.go:17 +0x42
github.com/urfave/cli/v2.ShowCommandCompletions(0xc0000a4500, 0x11d5476, 0x4)
	/Users/rob/src/go/pkg/mod/github.com/urfave/cli/v2@v2.0.0/help.go:252 +0x64
github.com/urfave/cli/v2.checkCommandCompletions(...)
	/Users/rob/src/go/pkg/mod/github.com/urfave/cli/v2@v2.0.0/help.go:366
github.com/urfave/cli/v2.(*Command).Run(0xc000120000, 0xc0000a4400, 0x1213800, 0xc0000a6c40)
	/Users/rob/src/go/pkg/mod/github.com/urfave/cli/v2@v2.0.0/command.go:107 +0x66e
github.com/urfave/cli/v2.(*App).Run(0xc0000ac180, 0xc0000a4000, 0x4, 0x4, 0x0, 0x0)
	/Users/rob/src/go/pkg/mod/github.com/urfave/cli/v2@v2.0.0/app.go:295 +0x6c2
github.com/urfave/cli/v2.(*App).RunAndExitOnError(0xc0000ac180)
	/Users/rob/src/go/pkg/mod/github.com/urfave/cli/v2@v2.0.0/app.go:316 +0x53
main.main()
	/Users/rob/Work/cli-panic/main.go:22 +0xc1
exit status 2
```

## Want to fix this yourself?

Happy to take a stab at if this is a valid use case, and some guidance on an approach to doing so would be much appreciated! Or maybe this is more of a documentation issue, and I should try something like what's done in [`DefaultCompleteWithFlags`](https://github.com/urfave/cli/blob/ee875e4a71871371ae9144b482248164698981f3/help.go#L162) instead of using `ctx.NArg()`? in which case I can send a PR to the manual.

```
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/rob/Library/Caches/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/Users/rob/src/go"
GOPROXY=""
GORACE=""
GOROOT="/usr/local/Cellar/go/1.12.5/libexec"
GOTMPDIR=""
GOTOOLDIR="/usr/local/Cellar/go/1.12.5/libexec/pkg/tool/darwin_amd64"
GCCGO="gccgo"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/rob/Work/cli-panic/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/mj/xhg1t1sx2tj_2g6pd0snwvf4k7p1w7/T/go-build964702892=/tmp/go-build -gno-record-gcc-switches -fno-common"
```

unRob:
Fixes #944

### DIFF
--- a/command.go
+++ b/command.go
@@ -100,7 +100,7 @@ func (c *Command) Run(ctx *Context) (err error) {
 		c.UseShortOptionHandling = true
 	}
 
-	set, err := c.parseFlags(ctx.Args())
+	set, err := c.parseFlags(ctx.Args(), ctx.shellComplete)
 
 	context := NewContext(ctx.App, set, ctx)
 	context.Command = c
@@ -174,7 +174,7 @@ func (c *Command) useShortOptionHandling() bool {
 	return c.UseShortOptionHandling
 }
 
-func (c *Command) parseFlags(args Args) (*flag.FlagSet, error) {
+func (c *Command) parseFlags(args Args, shellComplete bool) (*flag.FlagSet, error) {
 	set, err := c.newFlagSet()
 	if err != nil {
 		return nil, err
@@ -185,7 +185,8 @@ func (c *Command) parseFlags(args Args) (*flag.FlagSet, error) {
 	}
 
 	err = parseIter(set, c, args.Tail())
-	if err != nil {
+	// Continue parsing flags on failure during shell completion
+	if err != nil && !shellComplete {
 		return nil, err
 	}
 
--- a/command.go
+++ b/command.go
@@ -184,9 +184,8 @@ func (c *Command) parseFlags(args Args, shellComplete bool) (*flag.FlagSet, erro
 		return set, set.Parse(append([]string{"--"}, args.Tail()...))
 	}
 
-	err = parseIter(set, c, args.Tail())
-	// Continue parsing flags on failure during shell completion
-	if err != nil && !shellComplete {
+	err = parseIter(set, c, args.Tail(), shellComplete)
+	if err != nil {
 		return nil, err
 	}
 
--- a/parse.go
+++ b/parse.go
@@ -21,7 +21,7 @@ func parseIter(set *flag.FlagSet, ip iterativeParser, args []string, shellComple
 		err := set.Parse(args)
 		if !ip.useShortOptionHandling() || err == nil {
 			if shellComplete {
-				return nil
+				return err
 			}
 			return err
 		}
--- a/parse.go
+++ b/parse.go
@@ -11,19 +11,24 @@ type iterativeParser interface {
 }
 
 // To enable short-option handling (e.g., "-it" vs "-i -t") we have to
-// iteratively catch parsing errors.  This way we achieve LR parsing without
+// iteratively catch parsing errors. This way we achieve LR parsing without
 // transforming any arguments. Otherwise, there is no way we can discriminate
 // combined short options from common arguments that should be left untouched.
-func parseIter(set *flag.FlagSet, ip iterativeParser, args []string) error {
+// Pass `ignoreErrors` to continue parsing options on failure, for example
+// during shell completion when the user-supplied options may be incomplete.
+func parseIter(set *flag.FlagSet, ip iterativeParser, args []string, ignoreErrors bool) error {
 	for {
 		err := set.Parse(args)
 		if !ip.useShortOptionHandling() || err == nil {
+			if ignoreErrors {
+				return nil
+			}
 			return err
 		}
 
 		errStr := err.Error()
 		trimmed := strings.TrimPrefix(errStr, "flag provided but not defined: -")
-		if errStr == trimmed {
+		if !ignoreErrors && errStr == trimmed {
 			return err
 		}
 
--- a/parse.go
+++ b/parse.go
@@ -14,13 +14,13 @@ type iterativeParser interface {
 // iteratively catch parsing errors. This way we achieve LR parsing without
 // transforming any arguments. Otherwise, there is no way we can discriminate
 // combined short options from common arguments that should be left untouched.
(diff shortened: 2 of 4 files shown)


command.go
```diff
@@ -174,7 +174,7 @@ func (c *Command) useShortOptionHandling() bool {
 	return c.UseShortOptionHandling
 }
 
-func (c *Command) parseFlags(args Args) (*flag.FlagSet, error) {
+func (c *Command) parseFlags(args Args, shellComplete bool) (*flag.FlagSet, error) {
```
unRob: Not quite sure if this name/concept is right as an argument, kept it for the sake of consistency but happy to adjust if someone has a better idea!

command.go
```diff
@@ -185,7 +185,8 @@ func (c *Command) parseFlags(args Args) (*flag.FlagSet, error) {
 	}
 
 	err = parseIter(set, c, args.Tail())
-	if err != nil {
+	// Continue parsing flags on failure during shell completion
+	if err != nil && !shellComplete {
```
unRob: This basically allows for the context to continue to be parsed. Malformed flags will still be ignored and not part of the `set`, which is what v2 pre-release builds did.

lynncyrin: ⭐⭐⭐ _(must change)_ @unRob can you add in here a detailed description of what `parseIter`, and why it's ok to skip errors from it when doing shell completion?

FYI cc @rliebz

lynncyrin: OH! Actually! I have a better idea: can you change this code to pass `shellComplete` into the `parseIter` function, ignore the errors inside the function itself, and add the descriptions I just requested inside the function's docstring?

unRob: Great idea! 

- Changed the signature of `parseIter` to accept a fourth parameter `shellComplete bool` to deal with nil errors within.
- Decided to keep the `shellComplete` name, played with `ignoreErrors` instead, but it felt like a misnomer since `set.Parse` won't actually error on it.
- Modified a couple of consumers to pass `shellComplete` from their own context as well.
- Added a couple of lines to the docstring, hope my wording conveys the use for passing that argument.

parse.go
```diff
@@ -11,13 +11,18 @@ type iterativeParser interface {
...
 // combined short options from common arguments that should be left untouched.
-func parseIter(set *flag.FlagSet, ip iterativeParser, args []string) error {
+// Pass `shellComplete` to continue parsing options on failure during shell
+// completion when, the user-supplied options may be incomplete.
+func parseIter(set *flag.FlagSet, ip iterativeParser, args []string, shellComplete bool) error {
 	for {
 		err := set.Parse(args)
 		if !ip.useShortOptionHandling() || err == nil {
+			if shellComplete {
+				return nil
```
lynncyrin: I'm going to briefly change this line, I want to see the tests fail in CI 🙂

PR #951, opened 2019-11-28
Ask users to describe their packaging setup

lynncyrin:
## Changes

Expands the new issue template to include questions about the user's packaging setup

## Motivation

Many issues would be resolved much faster if we could immediately identify that a user isn't using go modules or vendoring.

## Desired Feedback

Is **Packaging Setup** good wording here? I'm unsure about it


asahasrabuddhe:
I think **Dependency Management** would sound better. Otherwise this is good to go :+1:

PR #953, opened 2019-11-28
Recieve and handle ctrl + c and exit application

### LINKED ISSUE #945: v2 feature: Allow registering ctrl-c with the app
## Checklist

## Describe the bug
Thank you for the v2 release.

Expecting the application to handle the `ctrl-c` but it doesn't. So the question then becomes:
- How to cleanly register a `ctrl-c` handler with the current app? ( I saw in the code `NewContext` but the signal is not getting handled apparently?

## To reproduce

Run the following snippet and press `ctrl-c`
```golang
package cli

import (
	"fmt"
	"log"
	"os"
	"time"

	"github.com/urfave/cli/v2"
)

func RealMain() {
	// ctx, cancel := context.WithCancel(context.Background())
	// defer cancel()

	app := &cli.App{
		Name:  "boom",
		Usage: "make an explosive entrance",
			fmt.Println("boom! I say!")
			time.Sleep(time.Duration(time.Second * 10))
			return nil
		},
	}

	// c := make(chan os.Signal, 2)
	// signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
	// go func() {
	// 	<-c
	// 	fmt.Println("\r- Ctrl+C pressed in Terminal")
	// 	os.Exit(0)
	// }()

	err := app.Run(os.Args)
	if err != nil {
		log.Fatal(err)
	}
}
```

## Expected behavior

Expecting the application to stop.

## Version of go
go version go1.13.3 linux/amd64

asahasrabuddhe:
Fixes #945 

We were correctly receiving `ctrl+c` signal and cancelling our application context. The missing piece of the puzzle was not listening for context cancellation and exiting the program.

Since `ctrl+c` is a non standard program termination, I have used a non zero exit code i.e. 1 for the moment. Please let me know if we want to give the user an option to set a exit code themselves in case of `ctrl+c` exit. We can then add an option in the `App` for that then.

### DIFF
--- a/app.go
+++ b/app.go
@@ -1,7 +1,6 @@
 package cli
 
 import (
-	"context"
 	"flag"
 	"fmt"
 	"io"
@@ -119,7 +118,7 @@ func NewApp() *App {
 		Action:           helpCommand.Action,
 		Compiled:         compileTime(),
 		Writer:           os.Stdout,
-		InterruptHandler: func(*Context, context.CancelFunc) {},
+		InterruptHandler: CancelContextOnInterrupt,
 	}
 }
 
@@ -158,7 +157,7 @@ func (a *App) Setup() {
 	}
 
 	if a.InterruptHandler == nil {
-		a.InterruptHandler = func(*Context, context.CancelFunc) {}
+		a.InterruptHandler = CancelContextOnInterrupt
 	}
 
 	if a.Compiled == (time.Time{}) {
--- a/app.go
+++ b/app.go
@@ -119,6 +119,7 @@ func NewApp() *App {
 		Action:       helpCommand.Action,
 		Compiled:     compileTime(),
 		Writer:       os.Stdout,
+		InterruptHandlerFunc: func(c *Context) {},
 	}
 }
 
@@ -154,6 +155,10 @@ func (a *App) Setup() {
 
 	if a.Action == nil {
 		a.Action = helpCommand.Action
+	}
+
+	if a.InterruptHandlerFunc == nil {
+		a.InterruptHandlerFunc = func(context *Context) {}
 	}
 
 	if a.Compiled == (time.Time{}) {
--- a/app.go
+++ b/app.go
@@ -88,6 +88,10 @@ type App struct {
 	// single-character bool arguments into one
 	// i.e. foobar -o -v -> foobar -ov
 	UseShortOptionHandling bool
+	// This function is automatically executed when the context is cancelled due to the program
+	// receiving an interrupt signal (Control+C). This function can be overridden to implement
+	// desired behaviour in such a scenario
+	InterruptHandlerFunc InterruptHandlerFunc
 
 	didSetup bool
 }
--- a/app.go
+++ b/app.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"io"
@@ -91,7 +92,7 @@ type App struct {
 	// This function is automatically executed when the context is cancelled due to the program
 	// receiving an interrupt signal (Control+C). This function can be overridden to implement
 	// desired behaviour in such a scenario
-	InterruptHandlerFunc InterruptHandlerFunc
+	InterruptHandler InterruptHandlerFunc
 
 	didSetup bool
 }
@@ -110,15 +111,15 @@ func compileTime() time.Time {
 // Usage, Version and Action.
 func NewApp() *App {
 	return &App{
-		Name:         filepath.Base(os.Args[0]),
-		HelpName:     filepath.Base(os.Args[0]),
-		Usage:        "A new cli application",
-		UsageText:    "",
-		BashComplete: DefaultAppComplete,
-		Action:       helpCommand.Action,
-		Compiled:     compileTime(),
-		Writer:       os.Stdout,
-		InterruptHandlerFunc: func(c *Context) {},
+		Name:             filepath.Base(os.Args[0]),
+		HelpName:         filepath.Base(os.Args[0]),
+		Usage:            "A new cli application",
+		UsageText:        "",
+		BashComplete:     DefaultAppComplete,
+		Action:           helpCommand.Action,
+		Compiled:         compileTime(),
+		Writer:           os.Stdout,
+		InterruptHandler: func(*Context, context.CancelFunc) {},
 	}
 }
 
@@ -156,8 +157,8 @@ func (a *App) Setup() {
 		a.Action = helpCommand.Action
 	}
 
-	if a.InterruptHandlerFunc == nil {
-		a.InterruptHandlerFunc = func(context *Context) {}
+	if a.InterruptHandler == nil {
+		a.InterruptHandler = func(*Context, context.CancelFunc) {}
 	}
 
 	if a.Compiled == (time.Time{}) {
--- a/context.go
+++ b/context.go
@@ -37,16 +37,13 @@ func NewContext(app *App, set *flag.FlagSet, parentCtx *Context) *Context {
 
 	if c.Context == nil {
 		ctx, cancel := context.WithCancel(context.Background())
+		signals := make(chan os.Signal, 1)
 		go func() {
-			// wait for context cancellation
-			<-ctx.Done()
-			app.InterruptHandlerFunc(c)
-		}()
-		go func() {
-			defer cancel()
-			signals := make(chan os.Signal, 1)
-			signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
-			<-signals
+			for {
+				signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
+				<-signals
+				c.App.InterruptHandler(c, cancel)
+			}
 		}()
 		c.Context = ctx
 	}
--- a/context.go
+++ b/context.go
@@ -18,6 +18,7 @@ import (
 type Context struct {
 	context.Context
 	App           *App
+	Cancel        context.CancelFunc
 	Command       *Command
 	shellComplete bool
 	setFlags      map[string]bool
@@ -38,14 +39,15 @@ func NewContext(app *App, set *flag.FlagSet, parentCtx *Context) *Context {
 	if c.Context == nil {
 		ctx, cancel := context.WithCancel(context.Background())
 		signals := make(chan os.Signal, 1)
+		signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
 		go func() {
 			for {
-				signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
 				<-signals
-				c.App.InterruptHandler(c, cancel)
+				c.App.InterruptHandler(c)
 			}
 		}()
 		c.Context = ctx
+		c.Cancel = cancel
 	}
 
 	return c
--- a/context.go
+++ b/context.go
@@ -40,7 +40,7 @@ func NewContext(app *App, set *flag.FlagSet, parentCtx *Context) *Context {
 		go func() {
 			// wait for context cancellation
 			<-ctx.Done()
-			os.Exit(1)
+			app.InterruptHandlerFunc(c)
 		}()
 		go func() {
 			defer cancel()
--- a/context.go
+++ b/context.go
@@ -38,10 +38,15 @@ func NewContext(app *App, set *flag.FlagSet, parentCtx *Context) *Context {
 	if c.Context == nil {
 		ctx, cancel := context.WithCancel(context.Background())
 		go func() {
+			// wait for context cancellation
+			<-ctx.Done()
+			os.Exit(1)
+		}()
+		go func() {
 			defer cancel()
-			sigs := make(chan os.Signal, 1)
-			signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
-			<-sigs
+			signals := make(chan os.Signal, 1)
+			signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
+			<-signals
 		}()
 		c.Context = ctx
 	}
--- a/funcs.go
+++ b/funcs.go
@@ -43,6 +43,6 @@ type FlagEnvHintFunc func(envVars []string, str string) string
 // with the file path details.
 type FlagFileHintFunc func(filePath, str string) string
 
-// InterruptHandler is used to define the behaviour when a program is interrupted
+// InterruptHandlerFunc is used to define the behaviour when a program is interrupted
 // using Control+C
 type InterruptHandlerFunc func(*Context)
--- a/funcs.go
+++ b/funcs.go
@@ -1,5 +1,7 @@
 package cli
 
+import "context"
+
 // BashCompleteFunc is an action to execute when the shell completion flag is set
 type BashCompleteFunc func(*Context)
 
@@ -43,6 +45,6 @@ type FlagEnvHintFunc func(envVars []string, str string) string
 // with the file path details.
 type FlagFileHintFunc func(filePath, str string) string
 
-// InterruptHandlerFunc is used to define the behaviour when a program is interrupted
+// InterruptHandler is used to define the behaviour when a program is interrupted
 // using Control+C
-type InterruptHandlerFunc func(*Context)
+type InterruptHandlerFunc func(*Context, context.CancelFunc)
--- a/funcs.go
+++ b/funcs.go
@@ -42,3 +42,7 @@ type FlagEnvHintFunc func(envVars []string, str string) string
 // FlagFileHintFunc is used by the default FlagStringFunc to annotate flag help
 // with the file path details.
 type FlagFileHintFunc func(filePath, str string) string
+
+// InterruptHandlerFunc is used to define the behaviour when a program is interrupted
+// using Control+C
+type InterruptHandlerFunc func(*Context)
--- a/funcs.go
+++ b/funcs.go
@@ -1,7 +1,5 @@
 package cli
 
-import "context"
-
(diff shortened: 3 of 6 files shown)


AudriusButkevicius:
As I explained, we should not do this. This takes away the opportunity from the user to clean up.

lynncyrin:
@AudriusButkevicius & @ansrivas can you two have a conversation about what the ideal solution is here? This is one potential solution for https://github.com/urfave/cli/issues/945.

asahasrabuddhe:
@AudriusButkevicius I think the real problem here is the lack of a sensible default. For most use cases, `os.Exit` is enough. I think we can probably have a handler function that would exit by default but would give advanced users a chance to write their own implementation for cleaning their stuff up.

AudriusButkevicius:
An option with os.Exit encourages people to design their applications poorly, not performing cleanup.

This library should be about argument parsing and based on the arguments calling functions, not about lifecycle management of application.

I suspect you won't find a library in other languages that does os.Exit unasked
What if people want to implement "Are you sure you want to cancel current operation?" prompt?

asahasrabuddhe:
If we have a default, they can read about it in the documentation and override it! The implementation right now provides no way to do this.

asahasrabuddhe:
Or we could simply have that handler print a message about intercepting the ctrl + c and request the user to override it!

AudriusButkevicius:
I think the default should be to do nothing, as thats least surprising from the API perspective with a out-of-the-box available handler that does os.Exit that users have to setup.

Perhaps the default handler should cancel the context on interrupt, which if the user implemented his app as expected (propagating contexts all the way down), would lead to the same effect.

asahasrabuddhe:
> I think the default should be to do nothing, as thats least surprising from the API perspective with a out-of-the-box available handler that does os.Exit that users have to setup.
> 
> Perhaps the default handler should cancel the context on interrupt, which if the user implemented his app as expected (propagating contexts all the way down), would lead to the same effect.

I agree. Let me implement this.

asahasrabuddhe:
@AudriusButkevicius Is this better now?

context.go
```diff
@@ -37,11 +37,16 @@ func NewContext(app *App, set *flag.FlagSet, parentCtx *Context) *Context {
...
+			app.InterruptHandlerFunc(c)
+		}()
 		go func() {
 			defer cancel()
-			sigs := make(chan os.Signal, 1)
-			signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
-			<-sigs
+			signals := make(chan os.Signal, 1)
+			signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
+			<-signals
```
AudriusButkevicius: I think this should be in a loop to deliver the signal more than once, and this should call the handler repeatedly.

context.go
```diff
@@ -37,11 +37,16 @@ func NewContext(app *App, set *flag.FlagSet, parentCtx *Context) *Context {
...
 	if c.Context == nil {
 		ctx, cancel := context.WithCancel(context.Background())
+		go func() {
+			// wait for context cancellation
+			<-ctx.Done()
+			app.InterruptHandlerFunc(c)
+		}()
 		go func() {
 			defer cancel()
-			sigs := make(chan os.Signal, 1)
```
AudriusButkevicius: The cancellation of the context should happen in the handler I think. You might want to completely ignore ctrl+c and that not to have any effect on the application.

ansrivas:
> Perhaps the default handler should cancel the context on interrupt, which if the user implemented his app as expected (propagating contexts all the way down), would lead to the same effect.
This totally makes sense here as this was exactly the confusion btw in #945 - how to propagate a context further for a cleanup.

context.go
AudriusButkevicius: This probably requires a ctx.Cancel() or something like that to be exposed, or pass cancellation function as an argument to the handler.

app.go
```diff
@@ -114,6 +118,7 @@ func NewApp() *App {
 		Action:       helpCommand.Action,
 		Compiled:     compileTime(),
 		Writer:       os.Stdout,
+		InterruptHandlerFunc: func(c *Context) {},
```
AudriusButkevicius: The struct member should be called InterruptHandler, the type can be called func. Also I guess we could drop func from the type name too, but we should check what action does.


 I think the default should be something that cancels the context, and have an ExitOnInterrupt handler freely available in the library.

asahasrabuddhe:
@ansrivas I think now you'd be able to do this:

```go
	app.InterruptHandlerFunc = func(c *cli.Context) {
		// received interrupt

                // cleanup
	}
```

app.go
```diff
@@ -88,6 +88,10 @@ type App struct {
 	// single-character bool arguments into one
 	// i.e. foobar -o -v -> foobar -ov
 	UseShortOptionHandling bool
+	// This function is automatically executed when the context is cancelled due to the program
+	// receiving an interrupt signal (Control+C). This function can be overridden to implement
+	// desired behaviour in such a scenario
+	InterruptHandlerFunc InterruptHandlerFunc
```
AudriusButkevicius: This function should only be executed on receipt of signal. This function should be the decision point if we want to cancel the context or not.

asahasrabuddhe:
Hi @AudriusButkevicius I have made the necessary changes. Here's an example illustrating the use of the feature:

```go
package main

import (
	"context"
	"fmt"
	"log"
	"os"
	"time"

	"github.com/urfave/cli/v2"
)

func main() {
	app := &cli.App{
		Name:  "long",
		Usage: "this takes a long time to finish",
			fmt.Println("working...")
			time.Sleep(10 * time.Second)
			return nil
		},
	}

	app.Flags = []cli.Flag{
		&cli.IntFlag{
			Name: "count",
		},
	}

	signalCount := 1
	app.InterruptHandler = func(c *cli.Context, cancelFunc context.CancelFunc) {
		go func() {
			<-c.Done()
			os.Exit(1)
		}()

		count := c.Int("count")
		if signalCount < count {
			fmt.Println("received interrupt")
		} else {
			fmt.Println("signal count exceeded, quitting")
			cancelFunc()
		}

		signalCount++
	}

	err := app.Run(os.Args)
	if err != nil {
		log.Fatal(err)
	}
}

```

In the above example, I am waiting for the interrupt signal to be received 5 times before the context is cancelled.

asahasrabuddhe:
The CI is failing because `gfmrun` is not able to find `app.InterruptHandler` in the current codebase. All the other test cases have passed. Please consider merging this.

context.go
```diff
@@ -37,16 +37,13 @@ func NewContext(app *App, set *flag.FlagSet, parentCtx *Context) *Context {
 
 	if c.Context == nil {
 		ctx, cancel := context.WithCancel(context.Background())
+		signals := make(chan os.Signal, 1)
```
AudriusButkevicius: Does this need to live outside of the closure?

app.go
```diff
@@ -110,15 +111,15 @@ func compileTime() time.Time {
...
-		InterruptHandlerFunc: func(c *Context) {},
+		Name:             filepath.Base(os.Args[0]),
+		HelpName:         filepath.Base(os.Args[0]),
+		Usage:            "A new cli application",
+		UsageText:        "",
+		BashComplete:     DefaultAppComplete,
+		Action:           helpCommand.Action,
+		Compiled:         compileTime(),
+		Writer:           os.Stdout,
+		InterruptHandler: func(*Context, context.CancelFunc) {},
```
AudriusButkevicius: I think the default behaviour should cancel the context, and we should have it as a public type available in the library, together with an alternative publicly available exit on interrupt handler.

app.go
```diff
@@ -156,8 +157,8 @@ func (a *App) Setup() {
 		a.Action = helpCommand.Action
 	}
 
-	if a.InterruptHandlerFunc == nil {
-		a.InterruptHandlerFunc = func(context *Context) {}
+	if a.InterruptHandler == nil {
+		a.InterruptHandler = func(*Context, context.CancelFunc) {}
```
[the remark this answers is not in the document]
AudriusButkevicius: I guess if this is to avoid having to do a null check then it's fine, but it should stay a noop.

context.go
```diff
@@ -37,16 +37,13 @@ func NewContext(app *App, set *flag.FlagSet, parentCtx *Context) *Context {
...
-		}()
-		go func() {
-			defer cancel()
-			signals := make(chan os.Signal, 1)
-			signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
-			<-signals
+			for {
+				signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
+				<-signals
+				c.App.InterruptHandler(c, cancel)
```
AudriusButkevicius: I am still wonder if the API is cleaner if we add a ctx.Cancel() and just pass a single thing, what do you think?

funcs.go
```diff
@@ -43,6 +45,6 @@ type FlagEnvHintFunc func(envVars []string, str string) string
 // with the file path details.
 type FlagFileHintFunc func(filePath, str string) string
 
-// InterruptHandlerFunc is used to define the behaviour when a program is interrupted
+// InterruptHandler is used to define the behaviour when a program is interrupted
```
AudriusButkevicius: The docs should till have the Func in the name

app.go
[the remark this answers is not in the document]
asahasrabuddhe: This is to avoid a nil pointer panic but it essentially remains noop as the body of the function is empty

app.go
asahasrabuddhe: The `ctx.Done()` channel is already publicly available through `cli.Context` which we are going to pass to the `InterruptHandler`. I just wanted to give more control to the user by allowing them to decide whether or not to decide the context.

If we cancel the context ourselves and then call the `InterruptHandler` with `cli.Context` then the `ctx.Done()` channel won't make much sense as the context would already have been cancelled. Waiting for that channel would essentially become a noop.

funcs.go
asahasrabuddhe: Will update this

context.go
asahasrabuddhe: The API is clean either way. What we need to decide is how much of power to hand over to the user? If we call the cancel function ourselves, the user would not be able to do things like wait a certain number of times before processing the `SIGINT`. 

Consider this example.  The `docker-compose` CLI has two methods of handling the `SIGINT`. When we do a `Ctrl+C` the first time, the CLI would start shutting down the containers gracefully. However if we do a `Ctrl+C` the second time, it would immediately and forcibly shutdown the containers.

If we wish that our users build similar behaviour, I think we need to give them the cancel function and let them decide when to cancel the Context.

app.go
AudriusButkevicius: Sure, but I am arguing that default Ctrl+c behaviour in the library should be cancellatiom of the context, which it does not do now.

I am also arguing that some common behaviours should be readily available in the library, so there should be a handled for ExitOnInterrupt and CancelContextOnInterrupt functions readily available in the library.

asahasrabuddhe: Common behaviours out of the box is intriguing

context.go
AudriusButkevicius: Sure, but that's just custom interrupt handler which the user can implement.

What I am arguing for is instead of passing cancellation function as an argument, we expose the cancellation function as a function of the context, so the user can cancel ot from wherever he wants.

asahasrabuddhe: There is no need to expose the function. The `cli.Context` embeds `context.Context`. So, anywhere the user is using the `cli.Context` they can simply call the `Done()` method to get access to the channel.

I think I understand where you are coming from. Let me work on this and let's discuss this again tomorrow

AudriusButkevicius: Done is a method that returns a channel the user should check, it does not terminate the context.

context.go
asahasrabuddhe: Fixed

app.go
```diff
@@ -151,6 +156,10 @@ func (a *App) Setup() {
 		a.Action = helpCommand.Action
 	}
 
+	if a.InterruptHandler == nil {
+		a.InterruptHandler = CancelContextOnInterrupt
+	}
```
AudriusButkevicius: Should not do this, user might set it to null not to get the context cancelled.

app.go
```diff
@@ -88,6 +88,10 @@ type App struct {
 	// single-character bool arguments into one
 	// i.e. foobar -o -v -> foobar -ov
 	UseShortOptionHandling bool
+	// This function is automatically executed when the context is cancelled due to the program
+	// receiving an interrupt signal (Control+C). This function can be overridden to implement
+	// desired behaviour in such a scenario
+	InterruptHandler InterruptHandlerFunc
```
AudriusButkevicius: The function is executed when the application receives and interuption signal, what that does is configurable based on what the handler is doing.

context.go
```diff
@@ -37,13 +38,16 @@ func NewContext(app *App, set *flag.FlagSet, parentCtx *Context) *Context {
...
+		signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
 		go func() {
-			defer cancel()
-			sigs := make(chan os.Signal, 1)
-			signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
-			<-sigs
+			for {
+				<-signals
+				c.App.InterruptHandler(c)
+			}
```
AudriusButkevicius: for signal := range signals {}

context.go
```diff
@@ -37,13 +38,16 @@ func NewContext(app *App, set *flag.FlagSet, parentCtx *Context) *Context {
 
 	if c.Context == nil {
 		ctx, cancel := context.WithCancel(context.Background())
+		signals := make(chan os.Signal, 1)
+		signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
```
AudriusButkevicius: Because we capture TERM, perhaps we should not call it an interrupt handler but call it signal handler. I did not see that we capture more than a single signal before. Also, because we capture multiple, perhaps we should pass the signal to the handler.

app.go
AudriusButkevicius: Essentially, if this is null, we should not even try to setup signal handling, and leave it up to the user.

app.go
AudriusButkevicius: If we change this to signal handler, we should change the comment quite heavily.

interrupt-handlers.go
```diff
@@ -0,0 +1,13 @@
+package cli
+
+import "os"
+
+var ExitOnInterrupt InterruptHandlerFunc = func(ctx *Context) {
+	CancelContextOnInterrupt(ctx)
+	<-ctx.Done()
+	os.Exit(1)
```
AudriusButkevicius: I don't think we should cancel the context in this case.

AudriusButkevicius: To be honest, if we start passing down signals, I think default should be os.Exit on sigterm, cancel context on sigint

rliebz:
I'd like to see testing around this. Specifically, I think cases should include tests for verifying that custom interrupt handlers (or signal handlers) are called when specified, tests for what happens with the defaults, and tests for what happens when a `nil` handler is specified.

asahasrabuddhe:
I like the direction in which this is headed to. Let me further improve this :)

AudriusButkevicius:
Sorry for changing the course so many times.

asahasrabuddhe:
@AudriusButkevicius Please don't be sorry! I would change to course a million times to produce good software 🙂

app.go
asahasrabuddhe: I think if it's set to nil, we should actually set it to the Noop handler. We can add that to the documentation as well. It's easier that way to not cause nil panics

lynncyrin:
Is this still relevant as of https://github.com/urfave/cli/pull/975?

(I originally made this comment on https://github.com/urfave/cli/pull/973, and https://github.com/urfave/cli/pull/974 by mistake 😭)

burdiyan:
I have struggled with developers not handling signals properly so much, so I'm really grateful that will come out of the box with `cli`.

I do strongly believe though, that cli must not run os.Exit. Simply return the error from `app.Run()` and then the user will do the os.Exit as they wish. And the commands themselves within the Action could just listen for context cancelation, giving the full control to the user in this case.

Basically the original approach (which is currently tagged as v2) worked pretty fine in a setup similar to this one:

```go
app := &cli.App{
    Name:    "say",
    Version: "0.1.0",
    Commands: []*cli.Command{
        {
            Name:        "hello",
            Aliases:     []string{"hi"},
            Usage:       "use it to see a description",
            Description: "This is how we describe hello the function",
                t := time.NewTicker(1 * time.Second)
                defer func() {
                    t.Stop()
                    fmt.Println("Gracefully shutting down...")
                    time.Sleep(2 * time.Second)
                }()

                fmt.Println("Timer started")
                for {
                    select {
                    case tt := <-t.C:
                        fmt.Println(tt.String())
                    case <-ctx.Done():
                        return ctx.Err()
                    }
                }
            },
        },
    },
}

    fmt.Printf("%+v\n", err)
    os.Exit(1)
}
```

So this way everything works perfectly. The only thing missing though is normally you can force close the app pressing another `ctrl+c`, and to do that you have to `signal.Stop(ch)` on a channel that was listening signals.

I think this would be the best of both worlds: user control and handling signals by default.

AudriusButkevicius:
I think this PR is dead. We've decided that the argument parsing library is not in the position to do signal handling, and instead allow users to pass a context to the cli which they can now cancel (by doing their own signal handling).

burdiyan:
This would also be perfect. Having a way to integrate context easily is one of the reasons I came to cli from cobra (my PR https://github.com/spf13/cobra/pull/893 there implementing this for some reason is not getting response from maintainers).

Issue #957, opened 2019-11-28
v2 bug: (go get github.com\urfave\cli\v2 error )


cw1997:
## my urfave/cli version is v2

_**( Put the version of urfave/cli that you are using here )**_

## Checklist

* [ ] Are you running the latest v2 release? The list of releases is [here](https://github.com/urfave/cli/releases).
* [x] Did you check the manual for your release? The v2 manual is [here](https://github.com/urfave/cli/blob/master/docs/v2/manual.md)
* [x] Did you perform a search about this problem? Here's the [Github guide](https://help.github.com/en/github/managing-your-work-on-github/using-search-to-filter-issues-and-pull-requests) about searching.

## Packaging Setup

- [ ] My project is using go modules.
- [ ] My project is using vendoring.
- [ ] My project is automatically downloading the latest version.
- [x] I am unsure of what my packaging setup is.

## Describe the bug

I can't install and go get the package.
When I run, show this error:

```
go get github.com/urfave/cli/v2
package github.com/urfave/cli/v2: cannot find package "github.com/urfave/cli/v2" in any of:
        C:\Go\src\github.com\urfave\cli\v2 (from $GOROOT)
        E:\cs\golang\gopath\src\github.com\urfave\cli\v2 (from $GOPATH)
```
Although I can find these files at %GOPATH%\src\github.com\urfave\cli, 
and I use `import "github.com/urfave/cli/v2"`
show this error:
```
main.go:10:2: cannot find package "github.com/urfave/cli/v2" in any of:
	C:\Go\src\github.com\urfave\cli\v2 (from $GOROOT)
	E:\cs\golang\gopath\src\github.com\urfave\cli\v2 (from $GOPATH)
```

## To reproduce

Describe the steps or code required to reproduce the behavior

## Additional context

If the issue relates to a specific open source Github repo, please link that repo here.

If you can reproduce this issue with a public CI system, please link a failing build here.

## Want to fix this yourself?

We'd love to have more contributors on this project! If the fix for this bug is easily explained and very small, free free to create a pull request for it.

## Run `go env` and paste its output here

```
E:\cs\golang\gopath\src\github.com\cw1997\all2mif\src>go env
set GO111MODULE=
set GOARCH=amd64
set GOBIN=
set GOCACHE=C:\Users\changweimbp\AppData\Local\go-build
set GOENV=C:\Users\changweimbp\AppData\Roaming\go\env
set GOEXE=.exe
set GOFLAGS=
[...]

Issue #952, opened 2019-11-28
Support users who are using GOPATH mode


lynncyrin:
## Desired Outcome

I would like to use the methodology described here in this blog post => https://blog.golang.org/v2-go-modules to support users who aren't using go modules.

Specifically, that would mean moving to a directory setup like so:

```
github.com/urfave/cli @ master branch
/go.mod    → module github.com/urfave/cli
/v2/go.mod → module github.com/urfave/cli/v2
```

## Risks

This proposed change will make the git repo most "noisy", in that there will be more total files in the repo. It will be a bit more confusing to direct new contributors to the right place to add code. I think both those issues are easily solved with strong documentation, though!

## Motivation

The motivation here is the many many cases where we broke someone who wasn't using go modules, or were the reason why someone had to start using go modules. Examples:

- https://github.com/urfave/cli/issues/925#issuecomment-552683840
- https://github.com/urfave/cli/issues/925#issuecomment-553173182
- https://github.com/urfave/cli/issues/925#issuecomment-559121697
- https://github.com/urfave/cli/issues/926#issuecomment-552679445
- https://github.com/urfave/cli/issues/947#issue-528380907
- https://github.com/anubhavmishra/envoy-consul-sds/pull/4
- https://github.com/ethereum/go-ethereum/issues/20164

## Non-goals

Even with this proposal, I would still consider v1 go be "locked" at `1.22.X`. Which is to say, I am by no means suggesting that we start merging feature PRs into v1! v1 would remain in its "critical bug fixes only" state, it would simply live in the `master` branch.

## Our Paths Forward

This section describes how our paths forward from here would effect each of the following types of users:

1. users using GOPATH mode, who have already updated
2. users using GOPATH mode, who have not updated yet
3. users using vendoring
4. users using go modules

### If we don't make this change

> 1. users using GOPATH mode, who have already updated

These users will be broken again when we update the package to v3 https://github.com/urfave/cli/issues/833. This will also be broken **again** if we update to v4, or v5, etc...

> 2. users using GOPATH mode, who have not updated yet

All of these users are currently broken, and will remain broken unless they take some action. The most common actions people take are:

1. swap to using go modules, or
2. update to v2

> 3. users using vendoring

These users are currently ok, and will remain ok

> 4. users using go modules

These users are currently ok, and will remain ok

### If we make this change

> 1. users using GOPATH mode, who have already updated

These users will be broken once, because they will update to downgrade from v2 => v1

[...]

PR #966, opened 2019-11-29
Add GO111MODULE=on to readme

lynncyrin:
Discovered via https://github.com/urfave/cli/issues/957 and the research I did for https://github.com/urfave/cli/issues/952, `GO111MODULE=on` is required for some users when downloading v2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Usage documentation exists for each major version. Don't know what version you'r
 **Warning**: `v2` is in a beta state.
 
 ```
-$ go get github.com/urfave/cli/v2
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
To be honest, perhaps we should specify the go version required.

lynncyrin:
> Perhaps we should specify the go version required

We do, but the wording makes it a bit easy to miss IMO. Here's the readme right now =>

> Make sure you have a working Go environment. Go version 1.10+ is supported. See the install instructions for Go.

I think I'll take a pass on updating it for visibility.

Issue #931, opened 2019-11-14
Switch CI from travis?


lynncyrin:
My impression of CI systems in the recent past (eg. end of 2019) is that circleci has begun to surpass travis as the first choice for open source projects? Primarily because travis has started having very very long build queues.

This issue is about considering a switch from travis to circleci, all discussion and feedback is welcome!

AudriusButkevicius:
There is also github actions, which can do builds and all of that stuff.

lynncyrin:
Ah, and Github Actions just hit general availability this week! I'm super interested in that 🔍

saschagrunert:
I’d be still in favor of CircleCI over GitHub actions because of the performance. But both are way better than Travis. 👍

lynncyrin:
I think I'm gonna test out Github Actions for release automation (eg. https://github.com/urfave/cli/issues/923), it's a good fit and not performance intensive (since releases are rare).

lynncyrin:
closed via https://github.com/urfave/cli/pull/970

PR #970, opened 2019-12-01
GitHub actions

asahasrabuddhe:
## Motivation

To replace currently used TravisCI system with GitHub Actions

Relevant to #931

## Changes
 - Add a `v1.yml` and `v2.yml` file to `.github/workflows` directory to manage workflows for the `v1` branch and `master` branch respectively.
 - Replace `t.Log` and `t.Logf` functions with `log.Print` and `log.Printf` as using the previous function triggered a test failure in github actions environment

## Reviewer Guidelines

I am going to still keep the appveyor.yml as for some reason the build fails in Windows. I am looking into it.


asahasrabuddhe:
Isn’t this embarrassing! This same pipeline ran flawlessly on my fork. I’ll check this tomorrow but you’d kind of get an idea or what I’m trying to do

app_test.go
```diff
@@ -1422,7 +1423,7 @@ func TestApp_Run_CommandWithSubcommandHasHelpTopic(t *testing.T) {
 	}
 
 	for _, flagSet := range subcommandHelpTopics {
-		t.Logf("==> checking with flags %v", flagSet)
+		log.Printf("==> checking with flags %v", flagSet)
```
AudriusButkevicius: Why this change? I don't think its needed?

asahasrabuddhe: It’s needed because printing with t.Log causes the test case to fail on GitHub actions. Will share screenshots in a moment as I’m on mobile now.

This change doesn’t cause a change in behaviour but it does not cause a failure.

asahasrabuddhe: gherror

Please check this screenshot. You can see that the `t.Log` and `t.Logf` are treated as error and are causing the tests to fail. Hence the change.

asahasrabuddhe:
I have updated this PR with a few changes. We have two files now: `v1.yml` and `v2.yml` that run on PRs to `v1` and `master` branch respectively. Currently the pipeline replicates our Travis CI setup and runs all tests and pushes update to Codecov. 

@AudriusButkevicius I would need your help to setup the token in the secrets of our repo.

The future plan is to integrate a release note generating system and also automate the release process when a new tag is pushed.

lynncyrin:
Immediately after this gets 2 approvals, I or `@AudriusButkevicius` can change the repo setting to remove expected travis check:

> continuous-integration/travis-ci Expected — Waiting for status to be reported

asahasrabuddhe:
> Immediately after this gets 2 approvals, I or `@AudriusButkevicius` can change the repo setting to remove expected travis check:
> 
> > continuous-integration/travis-ci Expected — Waiting for status to be reported

I would also need one of your help to set the Codecov Token

app_test.go
```diff
@@ -1451,7 +1452,7 @@ func TestApp_Run_CommandWithSubcommandHasHelpTopic(t *testing.T) {
 		}
 
 		output := buf.String()
-		//t.Logf("output: %q\n", buf.Bytes())
+		//log.Printf("output: %q\n", buf.Bytes())
```
saschagrunert: I'd prefer to remove these comment if you don't mind.

app_test.go
```diff
@@ -1718,7 +1719,7 @@ func TestApp_Run_Categories(t *testing.T) {
 	}
 
 	output := buf.String()
-	//t.Logf("output: %q\n", buf.Bytes())
+	//log.Printf("output: %q\n", buf.Bytes())
```
saschagrunert: The same for this.

.github/workflows/v1.yml
```diff
@@ -0,0 +1,55 @@
+name: Run Tests
```
saschagrunert: (Non blocking and maybe something for the future)

Can we parameterize v1 and v2 yml somehow to reuse the  logic in there? The diff seems not very rich:
```diff
6c6
<       v1
---
>       master
48,49c48,49
<           go run build.go gfmrun docs/v1/manual.md
<           go run build.go toc docs/v1/manual.md
---
>           go run build.go gfmrun docs/v2/manual.md
>           go run build.go toc docs/v2/manual.md
```

app_test.go
rliebz: `t.Log` and `t.Logf` do not cause tests to fail, they just attach the logs to a particular test. [From the docs](https://golang.org/pkg/testing/#T.Log):

> For tests, the text will be printed only if the test fails or the -test.v flag is set.

I'm not sure where the `##[error]` formatting is coming from, but my guess is that we're just looking at the `stderr` stream and the test was failing because of something else further down. You can see that the error lines being printed still have a `PASS` underneath, although I'm not sure where to find that log in particular.

I've noticed that this project tends to print out a lot of things when any test fails, which is something we should address at some point, but `t.Logf` is almost always the right call in tests.

.github/workflows/v1.yml
```diff
@@ -0,0 +1,55 @@
...
+
+on:
+  pull_request:
+    branches:
+      v1
+
+env:
+  GO111MODULE: on
+  GOPROXY: https://proxy.golang.org
```
rliebz: Any reason we're setting this explicitly?

The default is `https://proxy.golang.org,direct`, which seems better because we have a fallback if the proxy is down. From `go help modules`:

> The default setting for GOPROXY is "https://proxy.golang.org,direct", which means to try the
Go module mirror run by Google and fall back to a direct connection if the proxy reports that it does not have the module (HTTP error 404 or 410). See https://proxy.golang.org/privacy for the service's privacy policy. If GOPROXY is set to the string "direct", downloads use a direct connection to source control servers.

.github/workflows/v1.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Checkout Code
+        uses: actions/checkout@v1
+        with:
+          ref: ${{ github.ref }}
+
+      - name: Install Dependencies
+        run: |
+          go get github.com/urfave/gfmrun/cmd/gfmrun
+          go get golang.org/x/tools/cmd/goimports
+          go mod tidy
```
rliebz: I'm assuming the reason to run `go mod tidy` is because installing the CLI tools inside the working directory messes up the `go.mod` file. The downside to this pattern is we're making unnecessary modifications to `go.mod`, and after this point, we can't verify that the `go.mod` was actually tidied before the PR, which is probably something we want to do.

There's a few different ways to handle this:
- Install binaries instead of building from source. I don't think `goimports` distributes a binary, but we can probably add one to `gfmrun` at some point. I prefer to do this when possible.
- Install with `GO111MODULE=off`. This avoids writing the `go.mod` file, but has the downside of ignoring a dependency's `go.mod` if it exists. I typically do this if they don't have a binary, since it's pretty unusual for Go CLIs to break `go get` installations without providing an alternative installation method. 
- Create a temporary directory, do a `go mod init tmp` there, then do `go get` on CLI dependencies. This is a few extra lines to do, but it does respect the `go.mod` for transitive dependencies.

.github/workflows/v1.yml
asahasrabuddhe: It was a part of the old file which I simply carried over. I think @saschagrunert had added this and he would be the correct person to shed light on this.

.github/workflows/v1.yml
asahasrabuddhe: You are right in guessing the reason about using `go mod tidy`. However we do not need to worry about modifying the `go.mod` accidentally. We use `git diff --exit-code` at the end of our CI process which ensures that there are no files modified.

About `goimports`it is carried forward from a legacy build process. I think if everyone thinks that it's useless, I'd be more than happy to remove it. I although am in favour of updating the `gfmrun` project to build and release a binary artifact and then use that artifact in our build process.

Once that is done, I think even `go mod tidy` will become redundant.

.github/workflows/v1.yml
saschagrunert: It speeds things up on go 1.12 and is the default for 1.13, so I think we can remove it but it should not make much differences.

rliebz: Just checked the `1.12` docs. Default on `1.12` is effectively `direct`, and `1.13` is `https://proxy.golang.org,direct`. 

I'm happy to keep it in to speed things up for pre-1.13, but I'd recommend changing it to include the `,direct` at the end for a fallback in case the proxy is down for whatever reason.

rliebz: Oops, 1.12 doesn't support comma separated lists. This is fine if we want to keep it, I'm going to resolve this.

asahasrabuddhe: Alright!

.github/workflows/v1.yml
rliebz: I actually use `goimports` in all my projects and usually enforce it using `golangci-lint`, although you could enforce it by just running `goimports` and checking the diff. So I'm 👍 if we want to add it back into the pipeline.

A quick grep of the repo makes it look like we aren't actually using `goimports` anywhere, so we can either remove it or table it for now and deal with that separately if we want to keep the scope of this change set smaller.

.github/workflows/v1.yml
asahasrabuddhe: Done

app_test.go
asahasrabuddhe: Done

app_test.go
asahasrabuddhe: Done

.github/workflows/v1.yml
asahasrabuddhe: `goimports` has been removed and now we are using the pre-compiled binary for `gfmrun`

app_test.go
asahasrabuddhe: @rliebz Please check

> It’s needed because printing with t.Log causes the test case to fail on GitHub actions.

I am aware that `t.Log` and `t.Logf` does not cause the test cases to fail. The GitHub actions environment is interpreting it as such. The tests are just outputting stuff to be verbose. They are not real errors.

I tried running the test cases on my machine and they ran fine. It's just on the GitHub Actions environment where they are failing. Hence this change.

.github/workflows/cli.yml
```diff
@@ -0,0 +1,64 @@
...
+          ref: ${{ github.ref }}
+
+      - name: Install Dependencies
+        run: |
+          mkdir -p $GOPATH/bin
+          curl -s https://api.github.com/repos/urfave/gfmrun/releases/latest | grep gfmrun-linux-amd64 | grep download | cut -d '"' -f 4 | wget -O $GOPATH/bin/gfmrun -i -
+          npm install markdown-toc
+
+      - name: Run Tests (v1)
+        if: contains(github.ref, 'v1')
```
rliebz: Looks like the current setup doesn't actually ever execute the tests. Note the gray squares with `0s`:
image


I played around with this a bit on my fork, it looks like `github.ref` refers to the branch triggering the PR, and `github.base_ref` is the target. So if we're discriminating based on where the PR is supposed to land, we'll want to use `github.base_ref` to get tests to run

Source: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/contexts-and-expression-syntax-for-github-actions#github-context

app_test.go
rliebz: I couldn't find where the logs in the screenshot were coming from, so I had to play around with actions on my fork. I am not able to reproduce the behavior you're describing in this PR, although it appears we're only getting the ✅ on these because it's skipping the tests (see [my comment here](https://github.com/urfave/cli/pull/970/files#r354653289))

Whether or not we run `t.Logf` versus `log.Printf`, the tests are passing, but the error that's actually causing the test to fail is unrelated:

```
2019/12/06 03:36:42 exec: "gfmrun": executable file not found in $PATH
exit status 1
##[error]Process completed with exit code 1.
```

Since the `gfmrun` step only runs _after_ tests complete, I believes the unit tests are passing (albeit with a lot of extra noise), but the test step is failing because `gfmrun` is not available on the path. Turns out, we need to make it executable:

```diff
      - name: Install Dependencies
        run: |
          mkdir -p $GOPATH/bin
          curl -s https://api.github.com/repos/urfave/gfmrun/releases/latest | grep gfmrun-linux-amd64 | grep download | cut -d '"' -f 4 | wget -O $GOPATH/bin/gfmrun -i -
+         chmod +x $GOPATH/bin/gfmrun
          npm install markdown-toc
```

I then got even further, but ran into a strange error:
```
##[error]The operation was canceled.
```

Not sure what causes that. Because the symbol we're getting is a stop sign looking thing and not a big red X, I'm guessing the job is being terminated by GitHub actions for some reason--taking too long? Taking too much memory? Hard to say, Google didn't help me too much here. It passes sometimes and fails others when I re-run it. Note the symbols on the left:

image


The failures on macOS in that case are caused because we're still pulling in the linux version of `gfmrun`.

In terms of why logging could make it fail, from what I can tell, it's just bad UI on GitHub's part that gives misleading messages. GitHub actions is trying to be clever, and it considers anything that comes out of `stderr` to be an "error annotation". This is not really a good assumption to make, but it does not actually prevent the build from failing. In this case, the build passed with 10 error annotations:

image

I realize that's a pretty big information dump, but hopefully that's enough context to help figure out what to do from here. If you're interested in digging deeper into the exact version of the code I ran or the logs, you should be able to see everything here: https://github.com/rliebz/cli/pull/1/checks?check_run_id=336127776

.github/workflows/cli.yml
asahasrabuddhe: @rliebz I noticed this now. Will fix.

app_test.go
asahasrabuddhe: I think I have managed to fix it. Thanks for all your work.

app_test.go
```diff
@@ -2122,7 +2121,7 @@ func TestWhenExitSubCommandWithCodeThenAppQuitUnexpectedly(t *testing.T) {
 	var exitCodeFromExitErrHandler int
 	app.ExitErrHandler = func(c *Context, err error) {
 		if exitErr, ok := err.(ExitCoder); ok {
-			t.Log(exitErr)
+			log.Print(exitErr)
```
rliebz: Looks like there's one more `log.Print` hiding out here

.github/workflows/cli.yml
```diff
@@ -0,0 +1,74 @@
...
+      - name: Checkout Code
+        uses: actions/checkout@v1
+        with:
+          ref: ${{ github.ref }}
+
+      - name: Install Dependencies (Linux)
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          mkdir -p $GOPATH/bin
+          curl -s https://api.github.com/repos/urfave/gfmrun/releases/latest | grep gfmrun-linux-amd64 | grep download | cut -d '"' -f 4 | wget -O $GOPATH/bin/gfmrun -i -
```
rliebz: In case `gfmrun` ever makes a breaking change, we'd be better off pinning the dependency version. This way, we can control exactly when our pipeline breaks.

We can also avoid keeping two different install sections if we want by relying on `go env` to tell us what OS we're dealing with. These two things together give:

```bash
wget -O $GOPATH/bin/gfmrun https://github.com/urfave/gfmrun/releases/download/v1.2.14/gfmrun-$(go env GOOS)-amd64-v1.2.14
```

asahasrabuddhe: I doubt ‘gfmrun’ would make such a change as I’m one of the maintainers of ‘gfmrun’ but this makes sense

app_test.go
asahasrabuddhe: Nice eye!

lynncyrin:
```suggestion
			t.Log(exitErr)
```

asahasrabuddhe:
I would need someone's help in figuring out why the Windows builds are failing. I have a separate branch setup on my fork and you can see the results of the last run here:

https://github.com/asahasrabuddhe/cli/commit/fcdaa08dc8769979b8c1e9e8ee812060ff418069/checks?check_suite_id=345469919

Now the weird part is that the Windows tests fail when it comes to testing the outputs of ManPage and Markdown related functions. I tried downloading the logs and comparing the outputs (using diff) and they were identical! 

I would really appreciate someone who can try to reproduce this on Windows and help me out with the final step. Once this makes it, I think we can get rid of appveyor too :)

rliebz:
@asahasrabuddhe I've run into similar issues before on windows—there's a good chance it's a line endings issue.

I've currently got this in one of my code bases, just to make tests pass on windows:
```go
contents, err := ioutil.ReadFile(tt.path)
if err != nil {
	t.Fatal(err)
}

// Ignore windows line endings
got := bytes.ReplaceAll(contents, []byte("\r\n"), []byte("\n"))
```

A tool I've found incredibly useful for diffing large bodies of text is https://github.com/google/go-cmp. It prints out a line by line diff, so it's a lot clearer whether the reason they don't match is because the executable is named `myapp` vs. `myapp.exe`, the line endings are different, or there's a trailing newline in the test case, all of which are things I've definitely run into before.

asahasrabuddhe:
@AudriusButkevicius @lynncyrin Please remove the pending Travis CI check so that this can be merged. Alternatively, you could also consider giving me access so that I could do the same.

asahasrabuddhe:
> @asahasrabuddhe I've run into similar issues before on windows—there's a good chance it's a line endings issue.
> 
> I've currently got this in one of my code bases test files, just to make tests pass on windows:
> 
> ```go
> contents, err := ioutil.ReadFile(tt.path)
> if err != nil {
> 	t.Fatal(err)
> }
> 
> // Ignore windows line endings
> got := bytes.ReplaceAll(contents, []byte("\r\n"), []byte("\n"))
> ```
> 
> A tool I've found incredibly useful for diffing large bodies of text is https://github.com/google/go-cmp. It prints out a line by line diff, so it's a lot clearer whether the reason they don't match is because the executable is named `myapp` vs. `myapp.exe`, the line endings are different, or there's a trailing newline in the test case, all of which are things I've definitely run into before.

Awesome! Thanks a lot for your insight. I will work on this and come up with a separate PR.

lynncyrin:
> Please remove the pending Travis CI check so that this can be merged.

✔️ done!

Issue #968, opened 2019-11-29
Consider using a "stale bot"


lynncyrin:
The bot => https://github.com/marketplace/stale

Checking on stale issues / PRs is currently something I do manually, and I would prefer it be automated

@urfave/cli I would like to stop having to do this manually, so I'm gonna implement this sometime after Dec 1st unless I get a 👎 react. Do also feel free to let me know if you're 👍 !

asahasrabuddhe:
I think this is a good idea. It would certainly be more productive.

lynncyrin:
@AudriusButkevicius I believe you now have an email from Github about this ✉️

AudriusButkevicius:
I've approved it.

Issue #753, opened 2018-06-10
`Compiled` metadata is not used


genieplus:
The `Compiled` field in `App` and `Command` struct is not used neither printed with the default help template.
Shall it be removed ?

asahasrabuddhe:
@urfave/cli your opinions here?

lynncyrin:
I need this issue to have a bit more context on the field in question, and some possible issues that could arise from removing it or adding it to the version output.

Issue #744, opened 2018-05-18
Misleading user error message when using flags


MichaelMure:
Hi,

As far as I understand, `cli` expect the root/global flags to be provided between the binary name and the subcommand name, and the command flags to be provided at the end:

```
binary-name [global options] command [command options] [arguments...]
```

As this pattern is not that common, it's easy for a new user to try to provide the root flags at the end of the command. In this case, he will be greeted with a rather unhelpful error message:

```
Incorrect Usage: flag provided but not defined: -host
```
Note that the message is the same when trying a flag that doesn't exist.

Would it be possible to get a more explanatory message in these cases ? I tried to make a workaround using `OnUsageError` but it's impractical for several reasons:
- you have to match and parse the error message
- it's not a simple error message rewrite, it's a different code path inside `cli` and you have to emulate some behavior like printing the help text manually
- you have to plug this workaround in every single command

thedoc31:
Agreed. This format is not globally recognized and goes against how most CLIs work, where global options are not position-sensitive. This is extremely confusing, especially for experienced CLI end users.

lynncyrin:
I understand that's going on here 👍 I would be happy to help someone with a pull request for this issue!

lynncyrin:
I definitely agree that what we want is to allow global flags to be specified at any position. I'd love to see a PR for that, or for the issue of the misleading error message, or both!

AudriusButkevicius:
Thats not true, I think kubectl also has position sensitive global flags, so we're not the only ones.

lynncyrin:
Hey @MichaelMure & @thedoc31, this issue may have been fixed by our [v2 release](https://github.com/urfave/cli/releases/tag/v2.0.0)? Check it out and let me know what you think.

thedoc31:
@lynncyrin I'll see if the internal project using the framework has plans to upgrade to v2. Thanks for following up!

lynncyrin:
This issue is still waiting for someone to investigate if it's still a problem, and stand up a PR if so 🙏

carhartl:
> This issue is still waiting for someone to investigate if it's still a problem

Still seeing it with v2:

```
go run main.go cmd --global foo
Incorrect Usage: flag provided but not defined: -global

NAME:
   main cmd

USAGE:
   main cmd command [command options] [arguments...]
```

with:

```go
package main

import (
	"fmt"
	"log"
	"os"

	"github.com/urfave/cli/v2"
)

func main() {
	app := &cli.App{
[...]

Issue #707, opened 2018-01-30
Inconsistent error handling between App.OnUsageError and App.Before during App.Run


diliop:
When `OnUsageError` is set for a `cli.App` instance, one has full control of what is printed to stderr when an error is encountered. That is not the case when setting `Before`. 

Looking at how `OnUsageError` is handled in `App.Run` in [app.go](https://github.com/urfave/cli/blob/master/app.go) [[c6eb2a0](https://github.com/urfave/cli/commit/c6eb2a051026c083d4e33591f8d6e95d5f4189dc)]:
```
if a.OnUsageError != nil {
	err := a.OnUsageError(context, err, false)
	a.handleExitCoder(context, err)
	return err
}
```
and for `Before`:
```
if a.Before != nil {
	beforeErr := a.Before(context)
	if beforeErr != nil {
		fmt.Fprintf(a.Writer, "%v\n\n", beforeErr)
		ShowAppHelp(context)
		a.handleExitCoder(context, beforeErr)
		err = beforeErr
		return err
	}
}
```
The issue I'm having is with the call to `ShowAppHelp(..)` when `Before` is set. Essentially the experience to the user is different when a usage error is encountered and I want to print custom output versus when I am checking for lets say flag value validity before running any subcommands. Ideally I would prefer the 2 if statements above to be identical and the default case to call `ShowAppHelper(..)` as is currently the case  i.e. for `Before`:
```
if a.Before != nil {
	beforeErr := a.Before(context)
	if beforeErr != nil {
		a.handleExitCoder(context, beforeErr)
		err = beforeErr
		return err
	}
}
```
I've looked around to see if there is another way to modify this behavior but nothing obvious stands out.

jszwedko:
Hi @diliop,

Thanks for the detailed issue! If I had to summarize it, would it be fair to say that you want to consider errors in `Before` functions to be considered and handled in the same way as "usage errors"?

The difference when running a subcommand is also confusing.

I agree that these three cases should be unified.

diliop:
Yes @jszwedko that is correct! 

From a philosophical standpoint I want the error handling (and in this case error printing to stdout/stderr) to be the same across all types of errors that arise at different points in the execution of a command/subcommand. This essentially would be the user experience consistent. Currently I can set "usage errors" to print something like this:
```
cmd subcmd: --flag must be set
See 'cmd subcmd --help'.
```
[...]

PR #971, opened 2019-12-03
Add stale bot configuration

lynncyrin:
## Motivation

Fixes https://github.com/urfave/cli/issues/968

## Changes

Adds the configuration file for https://github.com/probot/stale

## Reviewer Guidelines

Of particular interest for reviews:
- should there be any wording changes to the bot messages?
- should the bot comment when closing stale issues?
- do the `daysUntilStale` and `daysUntilClose` values seem reasonable?

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,63 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 90
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 30
+
+# Only issues or pull requests with all of these labels are check if stale. Defaults to `[]` (disabled)
+onlyLabels: []
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
+  - pinned
+  - security
+  - "type: maintenance" 
+
+# Set to true to ignore issues in a project (defaults to false)
+exemptProjects: false
+
+# Set to true to ignore issues in a milestone (defaults to false)
+exemptMilestones: false
+
+# Set to true to ignore issues with an assignee (defaults to false)
+exemptAssignees: false
+
+# Label to use when marking as stale
+staleLabel: "status: stale"
+
+# Comment to post when marking as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. Please add a comment bumping this issue if you're still
+  interested in it's resolution! Thanks for your help, please let us know
+  if you need anything else.
+
+# Comment to post when removing the stale label.
+unmarkComment: >
+  This issue has been bumped and is no longer marked as stale! Feel free
+  to bump it again in the future, if it's still relevant.
+
+# Comment to post when closing a stale Issue or Pull Request.
+# closeComment: >
+#   Your comment here.
+
+# Limit the number of actions per hour, from 1-30. Default is 30
+limitPerRun: 30
+
+# Limit to only `issues` or `pulls`
+# only: issues
+
+# Optionally, specify configuration settings that are specific to just 'issues' or 'pulls':
+# pulls:
+#   daysUntilStale: 30
+#   markComment: >
+#     This pull request has been automatically marked as stale because it has not had
+#     recent activity. It will be closed if no further activity occurs. Thank you
+#     for your contributions.
+
+# issues:
+#   exemptLabels:
+#     - confirmed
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
Just wonder if this will just brutally blast genuine accepted issues as closed. How does one accept something is an issue and keep it open? Time does necesserily make it a non-issue.

.github/stale.yml
```diff
@@ -0,0 +1,63 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 90
```
rliebz: I've never actually worked extensively on a project with a stale bot, but this seems short to me. 

To put this into concrete terms:
- Only the most recent 15 of the currently open 62 issues were opened less than 90 days ago. I'm not sure how to easily tell how many of the remaining issues would become stale.
- For an issue that one person wanted to keep open, they'd have to comment 4 times a year to prevent it from being closed, unless we add a label to keep it open forever.

If we did something like 365 days, we'd instead end up with just over half the issues potentially being flagged—27/62 were opened in the last year. The upside is it gives issues more breathing room for things people care about without them feeling like they're nagging the maintainers. 

I'm not attached to any number in particular, and we can always change it, but it seems like starting conservatively and shortening if issues are piling up too quickly would be safer than starting with a small window.

lynncyrin:
@AudriusButkevicius 

> Just wonder if this will just brutally blast genuine accepted issues as closed.

Do you have a specific type of issue that you would like to keep open indefinitely? If so, can you describe it to me?

For context, I've already added the following types of labels as exceptions:

https://github.com/urfave/cli/blob/93911a08af70baab728da0637da8c5524210bb75/.github/stale.yml#L14-L17

---

> How does one accept something is an issue and keep it open?

1. Add a label designating it as an issue that should be kept open
2. Add a comment

---

> Time does necessarily make it a non-issue.

_(copying from here: https://github.com/probot/stale#is-closing-stale-issues-really-a-good-idea)_

**Is closing stale issues really a good idea?**

In an ideal world with infinite resources, there would be no need for this app.

But in any successful software project, there's always more work to do than people to do it. As more and more work piles up, it becomes paralyzing. Just making decisions about what work should and shouldn't get done can exhaust all available resources. In the experience of the maintainers of this app—and the hundreds of other projects and organizations that use it—focusing on issues that are actively affecting humans is an effective method for prioritizing work.

To some, a robot trying to close stale issues may seem inhospitable or offensive to contributors. But the alternative is to disrespect them by setting false expectations and implicitly ignoring their work. This app makes it explicit: if work is not progressing, then it's stale. A comment is all it takes to keep the conversation alive.

lynncyrin:
@rliebz

> My only concern is the time frame here.

For context, the default timeframe is

```
daysUntilStale: 60
daysUntilClose: 7
```

The timeframe as have as of this comment is

```
daysUntilStale: 90
daysUntilClose: 30
```

> Do we know how many of our open issues would be flagged with this configuration?

The term "flagged" is ambiguous here, so I'll respond with details about both "stale" and closed"

With the timeframe mentioned in this comment, which is `90 stale, 30 closed`, that would be:

- 6 issues would be marked as stale. I used [this search](https://github.com/urfave/cli/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+updated%3A%3C2019-09-01) via [these instructions](https://github.com/probot/stale#how-are-issues-and-pull-requests-considered-stale) which produced these results:
  - https://github.com/urfave/cli/issues/753
  - https://github.com/urfave/cli/issues/744
  - https://github.com/urfave/cli/issues/707
  - https://github.com/urfave/cli/issues/698
  - https://github.com/urfave/cli/issues/585
  - https://github.com/urfave/cli/issues/549
- 0 issues would be closed.  I used [this search](https://github.com/urfave/cli/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+updated%3A%3C2019-08-01) which at time of writing produces 0 results

.github/stale.yml
lynncyrin: @rliebz

> I've never actually worked extensively on a project with a stale bot

Everyone has different experiences ✨ this is the first project I've worked _(for longer than a few weeks)_ on that doesn't have "stale" handling of some sort.

> Only the most recent 15 of the currently open 62 issues were opened less than 90 days ago.

The stale bot uses the time of the issue's most recent **update**, not the **creation time**. More details here => https://github.com/probot/stale#how-are-issues-and-pull-requests-considered-stale

_(the rest of your comment likely needs to be reconsidered in light of this new information, so I'm going to skip replying to it for the moment)_

rliebz:
@lynncyrin—Six issues is a totally reasonable number to start out with, definitely less than I was expecting. That resolves my concern.

AudriusButkevicius:
I dunno, let's try it out, but I fear we are not on the same level as k8s, where we don't have enough volume of people on the issue tracker to keep "it's genuinely broken" issues alive.

PR #973, opened 2019-12-04
Inconsistent Name for Application from inside Subcommands (v1)

### LINKED ISSUE #783: Add an `App.ProgramName`
## Motivation & Context

I have the following code:

```
app := cli.NewApp()
app.Name = "myprogramname"
```

I then add a sub command with the name "foo", and in that sub command I run:

```
func fooAction(c *cli.Context) error {
	program := c.App.Name
	log.Printf("program is: %s", program)
}
```

Unfortunately this prints "program is: `myprogramname foo`" instead of leaving out the "`foo`".
This is unexpected, particularly because the "program" did not change. If it is desired to have some sort of variable that appends the sub command names on it, then it should be something different. I use this variable to ascertain the "base name" of the program reliably, for log messages, etc...

## Requested Change

Add an `app.ProgramName`, that is always `myprogramname`, regardless of whether or not a particular subcommand is being executed.

---

_heavily edited by @lynncyrin!_

asahasrabuddhe:
## Motivation

Fixes #783 

The motivation for doing this for v1 is to provide support, and a valuable feature that the users want to stick to v1 can appreciate.

## Release Notes

Adds `ProgramName` and `CommandName`. The `ProgramName` will always contain the name of the application as defined by the user. The `CommandName` will include the `ProgramName` with the name of the command, and it's parent commands.

## Changes

- `app.go` - Added the `ProgramName` and `CommandName` members to the `App` struct.
- `command.go` - Added logic to initialise and update the `ProgramName` and `CommandName` members.
- `command-test.go` - Added a test case to validate above change

## Testing

I used the following program to help test: 

```go
package main

import (
	"fmt"
	"log"
	"os"

	"github.com/urfave/cli"
)

func main() {
	app := cli.NewApp()
	app.Name = "myprogramname"
		fmt.Println("c.App.Name for app.Action is", c.App.Name)
		fmt.Println("c.App.ProgramName for app.Action is", c.App.ProgramName)
		fmt.Println("c.App.CommandName for app.Action is", c.App.CommandName)
		return nil
	}
	app.Commands = []cli.Command{
		{
			Name: "foo",
				fmt.Println("c.App.Name for app.Commands.Action is", c.App.Name)
				fmt.Println("c.App.ProgramName for app.Commands.Action is", c.App.ProgramName)
				fmt.Println("c.App.CommandName for app.Commands.Action is", c.App.CommandName)
				return nil
			},
			Subcommands: []cli.Command{
				{
					Name: "bar",
						fmt.Println("c.App.Name for App.Commands.Subcommands.Action is", c.App.Name)
						fmt.Println("c.App.ProgramName for App.Commands.Subcommands.Action is", c.App.ProgramName)
						fmt.Println("c.App.CommandName for App.Commands.Subcommands.Action is", c.App.CommandName)
						return nil
					},
					Subcommands: []cli.Command{
						{
							Name: "baz",
								fmt.Println("c.App.Name for App.Commands.Subcommands.Subcommands.Action is", c.App.Name)
								fmt.Println("c.App.ProgramName for App.Commands.Subcommands.Subcommands.Action is", c.App.ProgramName)
								fmt.Println("c.App.CommandName for App.Commands.Subcommands.Subcommands.Action is", c.App.CommandName)
								return nil
							},
						},
					},
				},
			},
		},
	}

	err := app.Run(os.Args)
	if err != nil {
		log.Fatal()
	}
}
```

The same program was modified to write the test case.

## Reviewer Guidelines

The PR is marked as a v2 feature. However, I thought that this, being a simple change, it would be worth our while to get this into v1 for the sake of sanity.


asahasrabuddhe:
This is still WIP

lynncyrin:
Similarly to https://github.com/urfave/cli/pull/974 --- I consider this PR WIP and not approve-able until consensus on the desired solution is reached, that conversation should happen here => https://github.com/urfave/cli/issues/783

lynncyrin:
Is this still relevant as of https://github.com/urfave/cli/pull/975?

lynncyrin:
For the sake of making this the most recent comment - this PR is blocked in discussion in => https://github.com/urfave/cli/issues/783

meatballhat:
In the end, I think I'd like to close this without merging so that the v1 branch can remain strictly for patches (1.22.x). I'll be opening a separate PR with only the CI config changes.

PR #974, opened 2019-12-04
Inconsistent Name for Application from inside Subcommands

### LINKED ISSUE #783: Add an `App.ProgramName`

asahasrabuddhe:
## Motivation

Fixes #783 

The current behavior of the `app.Name` is not consistent and results in a bad experience for the users. The PR attempts to fix this until such time a better solution is in place (v3)

## Release Notes

Adds `ProgramName` and `CommandName`. The `ProgramName` will always contain the name of the application as defined by the user. The `CommandName` will include the `ProgramName` with the name of the command, and it's parent commands.

## Changes

- `app.go` - Added the `ProgramName` and `CommandName` members to the `App` struct. Initialize `ProgramName` for the first time.
- `command.go` - Added logic to initialise and update the `ProgramName` and `CommandName` members. Make use of the context lineage to set the `ProgramName` correctly.
- `command-test.go` - Added a test case to validate above change'
- `helpers_test.go` - Added `equal` method to compare slices

## Testing

I used the following program to help test: 

```go

package main

import (
	"fmt"
	"log"
	"os"

	"github.com/urfave/cli/v2"
)

func main() {
	app := cli.NewApp()
	app.Name = "myprogramname"
		fmt.Println("c.App.Name for app.Action is", c.App.Name)
		fmt.Println("c.App.ProgramName for app.Action is", c.App.ProgramName)
		fmt.Println("c.App.CommandName for app.Action is", c.App.CommandName)
		return nil
	}
		{
			Name: "foo",
				fmt.Println("c.App.Name for app.Commands.Action is", c.App.Name)
				fmt.Println("c.App.ProgramName for app.Commands.Action is", c.App.ProgramName)
				fmt.Println("c.App.CommandName for app.Commands.Action is", c.App.CommandName)
				return nil
			},
			Subcommands: []*cli.Command{
				{
					Name: "bar",
						fmt.Println("c.App.Name for App.Commands.Subcommands.Action is", c.App.Name)
						fmt.Println("c.App.ProgramName for App.Commands.Subcommands.Action is", c.App.ProgramName)
						fmt.Println("c.App.CommandName for App.Commands.Subcommands.Action is", c.App.CommandName)
						return nil
					},
					Subcommands: []*cli.Command{
						{
							Name: "baz",
								fmt.Println("c.App.Name for App.Commands.Subcommands.Subcommands.Action is", c.App.Name)
								fmt.Println("c.App.ProgramName for App.Commands.Subcommands.Subcommands.Action is", c.App.ProgramName)
								fmt.Println("c.App.CommandName for App.Commands.Subcommands.Subcommands.Action is", c.App.CommandName)
								return nil
							},
						},
					},
				},
			},
		},
	}

	err := app.Run(os.Args)
	if err != nil {
		log.Fatal(err)
	}
}

```

The same program was modified to write the test case.

## Reviewer Guidelines

The PR is marked as a v2 feature. However, I thought that this, being a simple change, it would be worth our while to get this into v1 for the sake of sanity.


AudriusButkevicius:
I don't think we should add any public API, we should just fix App.Name to always be the same thing.
I think command can be retrieved from the context?

asahasrabuddhe:
> I don't think we should add any public API, we should just fix App.Name to always be the same thing.
> I think command can be retrieved from the context, or if the user is generating functions (the only reason why I can explain you'd need a name), he could curry (functionally speaking) the name into the acting function.

@lynncyrin  ☝️ 

@AudriusButkevicius You are right about currying through the context. We actually have a `ctx.Lineage` call available which would return a slice of contexts from the current till the parent which would help the user do the same.

lynncyrin:
> I don't think we should add any public API, we should just fix App.Name to always be the same thing.

It's unclear if anyone relies on the current behavior of `App.Name`. So until someone can provide evidence that `App.Name` was always intended to ever return the program name, I consider this change to be a feature request rather than a fix.

lynncyrin:
> So until someone can provide evidence

Also, this should be happening inside the issue => https://github.com/urfave/cli/issues/783 rather than inside this pull request. Pull requests aren't a good context for collecting information and discussion like this.

lynncyrin:
I consider this PR WIP and not approve-able until consensus on the desired solution is reached, that conversation should happen here => https://github.com/urfave/cli/issues/783

lynncyrin:
Similarly to https://github.com/urfave/cli/pull/973, is this still relevant as of https://github.com/urfave/cli/pull/975?

lynncyrin:
For the sake of making this the most recent comment - this PR is blocked in discussion in => https://github.com/urfave/cli/issues/783

asahasrabuddhe:
@meatballhat i see there’s a test failing. Will take a look at this later today.

dearchap:
@asahasrabuddhe Can you check if we still need this in latest release ? I've made some changes that actually fixes this.

dearchap:
Fixed in latest release.

PR #975, opened 2019-12-04
add RunWithContext + remove signal cancellation

marwan-at-work:
Fixes https://github.com/urfave/cli/issues/945

### DIFF
--- a/app.go
+++ b/app.go
@@ -208,13 +208,13 @@ func (a *App) useShortOptionHandling() bool {
 // Run is the entry point to the cli app. Parses the arguments slice and routes
 // to the proper flag/args combination
 func (a *App) Run(arguments []string) (err error) {
-	return a.RunWithContext(context.Background(), arguments)
+	return a.RunContext(context.Background(), arguments)
 }
 
-// RunWithContext is like Run except it takes a Context that will be
+// RunContext is like Run except it takes a Context that will be
 // passed to its commands and sub-commands. Through this, you can
 // propagate timeouts and cancellation requests
-func (a *App) RunWithContext(ctx context.Context, arguments []string) (err error) {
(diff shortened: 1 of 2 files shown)


app.go
```diff
@@ -224,7 +232,7 @@ func (a *App) Run(arguments []string) (err error) {
 
 	err = parseIter(set, a, arguments[1:], shellComplete)
 	nerr := normalizeFlags(a.Flags, set)
-	context := NewContext(a, set, nil)
+	context := NewContext(a, set, &Context{Context: ctx})
```
AudriusButkevicius: Will this not blow.up if user did `RunWithContext(nil, args)`?

AudriusButkevicius: Will this still work of user does RunWithContext(nil, args)

app.go
```diff
@@ -207,6 +208,13 @@ func (a *App) useShortOptionHandling() bool {
 // Run is the entry point to the cli app. Parses the arguments slice and routes
 // to the proper flag/args combination
 func (a *App) Run(arguments []string) (err error) {
+	return a.RunWithContext(context.Background(), arguments)
+}
+
+// RunWithContext is like Run except it takes a Context that will be
+// passed to its commands and sub-commands. Through this, you can
+// propagate timeouts and cancellation requests
+func (a *App) RunWithContext(ctx context.Context, arguments []string) (err error) {
```
AudriusButkevicius: Looking at go net lib, this should probably be RunContext

app.go
rliebz: Passing a nil Context is always a developer error.

[From the docs](https://golang.org/pkg/context/):

> Do not pass a nil Context, even if a function permits it. Pass context.TODO if you are unsure about which Context to use. 

We shouldn't need to support that

marwan-at-work:
@rliebz @AudriusButkevicius PR updated :)

PR #977, opened 2019-12-05
Add suggestions support

### LINKED ISSUE #895: "Did you mean" hint on wrong command or flag
We could tell the user what the closest command or argument match is by calculating the [levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) between them.

What to do you think?

saschagrunert:
## Motivation

The new option `app.Suggest` enables command and flag suggestions via the jaro-winkler distance algorithm. Flags are scoped to their appropriate commands whereas command suggestions are scoped to the current command level.  

Closes #895

## Release Notes

Add "Did you mean" suggestion support to flags and commands. This feature can be enabled by setting `app.Suggest = true`.

## Changes

* Added a new suggestions.go file where all the implementation details go in
* Changed critical paths to show-up the suggestions
* Added test cases as well

## Testing

Added unit tests :)

## Reviewer Guidelines

The scope of the suggestions should be correct, that we do not suggest flags from other sub-commands.


parse.go
```diff
@@ -67,6 +66,19 @@ func parseIter(set *flag.FlagSet, ip iterativeParser, args []string, shellComple
 	}
 }
 
+const providedButNotDefinedErrMsg = "flag provided but not defined: -"
```
asahasrabuddhe: How would this work with flags having two hyphens as a prefix?

saschagrunert: I tested it locally and it works the same. The error message is always this constant.

AudriusButkevicius:
Should this be an out of the box OnUsageError handler ? Just wondering, not proposing.

saschagrunert:
> Should this be an out of the box OnUsageError handler ? Just wondering, not proposing.

Seems a matter of taste to me, so I'm open for a vote :)

Code coverage seems broken for now, see https://github.com/urfave/cli/pull/985#issuecomment-562943707

saschagrunert:
The coverage now looks better :)
https://codecov.io/github/urfave/cli/commit/2cb8524de32183d2eed9d355e3007417bd6708ae

lynncyrin:
Minor request: can you avoid force pushing branches that I've already reviewed? I can code review faster when I can use Github's "commits since last push" feature, and force pushing breaks that.

saschagrunert:
> Minor request: can you avoid force pushing branches that I've already reviewed? I can code review faster when I can use Github's "commits since last push" feature, and force pushing breaks that.

Sure, I can do this.

suggestions.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	flags := a.Flags
+	if command != "" {
+		cmd := a.Command(command)
+		if cmd == nil {
+			return "", err
+		}
+		flags = cmd.Flags
+	}
```
lynncyrin: ⭐⭐ _(seriously consider)_ consider also adding global flags to the command flags?

suggestions.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func (a *App) suggestFlag(flags []Flag, provided string) (suggestion string) {
+	distance := 0.0
+
+	for _, flag := range flags {
+		flagNames := flag.Names()
+		if !a.HideHelp {
+			flagNames = append(flagNames, HelpFlag.Names()...)
+		}
```
lynncyrin: ⭐⭐ _(seriously consider)_ expanding on my command above - I feel like this would be more maintainable if it pulled from a list of global flags?

suggestions.go
```diff
@@ -0,0 +1,75 @@
...
+			newDistance := matchr.JaroWinkler(name, provided, true)
+			if newDistance > distance {
+				distance = newDistance
+				suggestion = name
+			}
+		}
+	}
+
+	return fmt.Sprintf(didYouMeanTemplate, suggestion)
+}
```
lynncyrin: ⭐ _(nice to have)_ maybe consider adding [this documentation](https://github.com/antzucaro/matchr/blob/cbc221335f3c9d9eff15d1d9d5983ddf70b906b6/jarowinkler.go#L119-L125) to the docstring? I mostly because I found the `if newDistance > distance` comparsion confusing, at first.

Although this is possibly a case where we might be better suited with a command that says "see matcher for implementation details". Hmmm.

...really I feel like this entire code block belongs in matcher, eg. via a `matchr.FindClosestMatch` function.

Feel free to ignore all of this 🙂

suggestions.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	return suggestion
+}
+
+// suggestCommand takes a list of commands and a provided string to suggest a
+// command name
+func suggestCommand(commands []*Command, provided string) (suggestion string) {
+	distance := 0.0
+	for _, command := range commands {
```
lynncyrin: ⭐ _(nice to have)_ I'm not sure about this suggestion, but: would it be a good idea to consider suggesting sub commands too? I'm imagining an example CLI that has exactly these 3 commands

```
docker info
docker info containers
docker info images
```

and I input this

```
docker containers
```

I would be expecting to get a suggestion like

```
did you mean: docker info containers?
```

whereas I believe this code with give me

```
did you mean: docker info?
```

is that a correct assumption?

suggestions.go
saschagrunert: I'm not sure If I get this correctly, but we don't have a distinction between global and non-global flags any more in v2, right? (so I'd go for `VisibleFlags()`).

lynncyrin: Oh, good question! I think so? I haven't actually had the chance to use v2 meaningfully, so I'm not 100% sure 😅

saschagrunert: Yeah I think we have no real support for flags added to every command, so the user has to specify it manually which should work fine with the current implementation. :)

suggestions.go
saschagrunert: I will resolve this conversation as well to keep track of the open items. :)

suggestions.go
saschagrunert: Hm, I will add the documentation to the dostring. I hope this makes the implementation more clear.

suggestions.go
saschagrunert: Yes it is correct that it would only suggest `docker info` in that case, because can only suggest one command with the current implementation. I'm also not sure if we should add such a _traversal suggestion_ to the users. It might be helpful but it also might result in confusing them. 😁

lynncyrin: ^^ fair!

lynncyrin:
👀

saschagrunert:
I rebased on top of the latest master. Do you think we can move forward with this, @lynncyrin @AudriusButkevicius @asahasrabuddhe ?

AudriusButkevicius:
So my only comment is that "Suggest" is a poor name, we have UseXYZ everywhere, so the property could be more descriptive.

saschagrunert:
> So my only comment is that "Suggest" is a poor name, we have UseXYZ everywhere, so the property could be more descriptive.

Hm, how about `EnableFlagAndCommandSuggestions` ? :thinking:

AudriusButkevicius:
Sounds better. Perhaps others have suggestions.

lynncyrin:
`EnableSuggestions`? I'm fine with any of the options mentioned, though!

lynncyrin:
I should mention as an FYI that adding this new dependency will increase the binary size, which will make stuff like https://github.com/urfave/cli/issues/1055 a bit more apparent. I'm not gonna block on that though 🙏

lynncyrin:
I believe test failures are just because master needs to be merged into this branch?

lynncyrin:
👀

coilysiren (dismissed):

mfinley3:
Hey, just curious what ended up happening with this feature. Looks like stale bot closed it but seems like the work was good. Getting suggestions on invalid commands would still be a nice feature to have

hojongs:
Hey, Why don't you merge this feature?

PR #981, opened 2019-12-08
String slice issue v1

### LINKED ISSUE #790: StringSliceFlags has strange behavior

asahasrabuddhe:
## Motivation

Fixes #790 

The issue describes a bug where the default values of a `Slice` type flag persist even after the user explicitly sets the flag. Consider the following code:

```go
package main

import (
	"fmt"
	"log"
	"os"

	"github.com/urfave/cli"
)

func main() {
	app := cli.NewApp()
	app.Flags = []cli.Flag{
		cli.StringSliceFlag{
			Name: "Hostnames, n",
			Usage: "`HOSTNAME`s to listen to.",
			Value: &cli.StringSlice{"localhost"},
		},
	}
		if c.IsSet("Hostnames") {
			// do something
		}

		if c.IsSet("n") {
			// do something else
		}

		return nil
	}

	err := app.Run(os.Args)
	if err != nil {
		log.Fatal()
	}
}
```

Suppose, we execute this code using `--Hostnames a.b.c --Hostnames e.f.g`, we would expect the `c.StringSlice("Hostnames")` function to return a `StringSlice` with length equal to 2 and containing `a.b.c` and `e.f.g`. As of writing this PR, the call to `c.StringSlice("Hostnames")` returns a `StringSlice` with length equal to 3 and  containing `localhost`, `a.b.c` and `e.f.g`. This behavior is incorrect.

The same behavior can be replicated in the `IntSliceFlag` and `Int64SliceFlag` also. This PR aims to fix this inconsistency.

## Release Notes

**[BUG FIX]** String flag no longer persists the default value if the flag is explicitly initialized.

## Changes

- `flag_int64_slice.go`: Update `lookupInt64Slice` function to extract default values of the flag and then subsequently remove them.
- `flag_int_slice.go`: Update `lookupIntSlice` function to extract default values of the flag and then subsequently remove them.
- `flag_string_slice.go`: Update `lookupStringSlice` function to extract default values of the flag and then subsequently remove them.
- `flag_test.go`: Add unit tests.

## Testing

Unit tests are added to cover the following scenarios:

1. Flags are initialized with default values and are not explicitly initialized by the user.
2. Flags are initialized with default values and are explicitly initialized by the user.
3. Flags are initialized with default values and are explicitly initialized by the user with values the same as the default values.

### DIFF
--- a/flag_int64_slice.go
+++ b/flag_int64_slice.go
@@ -137,24 +137,28 @@ func (c *Context) GlobalInt64Slice(name string) []int64 {
 func lookupInt64Slice(name string, set *flag.FlagSet) []int64 {
 	f := set.Lookup(name)
 	if f != nil {
-		parsed, err := (f.Value.(*Int64Slice)).Value(), error(nil)
-		if err != nil {
-			return nil
-		}
-		// extract default value from the flag
-		var defaultVal []int64
-		for _, v := range strings.Split(f.DefValue, ",") {
-			iV, _ := strconv.ParseInt(v, 10, 64)
-			defaultVal = append(defaultVal, iV)
-		}
-		// if the current value is not equal to the default value
-		// remove the default values from the flag
-		if !isInt64SliceEqual(parsed, defaultVal) {
-			for _, v := range defaultVal {
-				parsed = removeFromInt64Slice(parsed, v)
+		if value, ok := f.Value.(*Int64Slice); ok {
+			// extract the slice from asserted value
+			parsed := value.Value()
+
+			// extract default value from the flag
+			var defaultVal []int64
+			for _, v := range strings.Split(f.DefValue, ",") {
+				int64Value, err := strconv.ParseInt(v, 10, 64)
+				if err != nil {
+					panic(err)
+				}
+				defaultVal = append(defaultVal, int64Value)
 			}
+			// if the current value is not equal to the default value
+			// remove the default values from the flag
+			if !isInt64SliceEqual(parsed, defaultVal) {
+				for _, v := range defaultVal {
+					parsed = removeFromInt64Slice(parsed, v)
+				}
+			}
+			return parsed
 		}
-		return parsed
 	}
 	return nil
 }
--- a/flag_int64_slice.go
+++ b/flag_int64_slice.go
@@ -159,17 +159,13 @@ func lookupInt64Slice(name string, set *flag.FlagSet) []int64 {
 	return nil
 }
 
-func removeFromInt64Slice(slice []int64, val int64) (newVal []int64) {
-	var count int
-	for _, v := range slice {
+func removeFromInt64Slice(slice []int64, val int64) []int64 {
+	for i, v := range slice {
 		if v == val {
-			newVal = slice[count+1:]
-			return
+			return append(slice[:i], slice[i+1:]...)
 		}
-		newVal = append(newVal, v)
-		count++
 	}
-	return
+	return slice
 }
 
 func isInt64SliceEqual(newValue, defaultValue []int64) bool {
--- a/flag_int64_slice.go
+++ b/flag_int64_slice.go
@@ -137,28 +137,33 @@ func (c *Context) GlobalInt64Slice(name string) []int64 {
 func lookupInt64Slice(name string, set *flag.FlagSet) []int64 {
 	f := set.Lookup(name)
 	if f != nil {
-		if value, ok := f.Value.(*Int64Slice); ok {
-			// extract the slice from asserted value
-			parsed := value.Value()
+		value, ok := f.Value.(*Int64Slice)
+		if !ok {
+			return nil
+		}
 
-			// extract default value from the flag
-			var defaultVal []int64
-			for _, v := range strings.Split(f.DefValue, ",") {
+		// extract the slice from asserted value
+		parsed := value.Value()
+
+		// extract default value from the flag
+		var defaultVal []int64
+		for _, v := range strings.Split(f.DefValue, ",") {
+			if v != "" {
 				int64Value, err := strconv.ParseInt(v, 10, 64)
 				if err != nil {
 					panic(err)
 				}
 				defaultVal = append(defaultVal, int64Value)
 			}
-			// if the current value is not equal to the default value
-			// remove the default values from the flag
-			if !isInt64SliceEqual(parsed, defaultVal) {
-				for _, v := range defaultVal {
-					parsed = removeFromInt64Slice(parsed, v)
-				}
-			}
-			return parsed
 		}
+		// if the current value is not equal to the default value
+		// remove the default values from the flag
+		if !isInt64SliceEqual(parsed, defaultVal) {
+			for _, v := range defaultVal {
+				parsed = removeFromInt64Slice(parsed, v)
+			}
+		}
+		return parsed
 	}
 	return nil
 }
--- a/flag_int64_slice.go
+++ b/flag_int64_slice.go
@@ -22,7 +22,12 @@ func (f *Int64Slice) Set(value string) error {
 
 // String returns a readable representation of this value (for usage defaults)
 func (f *Int64Slice) String() string {
-	return fmt.Sprintf("%#v", *f)
+	slice := make([]string, len(*f))
+	for i, v := range *f {
+		slice[i] = strconv.FormatInt(v, 10)
+	}
+
+	return strings.Join(slice, ",")
 }
 
 // Value returns the slice of ints set by this flag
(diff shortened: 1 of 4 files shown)


asahasrabuddhe:
bump

flag_int64_slice.go
```diff
@@ -22,7 +22,12 @@ func (f *Int64Slice) Set(value string) error {
```
rliebz: Commenting at the top because there's nowhere else to put it—

It's not called out in the original issue, but there is also a `float64` slice that probably has the same issue here.

flag_string_slice.go
```diff
@@ -132,7 +132,51 @@ func lookupStringSlice(name string, set *flag.FlagSet) []string {
...
+	}
+
+	for i, v := range newValue {
+		if v != defaultValue[i] {
+			return false
+		}
+	}
+
+	return true
+}
```
rliebz: Request: Make sure to run `gofmt` on these files. 

Long term, this is something we should enforce in the pipeline as well.

flag_int64_slice.go
```diff
@@ -135,7 +141,52 @@ func lookupInt64Slice(name string, set *flag.FlagSet) []int64 {
...
+			}
+		}
 		return parsed
 	}
 	return nil
 }
+
+func removeFromInt64Slice(slice []int64, val int64) (newVal []int64) {
+	var count int
+	for _, v := range slice {
```
rliebz: We should be able to drop the `count` variable by using the slice index directly:

```go
for i, v := range slice {
```

flag_int64_slice.go
```diff
@@ -135,7 +141,52 @@ func lookupInt64Slice(name string, set *flag.FlagSet) []int64 {
...
+			for _, v := range defaultVal {
+				parsed = removeFromInt64Slice(parsed, v)
+			}
+		}
 		return parsed
 	}
 	return nil
 }
+
+func removeFromInt64Slice(slice []int64, val int64) (newVal []int64) {
```
rliebz: Nit: We shouldn't be using named returns unless the function signature is ambiguous, or we absolutely have to

There are caveats with the scope of a named variable that aren't always intuitive, and naked returns pretty much always obfuscate what code should be doing. It also encourages us to mutate an existing variable rather than returning directly.

I'd probably write the function like this:
```go
func removeFromInt64Slice(slice []int64, val int64) ([]int64) {
	for i, v := range slice {
		if v == val {
			return slice[count+1:]
		}
	}
	return slice
}
```

That said, there's an issue here, which I'm not sure if is intentional or not. This function removes all items up to and including the first occurrence of `val`, but the name implies that we would just be removing `val`. I don't think it manifests because of how this function is called, but we should either fix the implementation or rename the function to avoid future misuse of this helper.

flag_string_slice.go
lynncyrin: > Long term, this is something we should enforce in the pipeline as well.

=> https://github.com/urfave/cli/issues/993

asahasrabuddhe: This is done

asahasrabuddhe:
> I'm pausing review here because I think this PR might not be approaching the fix in the ideal way. Right now, we'd be accepting that the new values are getting appended incorrectly, and then removing the default based on their string representation after the fact, rather than preventing them from getting incorrectly joined in the first place.
> 
> The bigger issue I see is that there's already code to handle removing the defaults—this is in master:
> 
> https://github.com/urfave/cli/blob/cae7b0c5e15e1e19e0f11ab37195be7655afa41c/flag_string_slice.go#L22-L26
> 
> This means on the first setting of a slice flag, we should in theory be zeroing out the value. I'm not sure why that's not working (or no longer working, if it used to), but I think the better fix would be to address that issue at the root rather than patch it downstream on flag lookup.

I agree with your comment but unfortunately, this is right now not possible in the `v1` branch because the `StringSliceFlag` is simply an aliased `string` slice. To make it work like you say, I would need to rewrite the Slice Flags to make them into structs and then include the has been set.

flag_int64_slice.go
asahasrabuddhe: Float64 Slice is not on the v1

flag_int64_slice.go
asahasrabuddhe: Fixed

flag_int64_slice.go
asahasrabuddhe: I have updated the function to behave like it's named

```go
func removeFromInt64Slice(slice []int64, val int64) []int64 {
	for i, v := range slice {
		if v == val {
			return append(slice[:i], slice[i+1:]...)
		}
	}
	return slice
}
```

See it in action: https://play.golang.org/p/ZjJxp38tKNQ

flag_int_slice.go
```diff
@@ -136,7 +141,52 @@ func lookupIntSlice(name string, set *flag.FlagSet) []int {
...
+			for _, v := range defaultVal {
+				parsed = removeFromIntSlice(parsed, v)
+			}
+		}
 		return parsed
 	}
 	return nil
 }
+
+func removeFromIntSlice(slice []int, val int) (newVal []int) {
```
rliebz: Looks like this one still has the old definition—can we update this to match the other remove functions?

flag_int64_slice.go
```diff
@@ -135,7 +141,48 @@ func lookupInt64Slice(name string, set *flag.FlagSet) []int64 {
 		if err != nil {
 			return nil
 		}
+		// extract default value from the flag
+		var defaultVal []int64
+		for _, v := range strings.Split(f.DefValue, ",") {
+			iV, _ := strconv.ParseInt(v, 10, 64)
```
rliebz: Request: Even if we're confident that this parse should always work, I'd rather panic explicitly than silently ignore the error. Panics are for developer errors.

The reason for this is that if we or the user somehow make a mistake that allows a non-int to end up here, the current code will silently set this value to `0` and carry on into the territory of behavior we haven't planned for. If we panic instead, we get a clear message back to the end user that something is actually catastrophically wrong, and we stop the program.

```go
iV, err := strconv.ParseInt(v, 10, 64)
if err != nil {
	panic(err)
}
```

flag_int_slice.go
```diff
@@ -136,7 +141,52 @@ func lookupIntSlice(name string, set *flag.FlagSet) []int {
 		if err != nil {
 			return nil
 		}
+		// extract default value from the flag
+		var defaultVal []int
+		for _, v := range strings.Split(f.DefValue, ",") {
+			iV, _ := strconv.Atoi(v)
```
rliebz: Same note as above—we shouldn't ignore errors.

flag_test.go
```diff
@@ -1330,3 +1330,96 @@ func TestFlagFromFile(t *testing.T) {
 		}
 	}
 }
+
+func TestSliceFlag_WithDefaults(t *testing.T) {
```
rliebz: 🎉

flag_int_slice.go
asahasrabuddhe: Done

flag_int64_slice.go
asahasrabuddhe: I agree. I just noticed the other PR fixing this in the master and I would love to get it done right here.

flag_int64_slice.go
```diff
@@ -131,11 +137,56 @@ func (c *Context) GlobalInt64Slice(name string) []int64 {
 func lookupInt64Slice(name string, set *flag.FlagSet) []int64 {
 	f := set.Lookup(name)
 	if f != nil {
-		parsed, err := (f.Value.(*Int64Slice)).Value(), error(nil)
-		if err != nil {
-			return nil
+		if value, ok := f.Value.(*Int64Slice); ok {
```
rliebz: Nit/optional: Generally it's better to keep the error scenario in the conditional. We get to remove a level of indentation and reduce the overall cognitive overhead of the function:

```go
value, ok := f.Value.(*Int64Slice)
if !ok {
	return nil
}
// regular logic goes here ...
```

rliebz: Same optional note for the other two types.

flag_int64_slice.go
```diff
@@ -131,11 +137,61 @@ func (c *Context) GlobalInt64Slice(name string) []int64 {
...
+		if v == val {
+			return append(slice[:i], slice[i+1:]...)
+		}
+	}
+	return slice
+}
+
+func isInt64SliceEqual(newValue, defaultValue []int64) bool {
+	// If one is nil, the other must also be nil.
+	if (newValue == nil) != (defaultValue == nil) {
```
saschagrunert: This is the first time I every see an expression like this in golang. :eyes:

lynncyrin: Woooooooooah, love it!

PR #985, opened 2019-12-08
Fix codecov reports in GitHub actions

saschagrunert:
## Motivation

Coverage reports seems broken with the switch to GitHub actions. This PR should address this issue.

## Changes

We now use the official codecov GitHub action to report the coverage.

PR #986, opened 2019-12-09
Fix codecov reports in GitHub actions

saschagrunert:
## Motivation

Coverage reports seems broken with the switch to GitHub actions. This PR should address this issue.

Follup-up of https://github.com/urfave/cli/pull/985
## Changes

We now use the official codecov GitHub action to report the coverage.

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -61,5 +61,5 @@ jobs:
       - name: Send Coverage Report
         if: success() && matrix.go == 1.13 && matrix.os == 'ubuntu-latest'
         env:
-          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+          CODECOV_TOKEN: fa7ff1caed3c43c68ae994495a6469d8
         run: bash <(curl -s https://codecov.io/bash)
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -58,8 +58,9 @@ jobs:
           go run build.go gfmrun docs/v2/manual.md
           go run build.go toc docs/v2/manual.md
 
-      - name: Send Coverage Report
+      - name: Upload coverage to Codecov
         if: success() && matrix.go == 1.13 && matrix.os == 'ubuntu-latest'
-        env:
-          CODECOV_TOKEN: fa7ff1caed3c43c68ae994495a6469d8
-        run: bash <(curl -s https://codecov.io/bash)
+        uses: codecov/codecov-action@v1
+        with:
+          token: 0a8cc73b-bb7c-480b-8626-38a461643761
+          fail_ci_if_error: true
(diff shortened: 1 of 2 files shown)


saschagrunert:
It still does not seem to work :thinking:

saschagrunert:
I think the major issue is that we do not pass the secrets to PRs of forks:

asahasrabuddhe:
@saschagrunert create a separate secret in your fork. It will work

saschagrunert:
> @saschagrunert create a separate secret in your fork. It will work

Yes, but I guess we want to provide a working solution for every user out of the box. I see two ways to achieve this right now:
- Expose the token to the public and set it inside the script (should not be that harmful if the only drawback is that abusers can write the projects coverage)
- Switch to a different CI system for the code coverage

lynncyrin:
> Expose the token to the public and set it inside the script (should not be that harmful if the only drawback is that abusers can write the projects coverage)

I'm fine with this option ☝️ 

I can create create some pressure on codecov to remove the need to do any of this stuff.

AudriusButkevicius:
Surely CI should be done outside of the control of the committer. We should ask github/codecov what to do about this.

lynncyrin:
> Surely CI should be done outside of the control of the committer. We should ask github/codecov what to do about this.

@AudriusButkevicius can you do that?

AudriusButkevicius:
It seems we're stuffed: https://github.community/t5/GitHub-Actions/Make-secrets-available-to-builds-of-forks/td-p/30678

Yet it's not very clear.

I'd be interesting to see if a workflow modified in the fork gets executed in our environment (with out env vars etc). I suspect we take the PRs code and run our workflows against it, in which case we just need to have a single action that includes the token and calls no code from the PR to be safe.

Yet it sounds too easy.

saschagrunert:
Do we still want to copy the token inside as a current workaround?

lynncyrin:
> Do we still want to copy the token inside as a current workaround?

I'm still 👍 for this

saschagrunert:
> > Do we still want to copy the token inside as a current workaround?
> 
> I'm still  for this

Good :+1:, feel free to provide me the token somehow that I can push it into the branch.

lynncyrin:
👀 I pushed it to the branch

lynncyrin:
I just checked the codecov settings and there's a different token in there 👀

lynncyrin:
@saschagrunert I just sent it to your email address, so that it's not in the github comments 🙂

saschagrunert:
> I just checked the codecov settings and there's a different token in there

Ah, okay can you push the new token on top of the branch? We can also remove the `name` I guess.

saschagrunert:
Thanks, got the token and pushed it in the branch. :+1:

saschagrunert:
Re-pushed to trigger appveyor again.

saschagrunert:
Ready for review. :heavy_check_mark:

PR #987, opened 2019-12-09
timestamp flag

### LINKED ISSUE #696: TimeFlag
I'd like to add a `cli.TimeFlag`. Is this something that would be accepted as a pull request?

Also, I would like to be able to specify the format of the time in the flag.

``` go
cli.TimeFlag{
	Name:  "date",
	Usage: "date",
	Format: "2016-01-02",
}
```

However, I can't specify an extra field `flag-types.json`. Should I create custom flag type in flag.go or should I add the possibility of adding extra fields in `generate-flag-types`?

drov0:
Fixes #696 

## Changes

This is a simple flag to add the support for timestamps. At first this flag was using protobuf timestamps but I ended up editing it since I figured you would be more interested in something more general. But if you guys are interested I would love to make a protobuf timestamp as well. (saves a few lines on my app code)

## Reviewer Guidelines

I'm not sure if this code works with env variables. I tried to take inspiration from int64_slice and string_slice since those were the only flag that didn't use the default set from golang's `Flag`

I am open to any changes you see fit.


flag_timestamp.go
```diff
@@ -0,0 +1,130 @@
...
+)
+
+
+// timestamp wrap to satisfy golang's flag interface.
+type timestampWrap struct {
+	timestamp *time.Time
+	hasBeenSet bool
+	layout 	   string
+}
+
```
asahasrabuddhe: Add a constructor here

```go
func NewTimestampWrap(timestamp time.Time) *Timestamp {
	return &Timestamp{timestamp: &timestamp}
}
```

flag_timestamp.go
```diff
@@ -0,0 +1,130 @@
...
+import (
+	"flag"
+	"fmt"
+	"time"
+)
+
+
+// timestamp wrap to satisfy golang's flag interface.
+type timestampWrap struct {
```
asahasrabuddhe:
```suggestion
type Timestamp struct {
```

All the flags expose the wrapped type to allow setting a zero value default.

flag_timestamp.go
```diff
@@ -0,0 +1,130 @@
...
+type TimestampFlag struct {
+	Name        string
+	Aliases     []string
+	Usage       string
+	EnvVars     []string
+	FilePath    string
+	Required    bool
+	Hidden      bool
+	Layout 	    string
+	Value       timestampWrap
```
asahasrabuddhe:
```suggestion
	Value       *timestampWrap
```

This should be a pointer to the wrapped struct

flag_timestamp.go
```diff
@@ -0,0 +1,130 @@
...
+// GetValue returns the flag value
+func (f *TimestampFlag) GetValue() *time.Time {
+	return f.Value.timestamp
+}
+
+// Apply populates the flag given the flag set and environment
+func (f *TimestampFlag) Apply(set *flag.FlagSet) error {
+	for _, name := range f.Names() {
+		f.Value.SetLayout(f.Layout)
+		set.Var(&f.Value, name, f.Usage)
```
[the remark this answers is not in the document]
asahasrabuddhe: Please refer to the lines 123-133 in the `flag_int64_slice.go` and try to port that logic to enable reading values from env / file for the timestamp flag

flag_timestamp.go
```diff
@@ -0,0 +1,130 @@
...
+	return true
+}
+
+// GetUsage returns the usage string for the flag
+func (f *TimestampFlag) GetUsage() string {
+	return f.Usage
+}
+
+// GetValue returns the flag value
+func (f *TimestampFlag) GetValue() *time.Time {
```
asahasrabuddhe:
```suggestion
func (f *TimestampFlag) GetValue() string {
```

The timestamp flag currently does not implement the Docs generation flag.

flag_timestamp.go
```diff
@@ -0,0 +1,130 @@
...
+	return f.Value.timestamp
```
asahasrabuddhe:
```suggestion
	if f.Value != nil {
		return f.Value.String()
	}
	return ""
```

flag_timestamp.go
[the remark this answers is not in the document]
drov0: Okay so I end up with something like this 

```
func (f *TimestampFlag) Apply(set *flag.FlagSet) error {
	if val, ok := flagFromEnvOrFile(f.EnvVars, f.FilePath); ok {
		f.Value = &Timestamp{}

		if err := f.Value.Set(val); err != nil {
			return fmt.Errorf("could not parse %q as timestamp value for flag %s: %s", val, f.Name, err)
		}

		f.HasBeenSet = true
	}

		f.Value.SetLayout(f.Layout)
		set.Var(f.Value, name, f.Usage)
	}
	return nil
}
```

But looking at the doc I don't see how I can set env variables for both the layout and the value.

drov0:
@asahasrabuddhe

  > update the `manual.md` to reflect the new features.

Which part of the manual.md ? The full example ? There isn't much details on specific flags on the manual, It seems like most of the doc is auto generated through GoDoc

asahasrabuddhe:
> @asahasrabuddhe
> 
> > update the `manual.md` to reflect the new features.
> 
> Which part of the manual.md ? The full example ? There isn't much details on specific flags on the manual, It seems like most of the doc is auto generated through GoDoc

What I meant was that you add an example of using the timestamp flag as I think this is a unique flag, and users would appreciate an example on how to use it in their application.

asahasrabuddhe:
@drov0 If I initialize a timestamp flag without any default layout value, what layout is considered for the flag? Do you think having a default out of the box value for the layout would help?

asahasrabuddhe:
@drov0 If I initialize a timestamp flag without any default layout value, what layout is considered for the flag? Do you think having a default out of the box value for the layout would help?

drov0:
@asahasrabuddhe, okay for the manual, will add on monday. As for the layout It won't work if you don't set it, the layout is mandatory. I considered having a default value but since there are so many different layouts, I figured that people would almost always use a custom layout anyways. Kind of like `time.Parse` doesn't have a default layout.

asahasrabuddhe:
> @asahasrabuddhe, okay for the manual, will add on monday. As for the layout It won't work if you don't set it, the layout is mandatory. I considered having a default value but since there are so many different layouts, I figured that people would almost always use a custom layout anyways. Kind of like `time.Parse` doesn't have a default layout.

Fair enough. I think once we have an example, we are good to go 👍

drov0:
@asahasrabuddhe I added examples at the end in the full example, is that ok or would you rather that I add a complete new section for it ?

drov0:
Nevermind I did put it in its own block in the end. But tests are failing, this whole "the doc is actually compiling itself" is super confusing to me. I know my code snippet is correct yet it's failing for some reason, I cornered it down to the fact that it's somehow throwing an error when defining timestampFlag. I guess it's not pulling the right version ?

asahasrabuddhe:
> Nevermind I did put it in its own block in the end. But tests are failing, this whole "the doc is actually compiling itself" is super confusing to me. I know my code snippet is correct yet it's failing for some reason, I cornered it down to the fact that it's somehow throwing an error when defining timestampFlag. I guess it's not pulling the right version ?

You are right. The thing is that we have a utility that tries to execute the snippets in the documentation to verify their correctness. Now the tool is using the v2.0.0 of the library to verify your addition which is not yet released. Hence the failures.

I think too that the sample is correct and I recommend that this PR be merged.

drov0:
Awesome thanks !

lynncyrin:
Also, if I'm correct in my understanding that CI is failing because of the docs change, you can unblock this PR via:

- creating a new PR with just the docs changes
- removing the docs changes from this PR

drov0:
@lynncyrin done :) https://github.com/urfave/cli/pull/997

bkeyoumarsi:
@lynncyrin any chance we can get this into a minor release (2.0.1 perhaps)? Would love to use the timestamp flag in my application with the proper go.mod routine.

lynncyrin:
I'll draft a release soon 👍

lynncyrin:
=> https://github.com/urfave/cli/pull/1013

lynncyrin:
It's released 🚢 => https://github.com/urfave/cli/releases/tag/v2.1.0

PR #994, opened 2019-12-18
Run Github Actions on the master branch

lynncyrin:
## Motivation

Master merges are currently only running appveyor builds 
image

Whereas we would expect master merges to run build actions builds, like in this screnshot 
image

This is likely to cause a variety of problems, but noticed it via this error message from codecov on this page https://codecov.io/gh/urfave/cli/compare/cae7b0c5e15e1e19e0f11ab37195be7655afa41c...2cb8524de32183d2eed9d355e3007417bd6708ae 
image

So this PR changed the Github Actions config to run tests on master

## Documentation

https://help.github.com/en/actions/automating-your-workflow-with-github-actions/configuring-a-workflow#filtering-for-specific-branches-tags-and-paths

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -1,6 +1,9 @@
 name: Run Tests
 
 on:
+  push:
+    branches:
+      - master  
   pull_request:
     branches:
       - master


.github/workflows/cli.yml
```diff
@@ -1,6 +1,9 @@
 name: Run Tests
 
 on:
+  push:
+    branches:
+      - master
```
rliebz: Are we interested in running on merge to `v1` as well as `master`? Not a requirement for merge, just curious.

lynncyrin: Good question 🤔 I think so, I'll create another PR for that 👍

PR #997, opened 2019-12-18
Add timestampFlag docs

drov0:
Reason for this PR is that it makes the CI fail until #987 is merged

### DIFF
--- a/docs/v2/manual.md
+++ b/docs/v2/manual.md
@@ -1150,6 +1150,45 @@ The default version flag (`-v/--version`) is defined as `cli.VersionFlag`, which
 is checked by the cli internals in order to print the `App.Version` via
 `cli.VersionPrinter` and break execution.
 
+
+### Timestamp Flag
+
+Using the timestamp flag is simple, You can look at time.Parse to get layout examples : https://golang.org/pkg/time/#example_Parse 
+
+<!-- {
+  "args": ["&#45;&#45;meeting", "2019-08-12T15:04:05"],
+  "output": "\"2019-08-12 15:04:05 +0000 UTC\""
+} -->
+``` go
+package main
+
+import (
+  "fmt"
+  "log"
+  "os"
+
+  "github.com/urfave/cli/v2"
+)
+
+func main() {
+  app := &cli.App{
+    Flags: []cli.Flag {
+      &cli.TimestampFlag{Name: "meeting", Layout: "2006-01-02T15:04:05"},
+    },
+    Action: func(c *cli.Context) error {
+      fmt.Printf("%#v",c.Timestamp("meeting").String())
+      return nil
+    },
+  }
+
+  err := app.Run(os.Args)
+  if err != nil {
+    log.Fatal(err)
+  }
+}
+```
+
+
 #### Customization
 
 The default flag may be customized to something other than `-v/--version` by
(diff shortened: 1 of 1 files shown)


docs/v2/manual.md
```diff
@@ -1150,6 +1150,45 @@ The default version flag (`-v/--version`) is defined as `cli.VersionFlag`, which
 is checked by the cli internals in order to print the `App.Version` via
 `cli.VersionPrinter` and break execution.
 
+
+### Timestamp Flag
+
+Using the timestamp flag is simple, You can look at time.Parse to get layout examples : https://golang.org/pkg/time/#example_Parse 
+
+<!-- {
+  "args": ["&#45;&#45;meeting", "2019-08-12T15:04:05"],
```
lynncyrin: ⭐⭐⭐ _(must change)_ It took me a few seconds to figure out what I was reading here, can you add an example invoke to these docs? Like this:

```
myapp --meeting "2019-08-12T15:04:05"
```

asahasrabuddhe:
This won't pass until the changes that have been merged are tagged. When `gfmrun` tries to execute the code, it pulls the v2.0.0 which currently does not have this change. I am thinking of a way to fix this once and for all because everytime we have a new feature, this problem is going to recur.

Now we have two ways to move forward, make a release with a feature without having the documentation supporting the feature or, to just merge this as it's only documentation and we all know the exact reason for the CI failure.

lynncyrin:
>  I am thinking of a way to fix this once and for all because everytime we have a new feature, this problem is going to recur.

From a functionality point of view, this is what I would like => https://github.com/urfave/cli/issues/1006

> make a release with a feature without having the documentation supporting the feature

This is my preference for the time being 👍

asahasrabuddhe:
Finally this works! Phew

docs/v2/manual.md
```diff
@@ -1273,6 +1274,49 @@ func main() {
 }
 ```
 
+### Timestamp Flag
+
+Using the timestamp flag is simple, You can look at time.Parse to get layout examples : https://golang.org/pkg/time/#example_Parse
```
saschagrunert:
```suggestion
Using the timestamp flag is simple. Please refer to [`time.Parse`](https://golang.org/pkg/time/#example_Parse) to get possible formats.
```

docs/v2/manual.md
```diff
@@ -1273,6 +1274,49 @@ func main() {
...
+  }
+
+  err := app.Run(os.Args)
+  if err != nil {
+    log.Fatal(err)
+  }
+}
+```
+
+In this example the flag could be used like so :
```
saschagrunert:
```suggestion
In this example the flag could be used like this:
```

docs/v2/manual.md
```diff
@@ -1273,6 +1274,49 @@ func main() {
...
+    log.Fatal(err)
+  }
+}
+```
+
+In this example the flag could be used like so : 
+
+`myapp --meeting 2019-08-12T15:04:05`
+
+side note : quotes may be necessary around the date depending on your layout (if you have spaces for instance)
```
saschagrunert:
```suggestion
Side note: quotes may be necessary around the date depending on your layout (if you have spaces for instance)
```

PR #999, opened 2019-12-18
Use VisibleFlags during doc generation

subpop:
## Motivation

Hidden commands are excluded during doc generation but hidden flags are not. This results in generated documentation (man pages and markdown) including hidden flags in the output. Using `VisibleFlags()` instead of the `Flags` slice filters out hidden flags before preparing them for document output.

## Testing

I added a hidden flag to the testApp. Test output should remain unchanged, since hidden flags should not be included in documentation output.


saschagrunert:
Thanks for the fix, @subpop :)

subpop:
Ugh, sorry. Rebased this change onto master and it invalidated the approvals. :(

PR #1001, opened 2019-12-19
Update Github Actions to always run tests

lynncyrin:
## Motivation

Master is failing because the current Github Actions config only runs tests inside of PRs.


lynncyrin:
The test failure is unrelated, as far as I can tell. I'm looking into it 👀

PR #1023, opened 2019-12-27
Issue template copy updates

lynncyrin:
## Motivation

I used our issue templates to write up https://github.com/timonwong/vscode-shellcheck/issues/82, and noticed a few very minor ways they could be improved.

### DIFF
--- a/.github/ISSUE_TEMPLATE/v2-bug-report.md
+++ b/.github/ISSUE_TEMPLATE/v2-bug-report.md
@@ -28,13 +28,13 @@ _**( Put the version of urfave/cli that you are using here )**_
 
 A clear and concise description of what the bug is.
 
-## To reproduce
-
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/v2-bug-report.md
```diff
@@ -32,9 +32,9 @@ A clear and concise description of what the bug is.
 
 Describe the steps or code required to reproduce the behavior
 
-## Expected behavior
```
rliebz: Based on the `v1` template, I'm guessing we want both "Expected" and "Observed" here as well. Looks like "Expected" is being deleted.

lynncyrin: Ah, this was a copy paste issue on my part. Will fix, nice catch 🦅

PR #1027, opened 2019-12-30
Fix import order (goimports)

vkd:
## Motivation

https://github.com/golang/go/wiki/CodeReviewComments#imports

## Changes

Only the import order has changed.

## Reviewer Guidelines

This changes has produced by the [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports) tool.

### DIFF
--- a/altsrc/flag_generated.go
+++ b/altsrc/flag_generated.go
@@ -4,6 +4,7 @@ package altsrc
 
 import (
 	"flag"
+
 	"github.com/urfave/cli/v2"
 )
 
--- a/altsrc/flag_test.go
+++ b/altsrc/flag_test.go
@@ -3,12 +3,13 @@ package altsrc
 import (
 	"flag"
 	"fmt"
-	"github.com/urfave/cli/v2"
 	"os"
 	"runtime"
 	"strings"
 	"testing"
 	"time"
+
+	"github.com/urfave/cli/v2"
 )
 
 type testApplyInputSource struct {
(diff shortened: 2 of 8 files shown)


vkd:
@lynncyrin No, I don't. And I will try to investigate it.

rliebz:
There's two ways that I'm aware of. 

One would be to just run the goimports command (`goimports -w -local github.com/urfave/cli .`) and validate with `git` that no files have changed. 

The other would be to bring in [golangci-lint](https://github.com/golangci/golangci-lint), which is a linter that combines a bunch of smaller linters like `govet`, `golint`, `staticcheck`, and others. That's what I typically use in projects. 

The config to include this particular import check would be:

```toml
[linters]
enable = [
  "goimports",
]

[linters-settings.goimports]
local-prefixes = "github.com/urfave/cli"
```

As a litmus test, the default configuration for `golangci-lint` finds 4 ineffectual assignments, 5 instances of unused code that can be removed, one error that is not checked (in a test file), and one `nil` check that can be omitted. With [a super opinionated configuration that I've used on projects before](https://github.com/rliebz/tusk/blob/master/.golangci.toml), it flags 116 issues, the majority of which are style-related (e.g., cyclomatic complexity, line length, godoc comments, etc.).  We'd have the freedom to settle anywhere on that spectrum.

If there's interest, I can set this up for us. I also realize this is a comment on a merged PR, so I can open an issue as well if we want to continue discussion.

PR #1030, opened 2019-12-30
CI - add goimports check

vkd:
From the discussion on #1027

## Motivation

Auto check the code formatting in every PR.

## Reviewer Guidelines

`goimports -l ./` command returns a list of unformatted files.

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -45,6 +45,11 @@ jobs:
           curl -L -o $GOPATH/bin/gfmrun "https://github.com/urfave/gfmrun/releases/download/v1.2.14/gfmrun-$(go env GOOS)-amd64-v1.2.14"
           chmod +x $GOPATH/bin/gfmrun
           npm install markdown-toc
+          GO111MODULE=off go get golang.org/x/tools/cmd/goimports
+
+      - name: Check source code
+        run: |
+          make goimports-check
 
       - name: Run Tests
         run: |
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,12 @@
+GOFMT=goimports
+
+goimports:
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,12 @@
...
+goimports:
+	${GOFMT} -w ./
+
+goimports-check:
+	$(eval diff_files = $(shell ${GOFMT} -l ./))
+	@if [ -n "${diff_files}" ]; then \
+		echo "Please run 'make goimports' to fix the format errors in the files:"; \
+		echo "\t${diff_files}"; \
+		exit 1; \
+	fi;
```
saschagrunert: I recommend to check this via golangci-lint, like here:
https://github.com/cri-o/cri-o/blob/2dc77fc03c522237a353ca8dfbd9a69c1319bed6/Makefile#L236-L241

.github/workflows/cli.yml
```diff
@@ -45,6 +45,11 @@ jobs:
           curl -L -o $GOPATH/bin/gfmrun "https://github.com/urfave/gfmrun/releases/download/v1.2.14/gfmrun-$(go env GOOS)-amd64-v1.2.14"
           chmod +x $GOPATH/bin/gfmrun
           npm install markdown-toc
+          GO111MODULE=off go get golang.org/x/tools/cmd/goimports
+
+      - name: Check source code
+        run: |
+          make goimports-check
```
lynncyrin: We use `build.go` rather than a makefile, so please move this to `build.go` 🙂

vkd:
I see we want to check the `goimports` changes through the `golangci-lint` linter and use the `build.go` file instead of the `Makefile`.
Since this PR doesn't correspond to the assumptions of the project I will close it.

PR #1035, opened 2020-01-10
Use a more k8s/release friendly PR template.

j0n3lson:
## What type of PR is this?  
  /kind documentation
  
## What this PR does / why we need it:

Modifying the PR template will will enable easier parsing kubernetes release-notes tool (tracked in #965). The tool makes some assumptions about PR bodies so limiting drift between those assumptions and the PR template should make release note generation more reliable.

## Which issue(s) this PR fixes:

Partially addresses #965 

## Special notes for your reviewer:

Tried to keep the spirit of the original template but I've made some sections "Required" if they are important to release notes generation. Also I assume the list of "allowed" labels has been updated (or will be) based on the labels in this template.

PR #1037, opened 2020-01-10
Check if master is failing

lynncyrin:
Noticed via https://github.com/urfave/cli/pull/1035 and https://github.com/urfave/cli/pull/1036

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -28,6 +28,11 @@ jobs:
         with:
           go-version: ${{ matrix.go }}
 
+      - name: Use Node.js 10.x
+          uses: actions/setup-node@v1
+          with:
+            node-version: 10.x
+
       - name: Set GOPATH and PATH
         run: |
           echo "::set-env name=GOPATH::$(dirname $GITHUB_WORKSPACE)"
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -28,10 +28,10 @@ jobs:
         with:
           go-version: ${{ matrix.go }}
 
-      - name: Use Node.js 10.x
+      - name: Use Node.js 12.x
         uses: actions/setup-node@v1
         with:
-          node-version: 10.x
+          node-version: 12.x
 
       - name: Set GOPATH and PATH
         run: |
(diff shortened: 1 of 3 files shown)


j0n3lson:
For my own knowledge. What does that rest cover? What is it doing (supposed to do)?

lynncyrin:
> What does that **rest** cover?

What does "rest" mean here? OH! Is this supposed to say `test`?

j0n3lson:
Yes the *test* I looked at the test case and saw a markdown file and was
confused hehehe. Just was wondering what kind of test it was 😅

lynncyrin:
It processes the table of contents, via => https://github.com/jonschlinkert/markdown-toc

asahasrabuddhe:
@lynncyrin This is working now

.github/workflows/cli.yml
```diff
@@ -28,6 +28,11 @@ jobs:
         with:
           go-version: ${{ matrix.go }}
 
+      - name: Use Node.js 10.x
```
rliebz: Question: Any reason to prefer `10.x`? Latest LTS version is `12`. 

https://nodejs.org/en/about/releases/

.github/workflows/cli.yml
```diff
@@ -44,7 +49,7 @@ jobs:
           mkdir -p $GOPATH/bin
           curl -L -o $GOPATH/bin/gfmrun "https://github.com/urfave/gfmrun/releases/download/v1.2.14/gfmrun-$(go env GOOS)-amd64-v1.2.14"
           chmod +x $GOPATH/bin/gfmrun
-          npm install markdown-toc
+          npm install -g markdown-toc
```
rliebz: Request: We should pin this dependency version to prevent future breakages. 

```suggestion
          npm install -g markdown-toc@1.2.0
```

README.md
```diff
@@ -1,6 +1,8 @@
 cli
 ===
 
+checking to see if master is failing
```
rliebz: Request: This should be removed.

.github/workflows/cli.yml
asahasrabuddhe: I am using 10.x on my system where it was working so I used that. Let me try to change it to 12

README.md
asahasrabuddhe: Agreed

asahasrabuddhe:
@rliebz I have applied your suggestions

PR #1049, opened 2020-01-22
Fix possible panics in slice type assertion

saschagrunert:
## What type of PR is this?

- [x] bug
- [ ] cleanup  
- [ ] documentation
- [ ] feature

Before accessing the slices `.Value()` we should type check that it is
that actual type, otherwise we will end-up in a panic.

None

None

## Testing

Reproducer:
```go
package main

import (
	"os"

	"github.com/urfave/cli/v2"
)

func main() {
	app := cli.NewApp()
	app.Flags = []cli.Flag{&cli.StringFlag{Name: "test"}}
		ctx.StringSlice("test")
		return nil
	}
		os.Exit(1)
	}
}
```

## Release Notes

```release-note
- Fixed possible runtime panic in slice parsing
```

### DIFF
--- a/flag_float64_slice.go
+++ b/flag_float64_slice.go
@@ -155,11 +155,9 @@ func (c *Context) Float64Slice(name string) []float64 {
 func lookupFloat64Slice(name string, set *flag.FlagSet) []float64 {
 	f := set.Lookup(name)
 	if f != nil {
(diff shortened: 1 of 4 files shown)


asahasrabuddhe:
Looks good to me :+1:

Issue #833, opened 2019-08-03
Release 3.x meta/discussion


coilysiren:
This is a meta-issue meant to track the work in _other issues_ targeting the [Release 3.x milestone](https://github.com/urfave/cli/milestone/5).

<details>
  <summary>benevolent takeover notes</summary>

Hiii I'm working on getting many maintenance tasks addressed including
taking over this issue from @coilysiren 🙇🏼 who is busy with other life adventures ❤️ ~ @meatballhat

</details>

~I know 2.0 isn't out yet 🙂 (see https://github.com/urfave/cli/issues/826) but~ I wanted to create a tracking issue for the 3.x series.

Here's some breaking/difficult changes we want to consider for V3:

- [x] #1498 
- [x] #1489 
- [x] Replace code generation with generics where possible/maintainable (#1512 )
- [x] ~#1058~ 
- [x] #1055 
- [x] #1695 
- [x] ~Improve testability of `*cli.Context` (via https://github.com/urfave/cli/issues/833#issuecomment-574171962)~ (conflicts with #1587)
- [x] remove `EnableBashCompletion` in favor of `CompletionShells`
- [x] ~consider a different release method (eg. not just merging to `master`) via https://github.com/urfave/cli/issues/952#issuecomment-561062742~ (this was changed in v2)
- [x] ~make flags "global" by default, to some degree~ (this was added into v2)
- [x] #753 
- [x] #1071
- [x] ~Cleanup error handling. Make multiError implement ExitCoder~
- [x] #1663
- [x] #1890 [^1]
- [x] #1775
- [x] #1531
- [x] #1891

Please suggest more! I'll add them to this OP

**Update ~January 2023**: Progress! And yet, yeah, wish lists balanced with releasing usable software is hard to do. If you're the kind of human who finds themselves reading this, there's a `v3.0.0-alpha2` tag now, which should play nicely with `go get github.com/urfave/cli/v3@latest`. Feedback wanted! ❤️ ~ @meatballhat 

[^1]: The documentation-related code that depends on `github.com/russross/blackfriday/v2` and `github.com/cpuguy83/go-md2man/v2` has been [shown to add significant bulk to compiled outputs](https://github.com/urfave/cli/pull/1375#issuecomment-1109117786)

saschagrunert:
Regarding the API we could split it up into an internal and external one. Everything which should not be imported externally would go into the `internal` folder.

saschagrunert:
I would love to see fish shell completion support. 💚

[...]

PR #1052, opened 2020-01-25
Extract cli.Context into an interface

asahasrabuddhe:
## What type of PR is this?

- cleanup  

In response to the discussion here (https://github.com/urfave/cli/issues/833#issuecomment-574171962) this is an attempt to try to extract the `cli.Context` into its own interface. This is currently a WIP.

### DIFF
--- a/context.go
+++ b/context.go
@@ -1,7 +1,7 @@
 package cli
 
 import (
-	"context"
+	ctx "context"
 	"errors"
 	"flag"
 	"fmt"
@@ -18,7 +18,7 @@ type Context interface {
 	App() *App
 	SetCommand(command *Command)
 	Command() *Command
-	Context() context.Context
+	Context() ctx.Context
 	ParentContext() Context
 	SetShellComplete(shellComplete bool)
 	ShellComplete() bool
@@ -51,8 +51,8 @@ type Context interface {
 	Uint(name string) uint
 }
 
-type defaultContext struct {
-	context       context.Context
+type cliContext struct {
+	context       ctx.Context
 	app           *App
 	command       *Command
 	shellComplete bool
@@ -62,7 +62,7 @@ type defaultContext struct {
 
 // NewContext creates a new context. For use in when invoking an App or Command action.
 func NewContext(app *App, set *flag.FlagSet, parentCtx Context) Context {
-	c := &defaultContext{app: app, flagSet: set, parentContext: parentCtx}
+	c := &cliContext{app: app, flagSet: set, parentContext: parentCtx}
 	if parentCtx != nil {
 		c.context = parentCtx.Context()
 		c.shellComplete = parentCtx.ShellComplete()
@@ -74,74 +74,74 @@ func NewContext(app *App, set *flag.FlagSet, parentCtx Context) Context {
 	c.command = &Command{}
 
 	if c.context == nil {
-		c.context = context.Background()
+		c.context = ctx.Background()
 	}
 
 	return c
 }
 
-// NewWrappedContext creates a Context which wraps context.Context
-func NewWrappedContext(ctx context.Context) Context {
-	return &defaultContext{context: ctx}
+// NewWrappedContext creates a Context which wraps ctx.Context
+func NewWrappedContext(context ctx.Context) Context {
+	return &cliContext{context: context}
 }
 
 // Returns the App associated with the current context
-func (c *defaultContext) App() *App {
+func (c *cliContext) App() *App {
 	return c.app
 }
 
 // Associates a command with the current context
-func (c *defaultContext) SetCommand(command *Command) {
+func (c *cliContext) SetCommand(command *Command) {
 	c.command = command
 }
 
 // Returns the Command associated with the current context
-func (c *defaultContext) Command() *Command {
+func (c *cliContext) Command() *Command {
 	return c.command
 }
 
 // Returns the context.Context wrapped inside the current context
-func (c *defaultContext) Context() context.Context {
+func (c *cliContext) Context() ctx.Context {
 	return c.context
 }
 
 // Returns the parent of the current context
-func (c *defaultContext) ParentContext() Context {
+func (c *cliContext) ParentContext() Context {
 	return c.parentContext
 }
 
 // Sets the shellComplete boolean for the current context
-func (c *defaultContext) SetShellComplete(shellComplete bool) {
+func (c *cliContext) SetShellComplete(shellComplete bool) {
 	c.shellComplete = shellComplete
 }
 
 // Returns the value of shellComplete boolean for the current context
-func (c *defaultContext) ShellComplete() bool {
+func (c *cliContext) ShellComplete() bool {
 	return c.shellComplete
 }
 
 // Associates a flagset with the current context
-func (c *defaultContext) SetFlagSet(set *flag.FlagSet) {
+func (c *cliContext) SetFlagSet(set *flag.FlagSet) {
 	c.flagSet = set
 }
 
 // Returns the flagset associated with the current context
-func (c *defaultContext) FlagSet() *flag.FlagSet {
+func (c *cliContext) FlagSet() *flag.FlagSet {
 	return c.flagSet
 }
 
 // NumFlags returns the number of flags set
-func (c *defaultContext) NumFlags() int {
+func (c *cliContext) NumFlags() int {
 	return c.flagSet.NFlag()
 }
 
 // Set sets a context flag to a value.
-func (c *defaultContext) Set(name, value string) error {
+func (c *cliContext) Set(name, value string) error {
 	return c.flagSet.Set(name, value)
 }
 
 // IsSet determines if the flag was actually set
-func (c *defaultContext) IsSet(name string) bool {
+func (c *cliContext) IsSet(name string) bool {
 	if fs := lookupFlagSet(name, c); fs != nil {
 		if fs := lookupFlagSet(name, c); fs != nil {
 			isSet := false
@@ -167,7 +167,7 @@ func (c *defaultContext) IsSet(name string) bool {
 }
 
 // LocalFlagNames returns a slice of flag names used in this context.
-func (c *defaultContext) LocalFlagNames() []string {
+func (c *cliContext) LocalFlagNames() []string {
 	var names []string
 	c.flagSet.Visit(makeFlagNameVisitor(&names))
 	return names
@@ -175,7 +175,7 @@ func (c *defaultContext) LocalFlagNames() []string {
 
 // FlagNames returns a slice of flag names used by the this context and all of
 // its parent contexts.
-func (c *defaultContext) FlagNames() []string {
+func (c *cliContext) FlagNames() []string {
 	var names []string
 	for _, ctx := range c.Lineage() {
 		ctx.FlagSet().Visit(makeFlagNameVisitor(&names))
@@ -185,7 +185,7 @@ func (c *defaultContext) FlagNames() []string {
 
 // Lineage returns *this* context and all of its ancestor contexts in order from
 // child to parent
-func (c *defaultContext) Lineage() []Context {
+func (c *cliContext) Lineage() []Context {
 	var lineage []Context
 
 	for cur := Context(c); cur != nil; cur = cur.ParentContext() {
@@ -196,22 +196,22 @@ func (c *defaultContext) Lineage() []Context {
 }
 
 // Value returns the value of the flag corresponding to `name`
-func (c *defaultContext) Value(name string) interface{} {
+func (c *cliContext) Value(name string) interface{} {
 	return c.flagSet.Lookup(name).Value.(flag.Getter).Get()
 }
 
 // Args returns the command line arguments associated with the context.
-func (c *defaultContext) Args() Args {
+func (c *cliContext) Args() Args {
 	ret := args(c.flagSet.Args())
 	return &ret
 }
 
 // NArg returns the number of the command line arguments.
-func (c *defaultContext) NArg() int {
+func (c *cliContext) NArg() int {
 	return c.Args().Len()
 }
 
-func lookupFlag(name string, ctx *defaultContext) Flag {
+func lookupFlag(name string, ctx *cliContext) Flag {
 	for _, c := range ctx.Lineage() {
 		if c.Command() == nil {
 			continue
--- a/context.go
+++ b/context.go
@@ -22,7 +22,6 @@ type Context interface {
 	WithFlagset(set *flag.FlagSet) Context
 	// Functions to be able to manage Context type
 	App() *App
-	setCommand(command *Command)
 	Command() *Command
 	Context() context.Context
 	ParentContext() Context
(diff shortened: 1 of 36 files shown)


asahasrabuddhe:
@lynncyrin / @rliebz / @saschagrunert I would like to start a discussion over here about if you agree with the direction of this PR and what you would like to be changed

context.go
```diff
@@ -6,53 +6,142 @@ import (
...
-type Context struct {
-	context.Context
-	App           *App
-	Command       *Command
+type Context interface {
+	// Functions to be able to manage Context type
+	App() *App
+	SetCommand(command *Command)
+	Command() *Command
+	Context() context.Context
```
saschagrunert: Do we really need nested contexts?

context.go
```diff
@@ -6,53 +6,142 @@ import (
...
+	NumFlags() int
+	Set(name, value string) error
+	IsSet(name string) bool
+	LocalFlagNames() []string
+	FlagNames() []string
+	Lineage() []Context
+	Value(name string) interface{}
+	Args() Args
+	NArg() int
+	// Functions for each flag type supported by the cli
```
saschagrunert: We probably want to move this into a separate Interface. The context interface is already pretty huge so we could think about 1-3 logical groups of interfaces:

- flag values
- app settings
- context relations (parent, etc.) - I’m not sure if we need this at all.

context.go
rliebz: I think the names are unfortunate here, but this isn't really a nested context.

This field refers to the standard library's `context.Context`, so having this function is effectively the same as `http.Request` having a `Context()` function. We're attaching a standard library context to the thing that is already being passed through our application logic, which is our `cli.Context`.

context.go
```diff
@@ -6,53 +6,142 @@ import (
...
-	App           *App
-	Command       *Command
+type Context interface {
+	// Functions to be able to manage Context type
+	App() *App
+	SetCommand(command *Command)
+	Command() *Command
+	Context() context.Context
+	ParentContext() Context
+	SetShellComplete(shellComplete bool)
```
rliebz: I'm going to pick on this as an example, although I'm sure it's not the only one.

When creating an interface that we expose, we need to be conservative about what methods we're exposing publicly. Having a setter method here seems to be an artifact of how the code happened to have been written, but I can't think of a scenario in which a user would ever be calling this function.

In this case, since the one place where this is called is right after we create a new context, we could find a way to get this information into the context constructor and remove it from the interface. On the more general note though, I'd like to see this as small and immutable as we can possibly make it.

context.go
```diff
@@ -6,53 +6,142 @@ import (
...
 	shellComplete bool
 	flagSet       *flag.FlagSet
-	parentContext *Context
+	parentContext Context
 }
 
 // NewContext creates a new context. For use in when invoking an App or Command action.
-func NewContext(app *App, set *flag.FlagSet, parentCtx *Context) *Context {
-	c := &Context{App: app, flagSet: set, parentContext: parentCtx}
+func NewContext(app *App, set *flag.FlagSet, parentCtx Context) Context {
```
rliebz: One of the alternatives to this PR as a whole would be to revamp the `NewContext` function to actually be useful.

The problem it has now is that the interface makes it impossible to create a useful `cli.Context` outside of the library. Accepting a `*flag.FlagSet`, which is from the standard library, makes it look generic, but the issue is that the flag set creation logic we use internally is non-trivial and relies on unexported fields. Specifically, we:
- Use a private function to check if we're doing shell completion
- Use a private function to create a new flag set
- Use a private function to parse flags iteratively
- Use a private function to normalize flags
- Handle errors in a way that sometimes uses the context even if creating it has failed
- Modify private fields on the context immediately after its used

On top of this, the logic is similar but slightly different between doing those steps for an app versus a command.

I could see a much smaller breaking change solving the same problem by doing something like this:
```go
func (a *App) NewContext(args []string, parent Context) Context {
```

Switching from a `*flag.FlagSet` to `[]string` means we can continue to own the complicated flag parsing logic, while giving users a way to create new contexts in tests as simply as passing flags. It also means the `cli.Context` in tests doesn't need to be mocked, making tests slightly closer to what real code would be doing and potentially making test boilerplate smaller.

I'm not sure if that would be better than what this PR does, or what the implications of that change would be in terms of broader refactors, but I wanted to float that idea before we decide we're definitely going in this direction.

context.go
```diff
@@ -6,53 +6,142 @@ import (
...
 // can be used to retrieve context-specific args and
 // parsed command-line options.
-type Context struct {
-	context.Context
-	App           *App
-	Command       *Command
+type Context interface {
+	// Functions to be able to manage Context type
+	App() *App
+	SetCommand(command *Command)
```
rliebz: It looks like this is only used immediately after creating a new `cli.Context` for the command. Since this is something users won't ever be calling directly, we should see if we can't remove this from the public interface.

context.go
```diff
@@ -6,53 +6,142 @@ import (
...
+	FlagSet() *flag.FlagSet
+	NumFlags() int
+	Set(name, value string) error
+	IsSet(name string) bool
+	LocalFlagNames() []string
+	FlagNames() []string
+	Lineage() []Context
+	Value(name string) interface{}
+	Args() Args
+	NArg() int
```
rliebz: This is a great example of something we can probably remove to shrink our interface.

No reason to keep an `NArg()` function when we could just do `c.Args().Len()`

context.go
asahasrabuddhe: I am open to new names. I just extracted the interface out of what we had. As we are going for a v3, new names should not matter much.

context.go
asahasrabuddhe: Sounds good. Let me fix this up :)

context.go
asahasrabuddhe: I like where you are going with this. Having done this first step, it allowed me to gain a perspective over the grand scheme of things. I now will incorporate the feedback here slowly as we try to lessen the public face of the API.

asahasrabuddhe:
Thanks for the quick feedback guys especially on a Sunday! Let me get back to work and come back with some improvements :)

context.go
asahasrabuddhe: I agree. However, I think we must first begin with deprecating it in our first v3 release and then remove it in v3.1 which will allow people to upgrade without having to face issues.

Screenshot 2020-01-29 at 6 19 59 PM

The deprecation notice is quiet clear in most IDEs.

context.go
asahasrabuddhe: This is interesting. Making `NewContext` a part of `App` will lead to a large number of changes in the code and the tests. I would be willing to do it if we are left with no other option. I have been reading around to help me solve this and I think we can take inspiration from the standard library and have a number of functions like:

```go
func NewContext(context context.Context) Context
func WithParent(parent Context) Context
func WithApp(app *App) Context
func WithFlagset(set *flag.FlagSet) Context
```

Such kind of functions are available in `html/template` `text/template` and `context` standard library. I think these would allow us to be much flexible with how the context is created.

What do you think?

context.go
asahasrabuddhe: I have pushed new code with the `Context` interface split up.

asahasrabuddhe:
@saschagrunert  / @rliebz I have updated the code to include some of our discussed changes. Please review :) 

@lynncyrin  Although `GitHub` not selecting you as a reviewer for this PR, I would really appreciate if you could review this when time permits. Thanks!

context.go
```diff
@@ -6,53 +6,170 @@ import (
...
+	StringSlice(name string) []string
+	Int64(name string) int64
+	Float64Slice(name string) []float64
+	Duration(name string) time.Duration
+	IntSlice(name string) []int
+	Int(name string) int
+	Path(name string) string
+	Uint(name string) uint
+	Value(name string) interface{}
+}
```
saschagrunert: This is pretty clean, I like that :+1:

saschagrunert (dismissed):
That's a pretty good start, I think we should move forward with this :)

context.go
asahasrabuddhe: I'm glad you like it :)

context.go
rliebz: It's more important to adhere to semantic versioning here. Removing a deprecated feature is still a breaking change, meaning people will no longer safely be able to go from `3.0` to `3.1`.

If we're interested in marking it as deprecated, we should do so in the `v2` branch now, or mark it in `v3` and remove it in `v4`. That said, this kind of change is trivial enough we'd probably be fine just bundling it with the rest.

context.go
rliebz: I'm not sure this is a good fit here.

The biggest goals I would prioritize for a revamped `NewContext` function would be to make it as simple as possible to generate a valid `Context`, and to make it difficult or impossible to generate an invalid `Context`.

For the first goal, the reason I suggested the signature just taking a list of args is because that's an interface that users already are familiar with, and it's data they have access to. `app.Run` takes a list of args. The CLI itself effectively takes a list of args. `os.Args` gives easy access in their main program, and so on.

The challenge with chaining is that you get a `Context` back at every step. Is a `Context` valid without an app? Is it valid without any flag set? I believe we'd end up in a similar situation to now, where it would be easy to generate an invalid `Context`, and not necessarily simpler to create a valid one.

context.go
asahasrabuddhe: Hmm. What we can do is we deprecate things during development time and then remove deprecated things before making the final release.

context.go
asahasrabuddhe: The real question that I am thinking about is whether the `NewContext` function should even be exposed. I have not yet come across a scenario where I needed to create an instance of `cli.Context`. 

Also to answer your question, the `NewContext` function can be modified to set sane defaults.

rliebz: > I have not yet come across a scenario where I needed to create an instance of cli.Context.

If this is the case, what's the benefit in making `cli.Context` an interface? 

Interfaces are are less useful than structs in nearly every way. Compared to a struct, using an interface is more verbose because you have to access fields by function call rather than directly, harder to instantiate because you need a minimum of one additional type implemented to be able to use them, less performant because you're forcing an allocation, less clear because "go-to definition" in the editor now only reaches a signature rather than an implementation, and so on.

The benefit, though, is that you aren't tied to a single implementation. 

If you aren't reaching to create a `Context`, then I don't think this change brings us value in terms of public API. The big benefit of being able to provide an alternate implementation is irrelevant—there's no need to provide any implementation at all.

The reason I'm suggesting changing `NewContext` instead of making `cli.Context` an interface is because the specific goal I had in mind when suggesting making `Context` into an interface was to make a `Context` easier to create. For me, this is for unit testing.

Here is [an example of code where the only things I've needed out of a context are `NArg` and `IsSet`](https://github.com/rliebz/tusk/blob/f27bf1c90703ca30ea931bf583b56245d79e2adb/appcli/completion.go). It's awkwardly implemented because it was not possible with the current interface of `NewContext` for me to create a `Context` that was close enough to the real thing that it would be useful in tests. This awkward implementation does [let me create a context for myself](https://github.com/rliebz/tusk/blob/master/appcli/completion_test.go#L420-L437) that I can [instantiate in unit tests without having to manually parse arguments myself](https://github.com/rliebz/tusk/blob/f27bf1c90703ca30ea931bf583b56245d79e2adb/appcli/completion_test.go#L261-L264), because the logic to do so is [a private implementation detail of `urfave/cli` and incredibly non-trivial](https://github.com/urfave/cli/blob/fed41d6ef81daa62b81c858105cde642a31cd976/app.go#L217-L233). This PR would _definitely_ make some of those code snippets more straightforward and easier to test, but before we tie ourselves to this breaking change in particular, I want to explore if there are better ways to solve the same problems.

> Making NewContext a part of App will lead to a large number of changes in the code and the tests. 

The other point I want to make here is that I'm less concerned about the size of the change within `urfave/cli` and more concerned about the size of the change to dependent applications. I can spend a few hours on a weekend patching all the unit tests for `urfave/cli` if we make an internally-facing breaking change, no big deal. If we make an externally-facing breaking change, I'll still spend a few hours on a weekend patching some other applications, but so will everyone else who uses this project.

lynncyrin:
> If this is the case, what's the benefit in making cli.Context an interface?

☝️ I'm interested in the answer to this question! From here https://github.com/urfave/cli/pull/1052#discussion_r373327116

Nokel81:
Would it be possible instead of having a `Context() context.Context` method, to just have `cli.Context` extend `context.Context`? This would make using `cli.Context` easier.

The only problem with this is that the current `Value(key string) interface{}` method from `type FlagValues interface` conflicts with the `Value(key interface{}) interface{}` from `context.Context`. A possible solution to this could be making our function `FlagValue(key string) interface{}` instead.

rliebz:
> Would it be possible instead of having a Context() context.Context method, to just have cli.Context extend context.Context? 

The challenge with this is that the context interface was not designed to be extended. Consider a `CustomContext` type that extends the `context.Context` interface, which I'll use as a stand-in for the `cli.Context` type. What happens if we try to add a timeout, for example?

```go
package main

import (
	"context"
	"fmt"
	"time"
)

type CustomContext struct {
	context.Context
}

func main() {
	var ctx context.Context = CustomContext{context.TODO()}
	fmt.Printf("Type Before: %T\n", ctx)
	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
	defer cancel()
	fmt.Printf("Type After: %T\n", ctx)
}
```

This will print the following:
```
Type Before: main.CustomContext
Type After: *context.timerCtx
```

We create a new context that extends our previous one with no clear way to get our original type back. This means any custom functionality we have on our `cli.Context` is gone as soon as we try to perform operations on the context such as attaching deadlines or any `ContextWithFoo`-type methods that are common in third party libraries.

The problem we're solving is not dissimilar to the problem that the `http` package solves with `http.Request`—we have a struct that was already passed around from function to function containing metadata, and attaching a `context.Context` as a member of that object is probably the easiest way to work with it.

meatballhat:
@asahasrabuddhe I'm interesting in continuing this work if you're not able ✋🏼

asahasrabuddhe:
@meatballhat I can and would love to continue this. However, we need to decide on what direction to proceed. Based on earlier feedback, this PR was stuck. What do you think we should do?

meatballhat:
@asahasrabuddhe Hello! I'm so glad you're here and willing and able to continue this work! After more carefully reviewing the commentary, I think I am generally in agreement with @rliebz, which I'd restate (please correct me @rliebz!) as:

- trivially creating a new and valid `cli.Context` is mostly the concern of _this library's maintainers_
- the benefits of defining an interface that allows alternative implementations are unclear or otherwise not strongly evident based on real world library usage

and also notable, mentioned in another comment:

- standard library `context.Context` was not built with extension in mind

All of which is to say that I think redefining `cli.Context` as an interface won't improve the experience of real world production library usage.

What do you think about instead building out a sub-package that specifically targets testing concerns, like what is available in `net/http/httptest`? For example:

```go
// moved from current location
cCtx := clitest.NewContext(myApp, myFlagSet, nil)

// ...

cCtx := clitest.NewParentContext()

// ...

cCtx := clitest.NewContextWithFlagSet(myFlagSet)
```

dearchap:
@asahasrabuddhe @meatballhat  I'm closing this PR as we dont want to proceed in this direction. We can have a separate PR for subpackage

PR #1062, opened 2020-02-09
Fix zsh completion scripts

zhsj:
## What type of PR is this?

_(REQUIRED)_

- [x] bug
- [ ] cleanup  
- [ ] documentation
- [ ] feature

_(REQUIRED)_

This fix completion when argument startswith -
And prevent completing a space when no candidates

_(REQUIRED)_

## Release Notes

_(REQUIRED)_

```release-note
Fix zsh completion scripts
```


rliebz:
Thanks for fix!

Would you mind providing examples of what was broken before and what's changed? Between the PR description and the code changed, I'm having trouble understanding what exactly the issue was.

zhsj:
> Thanks for fix!
> 
> Would you mind providing examples of what was broken before and what's changed? Between the PR description and the code changed, I'm having trouble understanding what exactly the issue was.

Before the change, if you type `command -<tab>` in zsh, nothing happens. With this change, it will display the candicates(eg `-h`).

The problem here is the `words` variable, when type `<tab>` before `-`, the last word is `-`, not `""`.

The change is based on the code on bash completion script, 

https://github.com/urfave/cli/blob/cc46cf9c34062ce441eb9aa8d6804950f6d7e5a0/autocomplete/bash_autocomplete#L10-L14

PR #1078, opened 2020-02-28
Allow specifying a StringSlice destination for StringSliceFlag

### LINKED ISSUE #1075: v2 feature: Add Destination for StringSliceFlag
## Checklist

- My team want to start using the `Destination` values on flags so we can have our flag values strongly typed rather than stringly typed (i.e calling c.String("flag") etc.). I noticed that the `StringSliceFlag` does not have a `Destination` field like other flags do 
 
## Solution description

- Solution would be to add a `Destination` field to `StringSliceFlag` that works the same as all other `Destination` flag fields do.

davidsbond:
## What type of PR is this?

- [ ] bug
- [ ] cleanup  
- [ ] documentation
- [x] feature

Adds a `Destination` field to the `StringSliceFlag` of type `StringSlice` so that we can use a variable to access the value rather than calling `c.StringSlice`. This adds some additional type safety and can help to avoid scenarios where you have typos calling the `c.StringSlice("name")` method.

Fixes #1075 

## Testing

By running the code I've seen that the `Destination` gets set, I've also added in some tests for setting the destination value from env and from flags

## Release Notes

```release-note
Adds destination field to StringSliceFlag
```

### DIFF
--- a/flag_string_slice.go
+++ b/flag_string_slice.go
@@ -71,6 +71,7 @@ type StringSliceFlag struct {
 	Value       *StringSlice
 	DefaultText string
 	HasBeenSet  bool
+	Destination *StringSlice
 }
 
 // IsSet returns whether or not the flag has been set through env or file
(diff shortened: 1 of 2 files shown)


davidsbond:
I've found a bug in this code so far. The problem is that when you provide both environment variables and flags, you end up appending to the array instead of replacing one with the other. For example:

Environment: `[a, b, c]`
Flags: `[1, 2, 3]`
Result: `[a,b,c,1,2,3]`

davidsbond:
I think I've solved it, but I'm going to do some more testing.

davidsbond:
Based on my local testing this now works as expected, flags will override the slice set by env vars entirely.

davidsbond:
```go
package main

import (
	"fmt"
	"github.com/davidsbond/cli"
	"log"
	"os"
)

var (
	dest = &cli.StringSlice{}
)

func main() {
	os.Setenv("TEST_THIS_THING", "1,2,3")

	app := cli.NewApp()
	app.Action = run
	app.Flags = []cli.Flag{
		&cli.StringSliceFlag{
			Name:        "test",
			Destination: dest,
			EnvVars:     []string{"TEST_THIS_THING"},
		},
	}

		log.Fatal(err)
	}
}

func run(c *cli.Context) error {
	fmt.Println(dest.Value())
	return nil
}
```

```
go run main.go --test=a,b,c
[a,b,c]
go run main.go  
[1 2 3]
```

Issue #1057, opened 2020-01-30
maint: add binary size checking to CI


lynncyrin:
## What problem does this solve?

This fixes the problem of the binary size increasing over time, which is a non-starter for some users (like @oleorhagen!)

## Solution description

A script that does the following psuedo-code:

```
go build urfave/cli --output ./urfave-cli

binarySize = checkBinarySize ( "./urfave-cli" )
expectedBinarySize = 5000

if binarySize > expectedBinarySize then
    fail the build, reason: binary too large
if binarySize < expectedBinarySize * 0.70% then
    fail the build, reason: binary is much smaller now ✨\
        please decrease the expected size
```

## Potential Future Work

If we make the CLI modular (mentioned here https://github.com/urfave/cli/issues/833#issuecomment-579869829), we can assert different expected binary sizes for the "minimal" vs "complete" versions of the package.

## Context

This idea was inspired by https://github.com/urfave/cli/issues/1055

Issue #1055, opened 2020-01-28
Move man page compatibility to separate package


oleorhagen:
## What problem does this solve?

We have currently run into an issue where 30% of our binary size is the `urfave/cli` library. With the latest release `v2` it it 50%, which for us is not going to fly.

- Add a minimal version of the cli, using conditional compilation maybe?

## Solution description

Add a `minimal` build flag which does not include all the external dependencies like `md2man` etc, and simply contains the 'core' functionality of the library.

asahasrabuddhe:
This is an interesting observation. I think that we could look at making the library `modular` so that it is possible for the users to import certain modules that they want to use and ignore the others. This would considerably help in keeping the size down.

Having said that, I think this won't be possible for the v2.

But, I am glad to share with you that we are currently working on v3. I think I will add this to the list of things that we want to work on. Hopefully, very soon, you should be looking at a lean version of this library.

Thanks!

lynncyrin:
I think if nothing else, it would be a good idea to add some stats about binary size to CI. We can do that now, and look at reducing it for V3.

dearchap:
@meatballhat i believe we have builds which doesn't include docs

dearchap:
Will consider for v3 release

PR #1079, opened 2020-02-29
Add binary size bounds checking

lynncyrin:
## What type of PR is this?

- [ ] bug
- [ ] cleanup  
- [ ] documentation
- [ ] feature
- [x] maintenance / operations  

This PR adds binary size bounds checking, so that we can avoid more people having issues like that one described in https://github.com/urfave/cli/issues/1055.

Fixes https://github.com/urfave/cli/issues/1057

## Changes

- moved `build.go` to `internal/build.go`
- I didn't change any of the existing bits of build.go 😝 (except the top comment) the git diff is just betraying me
- adds a build script for checking binary size

Here's parts of this PR that I find interesting:

- the binary size diffs are different on ubuntu / mac / windows, and different on different go versions. as of https://github.com/urfave/cli/pull/1079/commits/a8a1ef04eea0e80585590153eb3f04dcb444285f they're `2.0`, `2.1`.
- ~the specific values for the min and max sizes are interesting to me, I think there being `0.5MB` between them is a comfortable distance. I would be open to decreasing the min and max to fit the current values more tightly (eg. min 4.7 max 4.9), though~ I changed the min and max to just be the min and max size diffs on the various platforms
- I assume that the next person to trigger test failures based on binary size won't be me, eg. it'll be some new open source contributor working on their 1st PR. For that person, do you think the included guidance is sufficient?

## Potential Future Work

I thought about adding either of these things:

- https://github.com/jondot/goweight
- https://github.com/google/bloaty

I think they'd be cool to add ✨ but I'm not _that invested_ in adding them personally.

## Trivia

People really care about binary sizes! https://github.com/golang/go/issues/6853

## Testing

I briefly changed the min and max values to make sure that they are actually exiting with a failure code and displaying the guidance properly.

## Release Notes

There's no user-facing changes in this PR ^_^


lynncyrin:
Possible future optimization: instead of the raw size, use the diff between main.go with "hello world" and a main.go with urfave/cli

internal/build/build.go
```diff
@@ -0,0 +1,266 @@
...
+		desiredMinBinarySize = 4.5
+		desiredMaxBinarySize = 5.0
+		badNewsEmoji         = "🚨"
+		goodNewsEmoji        = "✨"
+		checksPassedEmoji    = "✅"
+		mbStringFormatter    = "%.1fMB"
+	)
+
+	// build example binary
+	err = runCmd("go", "build", "-o", builtFilePath, sourceFilePath)
```
saschagrunert: Do we want to strip here, like `go build -ldflags '-s -w'`? (I think we should)

internal/build/build.go
```diff
@@ -0,0 +1,266 @@
...
+	if err != nil {
+		return err
+	}
+
+	// get file into
```
saschagrunert:
```suggestion
	// get file info
```

internal/build/build.go
```diff
@@ -0,0 +1,266 @@
...
+		fmt.Println("") // visual spacing
+		fmt.Println("     The binary is larger than the target max size.")
+		fmt.Println("     That means that your changes are increasing the binary size.")
+		fmt.Println("     The first thing you'll want to do is ask your yourself")
+		fmt.Println("     Is this change worth increasing the binary size?")
+		fmt.Println("     Larger binary sizes for this package can dissuade its use.")
+		fmt.Println("     If this change is worth the increase, then we can up the")
+		fmt.Println("     desired max binary size. To do that you'll want to go into")
+		fmt.Println("     ./internal/build/build.go and increase the desiredMaxBinarySize,")
+		fmt.Println("     and increase the desiredMinBinarySize by the same amount.")
```
saschagrunert: We probably want to add more details about what the binary size increases, like:
- Added dependencies
- Added const values (.text, .rodata)
- What else?

lynncyrin: I think if anything this console message should get smaller (because it's annoying to update) and we should point to a long-ish thing in the `docs/` folder that describes all the various ways binary size could increase. I'm imagining there being a `docs/developement.md#binary-size` section that's like 10 paragraphs long 😄 

Does that make sense? Or should we put something shorter into this error message?

@saschagrunert

internal/build/build.go
lynncyrin: done!

internal/build/build.go
saschagrunert: SGTM :)

PR #1080, opened 2020-03-02
Refactor '(default: …)' flag string into utility function

saschagrunert:
## What type of PR is this?

- [ ] bug
- [x] cleanup  
- [ ] documentation
- [ ] feature

The usage of the `(default: …)` strings are a bit error prone so we now
refactor them into a dedicated helper function.

None

## Release Notes

_(REQUIRED)_

```release-note
None
```

### DIFF
--- a/flag.go
+++ b/flag.go
@@ -244,6 +244,10 @@ func flagValue(f Flag) reflect.Value {
 	return fv
 }
 
+func formatDefault(format string) string {
+	return " (default: " + format + ")"
+}
+
 func stringifyFlag(f Flag) string {
 	fv := flagValue(f)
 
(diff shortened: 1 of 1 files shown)


flag.go
```diff
@@ -244,6 +244,10 @@ func flagValue(f Flag) reflect.Value {
 	return fv
 }
 
+func formatDefault(format string) string {
+	return " (default: " + format + ")"
+}
```
lynncyrin: _(non-blocking comment)_ do we know offhand why the different calls need different formatting verbs?

saschagrunert: Yeah we quote strings, but not bools for example. Because if the flag is a bool, then passing `"true"` will be interpreted as string and not bool.

Btw, I was wondering why we not print the default (like `""` for string flags) values. 🤔

rliebz: I think not printing the zero value makes sense because it's not unlikely that an application has logic that treats the zero value as something other than `""`.

For example, I have a CLI app with several string flags, including `--file <path>` and `--install-completion <shell-name>`. The "default" value for the config file is actually to recursively search up directories until it finds a file with a matching name. An empty string isn't accurate in this case, but that's how my app sees it. For `--install-completion`, the options are `bash` or `zsh`, but if nothing is passed, the flag doesn't run. 

An empty string in both of those cases means something to me as an application developer, but not to the consumers of my CLI application, so I like the current behavior.

PR #1082, opened 2020-03-02
Make help message optional if error happen in BeforeFunc

### LINKED ISSUE #1081: v2 feature: Make error message optional in BeforeFunc
## Checklist

Displaying the help message in case of error in BeforeFunc should be optional. In my case I am using the BeforeFunc to validate command context and fails if something goes wrong, and I don't want any help message displayed.

## Solution description

I have add a **HideHelpOnBeforeError** flag to allow to hide help message in case error happens in the BeforeFunc.

creekorful:
## What type of PR is this?

- [ ] bug
- [ ] cleanup  
- [ ] documentation
- [X] feature

Displaying the help message in case of error in BeforeFunc should be optional. In my case I am using the BeforeFunc to validate command context and fails if something goes wrong, and I don't want any help message displayed.

Having a flag will allow user to customize depending on his use case.

This PR is not a breaking change, since without the flag the behavior is the same as before.

Fixes #1081

## Release Notes

```release-note
Add a HideHelpOnBeforeError boolean to hide error display if something goes wrong in BeforeFunc.
```

### DIFF
--- a/app.go
+++ b/app.go
@@ -57,8 +57,8 @@ type App struct {
 	// An action to execute after any subcommands are run, but after the subcommand has finished
 	// It is run even if Action() panics
 	After AfterFunc
-	// Should the help message be displayed in case of error in the Before action
-	ShowHelpOnBeforeError bool
+	// Should the help message be hidden in case of error in the Before action
+	HideHelpOnBeforeError bool
 	// The action to execute when no subcommands are specified
 	Action ActionFunc
 	// Execute this function if the proper command cannot be found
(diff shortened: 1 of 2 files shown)


app.go
```diff
@@ -57,6 +57,8 @@ type App struct {
 	// An action to execute after any subcommands are run, but after the subcommand has finished
 	// It is run even if Action() panics
 	After AfterFunc
+	// Should the help message be hidden in case of error in the Before action
+	HideHelpOnBeforeError bool
```
lynncyrin: This seems too fine-grain for a top level API switch? I feel like there must be some other way to accomplish this without expanding the public API.

creekorful: Hello. I have feel the same while implementing this solution, however it's the only solution I've come up with to prevent behaviour change.

The other idea I have is to simply remove the help message, and let the user decide if he want to show help or not. But this change would break the old behaviour.

lynncyrin: We should discuss this in the issue here => https://github.com/urfave/cli/issues/1081, our understanding of this problem isn't mature enough to warrant a PR at this time.

lynncyrin:
Closing this via https://github.com/urfave/cli/pull/1082#discussion_r389356169, @creekorful please followup with a proposal inside that issue, rather than creating a PR.

PR #1083, opened 2020-03-05
Add HideHelpCommand

### LINKED ISSUE #523: Don't want any commands in the CLI interface (including help)

AkihiroSuda:
## What type of PR is this?

- [X] feature

While `HideHelp` hides both `help` command and `--help` flag, `HideHelpCommand`
only hides `help` command and leave `--help` flag as-is.

The behavior of `HideHelp` is untouched in this commit.

Fixes #523
Replaces #636

## Testing

`go test -v -run TestHideHelpCommand`

## Release Notes

```release-note
Added `HideHelpCommand`. While `HideHelp` hides both `help` command and `--help` flag, `HideHelpCommand` only hides `help` command and leave `--help` flag as-is.
```

### DIFF
--- a/app.go
+++ b/app.go
@@ -43,8 +43,11 @@ type App struct {
 	Flags []Flag
 	// Boolean to enable bash completion commands
 	EnableBashCompletion bool
-	// Boolean to hide built-in help command
+	// Boolean to hide built-in help command and help flag
 	HideHelp bool
+	// Boolean to hide built-in help command but keep help flag.
+	// Ignored if HideHelp is true.
+	HideHelpCommand bool
 	// Boolean to hide built-in version flag and the VERSION section of help
 	HideVersion bool
 	// categories contains the categorized commands and is populated on app startup
@@ -170,7 +173,9 @@ func (a *App) Setup() {
 	a.Commands = newCommands
 
 	if a.Command(helpCommand.Name) == nil && !a.HideHelp {
-		a.appendCommand(helpCommand)
+		if !a.HideHelpCommand {
+			a.appendCommand(helpCommand)
+		}
 
 		if HelpFlag != nil {
 			a.appendFlag(HelpFlag)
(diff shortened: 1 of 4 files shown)


AkihiroSuda:
Any chance to get v2.2.0 released soon with this?

app.go
```diff
@@ -333,7 +338,9 @@ func (a *App) RunAsSubcommand(ctx *Context) (err error) {
 	// append help to commands
 	if len(a.Commands) > 0 {
 		if a.Command(helpCommand.Name) == nil && !a.HideHelp {
-			a.appendCommand(helpCommand)
+			if !a.HideHelpCommand {
```
rliebz: Looks like the reason the PR checks are failing is because this condition isn't being checked anywhere in the tests. See: https://codecov.io/gh/urfave/cli/pull/1083

AkihiroSuda: Added tests for `RunAsSubcommand`, but codecov is still failing

AkihiroSuda: now green

app.go
```diff
@@ -328,19 +333,9 @@ func (a *App) RunAndExitOnError() {
...
-	// append help to commands
-	if len(a.Commands) > 0 {
-		if a.Command(helpCommand.Name) == nil && !a.HideHelp {
-			a.appendCommand(helpCommand)
-
-			if HelpFlag != nil {
-				a.appendFlag(HelpFlag)
-			}
-		}
-	}
```
lynncyrin: ❓ _(blocking question)_ this code is being removed because it was duplicated in `Setup` above, yeah?

lynncyrin:
> Any chance to get v2.2.0 released soon with this?

I can write up the release notes this weekend probs

app.go
AkihiroSuda: Yes, and because of the duplication, this code path was not executed and could not be covered with tests: https://github.com/urfave/cli/pull/1083#discussion_r388284552

lynncyrin: Nice find!

Issue #1089, opened 2020-03-23
v2 bug: Sub commands call "HandleExitCoder"


Nokel81:
## my urfave/cli version is

v2.2.0

## Checklist

* [x] Are you running the latest v2 release? The list of releases is [here](https://github.com/urfave/cli/releases).
* [x] Did you check the manual for your release? The v2 manual is [here](https://github.com/urfave/cli/blob/master/docs/v2/manual.md)
* [x] Did you perform a search about this problem? Here's the [Github guide](https://help.github.com/en/github/managing-your-work-on-github/using-search-to-filter-issues-and-pull-requests) about searching.

## Dependency Management

- [x] My project is using go modules.
- [x] My project is using vendoring.
- [ ] My project is automatically downloading the latest version.
- [ ] I am unsure of what my dependency management setup is.

## Describe the bug

If a subcommand's Action function returns an `ExitCoder` error then the code that executes the subcommand calls `HandleExitCoder`. This makes unit testing harder and makes sub commands act differently to `*cli.App`.

## To reproduce

1. Have a sub command return an `ExitCoder`
2. Try and write a unit test that uses the whole cli library to set up the flags correctly.

## Observed behavior

The test exits with the `ExitCoder`'s exit code

## Expected behavior

The `ExitCoder` instance to be returned.

## Additional context

I have looked at the other issues on this subject (from 2017) and it looks like there was previous discussion on this topic. 

There is a work-around, that is to define an identify function on the app itself `ExitErrHandler` field. This is suboptimal because spills testing requirements into the production code itself. It also seems weird that this would be an app wide setting and not a command setting.

Given that the current documentation hints that the user should manually call `HandleExitCoder` from the return value of `*cli.App.Run` it seems strange that a subcommand would call it for you.

rliebz:
This seems to me like the correct behavior. An `ExitCoder` is an interface meant to specifically trigger the behavior for an application to exit with a particular exit code. Returning an `ExitCoder` signals that it should happen, and if it shouldn't happen, an `ExitCoder` shouldn't be returned.

There's a default behavior, the `HandleExitCoder` function, which can be overridden at the `cli.App` level either to change that behavior or to allow for testing. A test could potentially look like this:

```go
func TestMyAppExitCode(t *testing.T) {
  myApp := mypackage.NewProductionApp()

  myApp.ExitErrHandler = func(_ *cli.Context, err error) {
    // Any custom test logic, or assertions
[...]

PR #1091, opened 2020-03-25
Update docs and tests around cli.Exit

rliebz:
## What type of PR is this?

_(REQUIRED)_

- [ ] bug
- [ ] cleanup  
- [x] documentation
- [ ] feature

_(REQUIRED)_

Some of the docs were lacking or incomplete concerning how to properly
use cli.Exit, cli.HandleExitCoder, and the ExitErrHandler field on cli.App.
This change aims to clarify the usage of those pieces.

I also noticed that we have two identical functions now: cli.Exit and
cli.NewExitError. Since the latter seems to be more recent and less well
documented, I went ahead and marked it as deprecated so that we can keep
our public interface small. If there's a reason not to do that, I can punt that
discussion until later to get the rest of these changes through.

Also added a missing test case for behavior that's been around a while
but was not documented or tested.

_(REQUIRED)_

Related: #1089

## Testing

It's docs. I added a test case anyway to prove the behavior I was trying to document.

## Release Notes

_(REQUIRED)_

```release-note
Deprecated cli.NewExitError in favor of the identical function, cli.Exit
```

### DIFF
--- a/app.go
+++ b/app.go
@@ -76,8 +76,9 @@ type App struct {
 	Writer io.Writer
 	// ErrWriter writes error output
 	ErrWriter io.Writer
-	// Execute this function to handle ExitErrors. If not provided, HandleExitCoder is provided to
-	// function as a default, so this is optional.
+	// ExitErrHandler processes any error encountered while running an App before
+	// it is returned to the caller. If no function is provided, HandleExitCoder
+	// is used as the default behavior.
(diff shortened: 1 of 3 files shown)


rliebz:
@Nokel81 I'd love to get your thoughts on this as well if you have the chance to take a look.

Nokel81:
That looks much better, thanks.

PR #1100, opened 2020-03-31
Set App.ErrWriter in App.Setup()

### LINKED ISSUE #1096: q: App.Writer is being set twice
I was wondering why there is two instance of setting `*App.Writer` to `os.Stdout` in `*App.Setup()`? 

- https://github.com/urfave/cli/blob/master/app.go#L161
- https://github.com/urfave/cli/blob/master/app.go#L199

And `*App.ErrWriter` is not being set at all. Is that because it is going to be deprecated?

Nokel81:
## What type of PR is this?

_(REQUIRED)_

- [ ] bug
- [x] cleanup  
- [ ] documentation
- [ ] feature

_(REQUIRED)_

This change fixes `App.ErrWriter` not being initialized in `App.Setup()` like `App.Writer`. It also removes the `App.errWriter()` private method since `App.ErrWriter` is now set directly.

_(REQUIRED)_

Fixes #1096 

## Release Notes

_(REQUIRED)_

```release-note
- `App.ErrWriter` now defaults to `os.Stderr` if not set.
```

### DIFF
--- a/app.go
+++ b/app.go
@@ -118,6 +118,7 @@ func NewApp() *App {
 		Action:       helpCommand.Action,
 		Compiled:     compileTime(),
 		Writer:       os.Stdout,
+		ErrWriter:    os.Stderr,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


lynncyrin:
✨thanks! ✨

Issue #459, opened 2016-06-17
cannot use cli.IntFlag literal (type cli.IntFlag) as type cli.Flag in array or slice literal: cli.IntFlag does not implement cli.Flag (Apply method has pointer receiver)


andreynering:
_**Here is the fix for this issue! => https://github.com/urfave/cli/issues/459#issuecomment-555077623**_

---

I updated today and can't build my app (luckily I do vendoring). Am I missing any API change? The error is below:

``` go
cannot use cli.IntFlag literal (type cli.IntFlag) as type cli.Flag in array or slice literal:
        cli.IntFlag does not implement cli.Flag (Apply method has pointer receiver)
```

---

Also, why the change from `codegangsta/cli` to `urfave/cli`? I couln't find any information on README or anywhere.

Cheers.

---

_edited by @lynncyrin!_

epaulet:
Also getting it for other flag types as well (although it builds locally for me so it must be a recent change?)

```
cannot use cli.StringFlag literal (type cli.StringFlag) as type cli.Flag in array element:
	cli.StringFlag does not implement cli.Flag (Apply method has pointer receiver)
```

```
cannot use cli.StringSliceFlag literal (type cli.StringSliceFlag) as type cli.Flag in array element:
	cli.StringSliceFlag does not implement cli.Flag (Apply method has pointer receiver)
```

meatballhat:
Are both of you pinning to the v2 branch?

epaulet:
Yes. I forgot to mention that in my original post.

andreynering:
Yes, v2

meatballhat:
Ah cool. Sorry about the surprise.  The v2 branch is volatile, and recently
received a patch to introduce a lot more pointer receivers.  Switching from
struct literals to pointers to struct literals should do the trick.  The
change log for v2 changes should cover everything, and please yell if it
doesn't.  There is also a migration script in the repo root that may help.
Please provide feedback!  :smile_cat:

meatballhat:
related: thoughts on renaming the `v2` branch to `v2pre` or equiv?

andreynering:
My fault that I thought the `v2` was stable. Downgrading to `v1` fixed for me. Thank you.

meatballhat:
@andreynering :+1:

Tengoru:
Hi, I have the same problem but with version 1. I tried to copy the code snippets from the documentation but no luck... Any ideas what the problem could be?

Tengoru:
[...]

PR #1115, opened 2020-04-27
Fix v1 syntax documentaiton regarding `Flags`

hoijui:
As described in [this issue](https://github.com/urfave/cli/issues/459).
Marking issue as closed without updating documentation ... HMMM!!! ;-)
YES it also changed in v1, as noted in the issue in Nov 2019.

... somehow, stuff from v2 started seeping into v1?

- [x] bug
- [ ] cleanup  
- [x] documentation
- [ ] feature

### DIFF
--- a/docs/v1/manual.md
+++ b/docs/v1/manual.md
@@ -221,7 +221,7 @@ func main() {
   app := cli.NewApp()
 
   app.Flags = []cli.Flag {
-    cli.StringFlag{
+    &cli.StringFlag{
       Name: "lang",
       Value: "english",
       Usage: "language for the greeting",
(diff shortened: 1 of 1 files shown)


rliebz:
From what I can tell, our docs are currently accurate—I have a module using the latest release of `v1` (`1.22.4`), and flags do not have pointer receivers. You can verify that that's the case here: https://github.com/urfave/cli/blob/v1.22.4/flag_bool.go

If you're not using modules, be aware that pulling the latest from `master` will get `v2` code. We don't officially support any installation method other than modules, but if you do get it working from vendoring or a third party package management tool, you'll almost certainly need to pin your version explicitly. 

If you are using go modules, and you continue to have this problem, please open an issue—I'm less interested in having an official ticket, and more interested in particular in the build environment that causes this problem, and I think our issue template does a pretty good job of capturing that.

hoijui:
ahh.. thank you @rliebz !
yeah I am sure it is as you said, that I am getting v2 (because I did not specify the version)
I am not a frequent Go dev.

PR #1118, opened 2020-04-27
Add support for comma separated IntSliceFlag and Int64SliceFlag

### LINKED ISSUE #1097: IntSliceFlag failes at parsing -flag=1,2,3
## my urfave/cli version is

_**2.2.0**_

## Checklist

## Dependency Management

## Describe the bug
When using `IntSliceFlag` it complain with:
`Incorrect Usage: invalid value "1,2,3" for flag -zones: strconv.ParseInt: parsing "1,2,3": invalid syntax`
This works for `StringSliceFlag`.

## To reproduce
Add a `IntFlagSlice{Name="flag"}` and call it with `-flag=1,2,3`.

## Observed behavior
It gives the error described above.

## Expected behavior

I expect it to parse this correctly and then be able to get it with `cli.Context.IntSlice("flag")`.

```
go version go1.14 linux/amd64
```

zemzale:
## What type of PR is this?

- [x] bug
- [ ] cleanup  
- [ ] documentation
- [ ] feature

IntSliceFlag can't parse comma separated  like this `-flag 1,2,3`. 
This PR adds support for that by : 
- Checking for comma in `IntSliceFlag.Set` and `Int64SliceFlag.Set` 
- If comma exists split the value string and add each number to the underlying slice 

Fixes #1097 

My only knowledge of this project is from using it.
I came up with a pretty naive solution to this problem, so I would be happy to know if the change is implemented in the right place.

## Testing

Added unit tests : 
- `TestParseMultiIntSliceCommaSeperated` 
- `TestParseMultiInt64SliceCommaSeperated`

These tests wouldn't fail if the error of `Run` isn't checked.
All the other parsing tests do the checking in `Action` but these tests won't even get there if the value is not valid.

## Release Notes

```release-note
Add support for comma separated `IntSliceFlag` and `Int64SliceFlag`
```

### DIFF
--- a/flag_int64_slice.go
+++ b/flag_int64_slice.go
@@ -33,6 +33,18 @@ func (i *Int64Slice) Set(value string) error {
 		return nil
 	}
 
+	if strings.Contains(value, ",") {
+		values := strings.Split(value, ",")
+		for _, v := range values {
+			tmp, err := strconv.ParseInt(v, 0, 64)
+			if err != nil {
+				return err
+			}
+			i.slice = append(i.slice, tmp)
+		}
+		return nil
+	}
+
 	tmp, err := strconv.ParseInt(value, 0, 64)
 	if err != nil {
 		return err
--- a/flag_int64_slice.go
+++ b/flag_int64_slice.go
@@ -33,25 +33,13 @@ func (i *Int64Slice) Set(value string) error {
 		return nil
 	}
 
-	if strings.Contains(value, ",") {
-		values := strings.Split(value, ",")
-		for _, v := range values {
-			tmp, err := strconv.ParseInt(v, 0, 64)
-			if err != nil {
-				return err
-			}
-			i.slice = append(i.slice, tmp)
+	for _, v := range strings.Split(value, ",") {
+		tmp, err := strconv.ParseInt(v, 0, 64)
+		if err != nil {
+			return err
 		}
-		return nil
+		i.slice = append(i.slice, tmp)
 	}
-
-	tmp, err := strconv.ParseInt(value, 0, 64)
-	if err != nil {
-		return err
-	}
-
-	i.slice = append(i.slice, tmp)
-
 	return nil
 }
 
--- a/flag_int_slice.go
+++ b/flag_int_slice.go
@@ -44,25 +44,13 @@ func (i *IntSlice) Set(value string) error {
 		return nil
 	}
 
-	if strings.Contains(value, ",") {
-		values := strings.Split(value, ",")
-		for _, v := range values {
-			tmp, err := strconv.ParseInt(v, 0, 64)
-			if err != nil {
-				return err
-			}
-			i.slice = append(i.slice, int(tmp))
+	for _, v := range strings.Split(value, ",") {
+		tmp, err := strconv.ParseInt(v, 0, 64)
+		if err != nil {
+			return err
 		}
-		return nil
+		i.slice = append(i.slice, int(tmp))
 	}
-
-	tmp, err := strconv.ParseInt(value, 0, 64)
-	if err != nil {
-		return err
-	}
-
-	i.slice = append(i.slice, int(tmp))
-
 	return nil
 }
 
--- a/flag_int_slice.go
+++ b/flag_int_slice.go
@@ -44,6 +44,18 @@ func (i *IntSlice) Set(value string) error {
 		return nil
 	}
 
+	if strings.Contains(value, ",") {
+		values := strings.Split(value, ",")
+		for _, v := range values {
+			tmp, err := strconv.ParseInt(v, 0, 64)
+			if err != nil {
+				return err
+			}
+			i.slice = append(i.slice, int(tmp))
+		}
+		return nil
+	}
+
 	tmp, err := strconv.ParseInt(value, 0, 64)
 	if err != nil {
 		return err
--- a/flag_test.go
+++ b/flag_test.go
@@ -1196,6 +1196,28 @@ func TestParseMultiIntSlice(t *testing.T) {
 	}).Run([]string{"run", "-s", "10", "-s", "20"})
 }
 
+func TestParseMultiIntSliceCommaSeperated(t *testing.T) {
+	//Need to check for error since the parsing fails before it gets into the Action
+	//Then this test will fail without being noticed
+	err := (&App{
+		Flags: []Flag{
+			&IntSliceFlag{Name: "serve", Aliases: []string{"s"}, Value: NewIntSlice()},
+		},
+		Action: func(ctx *Context) error {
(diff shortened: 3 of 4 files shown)


rliebz:
@lynncyrin The code currently errors on that input. Consider the following program:

```go
package main

import (
	"fmt"
	"log"
	"os"

	"github.com/urfave/cli/v2"
)

func main() {
	app := cli.NewApp()
	app.Flags = []cli.Flag{
		&cli.IntSliceFlag{
			Name: "val",
		},
	}

		fmt.Println(c.IntSlice("val"))
		return nil
	}

		log.Fatal(err)
	}
}
```

With `go run main.go -val 5000` we get `[5000]`. If we pass `5,000`, we get:

```
Incorrect Usage. invalid value "5,000" for flag -val: strconv.ParseInt: parsing "5,000": invalid syntax
```

That means we can add support for this feature without breaking backward compatibility.

I think that's a good call out that it allows potentially ambiguous inputs, but I'm okay with it in this case. We don't support that style of input in any other context (e.g., `IntFlag`), and it's consistent with how we handle slices of strings.

flag_int64_slice.go
```diff
@@ -33,6 +33,18 @@ func (i *Int64Slice) Set(value string) error {
 		return nil
 	}
 
+	if strings.Contains(value, ",") {
```
rliebz, line 36: Nit: I think the ergonomics around `strings.Contains` means we can skip the conditional and always range.

- `strings.Split("1", ",")` -> `[1]`
- `strings.Split("1,2", ",")` -> `[1 2]`

Meaning we can drop the `if` statement, outdent everything in there, and remove the logic for treating `value` as a non-list value.

flag_int_slice.go
```diff
@@ -44,6 +44,18 @@ func (i *IntSlice) Set(value string) error {
 		return nil
 	}
 
+	if strings.Contains(value, ",") {
```
rliebz, line 47: Nit: Same note as above—I don't believe we have to special case this.

zemzale:
> Should we be fixing float64 slices as well? From my initial look at the code, I think that would just be an identical change in flag_float64_slice.go, so that probably belongs in this PR.

Yeah, I can do that.  
Also searching through the tests, I can see that there are no tests `TestParseMultiFloat64Slice` like there are for `IntSlice` so I am gonna add those too while I am at it.

flag_int64_slice.go
zemzale: Fixed this

flag_int_slice.go
zemzale: Fixed this too

flag_test.go
```diff
@@ -1427,6 +1471,79 @@ func TestParseMultiFloat64FromEnvCascade(t *testing.T) {
...
 }
 
+func TestParseMultiFloat64Slice(t *testing.T) {
+	_ = (&App{
+		Flags: []Flag{
+			&Float64SliceFlag{Name: "intervals", Aliases: []string{"i"}, Value: NewFloat64Slice()},
+		},
+		Action: func(ctx *Context) error {
+			if !reflect.DeepEqual(ctx.Float64Slice("intervals"), []float64{0.1, -10.5}) {
+				t.Errorf("main name not set from env")
```
zemzale, line 1481: While copy pasting forgot to change the error messages. This might be a bit misleading so going to fix this

saschagrunert:
I think the string slice flag does not support such behavior, right? That's why I had to add a workaround in a urfave/cli consuming project. I know that comma separation could be intended even for string slice flags, but maybe we would like to add an option for such cases as well.

zemzale:
You are right that it doesn't support comma separated string slices from command line. 
If you pass string slice flag `-s foo,bar` the value will be `["foo,bar"]` not `["foo", "bar"]`.

saschagrunert:
> You are right that it doesn't support comma separated string slices from command line.
> If you pass string slice flag `-s foo,bar` the value will be `["foo,bar"]` not `["foo", "bar"]`.

We could add an option which enables that kind of parsing :thinking:

zemzale:
It could be done but I have no idea how I should implement that since this is the first time me touching this codebase aside from using it in my own applications :sweat_smile:

rliebz:
Sorry, I was under the impression that we were adding this to make it consistent with the behavior for strings.

The original issue (#1097) specifically cites:
> This works for StringSliceFlag.

If that's _not_ the behavior for string slices, then is #1097 really a bug, or just a misunderstanding of how the library was designed to work?

If we don't yet support this behavior for any sort of slice flag, I would actually caution against adding it at all. We can't actually make string slices behave this way at this point without breaking backward compatibility. Although it's not trivial, it is absolutely possible to get this behavior for very customized CLIs by implementing the `cli.Flag` interface manually.

So my recommendation here unfortunately would be not to add in this behavior right now. 

If the change is desired, the process should be to open a feature request rather than a bug ticket, weigh whether we actually want this behavior against the cost of making a breaking change for `StringSliceFlag`, and then should we accept it, make sure the PR that implements it covers all four of the flag slice types together.

saschagrunert:
I agree with @rliebz and I guess we should not automate any splitting. I think it would make sense in general to be added as optional feature.

zemzale:
> If that's not the behavior for string slices, then is #1097 really a bug, or just a misunderstanding of how the library was designed to work?

It must be a misunderstanding. Following programm :
```go 
package main

import (
        "fmt"
        "log"
        "os"

        "github.com/micro/cli"
)

func main() {
        app := cli.NewApp()
        app.Flags = []cli.Flag{
                &cli.StringSliceFlag{
                        Name: "val",
                },
        }

                fmt.Println(c.StringSlice("val"))
                return nil
        }

                log.Fatal(err)
        }
}
```
When run `go run main.go -val foo,bar` would return `[foo,bar]` instead of `[foo bar]`
It doesn't fail with a parse error like in the issue, it just reads the the given string as one value.

So this really is not a bug, but I would be glad to open up this as a feature request and implement it.

lynncyrin:
Here's another example that builds off the one above

```go
# main.go

package main

import (
	"fmt"
	"log"
	"os"

	"github.com/urfave/cli/v2"
)

func main() {
	app := cli.NewApp()
	app.Flags = []cli.Flag{
		&cli.StringSliceFlag{
			Name: "val",
		},
	}

		val := c.StringSlice("val")
		fmt.Println(val)
		fmt.Println(len(val))
		return nil
	}

		log.Fatal(err)
	}
}
```

```bash
$ go run main.go -val cat,hat
[cat,hat]
1
```

So yep, looks like `StringSliceFlag` doesn't split on `,`

zemzale:
Then I am closing this PR and gonna create a feature request for this since there are people that are interested in this kinda functionality

PR #1044, opened 2020-01-17
Add word-wrap support based on terminal size

saschagrunert:
Highly experimental and still WIP.

Refers to #287

Issue #287, opened 2015-10-20
Help text should automatically wrap based on term width


nikkolasg:
When we got a long text description in Usage, it should automatically wrap and make a nice display.
A bit like what "man" does, it automatically rewraps the text so it fits nicely in the current terminal session.

wikiwi:
+1

syossan27:
+1

immesys:
+1

hickeng:
and should be able to handle `\n` in the various command and flag descriptions, indenting appropriately.

JensRantil:
...or simply wrap at 80 characters, which is considered a common default and also adheres to [general readability best-practises](https://baymard.com/blog/line-length-readability).

lynncyrin:
This feature is now `in review` - @urfave/cli please add a 👍 or 👎 to the top post if you're in favor or against this feature being added!

saschagrunert:
@lynncyrin we should move forward with this one. Wrapping based on the term width seems a sane default. :)

lynncyrin:
👀 this is now ready for anyone to work on!

xordspar0:
> ...or simply wrap at 80 characters, which is considered a common default and also adheres to [general readability best-practises](https://baymard.com/blog/line-length-readability).

I agree that an 80 character limit would be easier to read. It would also be easier to implement. That said, I don't want to bikeshed this issue. Whoever actually does the work to implement this will have my vote.

mostynb:
I will try to make #1119 implement a basic version of this feature.

mostynb:
Bump.

mostynb:
Bump.

mostynb:
Bump.

dearchap:
@mostynb Are you up to making a PR for this ?

dearchap:
Duplicate of #674

mostynb:
I added line-wrapping support in #287, but the terminal's width detection is not automatic (to avoid adding new dependencies/adding binary size). Here's an example of how client code can specify the line-wrap length: https://github.com/urfave/cli/pull/1119/files#diff-4d3f36d62cad94424544bb84fc93b33b73ffe3491a4ddca81a7f825539afc19cR1161-R1169

PR #1119, opened 2020-04-28
Add word-wrap support, with wrap length provided by the user

mostynb:
## What type of PR is this?

- feature

The help text is difficult to read when the lines are long. This patch attempts to wrap long lines, if the user provides a terminal width to wrap at.

Updated version of #1044.

Refers to #287.

## Testing

Added a unit test (which can be expanded).

## Release Notes

```release-note
The help text is now wrapped to fit the terminal window.
```

### DIFF
--- a/help.go
+++ b/help.go
@@ -64,6 +64,11 @@ var HelpPrinter helpPrinter = printHelp
 // HelpPrinterCustom is a function that writes the help output. It is used as
 // the default implementation of HelpPrinter, and may be called directly if
 // the ExtraInfo field is set on an App.
+//
+// In the default implementation, if the customFuncs argument contains a
+// "wrapAt" key, which is a function which takes no arguments and returns
+// an int, this int value will be used to produce a "wrap" function used
+// by the default template to wrap long lines.
 var HelpPrinterCustom helpPrinterCustom = printHelpCustom
 
 // VersionPrinter prints the version for the App
@@ -286,12 +291,29 @@ func ShowCommandCompletions(ctx *Context, command string) {
 // The customFuncs map will be combined with a default template.FuncMap to
 // allow using arbitrary functions in template rendering.
 func printHelpCustom(out io.Writer, templ string, data interface{}, customFuncs map[string]interface{}) {
+
+	const maxLineLength = 10000
+
 	funcMap := template.FuncMap{
 		"join":    strings.Join,
 		"indent":  indent,
 		"nindent": nindent,
 		"trim":    strings.TrimSpace,
+		"wrap":    func(input string, offset int) string { return wrap(input, offset, maxLineLength) },
+		"offset":  offset,
 	}
+
+	if customFuncs["wrapAt"] != nil {
+		if wa, ok := customFuncs["wrapAt"]; ok {
+			if waf, ok := wa.(func() int); ok {
+				wrapAt := waf()
+				customFuncs["wrap"] = func(input string, offset int) string {
+					return wrap(input, offset, wrapAt)
+				}
+			}
+		}
+	}
+
 	for key, value := range customFuncs {
 		funcMap[key] = value
 	}
@@ -401,4 +423,56 @@ func indent(spaces int, v string) string {
 
 func nindent(spaces int, v string) string {
 	return "\n" + indent(spaces, v)
+}
+
+func wrap(input string, offset int, wrapAt int) string {
+	var sb strings.Builder
+
+	lines := strings.Split(input, "\n")
+
+	padding := strings.Repeat(" ", offset)
+
+	for i, line := range lines {
+		if i != 0 {
+			sb.WriteString(padding)
+		}
+
+		sb.WriteString(wrapLine(line, offset, wrapAt, padding))
+
+		if i != len(lines)-1 {
+			sb.WriteString("\n")
+		}
+	}
+
+	return sb.String()
+}
+
+func wrapLine(input string, offset int, wrapAt int, padding string) string {
+	if wrapAt <= offset || len(input) <= wrapAt-offset {
+		return input
+	}
+
+	lineWidth := wrapAt - offset
+	words := strings.Fields(input)
+	if len(words) == 0 {
+		return input
+	}
+
+	wrapped := words[0]
+	spaceLeft := lineWidth - len(wrapped)
+	for _, word := range words[1:] {
+		if len(word)+1 > spaceLeft {
+			wrapped += "\n" + padding + word
+			spaceLeft = lineWidth - len(word)
+		} else {
+			wrapped += " " + word
+			spaceLeft -= 1 + len(word)
+		}
+	}
+
+	return wrapped
+}
+
+func offset(input string, fixed int) int {
+	return len(input) + fixed
 }
--- a/template.go
+++ b/template.go
@@ -4,16 +4,16 @@ package cli
 // cli.go uses text/template to render templates. You can
 // render custom help text by setting this variable.
 var AppHelpTemplate = `NAME:
-   {{.Name}}{{if .Usage}} - {{.Usage}}{{end}}
+   {{$v := offset .Name 6}}{{wrap .Name 3}}{{if .Usage}} - {{wrap .Usage $v}}{{end}}
 
 USAGE:
-   {{if .UsageText}}{{.UsageText | nindent 3 | trim}}{{else}}{{.HelpName}} {{if .VisibleFlags}}[global options]{{end}}{{if .Commands}} command [command options]{{end}} {{if .ArgsUsage}}{{.ArgsUsage}}{{else}}[arguments...]{{end}}{{end}}{{if .Version}}{{if not .HideVersion}}
+   {{if .UsageText}}{{wrap .UsageText 3}}{{else}}{{.HelpName}} {{if .VisibleFlags}}[global options]{{end}}{{if .Commands}} command [command options]{{end}} {{if .ArgsUsage}}{{.ArgsUsage}}{{else}}[arguments...]{{end}}{{end}}{{if .Version}}{{if not .HideVersion}}
 
 VERSION:
    {{.Version}}{{end}}{{end}}{{if .Description}}
 
 DESCRIPTION:
-   {{.Description | nindent 3 | trim}}{{end}}{{if len .Authors}}
+   {{wrap .Description 3}}{{end}}{{if len .Authors}}
 
 AUTHOR{{with $length := len .Authors}}{{if ne 1 $length}}S{{end}}{{end}}:
    {{range $index, $author := .Authors}}{{if $index}}
@@ -31,26 +31,26 @@ GLOBAL OPTIONS:{{range .VisibleFlagCategories}}
 
 GLOBAL OPTIONS:
    {{range $index, $option := .VisibleFlags}}{{if $index}}
-   {{end}}{{$option}}{{end}}{{end}}{{end}}{{if .Copyright}}
+   {{end}}{{wrap $option.String 6}}{{end}}{{end}}{{end}}{{if .Copyright}}
 
 COPYRIGHT:
-   {{.Copyright}}{{end}}
+   {{wrap .Copyright 3}}{{end}}
 `
 
 // CommandHelpTemplate is the text template for the command help topic.
 // cli.go uses text/template to render templates. You can
 // render custom help text by setting this variable.
 var CommandHelpTemplate = `NAME:
-   {{.HelpName}} - {{.Usage}}
+   {{$v := offset .HelpName 6}}{{wrap .HelpName 3}}{{if .Usage}} - {{wrap .Usage $v}}{{end}}
 
 USAGE:
-   {{if .UsageText}}{{.UsageText | nindent 3 | trim}}{{else}}{{.HelpName}}{{if .VisibleFlags}} [command options]{{end}} {{if .ArgsUsage}}{{.ArgsUsage}}{{else}}[arguments...]{{end}}{{end}}{{if .Category}}
+   {{if .UsageText}}{{wrap .UsageText 3}}{{else}}{{.HelpName}}{{if .VisibleFlags}} [command options]{{end}} {{if .ArgsUsage}}{{.ArgsUsage}}{{else}}[arguments...]{{end}}{{end}}{{if .Category}}
 
 CATEGORY:
    {{.Category}}{{end}}{{if .Description}}
 
 DESCRIPTION:
-   {{.Description | nindent 3 | trim}}{{end}}{{if .VisibleFlagCategories}}
+   {{wrap .Description 3}}{{end}}{{if .VisibleFlagCategories}}
 
 OPTIONS:{{range .VisibleFlagCategories}}
    {{if .Name}}{{.Name}}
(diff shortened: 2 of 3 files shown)


internal/build/build.go
```diff
@@ -194,7 +194,7 @@ func checkBinarySizeActionFunc(c *cli.Context) (err error) {
 		helloSourceFilePath  = "./internal/example-hello-world/example-hello-world.go"
 		helloBuiltFilePath   = "./internal/example-hello-world/built-example"
 		desiredMinBinarySize = 2.0
-		desiredMaxBinarySize = 2.1
+		desiredMaxBinarySize = 2.2
```
lynncyrin, line 197: The binary size increasing and the new dependency (`golang.org/x/crypto`) concern me, can you rewrite this PR to remove those two changes?

mostynb: I can experiment with some platform-specific code, and see what is possible. Would it be OK if this feature only works on linux/mac/windows (for a small subset of each that I can easily test)?

lynncyrin: I needs to work on all of the platforms that we test on CI, and you don't need to support anything beyond that

termwidth_misc.go
```diff
@@ -0,0 +1,21 @@
+// +build !aix
+// +build !darwin
+// +build !dragonfly
+// +build !freebsd
+// +build !linux,!appengine
```
mostynb, line 5: I need to double-check this.

internal/build/build.go
mostynb: Even using the low-level `unix.IoctlGetWinsize`/`windows.GetConsoleScreenBufferInfo` calls, this increases the binary size by 0.1M. Though perhaps golang.org/x/sys/unix and golang.org/x/sys/windows are more likely to already be used by client programs than golang.org/x/crypto.

The only other option I can think of to not raise the binary size by this much would be for clients to figure out the terminal size themselves and provide it somehow (maybe set it as a new field in the App struct). What do you think?

lynncyrin: There's 2 options here that stand out for me:

1. import the code for the `unix.IoctlGetWinsize` and `windows.GetConsoleScreenBufferInfo` functions so that we aren't importing the whole package
2. allow people to manually set the terminal size, like you mention

Either way, we definitely want to avoid increasing the binary size for this feature. At least, not in a pre-v3 world where can't make this feature opt-in. (see https://github.com/urfave/cli/issues/833#issuecomment-579869829 for context)

lynncyrin: I would also say that it would probably be un-controversial to let people set the terminal size, and have a default terminal size of `80` as described here => https://github.com/urfave/cli/issues/287#issuecomment-320581746

mostynb: I would rather not copy+paste all that subtle syscall code, so I think we should go for (2) - it also makes this feature testable.

mostynb:
How does this look?

* Is it OK to add a new field to `helpPrinter`?
* The template probably needs tweaking to support wrapping more fields, but I don't know offhand which other fields to test.
* This makes no attempt to break long words.

go.mod
```diff
@@ -4,6 +4,8 @@ go 1.11
 
 require (
 	github.com/BurntSushi/toml v0.3.1
+	github.com/andreyvit/diff v0.0.0-20170406064948-c7f18ee00883
 	github.com/cpuguy83/go-md2man/v2 v2.0.0-20190314233015-f79a8a8ca69d
-	gopkg.in/yaml.v2 v2.2.2
+	github.com/sergi/go-diff v1.1.0 // indirect
```
lynncyrin, line 9: I hate to be a broken record, but please remove this change 🙏 I would rather this PR not add new requirements, even if they are only for testing.

mostynb: No problem- done.

lynncyrin:
I see you just force pushed, so I'm not sure what changed? Can you re-request my review when this is ready for eyes again, and please not force push until after the PR is ready to merge?

mostynb:
> I see you just force pushed, so I'm not sure what changed? Can you re-request my review when this is ready for eyes again, and please not force push until after the PR is ready to merge?

Sorry about that- I squashed the commit in an attempt to make it easy to verify that no new modules were added.

I have re-pushed the branch with [the original commit](https://github.com/urfave/cli/pull/1119/commits/be03ef5e9b4f9a8300e2e3ed5dd972c7840cbec0) followed by a [fixup that removes the newly added module](https://github.com/urfave/cli/pull/1119/commits/7c8032c545db2da51ab2ec4d85f567e5b71009ce).

lynncyrin:
I'll try to give this a look soon!

ericmillin:
@lynncyrin Would love to see this merged.

lynncyrin:
Oops 😄

template.go
```diff
@@ -4,16 +4,16 @@ package cli
 // cli.go uses text/template to render templates. You can
 // render custom help text by setting this variable.
 var AppHelpTemplate = `NAME:
-   {{.Name}}{{if .Usage}} - {{.Usage}}{{end}}
+   {{$v := offset .Name 6}}{{wrap .Name 3}}{{if .Usage}} - {{wrap .Usage $v}}{{end}}
 
 USAGE:
-   {{if .UsageText}}{{.UsageText}}{{else}}{{.HelpName}} {{if .VisibleFlags}}[global options]{{end}}{{if .Commands}} command [command options]{{end}} {{if .ArgsUsage}}{{.ArgsUsage}}{{else}}[arguments...]{{end}}{{end}}{{if .Version}}{{if not .HideVersion}}
+   {{if .UsageText}}{{wrap .UsageText 3}}{{else}}{{.HelpName}} {{if .VisibleFlags}}[global options]{{end}}{{if .Commands}} command [command options]{{end}} {{if .ArgsUsage}}{{.ArgsUsage}}{{else}}[arguments...]{{end}}{{end}}{{if .Version}}{{if not .HideVersion}}
```
lynncyrin, line 10: ❓ &nbsp; _(blocking question)_ what do the 6 and 3 here represent???

coilysiren (dismissed):
I have 1 blocking question, since I'm not following the `template.go` changes @mostynb

template.go
mostynb: They are arguments to the functions `wap` and `offset`, in the funcMap applied to the template.

`wrap` takes a string and an offset to indent wrapped lines, and returns a wrapped string (the second line onwards are all indented by the offset. `3` is the indent level used in the template.

`offset` returns the length of a string plus a given number, and is used to calculate an offset for the description field of the app name. I hope my terrible ascii art can explain where the `6` comes from (the sum of the two hard-coded `3`'s in this part of the template):

```
NAME:
   <appname> - <description blah blah blah...
               we want wrapped lines to line
               up here>
^-^^-------^^-^
 |    |      |
 |    |      another hardcoded 3
 |    |
 |   length of app name
 |
hardcoded 3

```

xordspar0:
Is this ready to be merged?

mostynb:
> Is this ready to be merged?

It's stalled, waiting for review.

help.go
```diff
@@ -42,10 +42,10 @@ var helpSubcommand = &Command{
 }
 
 // Prints help for the App or Command
-type helpPrinter func(w io.Writer, templ string, data interface{})
+type helpPrinter func(w io.Writer, templ string, data interface{}, wrapAt int)
```
rliebz, line 45: It's not entirely clear from what's exported here, but making a change to the signature of these functions is unfortunately a breaking change, as the exported variables `HelpPrinter` and `HelpPrinterCustom` both use these signatures. [One of the CLIs I've written](https://github.com/rliebz/tusk/blob/e58ee3109f2f99972a35a054329787aa411c6357/appcli/help.go#L19-L22), for example, would no longer compile if this change were merged in.

I love the feature idea, but making a breaking change to implement it is a non-starter.

rliebz (changes requested):

help.go
mostynb: Pushed an unpolished fixup which drops the `wrapAt` parameter, and instead allows the client to specify a "wrap" function to `HelpPrinterCustom`. If this seems reasonable I can add documentation, squash and rebase and perform some general cleanup etc.

asahasrabuddhe: This seems to be going in the correct direction to me

template.go
```diff
@@ -4,16 +4,16 @@ package cli
 // cli.go uses text/template to render templates. You can
 // render custom help text by setting this variable.
 var AppHelpTemplate = `NAME:
-   {{.Name}}{{if .Usage}} - {{.Usage}}{{end}}
+   {{$v := offset .Name 6}}{{wrap .Name 3}}{{if .Usage}} - {{wrap .Usage $v}}{{end}}
```
rliebz, line 7: Looks like we're wrapping most of the fields in the template, but not all of the fields. This PR also omits wrapping for command help and sub-command help.

Is there a reason _NOT_ to expand wrapping to all sections of all templates?

template.go
rliebz: I'm satisfied with this answer.

help.go
```diff
@@ -366,3 +368,53 @@ func checkCommandCompletions(c *Context, name string) bool {
...
+		} else {
+			wrapped += " " + word
+			spaceLeft -= 1 + len(word)
+		}
+	}
+
+	return wrapped
+}
+
+func offset(input string, fixed int) int {
```
rliebz, line 418: Probably not a big deal, but this function name is shadowed a few times in the other functions being introduced.

rliebz (commented):
Thanks for the contribution! 

This is a sweet feature, just had a few minor comments.

help.go
```diff
@@ -366,3 +368,53 @@ func checkCommandCompletions(c *Context, name string) bool {
...
+	return sb.String()
+}
+
+func wrapLine(input string, offset int, wrapAt int) string {
+	if wrapAt <= offset || len(input) <= wrapAt-offset {
+		return input
+	}
+
+	lineWidth := wrapAt - offset
+	words := strings.Fields(strings.TrimSpace(input))
```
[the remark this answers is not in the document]
mostynb: Thanks- removed.

help.go
```diff
@@ -366,3 +368,53 @@ func checkCommandCompletions(c *Context, name string) bool {
...
+	if len(words) == 0 {
+		return input
+	}
+
+	wrapped := words[0]
+	spaceLeft := lineWidth - len(wrapped)
+	for _, word := range words[1:] {
+		if len(word)+1 > spaceLeft {
+			wrapped += "\n" + strings.Repeat(" ", offset) + word
```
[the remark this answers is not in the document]
mostynb: I lifted this up into the calling function, since the same value was also used there multiple times.

template.go
mostynb: Mostly because I don't know all the variations that need to be tested. I suspect there will be some fields that aren't worth trying to wrap because they're only a problem in unusual cases, but command and sub-command help are worth supporting.

I will try to expand this a bit.

help.go
mostynb: If I understand correctly it should be safe, because there's only one `offset` function, and the other `offset`'s are `int` variables, but maybe it could be made less confusing. I could rename all the `offset` variables to `off` if you prefer?

rliebz: The current iteration doesn't seem to cause any problems, and doesn't seem to set off any linters I know of.

I think it should be alright as is, and renaming `offset` should be fine as well. Up to you.

mostynb:
I was looking into the conflicting files, and noticed that #1175 has landed, which implements a subset of the feature here.

Force-pushed a squashed version of this with two fixes appended, which makes `TestWrappedHelp` also test user specifed indenting, and another change which makes `Test_ShowAppHelp_MultiLineDescription` on master pass if I revert the code changes from be9c0378066dc5baf84a56005f90317cca6ecd84 and then apply this PR.

How should we proceed?

rliebz:
> I was looking into the conflicting files, and noticed that #1175 has landed, which implements a subset of the feature here.
> 
> Force-pushed a squashed version of this with two fixes appended, which makes `TestWrappedHelp` also test user specifed indenting, and another change which makes `Test_ShowAppHelp_MultiLineDescription` on master pass if I revert the code changes from [be9c037](https://github.com/urfave/cli/commit/be9c0378066dc5baf84a56005f90317cca6ecd84) and then apply this PR.
> 
> How should we proceed?

Those changes were contributed because someone needed them, and then tagged in a release. I think reverting them is off the table.

If you're still interested in getting this change in, you'll need to resolve conflicts. I'm totally fine if we change the template—I think folks using the default template are just signing up for whatever we decide makes sense. That said, the new template functions (indent, nindent, and trim) are now part of the public API, so those are something we shouldn't be removing.

mostynb:
> Those changes were contributed because someone needed them, and then tagged in a release. I think reverting them is off the table.
> 
> If you're still interested in getting this change in, you'll need to resolve conflicts. I'm totally fine if we change the template—I think folks using the default template are just signing up for whatever we decide makes sense. That said, the new template functions (indent, nindent, and trim) are now part of the public API, so those are something we shouldn't be removing.

Pushed a rebased and squashed version, which undoes the template changes from be9c0378066dc5baf84a56005f90317cca6ecd84, but leaves the `indent`, `nindent` and `trim` functions in the function map for anyone who started using them with a custom template.

mostynb:
Updated the template and added some tests to wrap commands and subcommands. There are probably other items that we can improve, but I think this is a good start.

mostynb:
How does this look now?

mostynb:
Ping.

meatballhat:
@mostynb Sorry for the delay!

meatballhat:
@mostynb Are you up for resolving the conflicts and moving this work forward? No worries if not. I'm happy to pick it up ❤️

mostynb:
I'll take a look and let you know if I run into trouble.

mostynb:
@meatballhat: the failing test also fails for me locally on the tip of main. The CI failure is also visible in this typo fix PR: #1402

meatballhat:
@mostynb this looks like the problem of the docs tests running against latest release instead of current working tree :weary:

bumping the priority of that work now...

meatballhat:
@mostynb Attempted fix for the test-docs step at #1403

mostynb:
@meatballhat: CI is green after rebasing on main.

help.go
```diff
@@ -286,12 +291,29 @@ func ShowCommandCompletions(ctx *Context, command string) {
...
 	funcMap := template.FuncMap{
 		"join":    strings.Join,
 		"indent":  indent,
 		"nindent": nindent,
 		"trim":    strings.TrimSpace,
+		"wrap":    func(input string, offset int) string { return wrap(input, offset, maxLineLength) },
+		"offset":  offset,
+	}
+
+	if customFuncs["wrapAt"] != nil {
```
dearchap, line 306: I think it is better to define a type for this function signature.

mostynb: I'm not quite sure what you mean- are you suggesting that we export a `WrapAtFunc` type alias, and then only use `customFuncs["wrapAt"]` if it is created as a `WrapAtFunc` explicitly?

mostynb: Gentle ping on this :)

dearchap: Yes correct

mostynb: This requires some boilerplate for the caller, but I'm not sure how it makes things better. How does this help? I couldn't find any examples of this pattern in a quick web search, but I'm not sure what to call this so maybe I'm missing something obvious.

meatballhat: fwiw I'm happy to defer resolution of this to a later PR, so I'm going to move forward with merging the current work 👍🏼

meatballhat (approved):

tonglil:
Is there any docs for this?

mostynb:
> Is there any docs for this?

Perhaps it could be more discoverable, but you need to set HelpPrinterCustom or HelpPrinter to add a `wrapAt` function to the funcMap. eg:
https://github.com/urfave/cli/pull/1119/files#diff-4d3f36d62cad94424544bb84fc93b33b73ffe3491a4ddca81a7f825539afc19cR1161-R1169

PR #1120, opened 2020-04-28
Improve README for newer go versions

vnepveu:
## What type of PR is this?

- [ ] bug
- [ ] cleanup  
- [x] documentation
- [ ] feature

* Improve installation instructions for newer go versions.

 It confused me that the installation was working but the module was not found, 
changing the GO__MODULE did the trick

## Release Notes

```release-note
Add installations instructions for newer go version
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,18 @@ Using this package requires a working Go environment. [See the install instructi
 
 Go Modules are required when using this package. [See the go blog guide on using Go Modules](https://blog.golang.org/using-go-modules).
 
-### Using `v2` releases
+### Golang version 1.14 
+
+- In the files using `cli`, use this import statement :
+```go
+...
+import (
+  "github.com/urfave/cli/v2" // imports as package "cli"
+)
+...
+```
+
+### Using `v2` releases with older go versions
 
 ```
 $ GO111MODULE=on go get github.com/urfave/cli/v2
@@ -41,7 +52,7 @@ import (
 ...
 ```
 
-### Using `v1` releases
+### Using `v1` releases with older go versions
 
 ```
 $ GO111MODULE=on go get github.com/urfave/cli
--- a/README.md
+++ b/README.md
@@ -27,9 +27,11 @@ Using this package requires a working Go environment. [See the install instructi
 
 Go Modules are required when using this package. [See the go blog guide on using Go Modules](https://blog.golang.org/using-go-modules).
 
-### Golang version 1.14 
+### Using `v2` releases
 
-- In the files using `cli`, use this import statement :
+To use `cli` in your project :
+
+- If the file `example.go` is using `cli`, use this import statement :
 ```go
 ...
 import (
@@ -37,26 +39,13 @@ import (
 )
 ...
 ```
-
-### Using `v2` releases with older go versions
-
-```
-$ GO111MODULE=on go get github.com/urfave/cli/v2
-```
-
-```go
-...
-import (
-  "github.com/urfave/cli/v2" // imports as package "cli"
-)
-...
+- Then, when running `go run example.go` for the first time, this should output :
+```bash
+go: finding module for package github.com/urfave/cli/v2
+go: found github.com/urfave/cli/v2 in github.com/urfave/cli/v2 v2.2.0
 ```
 
-### Using `v1` releases with older go versions
-
-```
-$ GO111MODULE=on go get github.com/urfave/cli
-```
+### Using `v1` releases 
 
 ```go
 ...


vnepveu:
Actually this GO__MODULE isn't documented anywhere except for GO111MODULE, so my modification is not supposed to work

vnepveu:
Made the instructions clearer for usage with go 1.14, as it was potentially confusing.

README.md
```diff
@@ -27,11 +27,46 @@ Using this package requires a working Go environment. [See the install instructi
 
 Go Modules are required when using this package. [See the go blog guide on using Go Modules](https://blog.golang.org/using-go-modules).
 
-### Using `v2` releases
+### Golang version 1.14 
+In Go 1.14, modules were made "production-ready" (see [wiki](https:/github.com/golang/go/wiki/Modules#modules)).
+Dependencies for projects were streamlined : there is no need to install
```
saschagrunert, line 32:
```suggestion
Dependencies for projects were streamlined and there is no need to manually install
```

README.md
```diff
@@ -27,11 +27,46 @@ Using this package requires a working Go environment. [See the install instructi
...
+(`example.com/user/projectName` is your module name)
+- In the files using `cli`, use this import statement :
+```go
+...
+import (
+  "github.com/urfave/cli/v2" // imports as package "cli"
+)
+...
+```
+- Then, install your project module :
```
saschagrunert, line 50:
```suggestion
Then, install your project module :
```

README.md
```diff
@@ -27,11 +27,46 @@ Using this package requires a working Go environment. [See the install instructi
...
+```bash
+go install example.com/user/projectName
+```
+(where `example.com/user/projectName` is your module name)
+- This should output :
```
saschagrunert, line 55:
```suggestion
This should output:
```

README.md
```diff
@@ -27,11 +27,46 @@ Using this package requires a working Go environment. [See the install instructi
...
+```bash
+go: finding module for package github.com/urfave/cli/v2
+go: found github.com/urfave/cli/v2 in github.com/urfave/cli/v2 v2.2.0
+```
+- The module has been installed and is ready to use anywhere on your system
```
saschagrunert, line 60:
```suggestion
The module has been installed and is ready to use anywhere on your system
```

README.md
```diff
@@ -27,7 +27,42 @@ Using this package requires a working Go environment. [See the install instructi
 
 Go Modules are required when using this package. [See the go blog guide on using Go Modules](https://blog.golang.org/using-go-modules).
 
-### Using `v2` releases
+### Golang version 1.14
```
rliebz, line 30: Whether or not people are using 1.14, this library only officially supports modules, which is explicitly stated already.

README.md
```diff
@@ -27,7 +27,42 @@ Using this package requires a working Go environment. [See the install instructi
...
+```bash
+go: finding module for package github.com/urfave/cli/v2
+go: found github.com/urfave/cli/v2 in github.com/urfave/cli/v2 v2.2.0
+```
+The module has been installed and is ready to use anywhere on your system
+
+For troubleshooting, this [golang doc](https://golang.org/doc/code.html) is useful
+
+
+### Using `v2` releases with older go versions
```
rliebz, line 65: This statement is not accurate. The instructions are not different between 1.14 and earlier versions. 

Turns out the name `GO111MODULE` is a bit misleading—it forces modules to be enabled, even in Go versions that do not enable them by default. See `go help modules` for more information, or the release notes for 1.14 to see what's changed: https://golang.org/doc/go1.14

vnepveu: The thing is, if your are using Go 1.14 the instructions simply don't work, using the GO111MODULE is sort of a band-aid from what I understood, and not the proper way to install modules at least for 1.14. So even though my PR might be a little too documented , it's better than giving a temporary solution that does not even work for 1.14.

README.md
vnepveu: But the installation instructions are different since the 1.14 changed the way modules are handled

README.md
```diff
@@ -27,7 +27,42 @@ Using this package requires a working Go environment. [See the install instructi
...
 
-### Using `v2` releases
+### Golang version 1.14 
+In Go 1.14, modules were made "production-ready" (see [wiki](https:/github.com/golang/go/wiki/Modules#modules)).
+Dependencies for projects were streamlined and there is no need to manually install
+go packages anymore, the go toolchain handles it alone.
+
+To use `cli` in your project :
+
+- If your project is not a module yet, initialize your project with
```
vnepveu, line 37: Possibly this info is too much, but IMO the others are fine.

README.md
rliebz: From `go help modules`:

> For more fine-grained control, the go command continues to respect
a temporary environment variable, GO111MODULE, which can be set to one
of three string values: off, on, or auto (the default).
If GO111MODULE=on, then the go command requires the use of modules,
never consulting GOPATH. We refer to this as the command
being module-aware or running in "module-aware mode".
If GO111MODULE=off, then the go command never uses
module support. Instead it looks in vendor directories and GOPATH
to find dependencies; we now refer to this as "GOPATH mode."
If GO111MODULE=auto or is unset, then the go command enables or disables
module support based on the current directory.
Module support is enabled only when the current directory contains a
go.mod file or is below a directory containing a go.mod file.

From my understanding, setting `GO111MODULE=on` doesn't enable modules for a project, it just requires the use of modules when running the `go` command. If a project isn't using modules already, you're  right that it probably won't work. But I think we cover that with the line immediately before this section:

> Go Modules are required when using this package. [See the go blog guide on using Go Modules](https://blog.golang.org/using-go-modules).

That said, I'm not entirely clear on what worked in Go 1.13 that no longer works in Go 1.14. If you could walk me through your local setup or point me to the part of the 1.14 docs that cover the difference, that might be helpful.

README.md
lynncyrin: At a basic level, the installation instructions are still `GO111MODULE=on go get github.com/urfave/cli/v2`. Go 1.14 hasn't fundamentally changed that.

vnepveu:
My point with this PR was simply to give **working** installation instructions for anyone using go 1.14. 
The instructions are different for 1.14 : as I mentioned in my original patch, there is no need to "install" packages anymore (see example https://golang.org/doc/code.html#ImportingRemote). I am not an expert in the intricacies of go modules, but both the blog example and test on my machine show that there is no need to _go get_ anything, or use the GO111MODULE variable on the user-side.

rliebz:
@vnepveu I did some testing here—you're correct that `go get` is no longer necessary. I actually tested with go 1.11, which is the first version to support modules, and the command is not necessary in that environment either. The first call to `go run`/`go install`/etc. will pick up the modules automatically.

There's some difference in how the `GO111MODULE` defaults are between 1.11 and 1.14 whether you are on the `GOPATH `or not, but I think at this point, instructions that say modules are required are probably sufficient on that front. We link to the official docs, and the `go` CLI gives you nice warnings if you try to do the wrong thing:

```
go: modules disabled inside GOPATH/src by GO111MODULE=auto; see 'go help modules'
```

Since we only support modules to begin with, I'd be happy removing the `GO111MODULE=on go get github.com/urfave/cli/v2` line from the docs and leaving just the note about using modules and import statements. As far as I can tell, we shouldn't need to special case Go 1.14.

@lynncyrin Curious if you had any thoughts on that as well.

lynncyrin:
I would be interested in changing the top level readme to a very very small go source code example + an example `go run ...` line. Because yeah, in my experience you definitely don't need the `go get` anymore.

vnepveu:
I have provided what I consider to be a minimal example for installation. Please let me know if you want any chahge.

vnepveu:
@lynncyrin are you ok with the changes ?

lynncyrin:
Not to be dismissive - but I'm struggling to see how this PR represents a meaningful improvement upon the current world?

vnepveu:
As previously stated, I am just suggesting up-to-date installation instructions, because the current ones are **not** accurate. It does not necessitate a PR, it's just that I happened to make one. Feel free to just copy its content. Or leave the installation instructions as they are (incorrect).

PR #1126, opened 2020-05-01
fix: Print the long options of the required flags missing

oleorhagen:
## What type of PR is this?

_(REQUIRED)_

- [ x] bug
- [ ] cleanup  
- [ ] documentation
- [ ] feature

This fixes a formatting bug in the v1 required flags error message

Formatting error in the requireflags error message

I at least think this is a bug. The short option behaviour is a little odd.

## Testing

Only tested manually, see the commit message for output.

## Release Notes

```release-note
Changes the error message printing the long option for the required flags missing, and not the alias.
```

### DIFF
--- a/context.go
+++ b/context.go
@@ -324,11 +324,12 @@ func checkRequiredFlags(flags []Flag, context *Context) requiredFlagsErr {
 			var flagPresent bool
 			var flagName string
 			for _, key := range strings.Split(f.GetName(), ",") {
+				key = strings.TrimSpace(key)
 				if len(key) > 1 {
 					flagName = key
 				}
 
-				if context.IsSet(strings.TrimSpace(key)) {
+				if context.IsSet(key) {
 					flagPresent = true
 				}
 			}
--- a/context_test.go
+++ b/context_test.go
@@ -576,6 +576,14 @@ func TestCheckRequiredFlags(t *testing.T) {
 			},
 			parseInput: []string{"-n", "asd", "-n", "qwe"},
 		},
+		{
+			testCase: "required_flag_with_short_alias_not_printed_on_error",
+			expectedAnError: true,
+			expectedErrorContents: []string{"Required flag \"names\" not set"},
+			flags: []Flag{
+				StringSliceFlag{Name: "names, n", Required: true},
+			},
+		},
 	}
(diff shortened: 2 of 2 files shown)


oleorhagen:
Also, would it be an idea to export the `requiredFlagsError` type?

context.go
```diff
@@ -324,11 +324,12 @@ func checkRequiredFlags(flags []Flag, context *Context) requiredFlagsErr {
 			var flagPresent bool
 			var flagName string
 			for _, key := range strings.Split(f.GetName(), ",") {
-				if len(key) > 1 {
-					flagName = key
+				_key := strings.TrimSpace(key)
```
rliebz, line 327: Nit: Prefixing variables with `_` is not a common pattern in Go, and I don't think we have any variables named like these in this project.

If `_key` has no semantic difference from `key`, it's fine to just do:

```go
key = strings.TrimSpace(key)
```

Since Go isn't going to enforce immutability, and the first value is no longer needed.

rliebz:
As a note, if we end up accepting this, I believe it is relevant for the v2/master branch as well.

oleorhagen:
> As a note, if we end up accepting this, I believe it is relevant for the v2/master branch as well.

Acrually, I was not able to reproduce this on master. The code has changed a little though, so the explicit string splitting is not done anylonger, and the name is returned directly, so the situation seems to be fine there. I did cherry the test over just to be sure, and it passed. Do you still want a PR for it on master (the test only that is).

context_test.go
```diff
@@ -576,6 +576,14 @@ func TestCheckRequiredFlags(t *testing.T) {
 			},
 			parseInput: []string{"-n", "asd", "-n", "qwe"},
 		},
+		{
+			testCase: "required_flag_with_short_alias_not_printed_on_error",
+			expectedAnError: true,
+			expectedErrorContents: []string{"Required flag \"names\" not set"},
```
oleorhagen, line 582: @rliebz this is the expected functionality. The old code would yield:
```
Required flag \" n\" not set
```

rliebz:
@oleorhagen Thanks for checking v2 as well. If you wanted to submit a PR to add the test case to master, that would be awesome!

PR #1128, opened 2020-05-02
Add ShowDefaultValue option to prevent default to be printed for BoolFlag

### LINKED ISSUE #1122: v2 feature: Add option to prevent default to be printed for BoolFlag
## Checklist

Since v2, the help output looks cluttered when bool flags are used, as these always get their default value appended with no apparent way to disable. The large majority of flags in applications are non-negatable or act as a form of command, so there's usually little benefit in showing the default.

From a current app of mine:
```
GLOBAL OPTIONS:
   --verbose, -v  Increase verbosity (default: false)
   --extract, -e  Extract only, do not execute the service (default: false)
   --keep, -k     Do not delete the temp dir after execution (default: false)
   --help, -h     show help (default: false)
   --version, -V  print the version (default: false)
```

## Solution description

An option within BoolFlag to disable printing the default would be helpful.

atif-konasl:
## What type of PR is this?

- feature

There is an issue #1122  which requires an extra option like (ShowDefaultVlaue) to prevent default to be printed for BoolFlag. 

Example 
```
GLOBAL OPTIONS:
   --verbose, -v  Increase verbosity (default: false)
   --extract, -e  Extract only, do not execute the service (default: false)
   --keep, -k     Do not delete the temp dir after execution (default: false)
   --help, -h     show help (default: false)
   --version, -V  print the version (default: false)
```

Needs an option for BoolFlag to prevent printing ```(default: false)```. 

Fixes #1122

## Testing
- Only effects on BoolFlag
- Set option ShowDefaultValue true then it shows default value in BoolFlag.
- Set option ShowDefaultValue false then it does not show default value in BoolFlag.
- Other flag behavior remains as is.

Before - 
```
GLOBAL OPTIONS:
   --verbose, -v  Increase verbosity (default: false)
   --extract, -e  Extract only, do not execute the service (default: false)
   --keep, -k     Do not delete the temp dir after execution (default: false)
   --help, -h     show help (default: false)
   --version, -V  print the version (default: false)
```
After - When sets ShowDefaultValue option false.
```
GLOBAL OPTIONS:
   --verbose, -v  Increase verbosity
   --extract, -e  Extract only, do not execute the service
   --keep, -k     Do not delete the temp dir after execution
   --help, -h     show help
   --version, -V  print the version
```

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -2187,20 +2187,21 @@ func TestSetupInitializesOnlyNilWriters(t *testing.T) {
 func TestApp_PreventDefaultPrint_When_BoolFlag(t *testing.T) {
 	var actualOutput bytes.Buffer
 
-	expectedOutputSlice := strings.Fields(`NAME:
-	greet - fight the loneliness!
-	   
-  USAGE:
-	cli.test [global options] command [command options] [arguments...]
-  
-  COMMANDS:
-	help, h  Shows a list of commands or help for one command
-  
-  GLOBAL OPTIONS:
-	--verbose, -v  Increase verbosity (default: default text)
-	--extract, -e  Extract only, do not execute the service
-	--keep, -k     Do not delete the temp dir after execution
-	--help, -h     show help`)
+	expectedOutput := `NAME:
+   greet - fight the loneliness!
+
+USAGE:
+   cli.test [global options] command [command options] [arguments...]
+
+COMMANDS:
+   help, h  Shows a list of commands or help for one command
+
+GLOBAL OPTIONS:
+   --verbose, -v  Increase verbosity
+   --extract, -e  Extract only, do not execute the service
+   --keep, -k     Do not delete the temp dir after execution (default: false)
+   --help, -h     show help (default: false)
+`
 
 	app := App{
 		Name:   "greet",
@@ -2211,14 +2212,14 @@ func TestApp_PreventDefaultPrint_When_BoolFlag(t *testing.T) {
 				Name:             "verbose",
 				Aliases:          []string{"v"},
 				Usage:            "Increase verbosity",
-				ShowDefaultValue: true,
+				HideDefaultValue: true,
 				DefaultText:      "default text",
 			},
 			&BoolFlag{
 				Name:             "extract",
 				Aliases:          []string{"e"},
 				Usage:            "Extract only, do not execute the service",
-				ShowDefaultValue: false,
+				HideDefaultValue: true,
 			},
 			&BoolFlag{
 				Name:    "keep",
@@ -2232,10 +2233,15 @@ func TestApp_PreventDefaultPrint_When_BoolFlag(t *testing.T) {
 		},
 	}
 
-	_ = app.Run([]string{"", "-h"})
+	err := app.Run([]string{"", "-h"})
+	if err != nil {
+		t.Errorf("Run returned unexpected error: %v", err)
+	}
 
-	actualOutputSlice := strings.Fields(actualOutput.String())
-	if !equal(actualOutputSlice, expectedOutputSlice, t) {
+	actualOutputLines := strings.Split(actualOutput.String(), "\n")
+	expectedOutputLines := strings.Split(expectedOutput, "\n")
+
+	if !equal(actualOutputLines, expectedOutputLines, t) {
 		t.Errorf("Actual output does not match with expected output")
 	}
 
@@ -2251,22 +2257,20 @@ func TestApp_PreventDefaultPrint_When_BoolFlag(t *testing.T) {
 			help, h  Shows a list of commands or help for one command
 
 			GLOBAL OPTIONS:
-			--verbose, -v  Increase verbosity (default: default text)
+			--verbose, -v  Increase verbosity
 			--extract, -e  Extract only, do not execute the service
-			--keep, -k     Do not delete the temp dir after execution
-			--help, -h     show help
+			--keep, -k     Do not delete the temp dir after execution (default: false)
+			--help, -h     show help (default: false)
 	*/
 }
 
-// Equal tells whether a and b contain the same elements.
-// A nil argument is equivalent to an empty slice.
-func equal(a, b []string, t *testing.T) bool {
-	if len(a) != len(b) {
-		t.Errorf("Length mismatch")
+func equal(actualOutputLines, expectedOutputLines []string, t *testing.T) bool {
+	if len(actualOutputLines) != len(expectedOutputLines) {
+		t.Errorf("Length mismatch when run TestApp_PreventDefaultPrint_When_BoolFlag")
 		return false
 	}
-	for i, v := range a {
-		if v != b[i] {
+	for i, line := range actualOutputLines {
+		if line != expectedOutputLines[i] {
 			return false
 		}
 	}
--- a/app_test.go
+++ b/app_test.go
@@ -2183,3 +2183,63 @@ func TestSetupInitializesOnlyNilWriters(t *testing.T) {
 		t.Errorf("expected a.Writer to be os.Stdout")
 	}
 }
+
+func TestApp_PreventDefaultPrint_When_BoolFlag(t *testing.T) {
+
+	app := App{
+		Name:  "greet",
+		Usage: "fight the loneliness!",
+		Flags: []Flag{
+			&BoolFlag{
+				Name:             "verbose",
+				Aliases:          []string{"v"},
+				Usage:            "Increase verbosity",
+				ShowDefaultValue: true,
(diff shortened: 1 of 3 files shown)


flag.go
```diff
@@ -271,10 +271,13 @@ func stringifyFlag(f Flag) string {
 	needsPlaceholder := false
 	defaultValueString := ""
 	val := fv.FieldByName("Value")
+	showDefaultValue := fv.FieldByName("ShowDefaultValue")
```
rliebz, line 274: Using reflection here opens us up to potential bugs after future refactors, since we don't get any compile time enforcement that this field actually exists.

Since we're already assuming below that this field is only relevant for a bool type, what we can do instead is do a type check:
```go
showDefaultValue := true
if boolFlag, ok := f.(BoolFlag); ok {
  showDefaultValue = boolFlag.ShowDefaultValue
}
```

We then have a value that represents whether we should show the default value that works consistently across all types and is slightly more refactor proof.

app_test.go
```diff
@@ -2183,3 +2183,63 @@ func TestSetupInitializesOnlyNilWriters(t *testing.T) {
...
+		Action: func(c *Context) error {
+			return nil
+		},
+	}
+
+	err := app.Run([]string{"", "-h"})
+	if err != nil {
+		t.Error(err)
+	}
+	/*
```
rliebz, line 2227: A test for this behavior needs to assert something other than having no errors during the run. Having this example is great, but what we really need is for that example to be proven and enforced programmatically.

In particular, if this is a change to help documentation, we should capture the output of the application (in a `bytes.Buffer` is probably easiest), then assert that the output looks like what's documented here.

atif-konasl: @rliebz I have used `bytes.Buffer` but i could not get proper output as expected. Actually i can not find out why `bytes.Buffer` can not convert multi line string.

would you please review the test code 
```
func TestApp_PreventDefaultPrint_When_BoolFlag(t *testing.T) {
	var actualOutput bytes.Buffer
	var expectedOutput bytes.Buffer

	expectedOutput.WriteString(
		`NAME: 
		greet - fight the loneliness!

		USAGE:
		debug.test [global options] command [command options] [arguments...]

		COMMANDS:

		GLOBAL OPTIONS:
		--verbose, -v  Increase verbosity (default: default text)
		--extract, -e  Extract only, do not execute the service
		--keep, -k     Do not delete the temp dir after execution
		--help, -h     show help
		`)

	app := App{
		Name:   "greet",
		Usage:  "fight the loneliness!",
		Writer: &actualOutput,
		Flags: []Flag{
			&BoolFlag{
				Name:             "verbose",
				Aliases:          []string{"v"},
				Usage:            "Increase verbosity",
				ShowDefaultValue: true,
				DefaultText:      "default text",
			},
			&BoolFlag{
				Name:             "extract",
				Aliases:          []string{"e"},
				Usage:            "Extract only, do not execute the service",
				ShowDefaultValue: false,
			},
			&BoolFlag{
				Name:    "keep",
				Aliases: []string{"k"},
				Usage:   " Do not delete the temp dir after execution",
			},
		},

		Action: func(c *Context) error {
			return nil
		},
	}

	_ = app.Run([]string{"", "-h"})

	actualOutputString := actualOutput.String()
	expectedOutputString := expectedOutput.String()
	expect(t, actualOutputString, expectedOutputString)
}
```

rliebz: `bytes.Buffer` doesn't do anything special with newlines, but Go's multi-line strings using the backtick character might not behave as you expect. Leading whitespace is preserved, so for multi-line strings, you'll have to outdent if you don't expect leading indentation.

Something like this:
```go
		expectedOutput.WriteString(
				`NAME: 
greet - fight the loneliness!

USAGE:
...
```

It definitely looks weird, but that's the way to represent them. Unfortunately our `expect` helper function doesn't do a great job of showing you what the difference is, but if you continue to run into issues, I'd recommend looking at a tool like https://github.com/google/go-cmp to get a good line-by-line comparison as you debug.

atif-konasl: Hi @rliebz, I used `strings.Fields(actualOutput.String())` to compare the multi-line strings which is actually great. Could you please review?

flag.go
```diff
@@ -271,9 +271,18 @@ func stringifyFlag(f Flag) string {
...
+	if boolFlag, ok := f.(*BoolFlag); ok {
+		showDefaultValue = boolFlag.ShowDefaultValue
+	}
+
 	if val.IsValid() {
 		needsPlaceholder = val.Kind() != reflect.Bool
 		defaultValueString = fmt.Sprintf(formatDefault("%v"), val.Interface())
+		if !showDefaultValue {
+			defaultValueString = ""
+		}
```
lynncyrin, line 285: Can you invert this conditional please? Like so

```suggestion
		if showDefaultValue {
			defaultValueString = fmt.Sprintf(formatDefault("%v"), val.Interface())
		}
```

flag.go
```diff
@@ -271,9 +271,18 @@ func stringifyFlag(f Flag) string {
...
 
 		if val.Kind() == reflect.String && val.String() != "" {
 			defaultValueString = fmt.Sprintf(formatDefault("%q"), val.String())
```
lynncyrin, line 288: It's unclear why this line doesn't also have a `showDefaultValue` check?

flag.go
```diff
@@ -271,9 +271,18 @@ func stringifyFlag(f Flag) string {
 	needsPlaceholder := false
 	defaultValueString := ""
 	val := fv.FieldByName("Value")
+
+	showDefaultValue := true
+	if boolFlag, ok := f.(*BoolFlag); ok {
+		showDefaultValue = boolFlag.ShowDefaultValue
+	}
```
lynncyrin, line 278: Unrelated tests are failing because the default value for the flag `ShowDefaultValue` is `false`, despite you setting `showDefaultValue := true` here. Given that "show" needs to be the default behavior, and the default value for `bool` if `false`, I think you need to change the flag option from `BoolFlag.ShowDefaultValue` to `BoolFlag.HideDefaultValue` (and change the logic accordingly).

app_test.go
rliebz: `strings.Fields` is not the best comparison for multi-line strings because it effectively ignores whitespace. 

If you're looking for a line-by-line breakdown, you could use `strings.Split(actualOutput.String(), "\n")`.

Also, we should remove the comment, since we have an actual test string that's being compared.

app_test.go
```diff
@@ -2183,3 +2183,92 @@ func TestSetupInitializesOnlyNilWriters(t *testing.T) {
...
+				Usage:   " Do not delete the temp dir after execution",
+			},
+		},
+
+		Action: func(c *Context) error {
+			return nil
+		},
+	}
+
+	_ = app.Run([]string{"", "-h"})
```
rliebz, line 2235: Can you also check to ensure this error is not nil?

atif-konasl: pls check now

flag.go
atif-konasl: pls check now

lynncyrin:
Looks like a test is still failing?

GregoryDosh:
Throwing this out there, it would be really nice if the `ShowDefaultValue` parameter existed for other options too. The reason I'm asking is because in a situation where a CLI app is picking up configuration from an environmental variable, it's too easy to accidentally expose that information.

I've got a situation where I'm using this library to create a containerized plugin for use within a CI environment, and it will get injected secrets through way of env variables, but if the user malforms their configuration, any of the set secrets will be exposed by the help menu. For now I'm just setting DefaultText to `N/A` so it doesn't appear, but this param would be nice to exist on all available flags.

meatballhat:
Given the recent changes to remove use of `reflect` and the (reasonable, imho) general opinion that further increasing API surface should be done sparingly, I think this should instead be accomplished (for now?) by wrapping `BoolFlag` in an admittedly awkward way like so:


```go
package main

import (
    "os"

    "github.com/urfave/cli/v2"
)

type quietBoolFlag struct {
    cli.BoolFlag
}

func (qbf *quietBoolFlag) String() string {
    return cli.FlagStringer(qbf)
}

func (qbf *quietBoolFlag) GetDefaultText() string {
    return ""
}

func main() {
    (&cli.App{
        Flags: []cli.Flag{
            &quietBoolFlag{
                BoolFlag: cli.BoolFlag{
                    Name:  "yep",
                    Usage: "declare yepness",
                },
            },
        },
    }).Run(os.Args)
}
```

```sh
go run 1128.go --help
```

```
NAME:
   1128 - A new cli application

USAGE:
   1128 [global options] command [command options] [arguments...]

COMMANDS:

GLOBAL OPTIONS:
   --yep       declare yepness
```

PR #1130, opened 2020-05-03
fix(zsh_autocomplete): List files on tab with no completion options

oleorhagen:
## What type of PR is this?

- [x ] bug
- [ ] cleanup  
- [ ] documentation
- [ ] feature

Aligns zsh auto-completion functionality with the bash auto-completion

## Testing

Tested manually

New functionality: https://asciinema.org/a/EAYRIEVGTGNSS2gCGwSJ4Zw1i

Old functionality: https://asciinema.org/a/BfOZz4BHUGwjXMFptbmDHZocH

## Release Notes

_(REQUIRED)_

```release-note
List files on tab with zsh auto-completion when no completion candidates are available.
```

### DIFF
--- a/autocomplete/zsh_autocomplete
+++ b/autocomplete/zsh_autocomplete
@@ -13,6 +13,8 @@ _cli_zsh_autocomplete() {
 
   if [[ "${opts[1]}" != "" ]]; then
     _describe 'values' opts
+  else
+    _files
   fi
 
   return


oleorhagen:
> Looks fine, although I would like if there was something akin to tests for this.

Yup, some tests for the auto-completion features (both bash and zsh) would have been nice. But I didn't see any in place as of now. I wonder how much effort it would be to bootstrap?

PR #1127, opened 2020-05-01
fix #1121(StringSliceFlag set default value into destination)

sgoroshko:
## What type of PR is this?

_(REQUIRED)_

- [x] bug
- [ ] cleanup  
- [ ] documentation
- [ ] feature

## What this PR does / why we need it:

default value does not set in Destination field in StringSliceFlag.

## Which issue(s) this PR fixes:

https://github.com/urfave/cli/issues/1121

## Release Notes

_(REQUIRED)_

```release-note
 - fix bug, default value does not set in Destination field in StringSliceFlag.
```

PR #1131, opened 2020-05-04
StringSlice flags set to default Value

darren-west:
StringSliceFlag types do not set the destination to the value if no env/arg is set.

For example if the argument `--arg` is not set and env variable `ARG` is not set you would expect the destination to be set to `cli.NewStringSlice("foo")`
```go
                 var arg = cli.NewStringSlice()
		...
                  &cli.StringSliceFlag{
			Name:        "arg",
			EnvVars:     []string{"ARG"},
			Value:       cli.NewStringSlice("foo"),
			Destination: arg,
		},
```

I have quickly looked at how the string flag achieves this and tried to do similar.

### DIFF
--- a/flag_string_slice.go
+++ b/flag_string_slice.go
@@ -148,7 +148,7 @@ func (f *StringSliceFlag) Apply(set *flag.FlagSet) error {
 		set.Var(f.Value, name, f.Usage)
 	}
 
-	if !f.HasBeenSet && f.Value != nil {
+	if !f.HasBeenSet && f.Value != nil && f.Destination != nil {
 		for _, v := range f.Value.Value() {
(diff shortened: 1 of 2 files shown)


darren-west:
I have added a test. The idea is that if the destination is set and the value is set but the flag is not set by env name or a program argument the value sets the destination. This is more consistent with how the other flags work (for instance StringFlag)

erikwilson:
It looks like this is a duplicate of #1127?

lynncyrin:
It does look like this is a duplicate!

meatballhat:
Confirmed already fixed via #1127

Thank you all the same @darren-west!

PR #1133, opened 2020-05-05
Backport zsh_autocomplete from master to the v1 branch

oleorhagen:
## What type of PR is this?

- [ ] bug
- [x] cleanup  
- [ ] documentation
- [ ] feature

Backports the `autocomplete/zsh_autocomplete` file from master to `v1` as there is nothing keeping these files from being mirrored.

No open issues for this.

## Testing
This is tested locally, and the auto-completion scripts have no tests at the moment.

## Release Notes

_(REQUIRED)_

```release-note
Backported the ZSH auto-completion script from master in order to:
* list files if no completions are present.
* auto-complete flags
```

### DIFF
--- a/autocomplete/zsh_autocomplete
+++ b/autocomplete/zsh_autocomplete
@@ -3,9 +3,19 @@
 _cli_zsh_autocomplete() {
 
   local -a opts
-  opts=("${(@f)$(_CLI_ZSH_AUTOCOMPLETE_HACK=1 ${words[@]:0:#words[@]-1} --generate-bash-completion)}")
+  local cur
+  cur=${words[-1]}
+  if [[ "$cur" == "-"* ]]; then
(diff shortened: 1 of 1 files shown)


oleorhagen:
> I'm not sure why CI is failing, can you re-request my review after you've look into that?
> 
> Also, v1 is intended to be in maintenance mode and isn't meant for getting new features. This is an ok exception since it's a simple change, but in general if you want new features then you need to migrate to v2.

Thank you @lynncyrin.

Yes, I understand the release cycle for semantic versioning. However, for some enterprise software the cycles can be loooong, thus we are stuck on v1 for some time still.

Also, I would not consider this a feature, but a fix, considering, what it is really doing is porting the buf-fixes from the auto-complete script in v2.

However, this is all up to you, on how you guys want to handle this :)

Thank you for a nice CLI library!

Issue #1092, opened 2020-03-25
String flag cannot be set to "-1"


AkihiroSuda:
## my urfave/cli version is

v1.22.1, v1.22.2, v1.22.3 (broken since v1.22.2)

## Checklist

* [X] Are you running the latest v1 release? The list of releases is [here](https://github.com/urfave/cli/releases).
* [X] Did you check the manual for your release? The v1 manual is [here](https://github.com/urfave/cli/blob/master/docs/v1/manual.md)
* [X] Did you perform a search about this problem? Here's the [Github guide](https://help.github.com/en/github/managing-your-work-on-github/using-search-to-filter-issues-and-pull-requests) about searching.

## Dependency Management

- [X] My project is using go modules.
- [X] My project is using vendoring.
- [ ] My project is automatically downloading the latest version.
- [ ] I am unsure of what my dependency management setup is.

## Describe the bug

This flag can be set to "-1" on v1.22.1 but cannot be set since v1.22.2.
```go
		cli.StringFlag{
			Name:  "memory-swap",
			Usage: "Total memory usage (memory + swap); set '-1' to enable unlimited swap",
		},
```
https://github.com/opencontainers/runc/blob/be51398a8af0b32b0b580e4da069cc6f704639bf/update.go#L111-L114

Regression in https://github.com/urfave/cli/compare/v1.22.1...v1.22.2

## To reproduce

Specify `urfave/cli` version in `go.mod` in the runc tree, run `go mod tidy`, `go mod vendor`, and `make integration`.

## Observed behavior

```
# runc update test_update --memory-swap -1 (status=1):
# Incorrect Usage: flag provided but not defined: -1
```

## Expected behavior

It should not fail.

## Additional context

https://github.com/opencontainers/runc/pull/2268#issuecomment-603444767

## Run `go version` and paste its output here
For vendoring:
```
go version go1.14.1 linux/amd64
```

For running the containerized test suite:
```
go version go1.12.17 linux/amd64
```

lynncyrin:
😅 looks like it thinks `-1` is a flag called `1`. This bug makes sense, thanks for the well written issue @AkihiroSuda!

🙏 PRs are welcome 🙏

thaJeztah:
Did some quick testing to do a minimal reproducer (LOL, I never worked with urfave/cli, so "code by example" 😅);

```go
package main

import (
	"fmt"
	"os"
	"strings"

	"github.com/urfave/cli"
)

func main() {
[...]

PR #1135, opened 2020-05-08
Fix regression in handling flag-values starting with hyphen

thaJeztah:
fixes https://github.com/urfave/cli/issues/1092

Fix for a regression between v1.22.1 and v1.22.2, where
flag values starting with a hyphen would be parsed as a flag:

    runc update test_update --memory-swap -1
    Incorrect Usage: flag provided but not defined: -1

This problem was caused by `reorderArgs()` not properly checking
if the next arg in the list was a value for the flag (and not
the next flag);

Before this patch:

    args before reordering: --opt,--opt-value,arg1
    args after reordering:  --opt,arg1,--opt-value
    args before reordering: --opt,--opt-value,arg1,arg2
    args after reordering:  --opt,arg1,--opt-value,arg2
    args before reordering: arg1,--opt,--opt-value,arg2
    args after reordering:  --opt,arg2,arg1,--opt-value

After this patch is applied:

    args before reordering: --opt,--opt-value,arg1
    args after reordering:  --opt,--opt-value,arg1
    args before reordering: --opt,--opt-value,arg1,arg2
    args after reordering:  --opt,--opt-value,arg1,arg2
    args before reordering: arg1,--opt,--opt-value,arg2
    args after reordering:  --opt,--opt-value,arg1,arg2

## What type of PR is this?

_(REQUIRED)_

- [x] bug
- [ ] cleanup  
- [ ] documentation
- [ ] feature

see description above

fixes https://github.com/urfave/cli/issues/1092

## Testing

Test was added

## Release Notes

```release-note
Fix a regression where flag values starting with a hyphen would be parsed as a flag instead of a value for a flag
```

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -546,8 +546,7 @@ func TestApp_CommandWithFlagBeforeTerminator(t *testing.T) {
 
 	expect(t, parsedOption, "my-option")
 	expect(t, args[0], "my-arg")
-	expect(t, args[1], "--")
-	expect(t, args[2], "--notARealFlag")
+	expect(t, args[1], "--notARealFlag")
 }
 
 func TestApp_CommandWithDash(t *testing.T) {
@@ -585,8 +584,7 @@ func TestApp_CommandWithNoFlagBeforeTerminator(t *testing.T) {
 	_ = app.Run([]string{"", "cmd", "my-arg", "--", "notAFlagAtAll"})
 
 	expect(t, args[0], "my-arg")
-	expect(t, args[1], "--")
-	expect(t, args[2], "notAFlagAtAll")
+	expect(t, args[1], "notAFlagAtAll")
 }
 
 func TestApp_VisibleCommands(t *testing.T) {
--- a/command.go
+++ b/command.go
@@ -226,18 +226,23 @@ func reorderArgs(commandFlags []Flag, args []string) []string {
 	nextIndexMayContainValue := false
 	for i, arg := range args {
 
-		// dont reorder any args after a --
-		// read about -- here:
-		// https://unix.stackexchange.com/questions/11376/what-does-double-dash-mean-also-known-as-bare-double-dash
-		if arg == "--" {
-			remainingArgs = append(remainingArgs, args[i:]...)
-			break
-
-			// checks if this arg is a value that should be re-ordered next to its associated flag
-		} else if nextIndexMayContainValue && !argIsFlag(commandFlags, arg) {
+		// if we're expecting an option-value, check if this arg is a value, in
+		// which case it should be re-ordered next to its associated flag
+		if nextIndexMayContainValue && !argIsFlag(commandFlags, arg) {
 			nextIndexMayContainValue = false
 			reorderedArgs = append(reorderedArgs, arg)
+		} else if arg == "--" {
+			// don't reorder any args after the -- delimiter As described in the POSIX spec:
+			// https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_02
+			// > Guideline 10:
+			// >   The first -- argument that is not an option-argument should be accepted
+			// >   as a delimiter indicating the end of options. Any following arguments
+			// >   should be treated as operands, even if they begin with the '-' character.
 
+			// make sure the "--" delimiter itself is at the start
+			remainingArgs = append([]string{"--"}, remainingArgs...)
+			remainingArgs = append(remainingArgs, args[i+1:]...)
+			break
 			// checks if this is an arg that should be re-ordered
 		} else if argIsFlag(commandFlags, arg) {
 			// we have determined that this is a flag that we should re-order
@@ -256,8 +261,9 @@ func reorderArgs(commandFlags []Flag, args []string) []string {
 
 // argIsFlag checks if an arg is one of our command flags
 func argIsFlag(commandFlags []Flag, arg string) bool {
-	// checks if this is just a `-`, and so definitely not a flag
-	if arg == "-" {
+	if arg == "-" || arg == "--"{
+		// `-` is never a flag
+		// `--` is an option-value when following a flag, and a delimiter indicating the end of options in other cases.
 		return false
 	}
 	// flags always start with a -
--- a/command.go
+++ b/command.go
@@ -234,7 +234,7 @@ func reorderArgs(commandFlags []Flag, args []string) []string {
 			break
 
 			// checks if this arg is a value that should be re-ordered next to its associated flag
-		} else if nextIndexMayContainValue && !strings.HasPrefix(arg, "-") {
+		} else if nextIndexMayContainValue && !argIsFlag(commandFlags, arg) {
 			nextIndexMayContainValue = false
 			reorderedArgs = append(reorderedArgs, arg)
 
--- a/command_test.go
+++ b/command_test.go
@@ -114,6 +114,71 @@ func TestParseAndRunShortOpts(t *testing.T) {
 	}
 }
 
+// test-case for https://github.com/urfave/cli/issues/1092
+func TestParseAndRunHyphenValues(t *testing.T) {
+	cases := []struct {
+		testArgs     []string
+		expectedArgs []string
+		expectedOpt string
+	}{
+		{[]string{"foo", "test", "argz"}, []string{"argz"}, ""},
+		{[]string{"foo", "test", "argz", "arga"}, []string{"argz", "arga"}, ""},
+		{[]string{"foo", "test", "--opt", "opt-value"}, []string{}, "opt-value"},
+		{[]string{"foo", "test", "--opt", "opt-value", "argz"}, []string{"argz"}, "opt-value"},
+		{[]string{"foo", "test", "--opt", "opt-value", "argz", "arga"}, []string{"argz", "arga"}, "opt-value"},
+		{[]string{"foo", "test", "argz", "--opt", "opt-value"}, []string{"argz"}, "opt-value"},
+		{[]string{"foo", "test", "argz", "--opt", "opt-value", "arga"}, []string{"argz", "arga"}, "opt-value"},
+
+		{[]string{"foo", "test", "--opt", "-opt-value"}, []string{}, "-opt-value"},
+		{[]string{"foo", "test", "--opt", "-opt-value", "argz"}, []string{"argz"}, "-opt-value"},
+		{[]string{"foo", "test", "--opt", "-opt-value", "argz", "arga"}, []string{"argz", "arga"}, "-opt-value"},
+		{[]string{"foo", "test", "argz", "--opt", "-opt-value"}, []string{"argz"}, "-opt-value"},
+		{[]string{"foo", "test", "argz", "--opt", "-opt-value", "arga"}, []string{"argz", "arga"}, "-opt-value"},
+
+		{[]string{"foo", "test", "--opt", "--opt-value"}, []string{}, "--opt-value"},
+		{[]string{"foo", "test", "--opt", "--opt-value", "argz"}, []string{"argz"}, "--opt-value"},
+		{[]string{"foo", "test", "--opt", "--opt-value", "argz", "arga"}, []string{"argz", "arga"}, "--opt-value"},
+		{[]string{"foo", "test", "argz", "--opt", "--opt-value"}, []string{"argz"}, "--opt-value"},
+		{[]string{"foo", "test", "argz", "--opt", "--opt-value", "arga"}, []string{"argz", "arga"}, "--opt-value"},
+	}
+
+	for _, tc := range cases {
+		tc := tc
+		t.Run(strings.Join(tc.testArgs, "_"), func(t *testing.T){
+			var (
+				args []string
+				opt string
+			)
+
+			cmd := Command{
+				Name:        "test",
+				Usage:       "this is for testing",
+				Description: "testing",
+				Action: func(c *Context) error {
+					args = c.Args()
+					opt = c.String("opt")
+					return nil
+				},
+				Flags: []Flag{
+					StringFlag{Name:  "opt"},
+				},
+			}
+
+			app := NewApp()
+			app.Writer = ioutil.Discard
+			app.ErrWriter = ioutil.Discard
+			app.Commands = []Command{cmd}
+
+			err := app.Run(tc.testArgs)
+
+			expect(t, err, nil)
+			expect(t, args, tc.expectedArgs)
+			expect(t, opt, tc.expectedOpt)
+		})
+
+	}
+}
+
 func TestCommand_Run_DoesNotOverwriteErrorFromBefore(t *testing.T) {
 	app := NewApp()
 	app.Commands = []Command{
--- a/command_test.go
+++ b/command_test.go
@@ -140,6 +140,17 @@ func TestParseAndRunHyphenValues(t *testing.T) {
 		{[]string{"foo", "test", "--opt", "--opt-value", "argz", "arga"}, []string{"argz", "arga"}, "--opt-value"},
 		{[]string{"foo", "test", "argz", "--opt", "--opt-value"}, []string{"argz"}, "--opt-value"},
 		{[]string{"foo", "test", "argz", "--opt", "--opt-value", "arga"}, []string{"argz", "arga"}, "--opt-value"},
+
+		// Tests below test handling of the "--" delimiter
+		{[]string{"foo", "test", "--", "--opt", "--opt-value", "argz"}, []string{"--opt", "--opt-value", "argz"}, ""},
+		{[]string{"foo", "test", "--", "argz", "--opt", "--opt-value", "arga"}, []string{"argz", "--opt", "--opt-value", "arga"}, ""},
+		{[]string{"foo", "test", "argz", "--", "--opt", "--opt-value", "arga"}, []string{"argz", "--opt", "--opt-value", "arga"}, ""},
+		{[]string{"foo", "test", "argz", "--opt", "--", "--opt-value", "arga"}, []string{"argz", "--opt-value", "arga"}, "--"},
+
+		// first "--" is an option-value for "--opt", second "--" is the delimiter indicating the end of options.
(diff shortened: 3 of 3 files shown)


thaJeztah:
wrote the test to a separate file (`regression_test.go`) (don't do this often, so 😅)

```bash
git bisect start
git bisect good v1.22.0
git bisect bad v1.22.3
Bisecting: 62 revisions left to test after this (roughly 6 steps)
[41eb645fa27833f6d76257a58e62ab9b09366c82] Merge branch 'master' into pass-through-regression


git bisect run go test -run TestParseAndRunHyphenValues
...
Bisecting: 0 revisions left to test after this (roughly 0 steps)
[bfeb4c88ec4344b456abce2b587d7bc0b957e879] Merge branch 'master' into pass-through-regression
running go test -run TestParseAndRunHyphenValues
PASS
ok  	github.com/urfave/cli	0.006s
7d0751fa1309a01b405cde327cc43275d13016a2 is the first bad commit
commit 7d0751fa1309a01b405cde327cc43275d13016a2
Author: Lynn Cyrin <lynn@textio.com>
Date:   Wed Sep 11 19:25:51 2019 -0700

    add argIsFlag check

:100644 100644 d33435e6dd71cca15fab4a44d8ee3289f333232c 6471f1005b3e9c33be475d6dc1f322eddabfd606 M	command.go
bisect run success
```

```bash
git bisect log

git bisect start
# good: [bfe2e925cfb6d44b40ad3a779165ea7e8aff9212] Merge pull request #882 from urfave/lynncyrin-patch-1
git bisect good bfe2e925cfb6d44b40ad3a779165ea7e8aff9212
# bad: [c354ceca3e45cda773c3846bcb0c67cec4046b03] Merge pull request #981 from urfave/string-slice-issue-v1
git bisect bad c354ceca3e45cda773c3846bcb0c67cec4046b03
# bad: [41eb645fa27833f6d76257a58e62ab9b09366c82] Merge branch 'master' into pass-through-regression
git bisect bad 41eb645fa27833f6d76257a58e62ab9b09366c82
# good: [ef47250cda5ff52a313118c01ad6b0c5b4877a70] Merge branch 'master' into asahasrabuddhe-patch-1
git bisect good ef47250cda5ff52a313118c01ad6b0c5b4877a70
# good: [2071bcfb8328cb13e5575a586f06a33563b04054] checkpoint
git bisect good 2071bcfb8328cb13e5575a586f06a33563b04054
# bad: [09cdbbfe281fd79e92616a28f1fdf701e9007c22] more docs
git bisect bad 09cdbbfe281fd79e92616a28f1fdf701e9007c22
# bad: [fcfc69e54ff0e870bae1eaf599b7d15b46abb86b] Merge remote-tracking branch 'origin/master' into pass-through-regression
git bisect bad fcfc69e54ff0e870bae1eaf599b7d15b46abb86b
# bad: [7d0751fa1309a01b405cde327cc43275d13016a2] add argIsFlag check
git bisect bad 7d0751fa1309a01b405cde327cc43275d13016a2
# good: [bfeb4c88ec4344b456abce2b587d7bc0b957e879] Merge branch 'master' into pass-through-regression
git bisect good bfeb4c88ec4344b456abce2b587d7bc0b957e879
# first bad commit: [7d0751fa1309a01b405cde327cc43275d13016a2] add argIsFlag check
```

Looks like the regression was in https://github.com/urfave/cli/commit/7d0751fa1309a01b405cde327cc43275d13016a2 (https://github.com/urfave/cli/pull/872)

command.go
```diff
@@ -265,13 +267,12 @@ func argIsFlag(commandFlags []Flag, arg string) bool {
 	if !strings.HasPrefix(arg, "-") {
 		return false
 	}
-	// this line turns `--flag` into `flag`
```
thaJeztah, line 268: Ignore these changes; this was just optimisation

command.go
```diff
@@ -189,9 +189,11 @@ func (c *Command) parseFlags(args Args, shellComplete bool) (*flag.FlagSet, erro
 		return set, set.Parse(append([]string{"--"}, args...))
 	}
 
+	fmt.Println("before reordering:", strings.Join(args, ","))
 	if !c.SkipArgReorder {
 		args = reorderArgs(c.Flags, args)
 	}
+	fmt.Println("after reordering: ", strings.Join(args, ","))
```
thaJeztah, line 196: Problem is in this area; 

```
=== RUN   TestParseAndRunHyphenValues/foo_test_--opt_-opt-value_arg1
before reordering: --opt,-opt-value,arg1
after reordering:  --opt,arg1,-opt-value
    TestParseAndRunHyphenValues/foo_test_--opt_-opt-value_arg1: helpers_test.go:20: (command_test.go:175) Expected <nil> (type <nil>) - Got flag provided but not defined: -opt-value (type *errors.errorString)
    TestParseAndRunHyphenValues/foo_test_--opt_-opt-value_arg1: helpers_test.go:20: (command_test.go:176) Expected [arg1] (type []string) - Got [] (type []string)
    TestParseAndRunHyphenValues/foo_test_--opt_-opt-value_arg1: helpers_test.go:20: (command_test.go:177) Expected -opt-value (type string) - Got  (type string)
```

looks like the `reorderArgs` does not take into account if an `arg` is a value for a flag, and now `arg1` becomes an argument for `--opt`;

```
after reordering:  --opt,arg1,-opt-value
```

thaJeztah:
Found the issue; ready for review @AkihiroSuda @lynncyrin PTAL

command_test.go
```diff
@@ -114,6 +114,72 @@ func TestParseAndRunShortOpts(t *testing.T) {
...
+		{[]string{"foo", "test", "--opt", "-opt-value", "arg1", "arg2"}, []string{"arg1", "arg2"}, "-opt-value"},
+		{[]string{"foo", "test", "arg1", "--opt", "-opt-value"}, []string{"arg1"}, "-opt-value"},
+		{[]string{"foo", "test", "arg1", "--opt", "-opt-value", "arg2"}, []string{"arg1", "arg2"}, "-opt-value"},
+
+		{[]string{"foo", "test", "--opt", "--opt-value"}, []string{}, "--opt-value"},
+		{[]string{"foo", "test", "--opt", "--opt-value", "arg1"}, []string{"arg1"}, "--opt-value"},
+		{[]string{"foo", "test", "--opt", "--opt-value", "arg1", "arg2"}, []string{"arg1", "arg2"}, "--opt-value"},
+		{[]string{"foo", "test", "arg1", "--opt", "--opt-value"}, []string{"arg1"}, "--opt-value"},
+		{[]string{"foo", "test", "arg1", "--opt", "--opt-value", "arg2"}, []string{"arg1", "arg2"}, "--opt-value"},
+	}
```
lynncyrin, line 143: I think it would be valuable to add a set of cases that includes the `--` syntax, what do you think? Here's an example

```suggestion
		{[]string{"foo", "test", "--opt", "--opt-value"}, []string{}, "--opt-value"},
		{[]string{"foo", "test", "--opt", "--opt-value", "arg1"}, []string{"arg1"}, "--opt-value"},
		{[]string{"foo", "test", "--opt", "--opt-value", "arg1", "arg2"}, []string{"arg1", "arg2"}, "--opt-value"},
		{[]string{"foo", "test", "arg1", "--opt", "--opt-value"}, []string{"arg1"}, "--opt-value"},
		{[]string{"foo", "test", "arg1", "--opt", "--opt-value", "arg2"}, []string{"arg1", "arg2"}, "--opt-value"},
				
		{[]string{"foo", "test", "--", "arg1", "--opt", "--opt-value", "arg2"}, []string{""}, ""},
		{[]string{"foo", "test", "arg1", "--", "--opt", "--opt-value", "arg2"}, []string{"arg1"}, ""},
		{[]string{"foo", "test", "arg1", "--opt", "--", "--opt-value", "arg2"}, []string{"arg1"}, ""},
		{[]string{"foo", "test", "arg1", "--opt", "--opt-value", "--", "arg2"}, []string{"arg1"}, "--opt-value"},
		{[]string{"foo", "test", "arg1", "--opt", "--opt-value", "arg2", "--"}, []string{"arg1", "arg2"}, "--opt-value"},
	}
```

Under normal circumstances I would think this was overkill, but this PR is a regression fix for a PR that was itself also a regression fix -- so better safe than sorry IMO!

lynncyrin:
@thaJeztah I have a requested change since this issue is particularly thorny!

Tangentially, I really appreciate how simple the fix is 🙏

command_test.go
thaJeztah: Oh, thanks! Good suggestion!

I was actually looking at the code handling `--` when debugging this thing, but adding a test case for it completely slipped my mind.

thaJeztah: After publishing the PR, I also thought of changing `arg1` and `arg2` to `argz` and `arga`, to be more explicit that things are ordered based on _position_, not alphabetically (and to catch any possible situation where we would be "sorting"). I'll include that as well, but can change back if you don't like that.

thaJeztah:
Thanks for your review 🤗 

> Tangentially, I really appreciate how simple the fix is 🙏

Yes! It was hiding in plain sight, yet so easy to miss.

app_test.go
```diff
@@ -546,8 +546,7 @@ func TestApp_CommandWithFlagBeforeTerminator(t *testing.T) {
 
 	expect(t, parsedOption, "my-option")
 	expect(t, args[0], "my-arg")
-	expect(t, args[1], "--")
```
thaJeztah, line 549: Could use extra eyes on this change; it's a result of the latest commit / fix, but could potentially be a breaking change for users (so _at least_ should be called out in the release notes); I _think_ I interpreted the POSIX spec correctly with my patch, but specs can sometimes be ambiguous, so could use the extra eyes.

I also haven't checked the `v2` branch/version yet (possibly the fix is needed there as well, or in another form)

thaJeztah:
(I'm ignoring codecov for now, as I think it's just a rounding error 😅)

app_test.go
thaJeztah: So far, the twitter-verse seems to consolidate to my interpretation being correct 😂  https://twitter.com/thaJeztah/status/1260921988102656002?s=20

thaJeztah:
@lynncyrin anything needed to move this forward?

app_test.go
lynncyrin: I'm having a hard time imagining is this is a mundane change or if it'll cause mass chaos for somebody somewhere. Is there a way to keep this behavior as-is???

lynncyrin:
@thaJeztah I have (hopefully only) 1 more question!

command.go
```diff
@@ -234,7 +234,7 @@ func reorderArgs(commandFlags []Flag, args []string) []string {
 			break
 
 			// checks if this arg is a value that should be re-ordered next to its associated flag
-		} else if nextIndexMayContainValue && !strings.HasPrefix(arg, "-") {
```
cyphar, line 237: This is still not quite correct if you wanted to do something like:

```
% runc run --bundle --some-real-runc-flag ctr
```

Where it would still re-order things. The least-ugly way I can think of solving this is that we change `nextIndexMayContainValue` to make use of `TakesValue()` (with some interface casting) and removing this "skip reordering flags that start with `-`" logic if it doesn't break anything.

I also hasten to point out that `urfave/cli` has actually had this behaviour for a *very long time* and we (speaking collectively in the containers community) have misunderstood how `urfave/cli` handles this case. I discussed this in #1152 (which I've marked as a duplicate of this) and I've had to work around this in umoci (see opencontainers/umoci#328 for how I ended up hotfixing this -- you just disable argument reordering). Since we want to avoid breaking other users of `urfave/cli` we might have to migrate runc, umoci, and containerd away from `urfave/cli` (maybe to `spf13/cobra` which is what Docker uses)...

cyphar: A possible patch to do the above might look like the following. This change does pass all of the existing tests...

```patch
diff --git a/command.go b/command.go
index f02d3589ff3a..3c8e971a1c53 100644
--- a/command.go
+++ b/command.go
@@ -223,7 +223,7 @@ func (c *Command) useShortOptionHandling() bool {
 func reorderArgs(commandFlags []Flag, args []string) []string {
        var remainingArgs, reorderedArgs []string
 
-       nextIndexMayContainValue := false
+       nextIndexIsValue := false
        for i, arg := range args {
 
                // dont reorder any args after a --
@@ -233,17 +233,29 @@ func reorderArgs(commandFlags []Flag, args []string) []string {
                        remainingArgs = append(remainingArgs, args[i:]...)
                        break
 
-                       // checks if this arg is a value that should be re-ordered next to its associated flag
-               } else if nextIndexMayContainValue && !strings.HasPrefix(arg, "-") {
-                       nextIndexMayContainValue = false
+                       // if this argument is a value then it isn't a flag and should be
+                       // kept with its associated flag (which was the last flag
+                       // reordered)
+               } else if nextIndexIsValue {
+                       nextIndexIsValue = false
                        reorderedArgs = append(reorderedArgs, arg)
 
                        // checks if this is an arg that should be re-ordered
-               } else if argIsFlag(commandFlags, arg) {
+               } else if flag := findFlagForArg(commandFlags, arg); flag != nil {
                        // we have determined that this is a flag that we should re-order
                        reorderedArgs = append(reorderedArgs, arg)
-                       // if this arg does not contain a "=", then the next index may contain the value for this flag
-                       nextIndexMayContainValue = !strings.Contains(arg, "=")
+
+                       // if the flag takes an argument and doesn't contain a "=" then the
+                       // next index contains a value for this flag -- unfortunately we
+                       // have to depend on the Flag implementing DocGenerationFlag for
+                       // TakesValue().
+                       if docFlag, ok := (*flag).(DocGenerationFlag); ok {
+                               nextIndexIsValue = docFlag.TakesValue() && !strings.Contains(arg, "=")
+                       } else {
+                               // assume that flags take arguments
+                               // XXX: This is quite ugly and I'm not sure what we should do here.
+                               nextIndexIsValue = !strings.Contains(arg, "=")
+                       }
 
                        // simply append any remaining args
                } else {
@@ -254,15 +266,16 @@ func reorderArgs(commandFlags []Flag, args []string) []string {
        return append(reorderedArgs, remainingArgs...)
 }
 
-// argIsFlag checks if an arg is one of our command flags
-func argIsFlag(commandFlags []Flag, arg string) bool {
+// findFlagForArg checks if an arg is one of our command flags and returns the
+// Flag if the arg is a flag (nil otherwise).
+func findFlagForArg(commandFlags []Flag, arg string) *Flag {
        // checks if this is just a `-`, and so definitely not a flag
        if arg == "-" {
-               return false
+               return nil
        }
        // flags always start with a -
        if !strings.HasPrefix(arg, "-") {
-               return false
+               return nil
        }
        // this line turns `--flag` into `flag`
        if strings.HasPrefix(arg, "--") {
@@ -279,12 +292,12 @@ func argIsFlag(commandFlags []Flag, arg string) bool {
                for _, key := range strings.Split(flag.GetName(), ",") {
                        key := strings.TrimSpace(key)
                        if key == arg {
-                               return true
+                               return &flag
                        }
                }
        }
-       // return false if this arg was not one of our flags
-       return false
+       // couldn't find the flag
+       return nil
 }
 
 // Names returns the names including short names and aliases.
```

thaJeztah: @cyphar I see you commented on the first commit; did you check with the second commit as well?

thaJeztah: (I was looking at the requested changes to preserve some of the existing behaviour)

cyphar: No, this was just talking about the behaviour of the first change (and not the fun `--` semantics that you quizzed everyone on in May). I /think/ that it should be fairly straight-forward (move the `--` check after the "is this an option-value check as in your current patch) but I agree with both of you that that particular change looks quite scary.

I think the `--opt -a` case is far more clear-cut.

thaJeztah: So, in your example case (`runc run --bundle --some-real-runc-flag`) both `--bundle` and `--some-real-runc-flag` are "known" flags to `runc`, but the second one should be treated as _value_ for `--bundle` (so `--bundle="--some-real-runc-flag"`) or as flag? In the former case, wouldn't it have to be used as `runc run --bundle --  --some-real-runc-flag`? 🤔)

cyphar: The former. The motivating use-case is (aside from POSIX-correctness and compatibility with `getopt(3)`) similar to `--`:

```
% runc run --bundle --foo ctr # where ./--foo is a directory containing an OCI bundle
```

(Well, for umoci it's more like `umoci config --image foo:bar --config.cmd sh --config.cmd -c` but that's a slightly more complicated example. But the root cause is the same as your original `--memory-limit -1` issue.)

Should mean that `--foo` is a value for `--bundle` regardless of whether `--foo` is a known command-line flag. Currently (and with your patches), `urfave/cli` will not reorder `--bundle --foo` (where `--foo` is not a valid flag for `runc`) but it *will* reorder the flags if you do `--bundle --systemd-cgroup` (where `--systemd-cgroup` *is* a supported flag for `runc`). Even more strangely, `--bundle=--systemd-cgroup` works without issue because the arguments don't get reordered.

The patch I posted above *should* make it so that `--bundle=--foo` acts exactly like `--bundle --foo` regardless of whether `--foo` is a valid `runc` flag.

AkihiroSuda: @thaJeztah PTAL?

AkihiroSuda: ping @thaJeztah :pray:

lynncyrin:
👀

cyphar:
I think my earlier comment (https://github.com/urfave/cli/pull/1135#discussion_r449498748) still stands. I can cook up a proper PR based on the patch I posted above which handles `--flagA --flagB` as "--flagB" is the value of `--flagA` regardless of whether `--flagB` is a known flag, if you'd still like that change.

brandond:
@cyphar is this still on your radar?

cyphar:
I think on paper this is still something we'd like to have fixed for runc and umoci, but disabling argument reordering fixed the issue well enough that it isn't bothering us at the moment (though it does mean that some of the other argument reordering features now don't work because we've disabled them). I can take a crack at implementing the behaviour I mentioned above (I think I might still have the code on my laptop) but I'd like to know whether there is interest in taking it -- you've mentioned before that touching this stuff can be quite hairy and it's not clear whether fixing a bug for one user will go on to cause bugs for another five (and as a fellow maintainer I get where you're coming from!).

meatballhat:
Superseded by #1356

PR #1141, opened 2020-05-21
Allow default values to be set for TimestampFlag

tonglil:
## What type of PR is this?

- [x] bug
- [ ] cleanup  
- [ ] documentation
- [ ] feature

Currently, TimestampFlags are not taking their default `Value` set in the Flag slice, and is nil even if `Value` is set to `cli.NewTimestamp(time.Now())`.

Used https://github.com/urfave/cli/blob/477292c8d462a3f51cd18bc77c0542193a62274d/flag_string.go#L58-L74 as a reference.

## Testing

Tested locally with a fork.

## Release Notes

```release-note
Fix setting a default `Value` for `TimestampFlag`.
```

### DIFF
--- a/flag_timestamp.go
+++ b/flag_timestamp.go
@@ -45,7 +45,7 @@ func (t *Timestamp) Set(value string) error {
 
 // String returns a readable representation of this value (for usage defaults)
 func (t *Timestamp) String() string {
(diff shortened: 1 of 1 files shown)


meatballhat:
@tonglil Sorry for the big delay! AFAICT this is no longer an issue since https://github.com/urfave/cli/commit/7ecfd216

PR #1153, opened 2020-06-17
Bugfix: Remove version flag from commands with subcommands

### LINKED ISSUE #1073: v2 bug: Do Not Make -v A Global Flag Or At Least Prevent Automatic Inclusion In Subcommands Of The Version Flag
## my urfave/cli version is

`v2.1.1`

## Checklist

## Dependency Management

## Describe the bug

The `-v` flag is used as a shorthand for `--version`. For example if I have a binary called `foo`, `foo -v` can be used to print the version. Now lets say I have a subcommand `bar` with a `verbose` flag that has a `v` alias. Running this subcommand will cause a panic, with an error message indicating that the `v` flag is redefined. Because the `version` flag is reimported for all commands and subcommands.

## To reproduce

Create a CLI binary that has a subcommand with any kind of flag that uses either `v` as the flag name, or as the flag alias. I would provide sample code but honestly im lazy and the project is closed-source lol

## Observed behavior

Program crashes with a panic

## Expected behavior

Allow me to invoke my `-v` aliased subcommands.

## Additional context

Personally I don't think the `-v` aliased `--version` flag should be imported to subcommands. Really it should *only* be available when the CLI binary is invoked without any commands.

With a binary `foo` `foo subcommand -v` should not be invoking the `--version` flag. I can't think of anyone in the history of computers who would invoke `-v` with a subcommand to see the version of the CLI binary. It is significantly more keystrokes to get the version by `foo subcommand -v` than it is to get the version by `foo -v`.

Additionally it is **extremely** common for CLI binaries to use `-v` with subcommands to signify `verbose` output. That is actually how I discovered the bug because one of our subcommands is attempting to use `-v` to signify verbose output.

Currently the solution for this is to alias the `verbose` flag with `vv` instead of `v`, but I would really like to be able to use `v` because typically unix/linux CLI uses increasing counts of `v` (ie `vv` and `vvv`) to signify increasingly verbose logs.

## Want to fix this yourself?

I would be willing to fix this myself but might need some guidance as to the part of the codebase to look at.

```
go version go1.13.4 linux/amd64
```

```
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/solidity/.cache/go-build"
GOENV="/home/solidity/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GONOPROXY="github.com/RTradeLtd/*"
GONOSUMDB="github.com/RTradeLtd/*"
GOOS="linux"
GOPATH="/home/solidity/go"
GOPRIVATE="github.com/RTradeLtd/*"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/home/solidity/Code/RTradeLtd/TemporalX/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build521563016=/tmp/go-build -gno-record-gcc-switches"
```

Irioth:
## What type of PR is this?

- [X] bug
- [ ] cleanup  
- [ ] documentation
- [ ] feature

This PR removes unexpected version flag on commands with subcommands.

Fixes #1073

## Testing

test added

## Release Notes

```release-note
fix: remove unexpected version flag on commands with subcommands
```

### DIFF
--- a/command_test.go
+++ b/command_test.go
@@ -389,9 +389,31 @@ func TestCommand_NoVersionFlagOnCommands(t *testing.T) {
 				HideHelp:    true,
 				Action: func(c *Context) error {
 					if len(c.App.VisibleFlags()) != 0 {
-						t.Fatalf("unexpected flag on command")
+						t.Fatal("unexpected flag on command")
 					}
 					return nil
+				},
+			},
+		},
+	}
+
+	err := app.Run([]string{"foo", "bar"})
+	expect(t, err, nil)
+}
+
+func TestCommand_CanAddVFlagOnCommands(t *testing.T) {
+	app := &App{
+		Version: "some version",
+		Writer:  ioutil.Discard,
+		Commands: []*Command{
+			{
+				Name:        "bar",
+				Usage:       "this is for testing",
+				Subcommands: []*Command{{}}, // some subcommand
+				Flags: []Flag{
+					&BoolFlag{
+						Name: "v",
+					},
 				},
 			},
 		},
--- a/command_test.go
+++ b/command_test.go
@@ -386,11 +386,10 @@ func TestCommand_NoVersionFlagOnCommands(t *testing.T) {
 				Name:        "bar",
 				Usage:       "this is for testing",
 				Subcommands: []*Command{{}}, // some subcommand
+				HideHelp:    true,
 				Action: func(c *Context) error {
-					for _, f := range c.App.VisibleFlags() {
-						if f == VersionFlag {
-							t.Fatalf("unexpected version flag")
-						}
+					if len(c.App.VisibleFlags()) != 0 {
+						t.Fatalf("unexpected flag on command")
 					}
 					return nil
 				},
--- a/command_test.go
+++ b/command_test.go
@@ -377,3 +377,27 @@ func TestCommand_Run_CustomShellCompleteAcceptsMalformedFlags(t *testing.T) {
 	}
 
 }
+
+func TestCommand_NoVersionFlagOnCommands(t *testing.T) {
+	app := &App{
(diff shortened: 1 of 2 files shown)


command_test.go
```diff
@@ -377,3 +377,26 @@ func TestCommand_Run_CustomShellCompleteAcceptsMalformedFlags(t *testing.T) {
...
+		Version: "some version",
+		Commands: []*Command{
+			{
+				Name:        "bar",
+				Usage:       "this is for testing",
+				Subcommands: []*Command{{}}, // some subcommand
+				HideHelp:    true,
+				Action: func(c *Context) error {
+					if len(c.App.VisibleFlags()) != 0 {
+						t.Fatalf("unexpected flag on command")
```
rliebz, line 392: Nit: Since there's no formatting here, `t.Fatal` is more appropriate.

Irioth: fixed

Irioth:
> Looks good, this change makes sense to me.
> 
> Since this is targeting #1073, would you mind adding a test case demonstrating that it is now possible to add a `-v` flag to sub-commands?

Sure, added such one.

PR #1163, opened 2020-07-22
make missing required flags an ExitCoder so that app exit code handling treats this as an error

cmstrickland:
## What type of PR is this?

- [x] bug
- [ ] cleanup  
- [ ] documentation
- [x] feature

modifies errRequiredFlags so that it can be used as an ExitCoder, and then changes the error handling in app.Run so that it does not bypass the osExit handling

I just picked a value to use for the exit status, it is hard coded
This patch is only against v1, I am not currently using v2, unsure if it is applicable there. 

## Testing

I added a simple  test to app_test that the handler is called when a required flag is not supplied

## Release Notes

```release-note
if a flag is defined as Required but is not present, app.Run will call os.Exit with 127 

```

### DIFF
--- a/app.go
+++ b/app.go
@@ -244,6 +244,7 @@ func (a *App) Run(arguments []string) (err error) {
 
 	cerr := checkRequiredFlags(a.Flags, context)
 	if cerr != nil {
+		a.handleExitCoder(context, cerr)
 		_ = ShowAppHelp(context)
 		return cerr
 	}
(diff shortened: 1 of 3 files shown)


cmstrickland:
> v1 is in maintenance mode and this is a very very small new feature, and a potentially breaking one at that. So I'm not sure it's good idea for us to merge this.
> 
> I could be convinced otherwise, though!

Thank you for the review! I can understand this point of view. I don't think this changes is much of a breaking one though, specific to this case, `Required` attribute is quite recent feature addition to 1.x (I think it is introduced with 1.22.1 flag rework) 

I do find it difficult to imagine who is relying on the specific implementation of required flags, particularly in expecting a successful exit code for an failure to run, but this is the nature of changing anything established, there's no way to predict every user's expectations, I grant. I am quite happy for it to be either merged or closed, and once again, thanks for your time and efforts.

Issue #1167, opened 2020-07-30
When parsing an argument from a file or env var, the error should mention the source


mostynb:
## Checklist

* [X] Are you running the latest v2 release? The list of releases is [here](https://github.com/urfave/cli/releases).
* [X] Did you check the manual for your release? The v2 manual is [here](https://github.com/urfave/cli/blob/master/docs/v2/manual.md)
* [X] Did you perform a search about this feature? Here's the [Github guide](https://help.github.com/en/github/managing-your-work-on-github/using-search-to-filter-issues-and-pull-requests) about searching.

## What problem does this solve?

I have had a few bug reports from users that run a binary that uses urfave/cli in orchestrated/containerised environments which automatically set some environment variables with default values, and those environment variable names match some of those used by cli in my application. The application exits with a "could not parse ..." error, but the user confused because they did not specify that value anywhere themselves.

## Solution description

When a parse error fails, we should state where the value was obtained (since multiple environment variables and files are allowed as the source of a single flag).

## Describe alternatives you've considered

We could remove the possibility of allowing multiple sources (environment variables/files) for a given flag, but that would break existing users.

mostynb:
Bump.

mostynb:
Bump.

mostynb:
Bump.

mostynb:
#1168 was merged (thanks!) - closing this.

PR #1168, opened 2020-07-30
Report the source of a value when we cannot parse it

mostynb:
## What type of PR is this?

- feature

## What this PR does / why we need it

If you allow a flag to be set from environment variables or files and a parse error occurs from one of them, it is very useful for the error message to mention where the value came from.

Without this, it can be difficult to notice an error caused by an unexpected environment variable being set.

Implements #1167.

The existing function `flagFromEnvOrFile` takes a slice of strings (environment variable names) and a file path, and checks each to see if they exist. If they do, then it returns the value and a bool which is true if a value was found.

This patch modifies it to also return a string description of where the value was found. Callers have been updated to use this extra value when reporting errors.

Implements #1167.

I haven't quite figured out what should be done in flag_path.go and flag_string.go yet- any advice?

## Testing

Existing tests have been updated to expect the new error strings.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/flag.go
+++ b/flag.go
@@ -372,17 +372,20 @@ func hasFlag(flags []Flag, fl Flag) bool {
 	return false
 }
 
-func flagFromEnvOrFile(envVars []string, filePath string) (val string, ok bool, source string) {
+// Return the first value from a list of environment variables and files
+// (which may or may not exist), a description of where the value was found,
+// and a boolean which is true if a value was found.
+func flagFromEnvOrFile(envVars []string, filePath string) (value string, fromWhere string, found bool) {
 	for _, envVar := range envVars {
 		envVar = strings.TrimSpace(envVar)
-		if val, ok := syscall.Getenv(envVar); ok {
-			return val, true, fmt.Sprintf("from environment variable %q", envVar)
+		if value, found := syscall.Getenv(envVar); found {
+			return value, fmt.Sprintf("environment variable %q", envVar), true
 		}
 	}
 	for _, fileVar := range strings.Split(filePath, ",") {
 		if data, err := ioutil.ReadFile(fileVar); err == nil {
-			return string(data), true, fmt.Sprintf("from file %q", filePath)
+			return string(data), fmt.Sprintf("file %q", filePath), true
 		}
 	}
-	return "", false, ""
+	return "", "", false
 }
--- a/flag.go
+++ b/flag.go
@@ -372,17 +372,17 @@ func hasFlag(flags []Flag, fl Flag) bool {
 	return false
 }
 
-func flagFromEnvOrFile(envVars []string, filePath string) (val string, ok bool) {
+func flagFromEnvOrFile(envVars []string, filePath string) (val string, ok bool, source string) {
 	for _, envVar := range envVars {
 		envVar = strings.TrimSpace(envVar)
 		if val, ok := syscall.Getenv(envVar); ok {
-			return val, true
+			return val, true, fmt.Sprintf("from environment variable %q", envVar)
 		}
 	}
 	for _, fileVar := range strings.Split(filePath, ",") {
 		if data, err := ioutil.ReadFile(fileVar); err == nil {
-			return string(data), true
+			return string(data), true, fmt.Sprintf("from file %q", filePath)
 		}
 	}
-	return "", false
+	return "", false, ""
 }
--- a/flag_path.go
+++ b/flag_path.go
@@ -56,7 +56,8 @@ func (f *PathFlag) GetValue() string {
 
 // Apply populates the flag given the flag set and environment
 func (f *PathFlag) Apply(set *flag.FlagSet) error {
-	if val, ok := flagFromEnvOrFile(f.EnvVars, f.FilePath); ok {
+	// TODO: how to report the source of parse errors?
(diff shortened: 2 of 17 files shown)


mostynb:
Bump.

asahasrabuddhe (changes requested):
Instead of returning a source from the function `flagFromEnvOrFile` why don't you directly return an error ? The function can return a value and error instead of value and boolean. This error can then directly returned by the `Apply` function.

mostynb:
> Instead of returning a source from the function `flagFromEnvOrFile` why don't you directly return an error ? The function can return a value and error instead of value and boolean. This error can then directly returned by the `Apply` function.

@asahasrabuddhe `flagFromEnvOrFile` just returns the first raw/unparsed string value from multiple possible sources that exist- the only possible error it can return with the current input type signature is "no value found in any of the possible sources", which is why it currently returns a bool instead of an error.

At the moment, the parse error checking is performed by the caller, eg flag_bool.go calls `flagFromEnvOrFile` and tries to parse the value as a bool, flag_float64.go tries to parse the value as a float64 and so on. If we want `flagFromEnvOrFile` to return a parse/invalid format error, it would need to take an empty interface argument, and have big type switch for every possible format (and hope that we don't forget any types). This doesn't sound nice to me- is this what you want?

flag.go
```diff
@@ -372,17 +372,17 @@ func hasFlag(flags []Flag, fl Flag) bool {
 	return false
 }
 
-func flagFromEnvOrFile(envVars []string, filePath string) (val string, ok bool) {
+func flagFromEnvOrFile(envVars []string, filePath string) (val string, ok bool, source string) {
```
rliebz, line 375: Much like returning errors, the "comma ok" idiom typically puts the `ok bool` as the _last_ value of a parameter:

```go
func flagFromEnvOrFile(envVars []string, filePath string) (val, source string, ok bool) {
```

flag_path.go
```diff
@@ -56,7 +56,8 @@ func (f *PathFlag) GetValue() string {
 
 // Apply populates the flag given the flag set and environment
 func (f *PathFlag) Apply(set *flag.FlagSet) error {
-	if val, ok := flagFromEnvOrFile(f.EnvVars, f.FilePath); ok {
+	// TODO: how to report the source of parse errors?
```
rliebz, line 59: My question here is—what parse errors?

It seems like the source is always reported as part of an error in the `Appy` function, but if this function doesn't return an error, I'm not sure we're meant to do anything.

flag_string.go
```diff
@@ -57,7 +57,8 @@ func (f *StringFlag) GetValue() string {
 
 // Apply populates the flag given the flag set and environment
 func (f *StringFlag) Apply(set *flag.FlagSet) error {
-	if val, ok := flagFromEnvOrFile(f.EnvVars, f.FilePath); ok {
+	// TODO: how to report source?
```
rliebz, line 60: Question: Report source in what situation? Like the above, if there aren't any errors here, I don't think we have anything to report.

rliebz:
I agree with the assessment that the boolean is more appropriate to return in this case. We're not interested in reporting an error if we fail to parse a value—we're interested in reporting an error if the value that we parsed in the helper function is invalid. 

The signature is a bit awkward as is, but I don't think there's a great way around it. If we're looking for something more well-structured, we could go with a custom type:

```go
type source struct {
  Type string // this could also be custom type, if we're really into custom types
  Name string
}

func (s source) String() string {
  // `environment variable "FOO"` or `file "bar.txt"`
  return fmt.Sprintf("%s %q", s.Type, s.Name)
}

func flagFromEnvOrFile(envVars []string, filePath string) (string, source, bool) {
  // ...
    return val, source{Type: "environment variable", Name: envVar}, true
  // ...
    return string(data), source{Type: "file", Name: filePath}, true
  // ...
  return "", source{}, false
}
```

It feels like a fair bit of ceremony, but it seems more obvious on a read why the signature looks the way it does. Just throwing the idea out there, not sure it's necessarily better.

flag_string.go
mostynb: Removed.

flag_path.go
mostynb: Removed.

flag.go
mostynb: Moved it to the end, and added a comment for the function.

mostynb:
> I agree with the assessment that the boolean is more appropriate to return in this case. We're not interested in reporting an error if we fail to parse a value—we're interested in reporting an error if the value that we parsed in the helper function is invalid.
> 
> The signature is a bit awkward as is, but I don't think there's a great way around it. If we're looking for something more well-structured, we could go with a custom type:
...
> It feels like a fair bit of ceremony, but it seems more obvious on a read why the signature looks the way it does. Just throwing the idea out there, not sure it's necessarily better.

I'm not sure what benefit that would give us. In the end all we want is a string description of where the value came from for an error message.

I added a function comment in the latest fixup, maybe that improves the readability while keeping the simpler form- what do you think?

rliebz:
> I'm not sure what benefit that would give us.

Types are documentation. But variable names are documentation, and documentation is documentation. I'm happy with the current iteration.

rliebz (dismissed):

mostynb:
Ping.

PR #1171, opened 2020-08-06
feat(docs): add UsageText to docs output for markdown and man page generation

clok:
## What type of PR is this?

_(REQUIRED)_

- [ ] bug
- [ ] cleanup  
- [x] documentation
- [X] feature

_(REQUIRED)_

As a tools developer, I find the `UsageText` field to be useful in providing details on what a command does and how to use it. A good example is with [gwsm](https://github.com/GoodwayGroup/gwsm)

For example, the command `gwsm diff` has as detailed description that helps the user know how to use the command.
```
$ gwsm diff --help
NAME:
   gwsm diff - View diff of local vs. namespace

USAGE:

View the diff of the local environment against a given command running on a pod within a namespace.

This will retrieve the stored secrets within AWS Secrets Manager and map them via
the secrets.yml file used by the 'summon' CLI tool to generate the current state of
Environment Variables for a given stage.

The AWS Secrets Manager names are assumed to be stored as '<SECRETS_GROUP>_NAME'
in the ConfigMap. Example: 'RDS_SECRET_NAME: rds/staging/service-yolo'

From the root of the service, the required files are typically found below:

The path to the configmap.yaml file is within the kubernetes deployment.
This is typically .kube/<stage>/05-configmap.yaml

The path to the secrets.yml is typically .docker/secrets.yaml

It will then grab current environment for a specific process running within a Pod in a given Namespace.

This is achieved by inspecting the /proc/<PID>/environ for the given process. This method uses
'/bin/bash -c' as the base command to perform the PID inspection via 'ps faux'.

The 'filter-prefix' flag will exclude any values that start with the flagged prefixes from display.

The 'exclude' flag will exclude any values where the KEY matches exactly from display.


OPTIONS:
   --secrets value, -s value        Path to secrets.yml (default: ".docker/secrets.yml")
   --configmap value, -c value      Path to configmap.yaml
   --namespace value, -n value      Kube Namespace list Pods from
   --cmd value                      Command to inspect (default: "node")
   --filter-prefix value, -f value  List of prefixes (csv) used to filter values from display. Set to "" to remove any filters. (default: "npm_,KUBERNETES_,API_PORT")
   --exclude value                  List (csv) of specific env vars to exclude values from display. Set to "" to remove any exclusions. (default: "PATH,SHLVL,HOSTNAME")
   --secret-suffix value            Suffix used to find ENV variables that denote the Secret Manager Secrets to lookup (default: "_NAME")
   --help, -h                       show help (default: false)
```

I would like for the docs generation (`ToMarkdown` and `ToMan`) to include this help text in the docs.

See the before and after here:

| | before | after |
|---|---|---|
| markdown | https://gist.github.com/clok/ca1452874e372b02fe597f8ad65ca7f2 |  https://gist.github.com/clok/fe5e00c973575778f8e548263b5dba44 |
| man | https://gist.github.com/clok/4ad90b19868709c86612ff76ee0e1350 | https://gist.github.com/clok/a810114a454beee90d5134e9c8c533e8 |

_(REQUIRED)_

N/A - New feature

## Testing

Updated a local copy of an existing tool ([gwsm](https://github.com/GoodwayGroup/gwsm)) with the delta applied and generated before/after docs.

See the before and after here:

| | before | after |
|---|---|---|
| markdown | https://gist.github.com/clok/ca1452874e372b02fe597f8ad65ca7f2 |  https://gist.github.com/clok/fe5e00c973575778f8e548263b5dba44 |
| man | https://gist.github.com/clok/4ad90b19868709c86612ff76ee0e1350 | https://gist.github.com/clok/a810114a454beee90d5134e9c8c533e8 |

- Updated test suite
- Generated local examples

## Release Notes

_(REQUIRED)_

```release-note
Add UsageText to docs output for markdown and man page generation
```

### DIFF
--- a/docs.go
+++ b/docs.go
@@ -68,15 +68,16 @@ func prepareCommands(commands []*Command, level int) []string {
 		if command.Hidden {
 			continue
 		}
-		usage := ""
-		if command.Usage != "" {
-			usage = command.Usage
-		}
 
-		prepared := fmt.Sprintf("%s %s\n\n%s\n",
+		usageText := prepareUsageText(command)
+
+		usage := prepareUsage(command, usageText)
+
+		prepared := fmt.Sprintf("%s %s\n\n%s%s",
 			strings.Repeat("#", level+2),
 			strings.Join(command.Names(), ", "),
 			usage,
+			usageText,
 		)
 
 		flags := prepareArgsWithValues(command.Flags)
@@ -154,4 +155,36 @@ func flagDetails(flag DocGenerationFlag) string {
 		description += " (default: " + value + ")"
 	}
 	return ": " + description
+}
+
+func prepareUsageText(command *Command) string {
+	usageText := ""
+	if command.UsageText != "" {
+		// Remove leading and trailing newlines
+		preparedUsageText := strings.TrimSuffix(command.UsageText, "\n")
+		preparedUsageText = strings.TrimPrefix(preparedUsageText, "\n")
+
+		if strings.Contains(preparedUsageText, "\n") {
+			// Format multi-line string as a code block
+			usageText = fmt.Sprintf("```\n%s\n```\n", preparedUsageText)
+		} else {
+			// Style a single line as a note
+			usageText = fmt.Sprintf(">%s\n", preparedUsageText)
+		}
+	}
+	return usageText
+}
+
+func prepareUsage(command *Command, usageText string) string {
+	usage := ""
+	if command.Usage != "" {
+		usage = fmt.Sprintf("%s\n", command.Usage)
+	}
+
+	// Add a newline to the Usage IFF there is a UsageText
+	if usageText != "" && usage != "" {
+		usage = fmt.Sprintf("%s\n", usage)
+	}
+
+	return usage
 }
--- a/docs.go
+++ b/docs.go
@@ -158,32 +158,37 @@ func flagDetails(flag DocGenerationFlag) string {
 }
 
 func prepareUsageText(command *Command) string {
-	usageText := ""
-	if command.UsageText != "" {
-		// Remove leading and trailing newlines
-		preparedUsageText := strings.TrimSuffix(command.UsageText, "\n")
-		preparedUsageText = strings.TrimPrefix(preparedUsageText, "\n")
-
-		if strings.Contains(preparedUsageText, "\n") {
-			// Format multi-line string as a code block
-			usageText = fmt.Sprintf("```\n%s\n```\n", preparedUsageText)
-		} else {
-			// Style a single line as a note
-			usageText = fmt.Sprintf(">%s\n", preparedUsageText)
-		}
+	if command.UsageText == "" {
+		return ""
 	}
+
+	// Remove leading and trailing newlines
+	preparedUsageText := strings.Trim(command.UsageText, "\n")
+
+	var usageText string
+	if strings.Contains(preparedUsageText, "\n") {
+		// Format multi-line string as a code block using the 4 space schema to allow for embedded markdown such
+		// that it will not break the continuous code block.
+		for _, ln := range strings.Split(preparedUsageText, "\n") {
+			usageText += fmt.Sprintf("    %s\n", ln)
+		}
+	} else {
+		// Style a single line as a note
+		usageText = fmt.Sprintf(">%s\n", preparedUsageText)
+	}
+
 	return usageText
 }
 
 func prepareUsage(command *Command, usageText string) string {
-	usage := ""
-	if command.Usage != "" {
-		usage = fmt.Sprintf("%s\n", command.Usage)
+	if command.Usage == "" {
+		return ""
 	}
 
+	usage := command.Usage + "\n"
 	// Add a newline to the Usage IFF there is a UsageText
-	if usageText != "" && usage != "" {
-		usage = fmt.Sprintf("%s\n", usage)
+	if usageText != "" {
+		usage += "\n"
 	}
 
 	return usage
--- a/docs_test.go
+++ b/docs_test.go
@@ -309,7 +309,7 @@ func Test_prepareUsage(t *testing.T) {
 		res := prepareUsage(&cmd, "")
 
 		// Then
-		expect(t, res, cmd.Usage + "\n")
+		expect(t, res, cmd.Usage+"\n")
 	})
 
 	t.Run("simple Usage with UsageText", func(t *testing.T) {
@@ -320,6 +320,6 @@ func Test_prepareUsage(t *testing.T) {
 		res := prepareUsage(&cmd, "a non-empty string")
 
 		// Then
-		expect(t, res, cmd.Usage + "\n\n")
+		expect(t, res, cmd.Usage+"\n\n")
 	})
 }
--- a/docs_test.go
+++ b/docs_test.go
@@ -67,6 +67,41 @@ func testApp() *App {
 	}, {
 		Name:   "hidden-command",
 		Hidden: true,
+	}, {
+		Aliases: []string{"u"},
+		Flags: []Flag{
+			&StringFlag{
+				Name:      "flag",
+				Aliases:   []string{"fl", "f"},
+				TakesFile: true,
+			},
+			&BoolFlag{
+				Name:    "another-flag",
+				Aliases: []string{"b"},
+				Usage:   "another usage text",
+			},
+		},
+		Name:  "usage",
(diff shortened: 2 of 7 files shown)


paulojblack (dismissed):

clok:
@rliebz and @lynncyrin any chance that this can get merged?

docs.go
```diff
@@ -146,3 +147,35 @@ func flagDetails(flag DocGenerationFlag) string {
 	}
 	return ": " + description
 }
+
+func prepareUsageText(command *Command) string {
+	usageText := ""
+	if command.UsageText != "" {
```
rliebz, line 153: Here we've got the entire body of a function inside a single `if` statement. If we flip the conditional with an early return, we'd be able to outdent the business logic and get the complexity down:

```go
func prepareUsageText(command *Command) string {
  if command.UsageText == "" {
    return ""
  }

  // Logic is now at the outermost level
  preparedUsageText := strings.TrimSuffix(command.UsageText, "\n")
  // ...

  return usageText
}

File: docs.go [hunk 2]
```diff
@@ -146,3 +147,35 @@ func flagDetails(flag DocGenerationFlag) string {
...
+		// Remove leading and trailing newlines
+		preparedUsageText := strings.TrimSuffix(command.UsageText, "\n")
+		preparedUsageText = strings.TrimPrefix(preparedUsageText, "\n")
+
+		if strings.Contains(preparedUsageText, "\n") {
+			// Format multi-line string as a code block
+			usageText = fmt.Sprintf("```\n%s\n```\n", preparedUsageText)
```
rliebz, line 160: A usage text that already contains a markdown code block will break here. Imagine something like this:

    See code block below:

    ```
    func() { ... }
    ```

    More text here ...


That would be rendered as follows, placing the prefix and suffix in code blocks, and having the inner code block be rendered as regular text.

    ```
    See code block below:

    ```
    func() { ... }
    ```

    More text here ...
    ```


Luckily the markdown spec provides an alternative—text indented to 4 spaces is considered part of a code block. That should allow us to safely turn an entire section into a block of code without worrying about special characters

docs.go
```diff
@@ -146,3 +147,35 @@ func flagDetails(flag DocGenerationFlag) string {
...
+			// Style a single line as a note
+			usageText = fmt.Sprintf(">%s\n", preparedUsageText)
+		}
+	}
+	return usageText
+}
+
+func prepareUsage(command *Command, usageText string) string {
+	usage := ""
+	if command.Usage != "" {
```
rliebz, line 171: Since both code blocks depend on `command.Usage` not being empty, we have the same situation as `prepareUsageText`, where the entire business logic of the function is nested under a single conditional branch. 

We can simplify the entire body to this:

```go
if command.Usage == "" {
  return ""
}

usage := command.Usage + "\n"
if usageText != "" {
  usage += "\n"
}

return usage
```

docs.go
```diff
@@ -146,3 +147,35 @@ func flagDetails(flag DocGenerationFlag) string {
 	}
 	return ": " + description
 }
+
+func prepareUsageText(command *Command) string {
```
rliebz, line 151: There's a fair bit of conditional logic in this PR for things like whether `usage` exists, whether `usage text` exists, and whether text spans multiple lines, but our current test suite doesn't capture that nuance.

Would you mind adding tests to cover the new logic?

rliebz (changes requested):
Thanks for the contribution, and sorry for the delay here. Got a few comments for you, but overall this looks great!

clok:
Thank you very much for the feedback and suggestions. I will get to updating tests and responding to the feedback. I had hoped to get to it last week, but life and work kept pulling me away. Sorry for the delay!

docs.go
```diff
@@ -146,3 +147,35 @@ func flagDetails(flag DocGenerationFlag) string {
...
+	if command.UsageText != "" {
+		// Remove leading and trailing newlines
+		preparedUsageText := strings.TrimSuffix(command.UsageText, "\n")
+		preparedUsageText = strings.TrimPrefix(preparedUsageText, "\n")
+
+		if strings.Contains(preparedUsageText, "\n") {
+			// Format multi-line string as a code block
+			usageText = fmt.Sprintf("```\n%s\n```\n", preparedUsageText)
+		} else {
+			// Style a single line as a note
```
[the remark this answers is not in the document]
clok: It was a styling opinion more than anything. The use case I was considering was a very long one-line comment. Say 200+ characters in length. Within markdown there is no styling for the wrapping rules on a pre-formatted text block. This is particularly pronounced on Github where the CSS rules for a text block are quite narrow. Take the case below as an example.

> This will retrieve the stored secrets within AWS Secrets Manager and map them via the secrets.yml file used by the 'summon' CLI tool to generate the current state of Environment Variables for a given stage.

versus

```
This will retrieve the stored secrets within AWS Secrets Manager and map them via the secrets.yml file used by the 'summon' CLI tool to generate the current state of Environment Variables for a given stage.
```

With a `>` note syntax, the markdown render will wrap the text accordingly. With the pre-formatted text block, it scrolls off screen making it difficult to read.

I dabbled with using `>` across many lines, but there were quite a few edge cases that made predicting the desired markdown format unruly.

docs.go
```diff
@@ -146,3 +147,35 @@ func flagDetails(flag DocGenerationFlag) string {
 	}
 	return ": " + description
 }
+
+func prepareUsageText(command *Command) string {
+	usageText := ""
+	if command.UsageText != "" {
+		// Remove leading and trailing newlines
+		preparedUsageText := strings.TrimSuffix(command.UsageText, "\n")
```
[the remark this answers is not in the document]
clok:
```go
preparedUsageText := strings.Trim(command.UsageText, "\n")
```
That is what I was looking for. Thank you for the suggestion! It was intended to remove all leading and trailing newlines.

docs.go
clok: Awesome catch. Updated with the 4 space alternative and it looks to be working well. I am wrapping up tests and should push soon.

docs.go
clok: Yep. Updated with new tests.

clok:
@rliebz Rebased with current `master` and updated with your feedback. Thank you again for the suggestions and comments.  I noticed that the github actions are failing, but I can't tell if it is due to the changes I pushed. I am happy to address any issues that I can. Just let me know.

clok:
I put together an extension of sorts that accomplishes this feature (and more). https://github.com/clok/cdocs

I would really like to put these features in the mainline as I feel they are useful for all `cli` users. The plan was to get this initial work through the process and then extend from there. 

I have updated the code, fixed the linting errors. Waiting for a review from maintainers. Thanks!

codefromthecrypt:
any chance this can be reconsidered? It seems a pain to have to do external markdown generation to accomplish the same meanwhile, even if there's an add-on

rliebz (approved):
Sorry again for the delays here! This looks great, thanks for the contribution!

clok:
Awesome! Happy to see it accepted.

codefromthecrypt:
I think this may have drifted the `-h` output.  As you'll notice, when multiple lines are present, it doesn't make a consistent indent.

```
USAGE:
   The '<version>' is from the "versions" command and installed if necessary.
The '<args>' are interpreted by Envoy.
The Envoy working directory is archived as $GETENVOY_HOME/runs/$epochtime.tar.gz upon termination.

Example:
$ getenvoy run 1.18.3 --config-path ./bootstrap.yaml
```

codefromthecrypt:
the above was from this..

```go
		Usage:     "Download and install a <version> of Envoy",
		ArgsUsage: "<version>",
		UsageText: fmt.Sprintf(`The '<version>' is from the "versions" command.
The Envoy <version> will be installed into $GETENVOY_HOME/versions/<version>

Example:
$ getenvoy install %s`, version.LastKnownEnvoy),
```

rliebz:
@codefromthecrypt Would you mind opening an issue for that? If there's a bug in the behavior here, it's easier to keep track of if we have an issue with the expected behavior and steps to reproduce.

codefromthecrypt:
@rliebz sure https://github.com/urfave/cli/issues/1278

rliebz:
Thanks!

clok:
Nice catch on the regression. I'm sorry about that unintended change. I'll take a look at the issue and if it hasn't been addressed yet, I'll work on a fix for it.

clok:
Got a PR up (#1279) to address the issue raised by @codefromthecrypt

PR #1174, opened 2020-08-17
Added Metadata to the `Command` struct.

imle:
Since App based metadata was already an option, it seemed like the correct way to add the same functionality to the Commands.

## What type of PR is this?

- [ ] bug
- [ ] cleanup  
- [ ] documentation
- [X] feature

This PR attempts to add a per-command metadata system to give commands created throughout a codebase the ability to specify info that is used by the app as a whole but only relevant to their execution. For instance, a system that takes an arbitrary interface of type `Startable` that needs to be managed in the background by the application root while a long term command is running.

No current issue is open for this.

## Testing

No testing required as the Metadata field is not used by the cli library itself.

## Release Notes

```release-note
Adds Metadata field to the command struct.
```

### DIFF
--- a/command.go
+++ b/command.go
@@ -61,6 +61,9 @@ type Command struct {
 	// cli.go uses text/template to render templates. You can
 	// render custom help text by setting this variable.
 	CustomHelpTemplate string
+
+	// Other custom info
+	Metadata map[string]interface{}
 }
 
 type Commands []*Command


rliebz:
I'm not convinced this is something that we want to add in. This is increasing the surface area of the public API, which we're extremely conservative about doing, but I don't think from the description here I have enough of an understanding of the use case to push this through.

In particular, I would like to know enough to either understand from a data flow perspective why this makes things easier, and what the alternatives look like. I'd recommend looking at the feature request issue template and making sure all the questions there are answered, since that's the traditional entry point for introducing new features like this.

saschagrunert:
I agree with @rliebz.

PR #1183, opened 2020-09-08
feat(help): add ShowSubcommandHelpAndExit

### LINKED ISSUE #1112: Add a ShowSubcommandHelpAndExit()
## Checklist

There is a `ShowAppHelpAndExit()` and a `ShowCommandHelpAndExit()` but there's currently no `ShowSubcommandHelpAndExit()`.

## Solution description

I usually use at the top of my command actions something like this:
```go
if c.NArg() < 2 {
        cli.ShowCommandHelpAndExit(c, c.Command.FullName(), 1)
}
```
So if someone doesn't supply the correct number of positional arguments they get the help and it exits with a non-zero code.

I tried to use sub-commands for the first time and found I can't do this.

I've tried implementing as:
```go
if c.NArg() < 2 {
        cli.ShowSubcommandHelp(c)
        return cli.NewExitError(nil, 1)
}
```
but that doesn't work very well.

knqyf263:
## What type of PR is this?

- [ ] bug
- [ ] cleanup  
- [ ] documentation
- [x] feature

- Add ShowSubcommandHelpAndExit to help.go
- It is also useful like ShowCommandHelpAndExit and ShowAppHelpAndExit

Fixes #1112 

I didn't find any test for XXXHelpAndExit.

## Testing

I prepared a sample program.

## Release Notes

```release-note
Added ShowSubcommandHelpAndExit to print help for the given subcommand and exit with exit code.
```

### DIFF
--- a/help.go
+++ b/help.go
@@ -214,6 +214,12 @@ func ShowCommandHelp(ctx *Context, command string) error {
 	return nil
 }
 
+// ShowSubcommandHelpAndExit - Prints help for the given subcommand and exits with exit code.
+func ShowSubcommandHelpAndExit(c *Context, exitCode int) {
+	_ = ShowSubcommandHelp(c)
+	os.Exit(exitCode)
+}
+
 // ShowSubcommandHelp prints help for the given subcommand
 func ShowSubcommandHelp(c *Context) error {
 	if c == nil {


rliebz:
I'm not a huge fan of increasing the surface area of the public API here to turn a two-liner into a one-liner. 

Is there a reason why doing these calls manually is problematic?

knqyf263:
This is because `urfave/cli` already has the following public APIs. I think they are useful even though it turns a two-liner into a one-liner.
https://github.com/urfave/cli/blob/74af85a00da15038fa915e6be00fb0679c113376/help.go#L67-L71
https://github.com/urfave/cli/blob/74af85a00da15038fa915e6be00fb0679c113376/help.go#L182-L186

If `ShowSubcommandHelpAndExit` doesn't make sense, I feel like the above APIs also don't make sense, just reducing one line. I thought they are the same, but is there any difference between `ShowSubcommandHelp` and  `ShowCommandHelp`?

rliebz (approved):
Sorry for the delay here. You're right—having 2/3 of these functions is an obvious omission.

Thanks for the contribution!

PR #1198, opened 2020-10-13
Support PathFlag.TakesFile in fish completion

### LINKED ISSUE #1156: v2 bug: fishAddFileFlag doesn't work with PathFlag
## my urfave/cli version is:

2.2.0

## Checklist

## Dependency Management

## Describe the bug

By default, [`fishAddFileFlag`](https://github.com/urfave/cli/blob/d648edd48d89ef3a841b1ec75c2ebbd4de5f748f/fish.go#L160-L177) adds `-f` to the `complete` command, indicating that the input to the flag shouldn't be a filename. Naturally, a `PathFlag` should accept a filename, but `fishAddFileFlag` only checks `GenericFlag`, `StringFlag`, and `StringSliceFlag`. 

## To reproduce

Create a `main.go`:

```Go
package main

import "fmt"
import "github.com/urfave/cli/v2"

func main() {
	app := &cli.App{
		Flags: []cli.Flag{
			&cli.PathFlag{
				Name:      "path-file",
				TakesFile: true,
			},
			&cli.StringFlag{
				Name:      "string-file",
				TakesFile: true,
			},
		},
	}

	output, _ := app.ToFishCompletion()
	fmt.Println(output)
}
```

## Observed behavior

run `go run ./main.go`. The output will include:

```
complete -c  -n '__fish__no_subcommand' -f -l path-file -r
complete -c  -n '__fish__no_subcommand' -l string-file -r
```

## Expected behavior

The first line, for `path-file`, shouldn't have `-f`.

## Want to fix this yourself?

Sure! I could add another case statement for `PathFlag`, but using `reflect` might be more future-proof. Do you have a preference?

## Run `go env` and `go version`

<details>
<summary>Output of those commands</summary>

### `go version`

```
go version go1.13.6 darwin/amd64
```

### `go env`

```
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/erincalling/Library/Caches/go-build"
GOENV="/Users/erincalling/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/erincalling/go/"
GOPRIVATE=""
GOROOT="/usr/local/Cellar/go/1.13.6/libexec"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/Cellar/go/1.13.6/libexec/pkg/tool/darwin_amd64"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/erincalling/code/futzing/cli-pathflag/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/bj/13jz98m53fl992t2sn3h0q2r0000gn/T/go-build047644300=/tmp/go-build -gno-record-gcc-switches -fno-common"
```
</details>

ErinCall:
## What type of PR is this?

- [x] bug
- [ ] cleanup  
- [ ] documentation
- [ ] feature

Flags of type `PathFlag` weren't processed correctly when generating fish completions—the generated `complete` command told fish that the flag doesn't take a filename, even when `TakesFile` is true.

* `fish.go` now respects the `TakesFile` field, regardless of the flag's underlying type.
  * The first commit on this branch (5bb54ac) is a simpler solution to the original bug: it simply appends `PathFlag` to the list of explicit types in the switch statement. 
  * The second commit refactors the switch statement, using reflection rather than explicit list of types. It's more flexible, but possibly harder to follow. I added some comments to offset that caveat, but I won't complain if you prefer the simpler design.
* `fish_test.go` has a test case for a `PathFlag` with `TakesFile = true`.
* `expected-fish-full.fish` has a corresponding change in its fixture data.

Fixes #1156 

@lynncyrin had suggested a lighter-weight form of reflection, but I wasn't able to make the type assertion work. AFAIK go's type system only deals in interfaces and specific, concrete types.

## Release Notes

```release-note
fish completion now respects TakesFile on PathFlags
```

### DIFF
--- a/fish.go
+++ b/fish.go
@@ -171,6 +171,10 @@ func fishAddFileFlag(flag Flag, completion *strings.Builder) {
 		if f.TakesFile {
 			return
 		}
+	case *PathFlag:
+		if f.TakesFile {
+			return
+		}
 	}
 	completion.WriteString(" -f")
 }
--- a/fish.go
+++ b/fish.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"reflect"
 	"strings"
 	"text/template"
 )
@@ -159,17 +158,24 @@ func (a *App) prepareFishFlags(flags []Flag, previousCommands []string) []string
 }
 
 func fishAddFileFlag(flag Flag, completion *strings.Builder) {
-	val := reflect.ValueOf(flag)
-	// if flag is a non-nil pointer to a struct...
-	if val.Kind() != reflect.Invalid && val.Elem().Kind() == reflect.Struct {
-		field := val.Elem().FieldByName("TakesFile")
-		// if flag's underlying type has a bool field called TakesFile, whose value is true...
-		if field.Kind() == reflect.Bool && field.Bool() {
-			// don't append '-f'
+	switch f := flag.(type) {
+	case *GenericFlag:
+		if f.TakesFile {
+			return
+		}
+	case *StringFlag:
+		if f.TakesFile {
+			return
+		}
+	case *StringSliceFlag:
+		if f.TakesFile {
+			return
+		}
+	case *PathFlag:
+		if f.TakesFile {
 			return
 		}
 	}
-	// append '-f', indicating that arguments to this flag are *not* meant to be file paths
 	completion.WriteString(" -f")
 }
 
--- a/fish.go
+++ b/fish.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"reflect"
 	"strings"
 	"text/template"
 )
@@ -158,24 +159,17 @@ func (a *App) prepareFishFlags(flags []Flag, previousCommands []string) []string
 }
 
 func fishAddFileFlag(flag Flag, completion *strings.Builder) {
-	switch f := flag.(type) {
-	case *GenericFlag:
-		if f.TakesFile {
-			return
-		}
-	case *StringFlag:
-		if f.TakesFile {
-			return
-		}
-	case *StringSliceFlag:
-		if f.TakesFile {
-			return
-		}
-	case *PathFlag:
-		if f.TakesFile {
+	val := reflect.ValueOf(flag)
+	// if flag is a non-nil pointer to a struct...
+	if val.Kind() != reflect.Invalid && val.Elem().Kind() == reflect.Struct {
+		field := val.Elem().FieldByName("TakesFile")
+		// if flag's underlying type has a bool field called TakesFile, whose value is true...
+		if field.Kind() == reflect.Bool && field.Bool() {
+			// don't append '-f'
 			return
 		}
 	}
+	// append '-f', indicating that arguments to this flag are *not* meant to be file paths
 	completion.WriteString(" -f")
 }
 
--- a/fish_test.go
+++ b/fish_test.go
@@ -7,6 +7,10 @@ import (
 func TestFishCompletion(t *testing.T) {
 	// Given
 	app := testApp()
+	app.Flags = append(app.Flags, &PathFlag{
+		Name:      "logfile",
+		TakesFile: true,
+	})
 
 	// When
 	res, err := app.ToFishCompletion()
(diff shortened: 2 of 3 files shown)


saschagrunert (dismissed):
Nice, LGTM

fish.go
```diff
@@ -158,20 +159,17 @@ func (a *App) prepareFishFlags(flags []Flag, previousCommands []string) []string
...
-		}
-	case *StringFlag:
-		if f.TakesFile {
-			return
-		}
-	case *StringSliceFlag:
-		if f.TakesFile {
+	val := reflect.ValueOf(flag)
+	// if flag is a non-nil pointer to a struct...
+	if val.Kind() != reflect.Invalid && val.Elem().Kind() == reflect.Struct {
```
rliebz, line 164: It's not safe to call `val.Elem()` unless we know that the type has an element. 

From the Godoc:
```go
// Elem returns the value that the interface v contains
// or that the pointer v points to.
// It panics if v's Kind is not Interface or Ptr.
// It returns the zero Value if v is nil.
```

And an example:

```go
// Basic version of this function for demonstration
func mightPanic(flag interface{}) {
	val := reflect.ValueOf(flag)
	if val.Kind() != reflect.Invalid && val.Elem().Kind() == reflect.Struct {
		field := val.Elem().FieldByName("TakesFile")
		if field.Kind() == reflect.Bool && field.Bool() {
			return
		}
	}

}


func main() {
	type impl struct {
		TakesFile bool
	}
	inst := impl{TakesFile: true}
	var empty *impl
	mightPanic(&inst)             // safe
	mightPanic(empty)             // safe
	mightPanic(map[string]bool{}) // panic
	mightPanic(inst)              // panic
	mightPanic(5)                 // panic
	mightPanic("oh no")           // panic
}
```

I believe the fix is as simple as swaping in `== reflect.Ptr` for `!= reflect.Invalid`—we need the value to be `reflect.Ptr` and the element to be `reflect.Struct`, and those both mean they are not invalid. 

Still, I have a few concerns with the implementation. 

The first is that the test cases did not catch the fact that any non-pointer implementation of `Flag` would cause a panic. The other is that while we're likely to catch additional internal implementations in the future, this is not a generic implementation, it's a check to see if the flag implementation happens to be a pointer to a struct that has a particular field with a particular name and type. The good news is it's unlikely that there are any implementations for which that isn't a safe assumption, and I do think that's more of a flaw with the current design rather than with this change.

I think there's two options going forward.
1. Revert back to the first commit in this MR that adds the `PathFlag` case, which we should be pretty safe to assume works, because it's effectively extending what we were already doing
2. Keep this pattern, but write sufficient test coverage such that we can be confident that for any implementation of a flag, this produces the correct output without panicking.

fish.go
```diff
@@ -158,20 +159,17 @@ func (a *App) prepareFishFlags(flags []Flag, previousCommands []string) []string
...
-			return
-		}
-	case *StringFlag:
-		if f.TakesFile {
-			return
-		}
-	case *StringSliceFlag:
-		if f.TakesFile {
+	val := reflect.ValueOf(flag)
+	// if flag is a non-nil pointer to a struct...
```
rliebz, line 163: Nit: I'm not a huge fan of inline comments that explain what each line of code is doing.

If it's a sufficiently complex line to warrant a comment, it's sufficiently complex to warrant extracting into a function with a descriptive name. I don't think any line in this function meets that bar.

rliebz (changes requested):
Thanks for putting this together! 

Left a couple comments with some options for how to go forward. Feel free to head down either path.

fish.go
ErinCall: Great note, will do :)

ErinCall: I pushed 279c84210adff5a98f53d221d72f57f9d16b9b67 to handle unusual types, but as I began expanding the tests to exercise more possible inputs I concluded that the reflection-based approach really just introduces more edge cases than it resolves. It might be worth revisiting if the problem comes up again with another Flag implementation, but for now I've reverted this branch to the switch/case version.

rliebz (approved):

asahasrabuddhe (approved):
LGTM

saschagrunert (approved):

PR #1204, opened 2020-11-14
errors and diagnostics should go to ErrWriter

alessio:
As per POSIX's recommendations, diagnostics and error
messages should always be printed to the /dev/stderr.

Forked from https://github.com/urfave/cli/pull/1203

## What type of PR is this?

- bug

- Errors and diagnostics should always go the defined ErrorWriter (default: stderr), not stdout.

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -189,11 +189,9 @@ func ExampleApp_Run_commandHelp() {
 }
 
 func ExampleApp_Run_noAction() {
-	output := new(bytes.Buffer)
-	app := App{ErrWriter: output}
+	app := App{ErrWriter: os.Stdout}
 	app.Name = "greet"
 	_ = app.Run([]string{"greet"})
-	fmt.Println(output.String())
 	// Output:
 	// NAME:
 	//    greet - A new cli application
--- a/app_test.go
+++ b/app_test.go
@@ -1262,7 +1262,7 @@ func TestApp_SetStdin_Subcommand(t *testing.T) {
 	}
 }
 
-func TestApp_SetStdout(t *testing.T) {
+func TestApp_SetErrWriter(t *testing.T) {
 	var w bytes.Buffer
 
 	app := &App{
--- a/app_test.go
+++ b/app_test.go
@@ -735,19 +735,43 @@ func TestApp_RunAsSubcommandParseFlags(t *testing.T) {
 				},
 				Flags: []Flag{
 					&StringFlag{
-						Name:  "lang",
-						Value: "english",
-						Usage: "language for the greeting",
+						Name:    "lang",
+						Aliases: []string{"language"},
+						Value:   "english",
+						Usage:   "language for the greeting",
 					},
 				},
 				Before: func(_ *Context) error { return nil },
 			},
 		},
 	}
-	_ = a.Run([]string{"", "foo", "--lang", "spanish", "abcd"})
+	err := a.Run([]string{"", "foo", "--lang", "spanish", "abcd"})
 
+	expect(t, err, nil)
 	expect(t, cCtx.Args().Get(0), "abcd")
 	expect(t, cCtx.String("lang"), "spanish")
+
+	err = a.Run([]string{"", "foo", "--lang", "spanish", "--language", "french", "abcd"})
+
+	expect(t, err.Error(), "Cannot use two forms of the same flag: language lang")
+}
+
+func TestApp_RunAsSubCommandIncorrectUsage(t *testing.T) {
+	a := App{
+		Name: "cmd",
+		Flags: []Flag{
+			&StringFlag{Name: "--foo"},
+		},
+		Writer: bytes.NewBufferString(""),
+	}
+
+	set := flag.NewFlagSet("", flag.ContinueOnError)
+	_ = set.Parse([]string{"", "---foo"})
+	c := &Context{flagSet: set}
+
+	err := a.RunAsSubcommand(c)
+
+	expect(t, err, errors.New("bad flag syntax: ---foo"))
 }
 
 func TestApp_CommandWithFlagBeforeTerminator(t *testing.T) {
--- a/app_test.go
+++ b/app_test.go
@@ -756,24 +756,6 @@ func TestApp_RunAsSubcommandParseFlags(t *testing.T) {
 	expect(t, err.Error(), "Cannot use two forms of the same flag: language lang")
 }
 
-func TestApp_RunAsSubCommandIncorrectUsage(t *testing.T) {
-	a := App{
-		Name: "cmd",
-		Flags: []Flag{
-			&StringFlag{Name: "--foo"},
-		},
-		Writer: bytes.NewBufferString(""),
-	}
-
-	set := flag.NewFlagSet("", flag.ContinueOnError)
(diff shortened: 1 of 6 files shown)


lynncyrin:
@alessio Can you pull in the default branch to fix the failing test please 🙏🏼

alessio:
Done @lynncyrin

lynncyrin:
@alessio 👀 can you increase the code coverage?

alessio:
@lynncyrin in all honesty, I don't have too much time to spare these days. codecov/project reports an unchanged figure (71.80%). Isn't that enough?

app_test.go
```diff
@@ -182,9 +184,11 @@ func ExampleApp_Run_commandHelp() {
 }
 
 func ExampleApp_Run_noAction() {
-	app := App{}
+	output := new(bytes.Buffer)
+	app := App{ErrWriter: output}
 	app.Name = "greet"
 	_ = app.Run([]string{"greet"})
+	fmt.Println(output.String())
```
rliebz, line 191: Nit: This looks like debug code—can we remove this line?

app_test.go
```diff
@@ -944,8 +954,8 @@ func TestApp_SetStdout(t *testing.T) {
 		t.Fatalf("Run error: %s", err)
 	}
 
-	if w.Len() == 0 {
-		t.Error("App did not write output to desired writer.")
+	if w.Len() != 0 {
+		t.Error("App wrote help screen to the wrong stream.")
```
rliebz, line 958: The negation of this test case makes me think it's no longer preserving the original intent. Before this change, the assertion was that the writer was written to, whereas now, the only thing we know is that a writer is _not_ written to. 

The naming on this test isn't great here, but I think the change that preserves the functionality tested would be to define the app like so:

```go
	app := &App{
		Name:   "test",
		ErrWriter: &w,
	}
```

And then keep the original assertion

rliebz:
I think it's not unreasonable for the nature of this change that the test coverage doesn't increase, so long as it also doesn't decrease, but I'll defer to @lynncyrin's judgment here.

app_test.go
alessio, line 191:
```suggestion
```

alessio: It's not debug code and this is a testable example, i.e. after removing this line the testable example fails as this line prints to stdout what originally goes to ErrWriter :+1:

app_test.go
alessio: This has been addressed. Please review again @rliebz

app_test.go
rliebz: Got it—the example requires we print to stdout, but we're now printing to stderr by default.

There's no need for a buffer to do this:

```go
app := App{ErrWriter: os.Stdout}
```

This should print to stdout like before without the buffer or the call to `fmt.Println`.

alessio: Done

lynncyrin (approved):
This LGTM fundamentally, although I'm not entirely sure that this isn't a breaking change 🤔

alessio:
@lynncyrin Messages that used to be printed to stdout will now by default be redirected to stderr. Thus this is an ABI breaking change.

lynncyrin (changes requested):
Given that this is a breaking change, it'll have to go into v3 rather than v2. Ideally it would go into v2 with a deprecation warning, but that sounds like a lot of work for a change of this size 🤔

alessio:
@lynncyrin how can I help?

app.go
```diff
@@ -247,7 +247,7 @@ func (a *App) RunContext(ctx context.Context, arguments []string) (err error) {
 	nerr := normalizeFlags(a.Flags, set)
 	context := NewContext(a, set, &Context{Context: ctx})
 	if nerr != nil {
-		_, _ = fmt.Fprintln(a.Writer, nerr)
+		_, _ = fmt.Fprintln(a.ErrWriter, nerr)
```
dearchap, line 250: ErrWriter should default to os.Stderr is user does not specify a value

app_test.go
```diff
@@ -99,6 +99,7 @@ func ExampleApp_Run_appHelp() {
 		Name:        "greet",
 		Version:     "0.1.0",
 		Description: "This is how we describe greet the app",
+		ErrWriter:   os.Stdout,
```
dearchap, line 102: change to os.Stderr.

app_test.go
```diff
@@ -182,7 +184,7 @@ func ExampleApp_Run_commandHelp() {
 }
 
 func ExampleApp_Run_noAction() {
-	app := App{}
+	app := App{ErrWriter: os.Stdout}
```
dearchap, line 187: os.Stderr

app_test.go
```diff
@@ -210,6 +212,7 @@ func ExampleApp_Run_subcommandNoAction() {
 				Description: "This is how we describe describeit the function",
 			},
 		},
+		ErrWriter: os.Stdout,
```
dearchap, line 215: Stderr

app_test.go
```diff
@@ -233,6 +236,7 @@ func ExampleApp_Run_bashComplete_withShortFlag() {
 	app := NewApp()
 	app.Name = "greet"
 	app.EnableBashCompletion = true
+	app.ErrWriter = os.Stdout
```
dearchap, line 239: Stderr

app_test.go
```diff
@@ -457,19 +465,25 @@ func TestApp_RunAsSubcommandParseFlags(t *testing.T) {
 				},
 				Flags: []Flag{
 					&StringFlag{
-						Name:  "lang",
-						Value: "english",
-						Usage: "language for the greeting",
+						Name:    "lang",
```
dearchap, line 468: how is this change related to Stderr functionaltity ?

app_test.go
```diff
@@ -933,8 +947,8 @@ func TestApp_SetStdout(t *testing.T) {
 	var w bytes.Buffer
 
 	app := &App{
-		Name:   "test",
-		Writer: &w,
+		Name:      "test",
+		ErrWriter: &w,
```
dearchap, line 951: This function is specifically testing Stdout. Either rename this to TestApp_SetStdErr or clone this function

dearchap:
@meatballhat Need your help on those doc errors. I changed the tag from "output" to "error" in the flags.md file but I am still getting errors.

dearchap:
@urfave/cli I've fixed all the compile/test issues. I want your opinion on whether this is a good candidate for merge into v2 or we get this into v3.

meatballhat:
@dearchap AFAICT the gfmrun errors are coming from examples that weren't altered as part of this patch. If you grep for the expected text in each error it should help you locate where changes are needed.

dearchap:
Okay so I've been thinking about this a bit more and the PR as it is wont work. It redirects all help output to stderr which is incorrect. Output to stderr should only be in the case when there is an error. So for example if an undefined flag is provided on the cmd line the output needs to be sent to stderr along with help

```
flag -foo provided but not defined
<help output>
```

if the user requests help via -h or --help the help output needs to be sent to stdout. 

@alessio Unless you want to make changes to ensure the above behaviour I am inclined to close this PR and open a separate PR with this functionality for v3.

PR #1211, opened 2020-12-06
Fix GitHub actions

jcgregorio:
## What type of PR is this?

_(REQUIRED)_

- [ x ] bug
- [ ] cleanup  
- [ ] documentation
- [ ] feature

_(REQUIRED)_

All PR checks currently fail because of this change in workflows:

       https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

The ::set-env and ::add-path commands no longer work and need to be replaced.
    

_(REQUIRED)_

NONE

## Release Notes

_(REQUIRED)_

```release-note
Fix for github workflow checks.
```

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -11,7 +11,6 @@ on:
       - v1
 
 jobs:
-
   test:
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


jcgregorio:
D'oh, wrong master.

PR #1212, opened 2020-12-06
Fix workflows/cli.yml.

jcgregorio:
The ::set-env and ::add-path commands no longer work and need to
to be replaced. See:

    https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

## What type of PR is this?

_(REQUIRED)_

- [ x ] bug
- [ ] cleanup  
- [ ] documentation
- [ ] feature

_(REQUIRED)_

PR checks currently don't run, this will fix that.

_(REQUIRED)_

NONE

## Release Notes

_(REQUIRED)_

```release-note
NONE
```

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -11,7 +11,6 @@ on:
       - v1
 
 jobs:
-
   test:
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


rliebz (approved):
Thanks for the fix!

asahasrabuddhe (approved):
LGTM

PR #1218, opened 2021-01-14
Another approach for zsh completion

z0rc:
## What type of PR is this?

- [ ] bug
- [ ] cleanup  
- [x] documentation
- [x] feature

I want to open discussion how this project provides completion for zsh. Current situation isn't right and promotes [cargo-cult approach](https://medium.com/@jzelinskie/please-dont-ship-binaries-with-shell-completion-as-commands-a8b1bcb8a0d0) to enable zsh completions.

Completion file shouldn't be sourced. It should provide only completion code (source of `_command` function) for `command`. Ideally this file should be just template/example, where developer should replace `$PROG` with name of binary and distribute it with `_command` name somewhere under project's contrib folder (alongside with man pages, vim syntax etc). As go build system isn't able to properly install this kind of stuff, it's up to package manager (homebrew, dpkg etc) to pick up this `_command` file and put it under correct `$fpath` where zsh can autoload it.

Currently this PR changes only zsh completion example. The documentation needs to be updated too. I can look into if there is agreement with current proposal.

There is fundamental problem with this project calling "completions" as "_auto_-completions", which are different terms. See https://www.gnu.org/software/bash/manual/html_node/Programmable-Completion.html and http://zsh.sourceforge.net/Doc/Release/Completion-System.html for correct terminology. You might want to fix this in future.

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -356,7 +356,7 @@ func ExampleApp_Run_bashComplete() {
 func ExampleApp_Run_zshComplete() {
 	// set args for examples sake
 	os.Args = []string{"greet", "--generate-bash-completion"}
-	_ = os.Setenv("_CLI_ZSH_AUTOCOMPLETE_HACK", "1")
+	_ = os.Setenv("SHELL", "/usr/bin/zsh")
 
 	app := NewApp()
 	app.Name = "greet"
--- a/autocomplete/zsh_autocomplete
+++ b/autocomplete/zsh_autocomplete
@@ -4,9 +4,9 @@ local -a opts
 local cur
 cur=${words[-1]}
 if [[ "$cur" == "-"* ]]; then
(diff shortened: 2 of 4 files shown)


lynncyrin (changes requested):
Can you pull in the default branch to fix the failing test please 🙏🏼

lynncyrin:
@z0rc Can you pull in the default branch to fix the failing test please 🙏🏼

z0rc:
@lynncyrin done.

This is work in progress, still requires consensus about proposal and documentation updates if proposal accepted. Marked as draft and awaiting your input.

z0rc:
Also refactored how zsh shell is detected. `$SHELL` shows login shell which is almost always user's current shell, unless they do something weird like launching zsh from bashrc.

PR #1221, opened 2021-01-16
Replace space only lines in help with only newline

cehoffman:
## What type of PR is this?

- bug

The changes to automatically indent description text added potential for empty lines that have only the padding text. As a consumer of this library, I have example tests to validate the help text appears as I intend. This was useful in helping to detect the change of description indent between 2.2 and 2.3. However I can no longer use these tests because it would require comment lines with trailing whitespace which is removed with tools like `go fmt` and `goimports`.

_(REQUIRED)_

## Testing

Made an example test with expected output.

## Release Notes

_(REQUIRED)_

```release-note
NONE
```


lynncyrin (changes requested):
Can you add a test case for this?

meatballhat:
@cehoffman Hello and sorry for the delay and auto-close via stalebot. Are you up for [adding a test for this change](https://github.com/urfave/cli/pull/1221#pullrequestreview-575061528)?

abitrolly:
I fixed the issue with trailing whitespace in https://github.com/urfave/cli/pull/1513

@cehoffman can you check that the fix works for you?

dearchap:
Closing

PR #1223, opened 2021-01-20
chore(ci): include upcoming Go 1.16 release build matrix

pjvds:
Go 1.16 beta1 is released not too long ago and we should also build against it in the ci pipeline.

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -15,13 +15,15 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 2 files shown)


lynncyrin (changes requested):
We should only include 1.16 when its out of beta, it adds extra toil to test against a beta release temporarily.

pjvds:
Thanks for the response. Go 16 will be released in a matter of a few weeks. I think we should build against it to make sure we are ready for that release. The overhead on the current build is imho something we can neglect.

lynncyrin:
> Go 16 will be released in a matter of a few weeks.

Feel free to check in again then 👍🏼

PR #1245, opened 2021-02-12
Allow same flag to be repeated by child commands. 

ipostelnik:
Reading the value or checking for presence looks at the most specific subcommand that includes the flag on the command line.

## What type of PR is this?

- feature

Add support for duplicate flags along subcommand chain. This allows the application and (sub)commands to define the same flag. The context for the ultimate action that runs will read the most specific value.

## Testing

Added a number of units tests to cover these cases. Corrected some of the existing tests to correctly mock operation of the production code.

## Release Notes

(sub)commands can now re-declare flags defined at parent command or the app. This allows these flags to appear in either location on the command-line. During action execution, context will read value provided at most-specific (rightmost) position.

```release-note

```

### DIFF
--- a/context.go
+++ b/context.go
@@ -17,6 +17,7 @@ type Context struct {
 	Command       *Command
 	shellComplete bool
 	flagSet       *flag.FlagSet
+	fromFlagSet   map[string]bool
 	parentContext *Context
 }
 
@@ -29,6 +30,14 @@ func NewContext(app *App, set *flag.FlagSet, parentCtx *Context) *Context {
 		if parentCtx.flagSet == nil {
 			parentCtx.flagSet = &flag.FlagSet{}
 		}
+	}
+
+	// pre-compute flag seen on the command line at this context
+	c.fromFlagSet = make(map[string]bool)
+	if set != nil {
+		set.Visit(func(f *flag.Flag) {
+			c.fromFlagSet[f.Name] = true
+		})
 	}
 
 	c.Command = &Command{}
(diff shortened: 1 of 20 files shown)


context.go
```diff
@@ -153,6 +177,25 @@ func lookupFlag(name string, ctx *Context) Flag {
...
+				// flag was defined, but value is not present among flags of the current context
+				// remember the most specific instance of the flag not from command line as fallback
+				src = f
+			}
+		}
+	}
+	return src
+}
+
 func lookupFlagSet(name string, ctx *Context) *flag.FlagSet {
```
dearchap, line 199: Couldnt you incorporate the resolveFlagDeep logic into this function to retrieve the corresponding flagset ? That would have meant zero changes to existing flag_xxxx src files. Moreover the current code handles the case of lookupFlagSet returning nil which the proposed changes dont handle .

context.go
```diff
@@ -48,35 +57,50 @@ func (c *Context) NumFlags() int {
 
 // Set sets a context flag to a value.
 func (c *Context) Set(name, value string) error {
-	return c.flagSet.Set(name, value)
+	err := c.flagSet.Set(name, value)
```
dearchap, line 60: combine into a single line, if err := c.flagSet.Set(..); err == nil { c.FromFlagSet[name]=true }

context.go
```diff
@@ -48,35 +57,50 @@ func (c *Context) NumFlags() int {
...
-
-		f := lookupFlag(name, c)
-		if f == nil {
-			return false
+		if ctx.flagOnCommandLine(name) {
+			return true
 		}
 
-		return f.IsSet()
+		// now see if value was set externally via environment
```
dearchap, line 80: why remove the lookupFlag call ?

dearchap (changes requested):
I dont see any tests written specifically for this change. Can you add some so that its clear that the said functionality works ?

meatballhat:
@ipostelnik Hello and sorry for the delay and auto-close via stalebot. Are you up for revisiting this work, resolving conflicts, and incorporating changes from feedback?

ipostelnik:
It has been a while, I will try to rebase this PR on current master and look into comments.

dearchap:
I am inclined to close this since we have merged persistent flag support in 3.x

PR #1251, opened 2021-03-07
Add Destination field to TimestampFlag

davidsbond:
## What type of PR is this?

- [ ] bug
- [ ] cleanup  
- [ ] documentation
- [x] feature

Adds a `Destination` field for the `TimestampFlag` type that allows you to specify a pointer to
a `Timestamp` rather than having to grab the `Timestamp` from the `cli.Context` using the flag
name.

Also updates the `expect` test helper to not use `runtime.Caller` etc directly. Instead, it calls `t.Helper` beforehand, which does the same thing. This is optional, but I think removes some unnecessary code. Happy to omit if preferred.

## Testing

Added a unit test `TestTimestampFlagApply_WithDestination`

## Release Notes

```release-note
Adds Destination field to TimestampFlag
```


rliebz (approved):
Thanks for putting this together!

asahasrabuddhe (approved):
LGTM

PR #1257, opened 2021-03-23
Add count option for bool flags

### LINKED ISSUE #757: Count number of occurrences of flag
Is it possible to have count the number of times a given (binary) flag is passed? This would mostly be used for verbosity flag sets, where e.g., `-v` and `-vvv` are varying levels of output. The README states "Note that giving two different forms of the same flag in the same command invocation is an error," so this doesn't exactly violate that principle.

dearchap:
## What type of PR is this?

- feature

This PR add a way of counting number of times a bool flag is set. This is useful when specifying the flag multiple times(-v -v -v etc). 

https://github.com/urfave/cli/issues/757

## Testing

Add test code in flag_test.go

go test -run=TestBoolFlagApply_SetsCount ./...

## Release Notes

Adds ability to count number of occurrences of bool flags


alexcb:
I was looking for this functionality and was considering implementing it, but it looks like it's already been done :tada: 

Is there any chance of getting this merged?

I'm happy to help if there's anything else that needs to be done and @dearchap is short on time.

lynncyrin:
I was just thinking about this feature the other day! Yay for a PR! I'll try to review it soon 🙏🏼

dearchap:
> I was just thinking about this feature the other day! Yay for a PR! I'll try to review it soon 🙏🏼

Can you review it soon @coilysiren . Its been more than a month since you suggested it. The code is pretty straightforward and also includes tests so the risk is very low.

dearchap:
Ping @coilysiren

meatballhat:
@dearchap sorry for the conflicts!

dearchap:
@meatballhat Merged my changes. There seems to be a circular dependency. The go generate wont run if it cant build the current code and the current code wont work unless the new spec has been generated.

meatballhat:
@dearchap You're absolutely right about the circular dependency 😭 I was hoping that we could "dogfood" the local version of urfave/cli in `./internal/genflags/cmd/` but maybe that's going to be more painful than is reasonable. I'd like to try out a nested `go.mod` approach with the version pinned to last release rather than removing urfave/cli usage entirely 🤔

meatballhat:
@dearchap lmkwyt? #1451

dearchap:
@meatballhat sure go for it.

go.mod
```diff
@@ -9,7 +9,4 @@ require (
 	gopkg.in/yaml.v3 v3.0.1
 )
 
-require (
-	github.com/russross/blackfriday/v2 v2.1.0 // indirect
-	golang.org/x/text v0.3.7 // indirect
```
meatballhat, line 14: 🤔 it's not clear to me why `golang.org/x/text` was automatically removed. It seems unrelated to this change.

meatballhat (approved):
I would love to see example usage in docs, too! 🎉

go.mod
dearchap: Reverted.

icholy:
I'm late to the party, but I would have expected the count to be exposed via a context method. 

``` go
func main() {
	app := &cli.App{
		Flags: []cli.Flag{
			&cli.BoolFlag{
				Name:  "foo",
				Usage: "foo greeting",
			},
		},
			fmt.Println("count", cCtx.Count("foo"))
			return nil
		},
	}
		log.Fatal(err)
	}
}
```

dearchap:
@meatballhat Added an example in docs but for some reason it is taking the "args" tag to execute the app

docs/v2/examples/flags.md
```diff
@@ -101,6 +101,46 @@ func main() {
 
 See full list of flags at https://pkg.go.dev/github.com/urfave/cli/v2
 
+For bool flags you can specify the flag multiple times to get a count(e.g -v -v -v or -vvv)
+
+<!-- {
+  "args": "--foo --foo",
```
meatballhat, line 107: it's a bit gnarly given the problem with using `--` inside an html comment, but this appears to work for me:

```suggestion
  "args": ["&#45;&#45;foo", "&#45;&#45;foo"],
```

dearchap:
@icholy I think you are correct. It would be good to have a ctx.Count("flagName"). That introduces some additional complexity as we would probably want to have a CountableFlag interface etc etc. Let me see what I can do

dearchap:
@icholy @meatballhat what do you think ?

docs/v2/examples/flags.md
```diff
@@ -550,6 +590,7 @@ Will result in help output like:
 #### Precedence
 
 The precedence for flag value sources is as follows (highest to lowest):
+eikdcclkkujudfjnhknkgruvlncbgvuckugignuhturk
```
icholy, line 593: What's going on here?

dearchap: 2FA key turned on by mistake.

PR #1259, opened 2021-03-27
Remove reflect calls for doc generation

### LINKED ISSUE #631: Eliminate reflection calls in flags
Reflection should not be used to check internal field values and instead should be part of the interface. For example in the visibleFlags func
```
func visibleFlags(fl []Flag) []Flag {
	visible := []Flag{}
	for _, flag := range fl {
		if !flagValue(flag).FieldByName("Hidden").Bool() {
			visible = append(visible, flag)
		}
	}
	return visible
}
```
there is a reflection call on "Hidden" requiring each implementation to have this field. Instead this should be part of the Flag interface to make it easier for users to implement their own custom flags and extend functionality.

```
type Flag interface {
	fmt.Stringer
	// Apply Flag settings to the given flag set
	Apply(*flag.FlagSet)
	GetName() string
        Hidden() bool
}

func visibleFlags(fl []Flag) []Flag {
	visible := []Flag{}
	for _, flag := range fl {
		if !flag.Hidden() {
			visible = append(visible, flag)
		}
	}
	return visible
}
```

dearchap:
## What type of PR is this?

- cleanup  

Instead of using reflection to figure out values of fields, the doc generation interface has been extended to include everything needed to stringify a flag. Custom flags need to properly implement the interface to get the valid synopsis for flags

Fixes #631 

Since the SliceFlags were type checked in the stringify(Flag) function it makes more sense to move those directly to the individual flags

## Testing

go test ./...

## Release Notes

DocGenerationFlag has bee extended and all currently defined flag type have been modified to use this interface. This allows doc generation to be customizable without having to resort to reflection

```release-note

```


dearchap:
bump

PR #1234, opened 2021-02-04
Add ChoiceFlag

Poweranimal:
A cli flag used for enums.
The flag displays all possible enum variations in the cli's help message and verifies the provided input value.

## What type of PR is this?

- feature

## What this PR does / why we need it:

This PR adds an enum flag implementation.

## Special notes for your reviewer:

I'm open for any suggestion to fulfill the job of supporting an enum flag better.

## Example
```
type LogFormat int32

const (
    LogFormatJSON LogFormat = iota
    LogFormatText
)

type LogMode in32

const (
    LogModeNoop LogMode = iota
    LogModeStdout
)

var logModeMap = map[LogMode]string{
    LogModeNoop: "noop",
    LogModeStdout: "stdout",
}

func (m LogMode) String() string {
    return logModeMap[m]
}

var (
    destFormat LogFormat
    destLevel string
)

var LogFormatFlag = &cli.ChoiceFlag{
    Name:         "log-format",
    Value:          LogFormatJSON,
    Choice:        cli.NewChoice(map[string]interface{}{
        "json": LogFormatJSON,
        "text":  LogFormatText,
    },
    Destination: &destFormat,
    EnvVars:      []string{"LOG_FORMAT"},
    Usage:         "The output formatting of the logger.",
    Required:     false,
}

var LogModeFlag = &cli.ChoiceFlag{
    Name:         "log-mode",
    Value:          LogModeStdout,
    Choice:        cli.NewStringerChoice(LogModeNoop, LogModeStdout),
    EnvVars:      []string{"LOG_MODE"},
    Usage:         "The log mode of the logger.",
}

var LogLevelFlag = &cli.ChoiceFlag{
    Name:         "log-level",
    Value:          "info",
    Choice:        cli.NewStringChoice("debug", "info", "warn", "error"),
    Destination: &destLevel,
    EnvVars:      []string{"LOG_LEVEL"},
    Usage:         "The log level of the logger.",
    Required:     false,
}
```

## Release Notes

```release-note

Added `cli.ChoiceFlag` for enum support.
```

## TODOS

- [x] Typed destination
- [ ] Add tests
- [ ] Add README

PR #1261, opened 2021-04-02
Add support for int/intslice flag validation

### LINKED ISSUE #786: Adding Flag Validators
I think it could be invaluable to have three features that instantly would be of use in a clean and succinct way when further examining my own use cases for this library around validation:

- Have the ability to access the original flags default value (perhaps this is already possible?)
- Add flag specific built-in validation attributes to flags (e.g. Min/Max values, Regex match, etc.)
- Add a custom validator attribute which can pass/fail a flag value in an anonymous or named function

I illustrate this and provide comments in the pseudocode in this example:

```go
package main

import (
	"fmt"
	"log"
	"os"

	"github.com/urfave/cli"
)

func main() {
	app := cli.NewApp()
	app.Action = rootAction
	app.Flags = []cli.Flag{
		cli.IntFlag{
			Name:  "timeout, t",
			// We would like to access this value later as the default value
			Value: 60,
			// These are "built-in" validators for the "int" type
			MinValue: 0,
			MaxValue: 600,
			// This is a custom anonymous or named function boolean validator
			CustomValidator: timeoutValidator,
		},
	}
	app.Name = "test-flag-value"

		log.Fatalf("An error occurred trying to run the application: %s", err)
	}
}

func rootAction(c *cli.Context) error {
	fmt.Printf("Flag TIMEOUT Value: %#v\n", c.GlobalInt("timeout"))
	fmt.Printf("Flag TIMEOUT Default Value: %#v\n", c.GlobalIntDefaultValue("timeout"))
	fmt.Printf("Flag TIMEOUT Default Min Value: %#v", c.GlobalIntDefaultMinValue("timeout"))
	fmt.Printf("Flag TIMEOUT Default Max Value: %#v", c.GlobalIntDefaultMaxValue("timeout"))

	if cli.GlobalInt("timeout") < 15 {
		// Access the original default value
		fmt.Printf("The supplied value is less than 15, resetting to: %s", c.GlobalIntDefaultValue("timeout"))
		cli.GlobalSet("timeout", c.GlobalIntDefaultValue("timeout"))
	}

	return nil
}

func timeoutValidator(c *cli.Context) bool {
	if c.GlobalInt("timeout") < 30 {
		return false
	}

	return true
}
```

Currently, as it stands, once the user supplies a value to the flag I have no way of knowing what the default value was associated with that flag as it seems to be overridden (I could be mistaken). 

Secondly it would be a nice-to-have to have not only some custom validators for each type of flag (int: min, max, non-zero, etc., string: regex, non-nil, etc. float: percision, etc.).

Lastly having the ability supply a custom callback validator could allow users to have a catch-all for validation of the flag values.

I started to take a look into doing this but, again if I am not mistaken, it looked like the `flagSet` was using the underlying go built-in `flag` library which may prohibit accessing the default value supplied without needing these features there first. If not keeping a copy of them before modification would be a workaround if that is possible.

I would be glad to help and add more to this discussion!

dearchap:
## What type of PR is this?

- feature

This PR adds basic support for flag validation. 

Fixes #786 

Added support ONLY for int and intslice flags. This is for proof of concept. If this design is acceptable I can add support for other types in a similar fashion. 

## Testing

New tests have been added
 * TestValidateFlags
 * TestIntValidatorFunc
 * TestIntSliceValidatorFunc

Following tests have been updated to include flag validators
 * TestFlagsFromEnv

## Release Notes

```release-note

```


instabledesign:
Yeah i like this feature, if the design is approve i can help implementing other validator (string, time date/duration, net.IP)

dearchap:
bump

meatballhat:
@dearchap (paraphrasing from discussion) This and #1234 are similar ideas, and given that you're actively working on this one and #1234 has been seemingly abandoned, I favor moving forward with this PR. _However_, I'm concerned about the additional API surface area introduced here, so what do you think about targeting the v3 series and using generics instead?

dearchap:
@meatballhat Yes that would be fine. However it might be worthwhile to add a light interface with implementation left to users for extending. And most common use cases can be provided with v3

meatballhat:
> @meatballhat Yes that would be fine. However it might be worthwhile to add a light interface with implementation left to users for extending. And most common use cases can be provided with v3

This sounds good to me 🎉

dearchap:
Will open a new PR with a small validation API

PR #1262, opened 2021-04-03
Allow -ve values for int, float & duration

### LINKED ISSUE #1199: Integer negative numbers are returned zero when parsed from yaml file
## my urfave/cli version is

v2.2.0

## Checklist

## Dependency Management

- [x] My project is automatically downloading the latest version.

## Describe the bug

Loading int negative values from a YAML file using altsrc returns zero.

## To reproduce

Parser an integer negative from an YAML file.
example:
```
        yaml file content:
        negative-number: -1

	app.Before = altsrc.InitInputSourceWithContext(
		flags,
		func(context *cli.Context) (altsrc.InputSourceContext, error) {
			if context.IsSet("load") {
				return altsrc.NewYamlSourceFromFile(context.String("load"))
			}

			return &altsrc.MapInputSource{}, nil
		},
	)

	app.Action = func(context *cli.Context) error {

		fmt.Print(context.Int("negative-number"))
		return nil
	}
```

## Observed behavior

Zero value.

## Expected behavior

the value that is defined in the variable of the YAML file.

## Additional context

The issue seems to be related to this condition

https://github.com/urfave/cli/blob/d648edd48d89ef3a841b1ec75c2ebbd4de5f748f/altsrc/flag.go#L200

## Want to fix this yourself?

The condition mentioned above, should not be greater than zero. It can not be applied for negative values.

```
go version go1.15.3 linux/amd64
```

```
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/renanrodrigues/.cache/go-build"
GOENV="/home/renanrodrigues/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/renanrodrigues/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/renanrodrigues/go"
GOPRIVATE=""
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/home/renanrodrigues/repos/telegram-sender/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build269715800=/tmp/go-build -gno-record-gcc-switches"
```

dearchap:
## What type of PR is this?

- bug

This PR allows negative values for int64, float64 and duration flags from alternate input sources

Fixes #1199 

## Testing

Added 3 tests
 *  TestIntApplyInputSourceMethodSetNegativeValue
 * TestDurationApplyInputSourceMethodSetNegativeValue
 * TestFloat64ApplyInputSourceMethodSetNegativeValue

## Release Notes

```release-note

```

### DIFF
--- a/altsrc/flag.go
+++ b/altsrc/flag.go
@@ -197,10 +197,8 @@ func (f *IntFlag) ApplyInputSourceValue(cCtx *cli.Context, isc InputSourceContex
 			if err != nil {
 				return err
 			}
-			if value > 0 {
-				for _, name := range f.Names() {
-					_ = f.set.Set(name, strconv.FormatInt(int64(value), 10))
-				}
+			for _, name := range f.Names() {
+				_ = f.set.Set(name, strconv.FormatInt(int64(value), 10))
 			}
 		}
 	}
@@ -215,10 +213,8 @@ func (f *DurationFlag) ApplyInputSourceValue(cCtx *cli.Context, isc InputSourceC
 			if err != nil {
 				return err
 			}
-			if value > 0 {
-				for _, name := range f.Names() {
-					_ = f.set.Set(name, value.String())
-				}
+			for _, name := range f.Names() {
+				_ = f.set.Set(name, value.String())
 			}
 		}
 	}
@@ -233,11 +229,9 @@ func (f *Float64Flag) ApplyInputSourceValue(cCtx *cli.Context, isc InputSourceCo
 			if err != nil {
 				return err
 			}
-			if value > 0 {
-				floatStr := float64ToString(value)
-				for _, name := range f.Names() {
-					_ = f.set.Set(name, floatStr)
-				}
+			floatStr := float64ToString(value)
+			for _, name := range f.Names() {
+				_ = f.set.Set(name, floatStr)
 			}
 		}
 	}
--- a/altsrc/flag_test.go
+++ b/altsrc/flag_test.go
@@ -234,6 +234,15 @@ func TestIntApplyInputSourceMethodSet(t *testing.T) {
 	expect(t, 15, c.Int("test"))
 }
 
+func TestIntApplyInputSourceMethodSetNegativeValue(t *testing.T) {
+	c := runTest(t, testApplyInputSource{
+		Flag:     NewIntFlag(&cli.IntFlag{Name: "test"}),
+		FlagName: "test",
+		MapValue: -1,
+	})
+	expect(t, -1, c.Int("test"))
+}
+
 func TestIntApplyInputSourceMethodContextSet(t *testing.T) {
 	c := runTest(t, testApplyInputSource{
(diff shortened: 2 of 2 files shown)


altsrc/flag.go
```diff
@@ -197,10 +197,8 @@ func (f *IntFlag) ApplyInputSourceValue(context *cli.Context, isc InputSourceCon
 			if err != nil {
 				return err
 			}
-			if value > 0 {
```
rliebz, line 200: It's hard to tell from the commit that introduced this check, because it happens to also be the commit that introduced YAML parsing, but I assume the intent here was to avoid calling `Set` on the flag set if the value wasn't explicitly specified. In an ideal world, I think we'd want `isc.Int` to return `value, ok, err`, so we would know whether the value was explicitly set as the zero value, versus just defaulting to zero. 

While it's also an imperfect solution, what are your thoughts on keeping a check for `value != 0` instead of `> 0`? We wouldn't know whether or not the value was actually set if it's zero, but the key difference in behavior is that we would report `false` for `IsSet`, which is closer to the current behavior, and, if I had to guess, more likely than the case where `0` is explicitly set.

dearchap: Actually thinking about it a bit more we want to check that the value from yaml is not the same as the default value for that flag and only then set it. Then this would take care of both the issues. i.e we would report IsSet only if the value from yaml is different from flag default and also allow for range of values. This way the value from yaml could be zero as well.

dearchap:
bump

albttx:
Can this be updated and merge ?! 

cc: @dearchap  @urfave

dearchap:
@albttx i'll take a look at it this week

altsrc/flag.go
dearchap: @rliebz Looking at it further the isc.XXX(f.Name) will return an error if the flag isnt found in the input source. So there is no chance of the Set being called in that case. Added a small unit test to verify that

meatballhat (approved):
Rad! Thanks! 🙇🏼 🎉

altsrc/flag.go
binwiederhier: FYI https://github.com/urfave/cli/issues/1373

I don't know if this is my code causing the problem, or the change in this PR, and I'm not knowledgeable enough to figure it out. Maybe you guys can find out?

albttx:
Amazing, thanks :+1:

Issue #1247, opened 2021-02-16
Can't run help without required global flags


0xch4z:
## my urfave/cli version is

v2.3.0

## Checklist

* [x] Are you running the latest v2 release? The list of releases is [here](https://github.com/urfave/cli/releases).
* [x] Did you check the manual for your release? The v2 manual is [here](https://github.com/urfave/cli/blob/master/docs/v2/manual.md)
* [x] Did you perform a search about this problem? Here's the [Github guide](https://help.github.com/en/github/managing-your-work-on-github/using-search-to-filter-issues-and-pull-requests) about searching.

## Dependency Management

- [x] My project is using go modules.
- [ ] My project is using vendoring.
- [ ] My project is automatically downloading the latest version.
- [ ] I am unsure of what my dependency management setup is.

## To reproduce

`> go run main.go help list`

```go
package main

import "github.com/urfave/cli/v2"

const cmdHelpTemplate = "should reach this point without error"

func main() {
	cli := &cli.App{
		Name: "test",
		Flags: []cli.Flag{
			&cli.StringFlag{
				Name:     "token",
				EnvVars:  []string{"API_TOKEN"},
				Required: true,
			},
		},
		Commands: []*cli.Command{
			{
				Name:  "list",
				Usage: "list all clusters",

				Flags: []cli.Flag{
					&cli.StringFlag{
						Name:     "filter",
						Required: true,
					},
				},

				Action: func(_ *cli.Context) error { return nil },

				CustomHelpTemplate: cmdHelpTemplate,
			},
                        // ... other misc. commands that all require a token
		},
	}
	cli.RunAndExitOnError()
}
```

## Observed behavior

The default app help text is rendered to the terminal.

## Expected behavior

The help text for the `list` command to be rendered to the terminal. Maybe the user wanted to learn about the `filter` flag. The user shouldn't have to fulfill global flag requirements just to display help for a subcommand.

---

would love to fix this myself but want to ensure that this is not the intended behavior.

## Run `go version` and paste its output here

```
go version go1.15.8 darwin/amd64
```

## Run `go env` and paste its output here

```

```

makeworld-the-better-one:
Also experiencing this, it's frustrating. Here is the code I used to get around it:

```go
	// Hack around issue where required flags are still required even for help
	// https://github.com/urfave/cli/issues/1247
	if len(os.Args) == 3 {
		if os.Args[1] == "h" || os.Args[1] == "help" {
			// Like: my-app help cmd
[...]

PR #1285, opened 2021-06-22
Fix help used as command returns errors when required flags are missing

jaimem88:
## What type of PR is this?

- bug

Using `app help` or `app h` without providing required flags causes `app.Run()` to return an `errRequiredFlags` error.
This does not happend if using `-help` or `-h`.

Added a function to check if any of the arguments matches `HelpFlag.Names()` to stop early and show the app help.

Fixes https://github.com/urfave/cli/issues/1247

I'm new to the code base so perhaps there is an easier way to check if the command matches `help` or `h`  before calling `checkRequiredFlags`. However, `context.Command.Name` is empty just before `checkRequiredFlags` so I couldn't just check that first.

## Testing

Added test `Test_Help_Required_Flags_Does_Not_Error` to `help_test.go`

## Release Notes

_(REQUIRED)_

```release-note
Fixed https://github.com/urfave/cli/issues/1247
```

### DIFF
--- a/app.go
+++ b/app.go
@@ -319,7 +319,7 @@ func (a *App) RunContext(ctx context.Context, arguments []string) (err error) {
 		}()
 	}
 
-	if !a.HideHelp && checkHelp(cCtx) {
+	if !a.HideHelp && (checkHelp(context) || checkHelpArguments(arguments)) {
 		_ = ShowAppHelp(cCtx)
 		return nil
 	}
--- a/app.go
+++ b/app.go
@@ -319,7 +319,7 @@ func (a *App) RunContext(ctx context.Context, arguments []string) (err error) {
 		}()
 	}
 
-	if !a.HideHelp && (checkHelp(context) || checkHelpArguments(arguments)) {
+	if !a.HideHelp && (checkHelp(cCtx) || checkHelpArguments(arguments)) {
 		_ = ShowAppHelp(cCtx)
 		return nil
 	}
--- a/help_test.go
+++ b/help_test.go
@@ -102,6 +102,39 @@ func Test_Help_Custom_Flags(t *testing.T) {
 	}
 }
 
+func Test_Help_Required_Flags_Does_Not_Error(t *testing.T) {
+	oldFlag := HelpFlag
+	defer func() {
+		HelpFlag = oldFlag
+	}()
+
+	app := App{
+		Flags: []Flag{
+			&StringFlag{Name: "foo", Required: true},
+		},
+	}
+
+	output := new(bytes.Buffer)
+	app.Writer = output
+
+	tests := []struct {
+		helpFlag string
+	}{
+		{"-help"},
+		{"-h"},
+		{"help"},
+		{"h"},
+	}
+	for _, tt := range tests {
+		t.Run(tt.helpFlag, func(t *testing.T) {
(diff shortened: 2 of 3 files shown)


help_test.go
```diff
@@ -101,6 +101,39 @@ func Test_Help_Custom_Flags(t *testing.T) {
 	}
 }
 
+func Test_Help_Required_Flags_Does_Not_Error(t *testing.T) {
+	oldFlag := HelpFlag
```
rliebz, line 105: It doesn't look like we actually need to save/reset this, since the test doesn't modify it

help_test.go
```diff
@@ -101,6 +101,39 @@ func Test_Help_Custom_Flags(t *testing.T) {
...
+		HelpFlag = oldFlag
+	}()
+
+	app := App{
+		Flags: []Flag{
+			&StringFlag{Name: "foo", Required: true},
+		},
+	}
+
+	output := new(bytes.Buffer)
```
rliebz, line 116: If we don't need to test the output here for any reason, we can just use `ioutil.Discard`

app.go
```diff
@@ -268,7 +268,7 @@ func (a *App) RunContext(ctx context.Context, arguments []string) (err error) {
 		return err
 	}
 
-	if !a.HideHelp && checkHelp(context) {
+	if !a.HideHelp && (checkHelp(context) || checkHelpArguments(arguments)) {
```
rliebz, line 271: The `arguments` variable here I believe is the unparsed arguments, rather than the parsed arguments.

What that means from a practical perspective is that if we were to have flags like this:

```
my-app --string-flag foo --int-flag 15 text1 text2
```

The unparsed args would be: `["my-app", "--string-flag", "foo", "--int-flag", "15", "text1", "text2"]`

The parsed args would be: `["text1", "text2"]`

The reason this is important is because we want to distinguish between when `help` is passed as the value of a flag (e.g., `--string-flag help` as opposed to as a command. 

I believe the way to do that would be to access `context.Args()`,  which under the hood uses a `flag.FlagSet` that has already parsed our flags. This also has the added benefit of letting us modify the definition of `checkHelp`, as opposed to creating a new function that must be called alongside it. But we'll definitely want to rely on good test cases to prove that it works as expected 🙂

rliebz (commented):
Thanks for the contribution! 

I think there's a couple small changes to make, but great work (and great test cases!)

codefromthecrypt:
will be great to see this landed as it is surprising behavior without a fix. @jaimem88 let me know if you ran out of steam!

codefromthecrypt:
I think this also hits when help is default. Ex. if your app is named "foo", `foo` and `foo help` will trigger the unnecessarily arg validation

app.go
jaimem88: ah! this is the kind of pointer I needed! I'll rework this part a bit. I just ran into a bit of trouble with `TestShowCommandHelp_HelpPrinter` showing the wrong template. I think we just need to know if there is a command being called or not. I'll dig a bit further soon 😄

dearchap:
Issue not present in latest releases. Closing.

dearchap:
No movement for this for a long time and fix isn't sufficient. Closing for now

dearchap:
Closing. Will revisit when new v3 parser is introduced.

PR #1297, opened 2021-08-18
Ignore not configured flags in json file

kungf:
## What type of PR is this?

- bug

Flags has default value, no need config every flag in config file

_(REQUIRED)_

## Testing

## Release Notes

_(REQUIRED)_

```release-note

```

### DIFF
--- a/altsrc/json_source_context.go
+++ b/altsrc/json_source_context.go
@@ -8,7 +8,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/pkg/errors"
 	"github.com/urfave/cli/v2"
 )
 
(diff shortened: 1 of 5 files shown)


kungf:
hi, @rliebz @saschagrunert,  who can have a review please.

altsrc/json_source_context.go
```diff
@@ -195,7 +196,7 @@ func jsonGetValue(key string, m map[string]interface{}) (interface{}, error) {
 	keys := strings.Split(key, ".")
 	for ix, k := range keys {
 		if ret, ok = working[k]; !ok {
-			return ret, fmt.Errorf("missing key %q", key)
+			return ret, syscall.ENOENT
```
meatballhat, line 199: Rather than returning a value that already has meaning attached to it, I think this should instead be a stable error that's declared at the package level, e.g. `errMissingKey`, which is then checked via `errors.Is` rather than an equality check.

meatballhat (changes requested):

altsrc/json_source_context.go
dearchap: @meatballhat I've made the change

meatballhat (commented):
Given that this is targeting the `v2-maint` branch which should only be receiving bug fixes, I don't like the idea of introducing a dependency on `github.com/pkg/errors`. Can this instead use `fmt.Errorf` with `%w` to wrap the error?

dearchap:
Hmm actually isc.IsSet checks if the value is defined in the key or not so we dont need this feature.

PR #1300, opened 2021-09-08
Remove stalebot

rliebz:
## What type of PR is this?

_(REQUIRED)_

- [ ] bug
- [ ] cleanup  
- [ ] documentation
- [ ] feature
- [x] maintenance

_(REQUIRED)_

Stalebot is aggressively closing real issues because we have not been able to keep up. Rather than deciding that these are no longer issues, we should keep them open.


saschagrunert (approved):
LGTM

PR #1311, opened 2021-10-27
Remove extraneous line from generated markdown

### LINKED ISSUE #1275: v2 bug: Man page section interferes with Markdown rendering
## my urfave/cli version is

_**( Put the version of urfave/cli that you are using here )**_

## Checklist

* [x] Are you running the latest v2 release?
* [x] Did you check the manual for your release?
* [x] Did you perform a search about this problem?

## Dependency Management

## Describe the bug

Right now, the `MarkdownTemplate` is used for both markdown and man page generation. There's a section added in  https://github.com/urfave/cli/commit/ed2ee4bc4a8c56b64983e58e1916af40de019260 which helps for man, but is not helpful in markdown

```
% {{ .App.Name }} 8
```

This renders literally and reduces the clean appearance otherwise present.

## To reproduce

run a command like `m, err := c.App.ToMarkdown()` and try to render it directly or in something like hugo

## Observed behavior

Screen Shot 2021-05-21 at 9 16 14 AM

## Expected behavior

I would expect nothing leaked from man into markdown. In this case, it is ok to have the line removed when in markdown.

## Additional context

While not specifically annotated about this, the following may be someone else working around this https://github.com/google/skia-buildbot/blob/4c87209ff9378a9befc65d5f93c4040d4fdd5922/go/urfavecli/urfavecli.go#L34

## Want to fix this yourself?

To fix this, I would only add that header when man is in use. For example, we can default a new `ManTemplate` to `MarkdownTemplate` with that header

jolheiser:
## What type of PR is this?

_(REQUIRED)_

- [ ] bug
- [x] cleanup  
- [ ] documentation
- [ ] feature

_(REQUIRED)_

This PR removes the extraneous man-page line from markdown rendering. It does this by only including the line if `SectionNum > 0`

_(REQUIRED)_

Fixes #1275

## Testing

`go test -v ./...`

## Release Notes

_(REQUIRED)_

```release-note
None.
```

### DIFF
--- a/docs.go
+++ b/docs.go
@@ -15,7 +15,7 @@ import (
 // The function errors if either parsing or writing of the string fails.
 func (a *App) ToMarkdown() (string, error) {
 	var w bytes.Buffer
-	if err := a.writeDocTemplate(&w, 8); err != nil {
+	if err := a.writeDocTemplate(&w, 0); err != nil {
 		return "", err
 	}
 	return w.String(), nil
--- a/template.go
+++ b/template.go
@@ -74,9 +74,9 @@ OPTIONS:
    {{end}}{{end}}
 `
 
-var MarkdownDocTemplate = `% {{ .App.Name }} {{ .SectionNum }}
+var MarkdownDocTemplate = `{{if gt .SectionNum 0}}% {{ .App.Name }} {{ .SectionNum }}
 
-# NAME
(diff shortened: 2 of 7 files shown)


rliebz (approved):

codefromthecrypt:
thanks a lot!

PR #1312, opened 2021-11-06
Support Help Template overrides for Subcommands

### LINKED ISSUE #1303: Custom SubcommandHelpTemplate
## Is there a way to set a custom help template for subcommands?
_**I know it's totally possible to set custom templates for the app (CustomAppHelpTemplate) and for commands (CustomHelpTemplate) but I can't find how to override the SubcommandHelpTemplate. What am I missing?**_

ddouglas:
## What type of PR is this?

- bug

When using the Subcommands, it is currently no possible to override the Subcommand Template that is used due to the Command being overwritten when initializing Context inside of the `RunAsSubcommand` function. I outline everything in #1303. The comment in the code details how I came to the conclusion I came to. Ideally, the Command would not be overridden but set to the Parent Command that houses the Subcommands that are being executed. 

Fixes #1303

This is a bandaid at best. I do not expect this PR to be merged. I might update it later with an actual fix, but I just wanted to test my hypothesis, I did that, this works, but ideally, you give Subcommand Context to the `NewContext` Command should the `c.Command` is initialized with the parent command instead of a empty `*cli.Command`

## Release Notes

```release-note
Adds Support for overriding cli.SubcommandHelpTemplate
```

### DIFF
--- a/help.go
+++ b/help.go
@@ -189,7 +189,14 @@ func ShowCommandHelpAndExit(c *Context, command string, code int) {
 func ShowCommandHelp(ctx *Context, command string) error {
 	// show the subcommand help for a command with subcommands
 	if command == "" {
-		HelpPrinter(ctx.App.Writer, SubcommandHelpTemplate, ctx.App)
+		// If the command == "", this maybe a subcommand. Check to see if the context
+		// has a non nil *cli.App and if that has a template defined, use it.
+		templ := SubcommandHelpTemplate
+		if ctx.App != nil && ctx.App.CustomAppHelpTemplate != "" {
+			templ = ctx.App.CustomAppHelpTemplate
+		}
+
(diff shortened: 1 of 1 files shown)


meatballhat:
@ddouglas Sorry for the big delay 😭 What do you think about adding a new field to `App` such as `CustomSubcommandHelpTemplate` instead of what you have here? This would mean bumping from a "bug" to a "feature" given the increased API surface area, but I think it'd result in fewer surprises in the future.

dearchap:
@ddouglas Can you add a test case as well ?

dearchap (changes requested):
Can you update the unit tests for this ?

dearchap:
@ddouglas I'm going to close this and push another PR with the proper fix/tests.

PR #1346, opened 2022-04-19
Exclude hidden sub-command flags from docs

jolheiser:
## What type of PR is this?

- bug

Not sure I would hard-classify this as a "bug" per-se, but it expands on #999 by also excluding flags marked as `hidden` in sub-commands for docs generation.

Expands upon #999 

## Testing

I've changed `docs_test.go` and `go test ./...` passes.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/docs.go
+++ b/docs.go
@@ -80,14 +80,14 @@ func prepareCommands(commands []*Command, level int) []string {
 			usageText,
 		)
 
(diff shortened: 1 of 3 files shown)


meatballhat (approved):
Thank you! 🎉

PR #1363, opened 2022-04-22
Make `ToMarkdown` and `ToMan` optional

kolyshkin:
## What type of PR is this?

- cleanup

Their functionality is not always used, and it brings in quite a big
dependency (russross/blackfriday).

Amend the documentation to make it clear that cli_docs build tag is
needed to use these functions.

This reduces runc size by about 318 Kb, or 3.4% (comparing stripped
binaries).

none

none

## Release Notes

```release-note
Compatibility: if you are using `ToMan` or `ToMarkdown` methods, from now on
you need to add `cli_docs` build tag for your app.
```

### DIFF
--- a/docs.go
+++ b/docs.go
@@ -1,3 +1,6 @@
+//go:build cli_docs
+// +build cli_docs
+
 package cli
 
 import (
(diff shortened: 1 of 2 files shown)


meatballhat:
@kolyshkin What do you think about inverting the build constraint so that the default behavior remains the same? Is opting _out_ of the docs bits something that can be introduced in `runc` (and others)?

kolyshkin:
Yes, that's possible, too. I will redo.

kolyshkin:
Opened #1375 as a replacement

PR #1367, opened 2022-04-22
Simplified flag value access

### LINKED ISSUE #1316: Simplify accessing flag values
Would it be reasonable to add convenient value accessors to flags like `myFlag.Value(context)` instead of `context.String(myFlag.Name)`?

Currently, I have to choose the correct accessor (`ctx.String()`, `ctx.StringSlice()`) or use the correct type assertion (`ctx.Value(…).(String)`). I think this part could be moved into the flag which already knows its correct type.

toaster:
## What type of PR is this?

- feature

This PR implements convenience accessors to access flag values in a context.
Instead of writing `ctx.Int(f.Name)` or `ctx.Value(f.Name).(int)` one can write `f.Get(ctx)`.
Thus, you don’t have to use the correct type method or the appropriate type assertion. The flag picks the correct name and gives you the correct type.

Fixes #1316

~I’m happy to give the accessor a shorter name if one comes up with a good idea :).~ (done! ~ @meatballhat)

## Release Notes

```release-note
All built-in flags now have a convenience method to access a flag’s value within a `Context`.
With `f.Get(ctx)` you don’t have to provide the flag’s name or its value’s actual type.
```


meatballhat:
@toaster What do you think of claiming `Get`? This would be similarly as terse as `flag.Getter.Get` and future (v3+) versions could similarly define an interface method with return signature of `any` once we're beyond supporting go `< 1.18`.

toaster:
> @toaster What do you think of claiming `Get`? This would be similarly as terse as `flag.Getter.Get` ~and future (v3+) versions could similarly define an interface method with return signature of `any` once we're beyond supporting go `< 1.18`~.

I like the idea and added a respective commit.
However, I’m not deep enough into the project to assess whether this fits the API.

meatballhat (approved):
I like it! I'll wait a bit before merging to make sure no other @urfave/cli humans have issues with the increased API surface area 👍🏼

meatballhat:
@dearchap when you get time, I'd appreciate your thoughts on this 🙇🏼

dearchap:
@meatballhat I have no problem with the new interface. This API is not within most people's workflow wherein they get the flag value from context but if users wants to use flag.GetFromContext thats fine with me.

xoxys:
Is there a minimal example on how to use it? On a minimal example from the docs, how could the new approach be used to access the value of the `lang` flag from the example below?  

```
app := &cli.App{
    Flags: []cli.Flag{
      &cli.StringFlag{
        Name:    "lang",
        Aliases: []string{"l"},
        Value:   "english",
        Usage:   "language for the greeting",
        EnvVars: []string{"LEGACY_COMPAT_LANG", "APP_LANG", "LANG"},
      },
    },
  }
```

meatballhat:
@xoxys You will need a reference to the flag in order to use this type of access. For example, to take your code and put it into an example app to show the typical way of using `*cli.Context`, but with a modification demonstrating the source of bugs that's alleviated via direct flag access:

```go
package main

import (
    "log"
    "os"

    "github.com/urfave/cli/v2"
)

func main() {
    app := &cli.App{
        Flags: []cli.Flag{
            &cli.StringFlag{
                Name:    "lang",
                Aliases: []string{"l"},
                Value:   "english",
                Usage:   "language for the greeting",
                EnvVars: []string{"LEGACY_COMPAT_LANG", "APP_LANG", "LANG"},
            },
        },
            notLang := cCtx.String("1ang")
            log.Println("is this lang?:", notLang)

            lang := cCtx.String("lang")
            log.Println("is this lang?:", lang)

            return nil
        },
    }

        log.Fatal(err)
    }
}
```

and then modify it to use the access provided via the flag's `Get` method:

```go
package main

import (
    "log"
    "os"

    "github.com/urfave/cli/v2"
)

func main() {
    langFlag := &cli.StringFlag{
        Name:    "lang",
        Aliases: []string{"l"},
        Value:   "english",
        Usage:   "language for the greeting",
        EnvVars: []string{"LEGACY_COMPAT_LANG", "APP_LANG", "LANG"},
    }

    app := &cli.App{
        Flags: []cli.Flag{langFlag},
            notLang := langFlag.Get(cCtx)
            log.Println("is this lang?:", notLang)

            lang := langFlag.Get(cCtx)
            log.Println("is this lang?:", lang)

            return nil
        },
    }

        log.Fatal(err)
    }
}
```

Here's the diff, if helpful:

```diff
--- a/1367.go
+++ b/1367.go
@@ -8,21 +8,21 @@
 )

 func main() {
+       langFlag := &cli.StringFlag{
+               Name:    "lang",
+               Aliases: []string{"l"},
+               Value:   "english",
+               Usage:   "language for the greeting",
+               EnvVars: []string{"LEGACY_COMPAT_LANG", "APP_LANG", "LANG"},
+       }
+
        app := &cli.App{
-               Flags: []cli.Flag{
-                       &cli.StringFlag{
-                               Name:    "lang",
-                               Aliases: []string{"l"},
-                               Value:   "english",
-                               Usage:   "language for the greeting",
-                               EnvVars: []string{"LEGACY_COMPAT_LANG", "APP_LANG", "LANG"},
-                       },
-               },
+               Flags: []cli.Flag{langFlag},
                Action: func(cCtx *cli.Context) error {
-                       notLang := cCtx.String("1ang")
+                       notLang := langFlag.Get(cCtx)
                        log.Println("is this lang?:", notLang)

-                       lang := cCtx.String("lang")
+                       lang := langFlag.Get(cCtx)
                        log.Println("is this lang?:", lang)

                        return nil
```

I hope this is helpful 😁

xoxys:
> I hope this is helpful grin

That helps, thanks a lot!

PR #796, opened 2019-01-27
Add flag category support

michaeljs1990:
This could be better and needs some cleanup as well as tests. It gives you the basic ability that I wanted which lets you set a category on a flag and have it printed in the help to break up commands with super long help messages. Global options ignore this right now but could be made to use it if someone messed with the template.

Here is an example output with this patch applied.

```
NAME:
   collins-go-cli query - Search for assets in Collins

USAGE:
   collins-go-cli query [command options] [arguments...]

OPTIONS:
  
   -Z, --remote-lookup                Query remote datacenters for asset
   -T value, --type value             Only show asset with type value
   -n value, --nodeclass value        Assets in nodeclass value[,...]
   -p value, --pool value             Assets in pool value[,...]
   -s value, --size value             Number of assets to return per page (default: 100)
   --limit value                      Limit total results of assets (default: 0)
   -r value, --role value             Assets in primary role
   -R value, --secondary-role value   Assets in secondary role
   -i value, --ip-address value       Assets with IP address[es]
   -S value, --status value           Asset status (and optional state after :)
   -a value, --attribute value        Arbitrary attributes and values to match in query. : between key and value
   -H, --show-header                  Show header fields in output
   -c value, --columns value          Attributes to output as columns, comma separated (default: "tag,hostname,nodeclass,status,pool,primary_role,secondary_role")
   -x value, --extra-columns value    Show these columns in addition to the default columns, comma separated
   -f value, --field-separator value  Separator between columns in output
   -l, --link                         Output link to assets found in web UI
   -j, --json                         Output results in JSON
   -y, --yaml                         Output results in YAML
   
  Test
   -t value, --tag value  Assets with tag[s] value[,...]
   
```
It also adds go mod support so I can overwrite the package locally with the mod replace syntax.

PR #1368, opened 2022-04-22
Add flag category support (#796)

### LINKED ISSUE #259: Flag Groups
It might be nice to be able to have `command.Flags` be a `[][]Flag` for the purposes of being able to group flags in the command help output.
In addition instead of a straight up `[][]Flag`, a new `FlagGroup` struct type may be better, where we can define a name/other fields.

WDYT?

meatballhat:
## What type of PR is this?

- feature

See #796 

Closes #259 

This PR is a best-effort porting of #796 so detailed criticism is very much appreciated 🙇🏼 

## Release Notes

```release-note
add flag category support
```


category.go
```diff
@@ -77,3 +77,48 @@ func (c *commandCategory) VisibleCommands() []*Command {
 	}
 	return ret
 }
+
+// FlagCategories is a slice of *FlagCategory.
+type FlagCategories []*FlagCategory
+
+// FlagCategory is a category containing commands.
```
dearchap, line 84: containing flags

category.go
```diff
@@ -77,3 +77,48 @@ func (c *commandCategory) VisibleCommands() []*Command {
 	}
 	return ret
 }
+
+// FlagCategories is a slice of *FlagCategory.
+type FlagCategories []*FlagCategory
```
dearchap, line 82: map[category]Flag is better ?

dearchap (dismissed):

category.go
meatballhat: Agreed!

meatballhat: Welp, I started trying to switch the implementation to `map[string][]Flag`, and it seems non-trivial to me so far... but I'm also nervous about introducing `FlagCategories` and `FlagCategory` and being committed to their behaviors for the rest of the `v2` series 😐

flag.go
```diff
@@ -132,6 +132,14 @@ type VisibleFlag interface {
 	IsVisible() bool
 }
 
+// CategorizableFlag is an interface that allows us to potentially
+// use a flag in a categorized representation.
+type CategorizableFlag interface {
+	VisibleFlag
+
+	GetCategory() string
```
dearchap, line 140: Rename to just Category(). golang doesnt recommend using GetXXX

app.go
```diff
@@ -181,6 +183,15 @@ func (a *App) Setup() {
 		if c.HelpName == "" {
 			c.HelpName = fmt.Sprintf("%s %s", a.HelpName, c.Name)
 		}
+
+		fc := newFlagCategories()
```
dearchap, line 187: Move this out into a separate function, with a []Flag as argument

app.go
```diff
@@ -481,6 +499,14 @@ func (a *App) VisibleCommands() []*Command {
 	return ret
 }
 
+// VisibleFlagCategories returns a slice containing all the categories with the flags they contain
+func (a *App) VisibleFlagCategories() []VisibleFlagCategory {
+	if a.flagCategories == nil {
+		a.flagCategories = newFlagCategories()
```
dearchap, line 505: cant you return nil or [] if a.flagCategories is nil ? Its not nice behaviour to update flagCategories here

category.go
```diff
@@ -77,3 +79,82 @@ func (c *commandCategory) VisibleCommands() []*Command {
...
+
+func (f *defaultFlagCategories) AddFlag(category string, fl Flag) {
+	if _, ok := f.m[category]; !ok {
+		f.m[category] = &defaultVisibleFlagCategory{name: category, m: map[string]Flag{}}
+	}
+
+	f.m[category].m[fl.String()] = fl
+}
+
+func (f *defaultFlagCategories) VisibleCategories() []VisibleFlagCategory {
```
dearchap, line 109: where is the visibility of the category being checked here ?

category.go
```diff
@@ -77,3 +79,82 @@ func (c *commandCategory) VisibleCommands() []*Command {
...
+	catNames := []string{}
+	for name := range f.m {
+		catNames = append(catNames, name)
+	}
+
+	sort.Strings(catNames)
```
dearchap, line 115: isnt f.m already sorted ?

command.go
```diff
@@ -280,6 +281,14 @@ func (c *Command) startApp(ctx *Context) error {
 	return app.RunAsSubcommand(ctx)
 }
 
+// VisibleFlagCategories returns a slice containing all the visible flag categories with the flags they contain
+func (c *Command) VisibleFlagCategories() []VisibleFlagCategory {
+	if c.flagCategories == nil {
```
dearchap, line 286: not a good idea to initialize c.flagCategories here .

command.go
```diff
@@ -280,6 +281,14 @@ func (c *Command) startApp(ctx *Context) error {
...
+		c.flagCategories = newFlagCategories()
```
dearchap, line 287: return nil or empty slice here ?

flag_bool.go
```diff
@@ -19,6 +19,7 @@ type BoolFlag struct {
 	DefaultText string
 	Destination *bool
 	HasBeenSet  bool
+	Category    string
```
dearchap, line 22: @meatballhat How would this tie-in with your generator code ?

flag.go
meatballhat: What about the conflict with the field name `Category`? 🤔 I agree that `GetXXX` is generally bad style, but this is a language-level wart we're working around.

flag_bool.go
meatballhat: If all the flag types defined in this repo have `Category` fields, then it'd be part of the struct template. Or did you mean the potential merge conflict? (or something else?)

app.go
meatballhat: Excellent 👍🏼

app.go
meatballhat: I agree!

category.go
meatballhat: IIUC ranging over a `map` has undefined sort order: <https://go.dev/play/p/aR7JKg4ae8r> Or did you mean something else? 🤔

command.go
meatballhat: I agree!

PR #1375, opened 2022-04-25
Add `urfave_cli_no_docs` build tag

kolyshkin:
## What type of PR is this?

- feature

Add urfave_cli_no_docs build tag
  
This removes the resulting binary dependency on cpuguy83/md2man and
russross/blackfriday, which saves more than 400 KB (more than 300 KB
once stripped) from the resulting binary.

none (although see some discussions in e.g. https://github.com/urfave/cli/issues/1055)

none

## Testing

See last commit.

## Release Notes

```release-note
Added support for `urfave_cli_no_docs` tag to reduce the binary size by 300+ KB
in case your application is not using `ToMarkdown` and `ToMan` methods.
```


kolyshkin:
Result of check-binary-size CI job, defaults:

> 1.7MB is the current binary size

Same with `urfave_cli_no_docs`: 

> 1.4MB is the current binary size

Are we planning on v3 yet? In there, we can move docs to a separate package, so it won't be compiled in unless explicitly included.

meatballhat:
> Result of check-binary-size CI job, defaults:
> 
> > 1.7MB is the current binary size
> 
> Same with `urfave_cli_no_docs`:
> 
> > 1.4MB is the current binary size

🎉 🤘🏼 
 
> Are we planning on v3 yet? In there, we can move docs to a separate package, so it won't be compiled in unless explicitly included.

Added to [the list](https://github.com/urfave/cli/issues/833)!

meatballhat (approved):
🎉

kolyshkin:
v1 backport: #1383

PR #1376, opened 2022-04-26
Do not apply altsrc input values that are not set

### LINKED ISSUE #1373: v2.4.8 breaks duration default
## My urfave/cli version is

v2.4.8

## Checklist

## Dependency Management

## Describe the bug

The latest version breaks the duration parsing. Here's a sample pipeline that works with 2.4.7 but breaks with 2.4.8: https://github.com/binwiederhier/ntfy/runs/6150219289?check_suite_focus=true

Relevant option here: https://github.com/binwiederhier/ntfy/blob/main/cmd/serve.go#L34

The value for `keepalive-interval` now defaults to `0` instead of `server.DefaultKeepaliveInterval` which is `45 * time.Second`

## To reproduce

https://gist.github.com/binwiederhier/b925ee20822257f55ec04271e40a1437

## Observed behavior

`c.Duration()` returns 0

## Expected behavior

`c.Duration()` returns 45s

## Additional context
n/a

## Want to fix this yourself?
n/a

```
go version go1.18.1 linux/amd64

```

meatballhat:
## What type of PR is this?

- bug

This fixes the behavior of applied altsrc inputs so that they are only applied if the matching key exists in the input.

Fixes #1373

### DIFF
--- a/altsrc/flag.go
+++ b/altsrc/flag.go
@@ -64,7 +64,7 @@ func InitInputSourceWithContext(flags []cli.Flag, createInputSource func(cCtx *c
 
 // ApplyInputSourceValue applies a generic value to the flagSet if required
 func (f *GenericFlag) ApplyInputSourceValue(cCtx *cli.Context, isc InputSourceContext) error {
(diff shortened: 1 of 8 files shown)


meatballhat:
Working on the codecov bit now 👍🏼

PR #1241, opened 2021-02-08
Accept multi-value input on slice flags

vipally:
## What type of PR is this?

- bug

## What this PR does / why we need it:

fix v2 bug: accept multi-value input on sclice flags.

## Which issue(s) this PR fixes:

Fixes #1238

## Testing

TestFlagValue has cover this case.

## Release Notes

```
fix #1238: accept multi-value input on sclice flags
```

PR #1377, opened 2022-04-26
Accept multi-value input on slice flags (#1241)

meatballhat:
## What type of PR is this?

- bug

See #1241


dearchap (approved):
Excellent fix. Very comprehensive. Thank you for that

PR #1379, opened 2022-04-27
[v1] fix CI, add Go 1.17.x and 1.18.x

kolyshkin:
## What type of PR is this?

- cleanup

One test is failing with Go 1.17+ since it now panics instead of
returning an error. Catch the panic and consider the test passed.

With that out of the door, add Go 1.17.x and 1.18.x to the testing
matrix.

While at it, do not use && in run since it is executed by a shell with
set -e, thus making && redundant.

None (AFAICS)

None 

## Testing

That's what we have CI for!

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -15,13 +15,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        go:
-          - 1.16.x
-          # NOTE: tests fail with panic at
-          # TestApp_RunAsSubCommandIncorrectUsage on these
-          # versions:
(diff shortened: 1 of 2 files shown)


.github/workflows/cli.yml
```diff
@@ -38,6 +32,6 @@ jobs:
 
       - name: Run Tests
         run:
-          go run build.go vet &&
-          go run build.go test &&
+          go run build.go vet
+          go run build.go test
```
meatballhat, line 36: TIL that the default is to run with `set -e` 😯 !

meatballhat (approved):
Thank you! 🎉

PR #1388, opened 2022-05-06
feature: add DefaultCommand field to App

### LINKED ISSUE #1307: Allow default command if one is not passed
## Checklist

Say I have an `App` with one or two primary commands. Say I want it to default to one of those commands if a command arg isn't specifically passed, so that the app has easy and useful default behavior. 

Right now, there's no good way to do this; rather, there's no way to do it at all, if the "default" to be set takes any flags. 

Current app behavior: 

**Calling command "properly"**

`./app-binary do-thing --flag1 x -n 2`

No errors.

**Hacking together some stuff to try and have default command-like behavior**

(should default to whatever `do-thing`  is supposed to do, **including flag parsing**)

`./app-binary --flag1 x -n 2`

The above causes a UsageError, since flag parsing seems to happen at the `App` level. 

## Solution description

Implement a field on `App` similar to `Before`, `After`, and `Action`, which allows one to specify a command to be run if none are otherwise specified AND `-h/--help` isn't passed. If this field is present, flag parsing should happen at the `Command` level rather than the `App` level. 

¯\\\_(ツ)\_/¯

jalavosus:
See issue #1307 for context.

## What type of PR is this?

- feature

Enables a CLI application to run a default `Command` (by name) if no command name is passed as a cli argument.

Fixes #1307 

Genuinely unsure if I implemented this correctly, but the test cases look sane on my end and they pass. 

## Release Notes

```release-note
cli.App now has a `DefaultCommand` field, which, if set, enables a cli.App to run a specific command if no command is otherwise passed as an argument. 
```

### DIFF
--- a/app.go
+++ b/app.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"reflect"
 	"sort"
 	"time"
 )
@@ -43,6 +44,9 @@ type App struct {
 	Version string
 	// Description of the program
 	Description string
+	// DefaultCommand is the (optional) name of a command
+	// to run if no command names are passed as CLI arguments.
+	DefaultCommand string
 	// List of commands to execute
 	Commands []*Command
 	// List of flags to parse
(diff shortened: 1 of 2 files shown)


jalavosus:
Just had the thought that it might make sense to have the flag name check also check for default command's subcommands (should they exist), to make the whole thing real seamless.

dearchap:
I dont see a real use case for this. Are there any real world clis which make use of this functionality.

jalavosus:
> I dont see a real use case for this. Are there any real world clis which make use of this functionality.

@dearchap 

- https://github.com/ahmetb/kubectx (defaults to listing all configured kubectl contexts)
- yarn (defaults to `yarn install`)
- `python`/`python3` (defaults to loading a REPL if no arguments are specified)

Also, if the feature is able to be implemented (which, well, it clearly is) then why not implement it? It adds functional value to the package, and doesn't seem to be -- at least imo -- any sort of feature creep.

meatballhat:
@jalavosus Hello and sorry for the delay 😭 ! Are you up for resolving conflicts and getting tests happy? I agree with you regarding examples in the wild and I'd like to move forward with this work 👍🏼

meatballhat:
@jalavosus Thank you! Are you able to appease codecov, too? 😅

meatballhat (approved):
Lovely! 🤩

PR #1403, opened 2022-05-23
Run docs tests against current work tree

meatballhat:
## What type of PR is this?

- bug
- documentation

This change ensures the docs tests are run against the current work tree by creating a temporary go workspace to be used by gfmrun via `TMPDIR`.

N/A 😬

### DIFF
--- a/internal/build/build.go
+++ b/internal/build/build.go
@@ -182,6 +182,13 @@ func testCleanup(packages []string) error {
 func GfmrunActionFunc(cCtx *cli.Context) error {
 	top := cCtx.Path("top")
 
+	bash, err := exec.LookPath("bash")
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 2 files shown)


internal/build/build.go
```diff
@@ -179,8 +179,37 @@ func testCleanup(packages []string) error {
...
+	}
+
+	fmt.Fprintf(cCtx.App.ErrWriter, "# ---> workspace/TMPDIR is %q\n", tmpDir)
+
+	if err := runCmd("go", "work", "init", top); err != nil {
+		return err
+	}
+
+	os.Setenv("TMPDIR", tmpDir)
+	os.Setenv("SHELL", "bash")
```
mostynb, line 206: Not sure if it matters, but I think this is usually an absolute path.

mostynb (approved):

PR #1409, opened 2022-06-06
Add SliceFlag wrapper and fix bugs in existing implementations

### LINKED ISSUE #1410: IntSliceFlag and friends don't work correctly with destination (+ behavior quirks relating to value)
## My urfave/cli version is

v2.8.1

## Checklist

## Dependency Management

## Describe the bug

I've noticed a some issues / inconsistencies, while implementing https://github.com/urfave/cli/pull/1409. I'm creating this issue just to make it clear what they are.

- It appears `Destination` fields were added to the `IntSliceFlag`, `Int64SliceFlag`, and `Float64SliceFlag` structs, without actually implementing the necessary behavior
- The `StringSliceFlag.Value` (default value) field may be modified by values passed in by environment variables
- It doesn't seem like a problem, but the `Value` field seems like it should never be modified (e.g. don't initialise it if it's nil / use a local variable instead)

Past issues/PRs that may be relevant: https://github.com/urfave/cli/issues/1235, https://github.com/urfave/cli/pull/981, https://github.com/urfave/cli/pull/1178

Currently open issues that may be relevant: https://github.com/urfave/cli/issues/1224, https://github.com/urfave/cli/issues/1143

## To reproduce

N/A, but take a look at https://github.com/urfave/cli/pull/1409, particularly the existing test I modified (for the second issue)

## Observed behavior

See above

## Expected behavior

The value / destination support of `IntSliceFlag` etc should be equivalent to `StringSliceFlag`. That said, there's a bit of a question around handling the (set) empty string env var case, however. See the note on my PR.

```
go version go1.18.2 linux/amd64
```

```
GO111MODULE="on"
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/joey/.cache/go-build"
GOENV="/home/joey/.config/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/joey/under/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/joey/under/go"
GOPRIVATE=""
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOVCS=""
GOVERSION="go1.18.2"
GCCGO="gccgo"
GOAMD64="v1"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/f/dev/urfave-cli/go.mod"
GOWORK=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build1554591512=/tmp/go-build -gno-record-gcc-switches"
```

joeycumines:
## What type of PR is this?

- bug
- feature

The SliceFlag implementation and associated aliases (MultiStringFlag, etc) extend the existing slice implementations (StringSliceFlag, etc) to support actual slices as the flag value and destination.

This change also fixes various bugs in the existing implementation. Notably, the StringSliceFlag.Apply implementation would modify the input (default) Value, if an env var was set, and no destination was provided. The bugs fixed in the other three implementations were all already fixed in either StringSliceFlag, or in one case (ignoring empty env var) in Float64SliceFlag.

Fixes #1410

I'm not sure if the "ignoring empty env var" behavior is desirable or not. It seems like it should be made consistent, at least.

## Release Notes

```release-note
Add SliceFlag and aliases (MultiStringFlag, MultiIntFlag, MultiFloat64Flag, MultiInt64Flag).
Fix value/destination behavior for slice flags.
```

### DIFF
--- a/sliceflag.go
+++ b/sliceflag.go
@@ -1,3 +1,6 @@
+//go:build go1.18
+// +build go1.18
+
 package cli
 
 import (
@@ -225,4 +228,66 @@ func unwrapFlagValue(v flag.Value) flag.Value {
 		}
 		v = h.value
 	}
+}
+
+// NOTE: the methods below are in this file to make use of the build constraint
+
+func (f *Float64SliceFlag) SetValue(slice []float64) {
+	f.Value = newSliceFlagValue(NewFloat64Slice, slice)
+}
+
+func (f *Float64SliceFlag) SetDestination(slice []float64) {
+	f.Destination = newSliceFlagValue(NewFloat64Slice, slice)
+}
+
+func (f *Float64SliceFlag) GetDestination() []float64 {
+	if destination := f.Destination; destination != nil {
+		return destination.Value()
+	}
+	return nil
+}
+
+func (f *Int64SliceFlag) SetValue(slice []int64) {
+	f.Value = newSliceFlagValue(NewInt64Slice, slice)
+}
+
+func (f *Int64SliceFlag) SetDestination(slice []int64) {
+	f.Destination = newSliceFlagValue(NewInt64Slice, slice)
+}
+
+func (f *Int64SliceFlag) GetDestination() []int64 {
+	if destination := f.Destination; destination != nil {
+		return destination.Value()
+	}
+	return nil
+}
+
+func (f *IntSliceFlag) SetValue(slice []int) {
+	f.Value = newSliceFlagValue(NewIntSlice, slice)
+}
+
+func (f *IntSliceFlag) SetDestination(slice []int) {
+	f.Destination = newSliceFlagValue(NewIntSlice, slice)
+}
+
+func (f *IntSliceFlag) GetDestination() []int {
+	if destination := f.Destination; destination != nil {
+		return destination.Value()
+	}
+	return nil
+}
+
+func (f *StringSliceFlag) SetValue(slice []string) {
+	f.Value = newSliceFlagValue(NewStringSlice, slice)
+}
+
+func (f *StringSliceFlag) SetDestination(slice []string) {
+	f.Destination = newSliceFlagValue(NewStringSlice, slice)
+}
+
+func (f *StringSliceFlag) GetDestination() []string {
+	if destination := f.Destination; destination != nil {
+		return destination.Value()
+	}
+	return nil
 }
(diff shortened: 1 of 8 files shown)


joeycumines:
Thanks @meatballhat :)

I've added a build constraint + restructured it slightly, so it should no longer impact support for older versions of Go.

joeycumines:
Could I get an update on getting this merged?

I don't want to forget I have this PR open, is all :)

Looks like it passes CI now

meatballhat:
@joeycumines Sorry for my delay 😭 This week was 1000% more kiddo chaos than usual for me. Weekend mode: engage 🤘🏼

sliceflag.go
```diff
@@ -0,0 +1,293 @@
...
+	if value, ok := x.value.(Serializer); ok {
+		return value.Serialize()
+	}
+	return x.String()
+}
+
+// applyFlagValueHook wraps calls apply then wraps flags to call a hook function on update and after initial apply.
+func applyFlagValueHook(set *flag.FlagSet, apply func(set *flag.FlagSet) error, hook func()) error {
+	if apply == nil || set == nil || hook == nil {
+		panic(`invalid input`)
```
meatballhat, line 202: Given that this is a library and this function already has an `error` return signature, I think I'd rather this not `panic`. WDYT?

meatballhat (approved):
@joeycumines 🤯 Wow. Very wow.

sliceflag.go
joeycumines: IMO, panics are ideal for "misuse of API" cases, especially internal / unexported APIs, as, generally speaking, panics are more likely to be noticed, and you want to notice any such cases. The least-effort alternative is just having it panic once it tries to call a nil function, which is similar, but more likely to go unnoticed.

If there was a use case for having it return an error, that could always be added later.

PR #1415, opened 2022-06-18
Upgrade gopkg.in/yaml to v3

jszwedko:
## What type of PR is this?

- bug

Fixes vulnerability: https://github.com/advisories/GHSA-hp87-p4gw-j4gq

yaml.v3 deserializes nested maps as `map[string]interface{}` now. TOML still deserializes nested maps as `map[interface{}]interface{}`. I considered updating `MapInputSource` to store a `map[string]interface{}` for its `valueMap` but this would break the API of `NewMapInputSource` so I went with the approach you see here of handling both `map` types when looking up values.

## Testing

`go build ./...`
`go test ./...`

## Release Notes

```release-note
NONE
```

### DIFF
--- a/altsrc/map_input_source.go
+++ b/altsrc/map_input_source.go
@@ -35,11 +35,10 @@ func nestedVal(name string, tree map[interface{}]interface{}) (interface{}, bool
 
 			switch child := child.(type) {
 			case map[string]interface{}:
-				m := make(map[interface{}]interface{}, len(child))
+				node = make(map[interface{}]interface{}, len(child))
 				for k, v := range child {
-					m[k] = v
+					node[k] = v
 				}
-				node = m
 			case map[interface{}]interface{}:
 				node = child
 			default:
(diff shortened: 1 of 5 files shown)


altsrc/map_input_source.go
```diff
@@ -32,11 +32,19 @@ func nestedVal(name string, tree map[interface{}]interface{}) (interface{}, bool
 			if !ok {
 				return nil, false
 			}
-			ctype, ok := child.(map[interface{}]interface{})
-			if !ok {
+
+			switch child := child.(type) {
+			case map[string]interface{}:
+				m := make(map[interface{}]interface{}, len(child))
```
mostynb, line 38: Bit of a nitpick, but I guess you could use `node = make(...` here. It's not a big deal though.

mostynb (dismissed):

altsrc/map_input_source.go
jszwedko: Good point :slightly_smiling_face:

meatballhat (approved):

PR #1422, opened 2022-06-22
Add SetRequired method to Flag interface

### LINKED ISSUE #1329: Is there an easy way to toggle the Required attribute of a cli.Flag ?
Now I have to create separate `cli.Flag` flag instances for `Required` and `Optional`, which is quite tedious:

```golang
// HeightFlag ...
var HeightFlag = cli.IntFlag{
	Name:     "height",
	Usage:    "specify height",
	Required: true,
}

// OptionalHeightFlag ...
var OptionalHeightFlag = cli.IntFlag{
	Name:  "height",
	Usage: "specify optional height",
}
```

jalavosus:
Flags can now have their "Required" field set at the interface level.

## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

Fixes #1329 

## Release Notes

```release-note
Flags can now have their "Required" field set at the interface level with `SetRequired(bool)`.
```

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -2083,6 +2083,8 @@ func (c *customBoolFlag) IsSet() bool {
 	return false
 }
 
+func (c *customBoolFlag) SetRequired(bool) {}
+
 func TestCustomFlagsUnused(t *testing.T) {
 	app := &App{
 		Flags:  []Flag{&customBoolFlag{"custom"}},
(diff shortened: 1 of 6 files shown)


dearchap:
I dont think this is a good idea. Its not that the implemented is bad or anything, the problem is that so many users are already using Flag interface as is so if this is implemented anyone who has custom flag implementations would have to implement this. Plus there is a RequiredFlag interface already https://github.com/urfave/cli/blob/main/flag.go#L100 so this change can go in there.

dearchap:
@jalavosus I'm closing this as this is not the direction we want to go with Flag interface right now. Please consider contributing this to v3 which is ongoing. Thank you

PR #1432, opened 2022-07-09
accept timezone for timestamps

julian7:
## What type of PR is this?

- feature

When `TimestampFlag` is used with implicit timezone, default `time.Parse` behavior is assumed, which is parsing time in some timezone, and provide a UTC time. I haven't tested it myself, but I've read somewhere this works just with local timezone on windows, but it does this in UTC on linux. Either way, `time.Parse` assumes UTC times on linux and mac.

The suggested solution is to use `time.ParseInLocation` instead, and provide a `*time.Location` value for the default timezone.

This PR adds a `Timezone` member into `TimestampFlag` struct. Input parsing runs `time.ParseInLocation` when location is present, but keeps old behavior if Timezone is nil, to guarantee backwards compatibility.

I haven't opened a separate issue for this PR.

I haven't included v2approve output.

## Testing

This PR adds a new test, TestTimestampFlagApply_Timezoned (copied from TestTimestampFlagApply), which constructs a static timezone called PDT (UTC-7h), and expects the example UTC time provided in PDT timezone.

## Release Notes

```release-note
Add optional timezone (*time.Location) to TimezoneFlag to assume other timezones than UTC
```

### DIFF
--- a/docs/v2/manual.md
+++ b/docs/v2/manual.md
@@ -1413,6 +1413,19 @@ In this example the flag could be used like this:
 $ myapp --meeting 2019-08-12T15:04:05
 ```
 
+When the layout doesn't contain timezones, timestamp will render with UTC. To
+change behavior, a default timezone can be provided with flag definition:
+
+```go
+app := &cli.App{
+	Flags: []cli.Flag{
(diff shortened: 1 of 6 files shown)


meatballhat (approved):
Lovely! Thank you 🎉

Issue #1408, opened 2022-06-06
Strict flag lookup


icholy:
## Checklist

* [x] Are you running the latest v2 release? The list of releases is [here](https://github.com/urfave/cli/releases).
* [x] Did you check the manual for your release? The v2 manual is [here](https://github.com/urfave/cli/blob/main/docs/v2/manual.md).
* [x] Did you perform a search about this feature? Here's the [Github guide](https://help.github.com/en/github/managing-your-work-on-github/using-search-to-filter-issues-and-pull-requests) about searching.

## What problem does this solve?

I've had several cases where I've made a typo when accessing a flag value via the context.

## Solution description

I'd like the library to panic if I access an undefined flag.

``` go
func main() {
	app := &cli.App{
		Flags: []cli.Flag{
			&cli.StringFlag{
				Name: "foo",
			},
		},
		Action: func(c *cli.Context) error {
		
			// I want this to panic
			_ = c.String("bar")
			return nil
		},
	}
	app.Run([]string{"test"})
}
```

meatballhat:
@icholy you may want to consider using the recently-added (by @toaster 🙇🏼) ability to [get the value directly from a given flag](https://github.com/urfave/cli/pull/1367#issuecomment-1133478036)

icholy:
I'd much prefer to keep my existing code and just enable some type of "strict mode" flag.

dearchap:
@icholy This would be nice, are you willing to submit a PR ?

icholy:
@dearchap if you want to, go for it.

PR #1434, opened 2022-07-12
Add App.EnableStrictLookup option

icholy:
## What type of PR is this?

- feature

The current behaviour returns a zero value when accessing an undefined flag. This hides programmer errors and has resulted in several bugs on my end.

https://github.com/urfave/cli/issues/1408

## Testing

I added tests which make sure `Context.Set` and `Context.Value` panic when accessing an undefined flag.

## Release Notes

```release-note
A new `App.EnableStrictLookup` option which causes undefined flag lookups to panic
```

### DIFF
--- a/app.go
+++ b/app.go
@@ -53,6 +53,8 @@ type App struct {
 	Flags []Flag
 	// Boolean to enable bash completion commands
 	EnableBashCompletion bool
+	// Boolean to enable strict flag lookup
+	EnableStrictLookup bool
 	// Boolean to hide built-in help command and help flag
 	HideHelp bool
 	// Boolean to hide built-in help command but keep help flag.
--- a/context.go
+++ b/context.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"context"
 	"flag"
+	"fmt"
 	"strings"
 )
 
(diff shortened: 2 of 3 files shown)


meatballhat (changes requested):
Thank you for adding this!

I would like to see two changes:
- statically define an `error` type (`cli.ErrStrictFlagLookup`?) that can be reliably checked via funcs like `errors.Is`
- return this `error` instead of using `panic`

icholy:
Where would I return this error? The flag value getters don't return errors.

meatballhat:
@icholy yes! Sorry for not being clear about that distinction. What you propose LGTM 🤘🏼

julian7:
I'm also wondering whether handling this event via a hook would be more powerful.

Like, instead of having `App. EnableStrictLookup` boolean, we could have an `App.UnknownFlagHandler` or something like that. Then, the implementer can decide whether the app should throw `panic()`, whether it should log it, or perhaps even change context on such an event.

icholy:
@julian7 that's an interesting idea. Something like this?

``` go
type App struct {
    // ...
    UnknownFlagHandler func(ctx Context, name string)
}
```

meatballhat:
> I'm also wondering whether handling this event via a hook would be more powerful.

Oooh I like that idea a lot @julian7

PR #1446, opened 2022-08-02
Add App.InvalidFlagAccessHandler

icholy:
Follow up from #1434

### DIFF
--- a/app.go
+++ b/app.go
@@ -78,6 +78,8 @@ type App struct {
 	CommandNotFound CommandNotFoundFunc
 	// Execute this function if a usage error occurs
 	OnUsageError OnUsageErrorFunc
+	// Execute this function when an unknown flag is accessed from the context
+	UnknownFlagHandler UnknownFlagFunc
 	// Compilation date
 	Compiled time.Time
 	// List of all authors who contributed
--- a/app.go
+++ b/app.go
@@ -78,8 +78,8 @@ type App struct {
 	CommandNotFound CommandNotFoundFunc
 	// Execute this function if a usage error occurs
 	OnUsageError OnUsageErrorFunc
-	// Execute this function when an unknown flag is accessed from the context
-	UnknownFlagHandler UnknownFlagFunc
+	// Execute this function when an invalid flag is accessed from the context
+	InvalidFlagAccessHandler InvalidFlagAccessFunc
 	// Compilation date
 	Compiled time.Time
 	// List of all authors who contributed
--- a/context.go
+++ b/context.go
@@ -47,7 +47,7 @@ func (cCtx *Context) NumFlags() int {
 // Set sets a context flag to a value.
 func (cCtx *Context) Set(name, value string) error {
 	if cCtx.flagSet.Lookup(name) == nil {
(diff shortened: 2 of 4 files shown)


julian7:
I couldn't test it myself, but I like this PR.

icholy:
@julian7 I don't like the name. It sounds like it will get called when someone passes an unknown flag, ex `mycli --unknown-flag`

julian7:
> @julian7 I don't like the name. It sounds like it will get called when someone passes an unknown flag, ex `mycli --unknown-flag`

Hmm, reasonable. What about UndefinedFlagHandler?

icholy:
> > @julian7 I don't like the name. It sounds like it will get called when someone passes an unknown flag, ex `mycli --unknown-flag`
> 
> Hmm, reasonable. What about UndefinedFlagHandler?

That name has the same problem.

app.go
```diff
@@ -78,6 +78,8 @@ type App struct {
 	CommandNotFound CommandNotFoundFunc
 	// Execute this function if a usage error occurs
 	OnUsageError OnUsageErrorFunc
+	// Execute this function when an unknown flag is accessed from the context
+	UnknownFlagHandler UnknownFlagFunc
```
dearchap, line 82: I think InvalidFlagAccessHandler would be better

context_test.go
```diff
@@ -150,6 +150,19 @@ func TestContext_Value(t *testing.T) {
 	expect(t, c.Value("unknown-flag"), nil)
 }
 
+func TestContext_Value_InvalidFlagAccessHandler(t *testing.T) {
```
dearchap, line 153: Can you add a test with one app, one command and one subcommand.  ? I dont think it will work with the way you expect as of now since you have to set InvalidFlagAccessHandler in Command.startApp.

dearchap (changes requested):
Add more tests with commands and subcommands.

context_test.go
icholy: Fixed it.

context.go
```diff
@@ -158,7 +161,7 @@ func (cCtx *Context) lookupFlagSet(name string) *flag.FlagSet {
 			return c.flagSet
 		}
 	}
-
+	cCtx.onInvalidFlag(name)
```
dearchap, line 164: I would prefer if you move this call into Context.Value() . Thats the direct calling place from user so its better to be closer to there. lookupFlagSet is an internal function and so we shouldnt expect the user to get notified is we lookup something internally

dearchap (changes requested):
Move function call to Value

context.go
icholy: That won't work because the other getter functions don't call `Context.Value` internally.  https://github.com/urfave/cli/blob/v2.11.2/flag_uint.go#L75

icholy: @dearchap any ideas on a better place to do the check?

dearchap: let me look at it.

dearchap (approved):

dearchap:
@icholy lets leave it as it is. Thanks

PR #1448, opened 2022-08-11
Add support for alias in YAMLs

### LINKED ISSUE #1394: Support aliases in .yml files (altsrc.InitConfigFileInputSourceFunc)
## Checklist

* [ ] Did you perform a search about this feature? Here's the [Github guide](https://help.github.com/en/github/managing-your-work-on-github/using-search-to-filter-issues-and-pull-requests) about searching.

Thank you for this project. It's fantastic. I've been a long time user and I'm very happy.

I just added the ability to use the `Aliases` in my yml file when loading through `altsrc` (see https://github.com/binwiederhier/ntfy/commit/91594faf288d13bd128067436e11345dc7a9d583#diff-d1253876bc7a1b30f67245948f61e61430af2f3f0bf72780e3c0314a24d152deR43-R50). I was wondering if that was something you'd be interested in adding as a feature. 

If not, I'm alright too and you can close this ticket.

## Solution description

Example:

``` go
altsrc.NewStringFlag(&cli.StringFlag{Name: "listen-http", Aliases: []string{"listen_http", "l"},  ...}),
```

Should be loadable via the .yml file:

``` yaml
# Currently supported:
listen-http: ":80"

# Would also be supported with the suggested change (see Aliases above):
listen_http: ":80"
l: ":80"
```

james-prysm:
## What type of PR is this?

_(REQUIRED)_

- bug

Currently, one can not process aliased flags via yaml config file because only the original name is used in the flag lookup.

Fixes #1394

**First time trying to create a PR.
Supports all flag types defined in flag.go**

## Testing

Unit Tested

## Release Notes

```release-note
Supports the setting of flags through an alias name via yaml config file
```


Dokiys (dismissed):

altsrc/flag.go
```diff
@@ -121,28 +133,34 @@ func (f *IntSliceFlag) ApplyInputSourceValue(cCtx *cli.Context, isc InputSourceC
...
 // ApplyInputSourceValue applies a Bool value to the flagSet if required
 func (f *BoolFlag) ApplyInputSourceValue(cCtx *cli.Context, isc InputSourceContext) error {
-	if f.set != nil && !cCtx.IsSet(f.Name) && !isEnvVarSet(f.EnvVars) && isc.isSet(f.BoolFlag.Name) {
-		value, err := isc.Bool(f.BoolFlag.Name)
-		if err != nil {
-			return err
-		}
-		for _, name := range f.Names() {
-			_ = f.set.Set(name, strconv.FormatBool(value))
+	if f.set != nil && !cCtx.IsSet(f.Name) && !isEnvVarSet(f.EnvVars) {
```
Dokiys, line 136: @dearchap @james-prysm Do you think there is too mush nesting? Should we return early when don't need to handle. For example:
```
	if f.set == nil || cCtx.IsSet(f.Name) || isEnvVarSet(f.EnvVars) {
		return nil
	}
	for _, name := range f.BoolFlag.Names() {
		if !isc.isSet(name) {
			continue
		}
		value, err := isc.Bool(name)
		if err != nil {
			return err
		}
		for _, n := range f.Names() {
			_ = f.set.Set(n, strconv.FormatBool(value))
		}
	}
```

Dokiys (approved):

altsrc/flag.go
dearchap: @Dokiys definitely your approach is more readable. You want to make the changes and push ?

dearchap (approved):

PR #1451, opened 2022-08-14
Move genflags tool to cmd/ and pin to previous release

meatballhat:
## What type of PR is this?

- bug
- cleanup

The goal here is to to alleviate problems caused by the circular dependency of using the same code as a library that is potentially being generated to adhere to a different API.

Connected to #1257


dearchap (approved):
nice

PR #1454, opened 2022-08-14
issue_62: Make slice options more posix like

### LINKED ISSUE #62: Slice Flags help output is unexpected/confusing
Using a flag such as 

```
cli.StringSliceFlag{"lang", &cli.StringSlice{"english", "spanish"}, "language for the greeting"},
```

results in help output like this:

```
OPTIONS:
   --lang [english spanish] `-lang option -lang option` language for the greeting
```

What exactly is the intended meaning of "`-lang option -lang option`"? Am I just uninformed or is this a WIP? I'd be happy to work on it if I know what it's supposed to do! :)

dearchap:
## What type of PR is this?

Rearranges stringify slice code to within each slice file rather than in flag.go

- cleanup

Makes multi option more informative

## Testing

Unit tests run. 

## Release Notes

_(REQUIRED)_

```release-note

```

### DIFF
--- a/flag.go
+++ b/flag.go
@@ -7,7 +7,6 @@ import (
 	"io/ioutil"
 	"regexp"
 	"runtime"
-	"strconv"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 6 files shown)


flag.go
```diff
@@ -377,11 +329,12 @@ func stringifySliceFlag(usage string, names, defaultVals []string) string {
 	}
 
 	usageWithDefault := strings.TrimSpace(fmt.Sprintf("%s%s", usage, defaultVal))
-	multiInputString := "(accepts multiple inputs)"
+	/*multiInputString := fmt.Sprintf("[ %s ]", usage)
 	if usageWithDefault != "" {
 		multiInputString = "\t" + multiInputString
-	}
-	return fmt.Sprintf("%s\t%s%s", prefixedNames(names, placeholder), usageWithDefault, multiInputString)
+	}*/
```
meatballhat, line 335: Is this leftover commented-out stuff?

meatballhat (approved):
Lovely! I'm not sure that we explicitly say that help templates are not strictly part of the API, but I feel like we should 😁  In any case, I think this counts as a feature addition, so the version will bump by minor after this merges anyway.

flag.go
dearchap: Yes. Updated

PR #1456, opened 2022-08-16
Fix:(issue_1455) Allow bool flags from input altsrc

### LINKED ISSUE #1455: BoolFlag from file/altsrc default handling
## My urfave/cli version is

v2

## Checklist

- [ X] Are you running the latest v2 release? The list of releases is [here](https://github.com/urfave/cli/releases).
- [X ] Did you check the manual for your release? The v2 manual is [here](https://github.com/urfave/cli/blob/main/docs/v2/manual.md)
- [X ] Did you perform a search about this problem? Here's the [Github guide](https://help.github.com/en/github/managing-your-work-on-github/using-search-to-filter-issues-and-pull-requests) about searching.

## Dependency Management

## Describe the bug

I have a (nested) flag `enabled`
```
		altsrc.NewBoolFlag(&cli.BoolFlag{
			Name:        prefix + ".enabled",
			Usage:       "enable open telemetry tracing",
			Value:       true,
			EnvVars:     []string{"OPENTELEMETRY_ENABLED"},
			Destination: &ot.Enabled,
		}),
```
When I set this to `false` in the file, it seems to not be recognized (the value stays `true` in the final lookup).

I traced this to this function:
```
// ApplyInputSourceValue applies a Bool value to the flagSet if required
func (f *BoolFlag) ApplyInputSourceValue(cCtx *cli.Context, isc InputSourceContext) error {
	if f.set != nil && !cCtx.IsSet(f.Name) && !isEnvVarSet(f.EnvVars) && isc.isSet(f.BoolFlag.Name) {
		value, err := isc.Bool(f.BoolFlag.Name)
		if err != nil {
			return err
		}
		if value {
				_ = f.set.Set(name, strconv.FormatBool(value))
			}
		}
	}
	return nil
}
```

It seems there's an implicit requirement that the default value for BoolVal flags is `false` and you can only set them to `true`? I.e. `if value` has to be true (meaning the value from the FILE has to be `true` for the value to be `f.set.Set()` in the config struct. Is that correct?

## To reproduce

Create a boolVal flag with Default=False, set the value in a YAML file and read it

## Observed behavior

The boolVal is still True, instead of False.

## Expected behavior

The BoolVal should be False

## Want to fix this yourself?

this bug is easily explained and very small, free free to create a
pull request for it.

```
go version go1.19 darwin/arm64
```

```
GO111MODULE=""
GOARCH="arm64"
GOBIN=""
GOCACHE="/Users/jvilhuber/Library/Caches/go-build"
GOENV="/Users/jvilhuber/Library/Application Support/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="arm64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/jvilhuber/go/pkg/mod"
GONOPROXY="XXX.com/*"
GONOSUMDB="XXX.com/*"
GOOS="darwin"
GOPATH="/Users/jvilhuber/go"
GOPRIVATE="XXX.com/*"
GOROOT="/opt/homebrew/Cellar/go/1.19/libexec"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/opt/homebrew/Cellar/go/1.19/libexec/pkg/tool/darwin_arm64"
GOVCS=""
GOVERSION="go1.19"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/jvilhuber/PLAYG/foo/go.mod"
GOWORK=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -arch arm64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/s5/8qspwhrs4438zmz3q_yh6kh80000gs/T/go-build2516793279=/tmp/go-build -gno-record-gcc-switches -fno-common"
```

dearchap:
## What type of PR is this?

- bug

Allow bool values to be set from altsrc files

## Testing

Updated unit tests 

## Release Notes

_(REQUIRED)_

```release-note

```

### DIFF
--- a/altsrc/flag.go
+++ b/altsrc/flag.go
@@ -124,13 +124,14 @@ func (f *BoolFlag) ApplyInputSourceValue(cCtx *cli.Context, isc InputSourceConte
 	if f.set != nil && !cCtx.IsSet(f.Name) && !isEnvVarSet(f.EnvVars) && isc.isSet(f.BoolFlag.Name) {
 		value, err := isc.Bool(f.BoolFlag.Name)
 		if err != nil {
+			fmt.Println(err)
 			return err
 		}
-		if value {
-			for _, name := range f.Names() {
-				_ = f.set.Set(name, strconv.FormatBool(value))
-			}
+		for _, name := range f.Names() {
+			_ = f.set.Set(name, strconv.FormatBool(value))
 		}
+	} else {
+		fmt.Println("not fill")
 	}
 	return nil
 }
--- a/altsrc/flag.go
+++ b/altsrc/flag.go
@@ -124,14 +124,11 @@ func (f *BoolFlag) ApplyInputSourceValue(cCtx *cli.Context, isc InputSourceConte
 	if f.set != nil && !cCtx.IsSet(f.Name) && !isEnvVarSet(f.EnvVars) && isc.isSet(f.BoolFlag.Name) {
 		value, err := isc.Bool(f.BoolFlag.Name)
 		if err != nil {
-			fmt.Println(err)
 			return err
 		}
 		for _, name := range f.Names() {
 			_ = f.set.Set(name, strconv.FormatBool(value))
 		}
-	} else {
-		fmt.Println("not fill")
 	}
 	return nil
 }
--- a/altsrc/json_command_test.go
+++ b/altsrc/json_command_test.go
@@ -11,7 +11,7 @@ import (
 
 const (
 	fileName   = "current.json"
-	simpleJSON = `{"test": 15}`
+	simpleJSON = `{"test": 15, "testb": false}`
 	nestedJSON = `{"top": {"test": 15}}`
 )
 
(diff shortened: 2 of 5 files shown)


altsrc/flag.go
```diff
@@ -124,13 +124,14 @@ func (f *BoolFlag) ApplyInputSourceValue(cCtx *cli.Context, isc InputSourceConte
 	if f.set != nil && !cCtx.IsSet(f.Name) && !isEnvVarSet(f.EnvVars) && isc.isSet(f.BoolFlag.Name) {
 		value, err := isc.Bool(f.BoolFlag.Name)
 		if err != nil {
+			fmt.Println(err)
```
Dokiys, line 127: If printing is required here?

dearchap: Ah nice catch. Those are debug logs. I've removed them.

Dokiys (approved):

altsrc/flag.go
Dokiys: Seems to have the same problem at 
[*StringFlag.ApplyInputSourceValue](https://github.com/urfave/cli/blob/b4df361387226948b6cbaf53da539e26c7562cce/altsrc/flag.go#L145)

PR #1459, opened 2022-08-19
Fix:(issue_1094) Dont execute Before/After during shell completions

### LINKED ISSUE #1094: Before callback affects bash completions
## my urfave/cli version is

v1.22.2, v1.22.3

## Checklist

* [x] Are you running the latest v1 release? The list of releases is [here](https://github.com/urfave/cli/releases).
* [x] Did you check the manual for your release? The v1 manual is [here](https://github.com/urfave/cli/blob/master/docs/v1/manual.md)

## Dependency Management

## Describe the bug

I wanted to introduce pre-check (with `.Before` callback) that would be run if the command is being executed. But it seems like the pre-check is also executed when I want to only show the completions for a given command what breaks the completions output if the `.Before` has printed anything (in particular the error).

## To reproduce

```go
package main

import (
	"errors"
	"fmt"
	"log"
	"os"

	"github.com/urfave/cli"
)

func main() {
	app := cli.NewApp()
	app.Name = "example"
	app.Before = func(_ *cli.Context) error { return errors.New("uups") }
	app.Commands = []cli.Command{{
		Name:   "cmd",
		Action: func(c *cli.Context) error { return nil },
		Flags:  []cli.Flag{cli.StringFlag{Name: "abc"}},
		BashComplete: func(c *cli.Context) {
			fmt.Println("--abc")
		},
	}}
		fmt.Println("Hello friend!")
		return nil
	}

	err := app.Run(os.Args)
	if err != nil {
		log.Fatal(err)
	}
}
```

1. Build the code above.
2. Enable completions
3. `example cmd [TAB]` 

## Observed behavior

```console
$ example cmd ......2020/03/25 11:53:14 uups
$ example cmd
GLOBAL OPTIONS  COMMANDS  USAGE  NAME  --
                                                                                example - A new cli application                                           uups
   --help, -h  show help                                                        example [global options] command [command options] [arguments...]
   cmd                                                                          help, h  Shows a list of commands or help for one command
```

## Expected behavior

Show completions for the given cmd:
```
$ example cmd [TAB]
--abc
```

## Additional context

Looking at the code I've noticed that the commands are parsed/visited hierarchically and this also applies for the completions. But I'm not sure if the order should be as it is now. To me completions should be executed first even if the command is nested.

```
go version go1.14 darwin/amd64
```

dearchap:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

Before/After handlers shouldnt be invoked during shell completion

_(REQUIRED)_

## Testing

go test -run=TestApp_BeforeAfterFuncShellCompletion

## Release Notes

_(REQUIRED)_

```release-note

```

### DIFF
--- a/app.go
+++ b/app.go
@@ -248,7 +248,7 @@ func (a *App) Run(arguments []string) (err error) {
 		return cerr
 	}
 
-	if a.After != nil {
+	if a.After != nil && !context.shellComplete {
 		defer func() {
 			if afterErr := a.After(context); afterErr != nil {
 				if err != nil {
@@ -260,7 +260,7 @@ func (a *App) Run(arguments []string) (err error) {
 		}()
 	}
 
-	if a.Before != nil {
+	if a.Before != nil && !context.shellComplete {
 		beforeErr := a.Before(context)
 		if beforeErr != nil {
 			a.handleExitCoder(context, beforeErr)
@@ -374,7 +374,7 @@ func (a *App) RunAsSubcommand(ctx *Context) (err error) {
 		return cerr
 	}
 
-	if a.After != nil {
+	if a.After != nil && !context.shellComplete {
 		defer func() {
 			afterErr := a.After(context)
 			if afterErr != nil {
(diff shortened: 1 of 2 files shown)


VirrageS:
@dearchap LGTM. Tested and works :)

talarian1:
@dearchap, @Dokiys  We will be happy to see this improvement as well  :)

Dokiys (approved):

PR #1464, opened 2022-08-25
Add configurable Base to int, uint and uint64 flags

### LINKED ISSUE #1462: Allow to configure the base for Integer parsing
## Checklist

We pass a year and month as CLI flags/envs (each have their own flag/env var).
Unfortunately, when passing a month with preceding 0 like `08` for August it gets parsed as octal, because of `strconv.ParseInt(..., 0, ...) (see https://github.com/urfave/cli/blob/b4df361387226948b6cbaf53da539e26c7562cce/flag_int.go#L47) it's set to autodetect (https://pkg.go.dev/strconv#ParseInt).
In cases where the flag or env var is set by another tool with preceding 0, this can lead to unexpected issues like https://github.com/vshn/appuio-odoo-adapter/issues/57.

If we could configure which base the parser should use, it would solve our issue.
See playground here: https://go.dev/play/p/a8VllHp_Z6D

## Solution description

Add a property to all Int flags to let us explicitly configure the base for our parsing, e.g. `Base` or `ParseBase` (name tbd) that is then used in `ParseInt`.
Since Go's default value for Integers are 0, it would match what is currently used in `ParseInt`, thus I don't think it causes a breaking change.

* Configure the month flag as StringFlag and do the parsing explicitly within the Before/Action functions.

Other workarounds are all outside of urface/cli:

* Combine the year and month flag into a date flag with format ("yyyy-mm"), this would be a breaking change for the user.
* Force any external tool that provide the value for the month to trim zeros when invoking the CLI. Depending on the tool used this might not always be possible.

ccremer:
This allows users to configure the basis for integer parsing.

## What type of PR is this?

- feature

This allows users to configure the base for integer parsing.

Fixes #1462 

* I didn't find the location where I could add godoc for the `Base` field and its meaning.
Is this not possible? 
* I didn't yet add the same functionality for the IntSlice and UInt64Slice flags. I couldn't find a suitable test case where such a test fits. Do you think it's worth having the base configurable for the slice flags as well? If yes, which test case should I update?
* There seem to be several lookup functions, e.g. https://github.com/urfave/cli/blob/77a5feffee931936e8fb1a7aabf01fb63b1b3eb6/flag_int.go#L83-L93
I don't have access to the `base` field in there resp. the flagset. Is this going to be a problem?

## Testing

`make test`

## Release Notes

_(REQUIRED)_

```release-note
Allow to configure the base for integer parsing in int, uint and uint64 flags
```

### DIFF
--- a/flag-spec.yaml
+++ b/flag-spec.yaml
@@ -5,11 +5,19 @@
 flag_types:
   bool: {}
   float64: {}
-  int64: {}
-  int: {}
+  int64:
+    struct_fields:
+      - { name: Base, type: int }
+  int:
+    struct_fields:
+      - { name: Base, type: int }
   time.Duration: {}
-  uint64: {}
-  uint: {}
+  uint64:
+    struct_fields:
+      - { name: Base, type: int }
+  uint:
+    struct_fields:
+      - { name: Base, type: int }
 
   string:
     struct_fields:
--- a/flag_int.go
+++ b/flag_int.go
@@ -44,7 +44,7 @@ func (f *IntFlag) GetEnvVars() []string {
 func (f *IntFlag) Apply(set *flag.FlagSet) error {
 	if val, source, found := flagFromEnvOrFile(f.EnvVars, f.FilePath); found {
 		if val != "" {
-			valInt, err := strconv.ParseInt(val, 0, 64)
+			valInt, err := strconv.ParseInt(val, f.Base, 64)
 
 			if err != nil {
 				return fmt.Errorf("could not parse %q as int value from %s for flag %s: %s", val, source, f.Name, err)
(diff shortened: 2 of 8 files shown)


dearchap (commented):
LGTM. do you think we need to specify somewhere in flag help that the base is X ?

ccremer:
> LGTM. do you think we need to specify somewhere in flag help that the base is X ?

A bit of godoc would certainly help the developers that are setting up the flags, otherwise they'd have to read the source code to understand its meaning and the accepted values. Might also be enough to mention it in the docs, like the timestamp flag (https://cli.urfave.org/v2/#timestamp-flag). Maybe the flag-spec yaml could receive an additional `godoc` property to render it in https://github.com/urfave/cli/blob/main/cmd/urfave-cli-genflags/generated.gotmpl#L26 ?

Whether the base should also be mentioned in `--help` (if that was your question) I don't have strong opinions there.

dearchap (approved):

PR #1490, opened 2022-09-11
Move more functionality into internal/build/build.go

meatballhat:
## What type of PR is this?

- cleanup
- maintenance

Moves more functionality into the internal "build script", uses make targets in CI and passes flag spec YAML through yq if available.

This patch also includes the result of running `make v2approve` since the `v2diff` command wasn't working correctly.

NA 😬

### DIFF
--- a/internal/build/build.go
+++ b/internal/build/build.go
@@ -5,12 +5,17 @@ package main
 import (
 	"bufio"
 	"bytes"
+	"errors"
 	"fmt"
+	"io"
 	"log"
 	"math"
+	"net/http"
+	"net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/urfave/cli/v2"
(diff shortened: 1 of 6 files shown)


dearchap (approved):

internal/build/build.go
```diff
@@ -131,6 +191,43 @@ func runCmd(arg string, args ...string) error {
...
+	}
+
+	if err := os.MkdirAll(filepath.Dir(dest), dirPerm); err != nil {
+		return err
+	}
+
+	out, err := os.Create(dest)
+	if err != nil {
+		return err
+	}
```
icholy, line 218: defer close here.

internal/build/build.go
```diff
@@ -411,6 +513,125 @@ func GenerateActionFunc(cCtx *cli.Context) error {
...
+	if err != nil {
+		if !cCtx.Bool("strict") {
+			fmt.Fprintln(cCtx.App.ErrWriter, "# ---> no yq found; skipping")
+			return nil
+		}
+
+		return err
+	}
+
+	os.Chdir(cCtx.Path("top"))
```
icholy, line 527: Missing error check? (there's a few of these).

PR #1494, opened 2022-09-12
Bump declared module and internal usage to `/v3/`

meatballhat:
## What type of PR is this?

- cleanup

This change ensures that tags pushed from the `v3-dev-main` branch will declare a different package name than the `main` (v2) series.

Connected to #1507


go.mod
```diff
@@ -1,6 +1,6 @@
-module github.com/urfave/cli/v2
+module github.com/urfave/cli/v3
 
-go 1.18
+go 1.19
```
dearchap, line 3: We are targeting min 1.19 ?

dearchap (approved):

dearchap (dismissed):

go.mod
meatballhat: I don't remember why I did this 😐 I agree this value should be `1.18`

hay-kot (approved):

PR #1498, opened 2022-09-20
Cleanup redundant code in App/Command

### LINKED ISSUE #554: Refactor App and Command to share logic
Currently, there is a lot of duplication between the `App` and `Command` code paths. We should take the opportunity to clean this up for the <ins>next</ins> `v2` release.

My thoughts are to embed `Command` in `App` and remove as many shared fields as is reasonable.

dearchap:
## What type of PR is this?

_(REQUIRED)_

- cleanup

_(REQUIRED)_

Current code path requires almost similar code is three separate places. App.RunContext, App.RunAsSubcommand, and Command.Run. Any bug in flag handling/help etc would need changes in all these places. To make it more manageable the proposal is to condense it into a single place in Command.Run. To enable this we simulate the main App instance as a "rootCommand". 

This is a draft proposal. It still requires some work to have all unit tests pass but I wanted some buy in from the community

_(REQUIRED)_

Fixes #554 

## Testing

## Release Notes

_(REQUIRED)_

```release-note

```

### DIFF
--- a/app.go
+++ b/app.go
@@ -305,6 +305,12 @@ func (a *App) RunContext(ctx context.Context, arguments []string) (err error) {
 	return a.rootCommand.Run(cCtx, arguments...)
 }
 
+// This is a stub function to keep public API unchanged from old code
+// No one should really use this. Always use a.Run to execute app
+func (a *App) RunAsSubcommand(ctx *Context) (err error) {
+	return a.RunContext(ctx.Context, ctx.Args().Slice())
+}
+
 func (a *App) suggestFlagFromError(err error, command string) (string, error) {
 	flag, parseErr := flagFromError(err)
 	if parseErr != nil {
--- a/app.go
+++ b/app.go
@@ -306,7 +306,8 @@ func (a *App) RunContext(ctx context.Context, arguments []string) (err error) {
 }
 
 // This is a stub function to keep public API unchanged from old code
-// No one should really use this. Always use a.Run to execute app
+// 
+// Deprecated: use App.Run or App.RunContext
 func (a *App) RunAsSubcommand(ctx *Context) (err error) {
 	return a.RunContext(ctx.Context, ctx.Args().Slice())
 }
--- a/app.go
+++ b/app.go
@@ -241,6 +241,31 @@ func (a *App) Setup() {
 	}
 }
 
+func (a *App) newRootCommand() *Command {
+	return &Command{
+		Name:                   a.Name,
+		Usage:                  a.Usage,
+		UsageText:              a.UsageText,
+		Description:            a.Description,
+		ArgsUsage:              a.ArgsUsage,
+		BashComplete:           a.BashComplete,
+		Before:                 a.Before,
+		After:                  a.After,
+		Action:                 a.Action,
+		OnUsageError:           a.OnUsageError,
+		Subcommands:            a.Commands,
+		Flags:                  a.Flags,
+		flagCategories:         a.flagCategories,
+		HideHelp:               a.HideHelp,
+		HideHelpCommand:        a.HideHelpCommand,
+		UseShortOptionHandling: a.UseShortOptionHandling,
+		HelpName:               a.HelpName,
+		CustomHelpTemplate:     a.CustomAppHelpTemplate,
+		categories:             a.categories,
+		isRoot:                 true,
+	}
+}
+
 func (a *App) newFlagSet() (*flag.FlagSet, error) {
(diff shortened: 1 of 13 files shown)


command.go
```diff
@@ -104,46 +147,72 @@ func (c *Command) Run(ctx *Context) (err error) {
...
+	for _, scmd := range c.Subcommands {
+		if scmd.HelpName == "" {
+			scmd.HelpName = fmt.Sprintf("%s %s", c.HelpName, scmd.Name)
+		}
+		newCmds = append(newCmds, scmd)
+	}
+	c.Subcommands = newCmds
+}
+
+func (c *Command) Run(cCtx *Context, arguments []string) (err error) {
```
meatballhat, line 166: This method signature is an API change. Could it instead use variadic for `arguments` so that existing code can still compile?

```go
func (c *Command) Run(cCtx *Context, arguments ...string) (err error) {
```

meatballhat (changes requested):

meatballhat:
fwiw the `test-docs` check failure seems to be revealing something that's unhappy _somewhere_

https://github.com/urfave/cli/actions/runs/3215687734/jobs/5256958204#step:7:244

```
# ---- >8 ----
time="2022-10-09T21:52:33Z" level=info msg=done error_count=1 example_count=29 fields.time=13.723986848s source_count=15
time="2022-10-09T21:52:33Z" level=error msg="expected output does not match actual: \"made it!\\nPhew!\" != \"\""
```

command.go
dearchap: Makes sense. Updated.

app_test.go
```diff
@@ -1916,7 +1901,6 @@ func TestApp_Run_CommandHelpName(t *testing.T) {
 	}
 	cmd := &Command{
 		Name:        "foo",
-		HelpName:    "custom",
```
meatballhat, line 1919: Were there assertions around this `HelpName` value that were breaking, or did you need to remove it for some other reason? (or something else?)

command.go
```diff
@@ -89,10 +96,46 @@ func (c *Command) FullName() string {
...
+		HideHelpCommand:        a.HideHelpCommand,
+		OnUsageError:           a.OnUsageError,
+		CustomHelpTemplate:     a.CustomAppHelpTemplate,
+		Usage:                  a.Usage,
+		UsageText:              a.UsageText,
+		Description:            a.Description,
+		ArgsUsage:              a.ArgsUsage,
+		BashComplete:           a.BashComplete,
+		categories:             a.categories,
+		isRoot:                 true,
```
meatballhat, line 130: This is very much a nit, but can these be sorted?

help_test.go
```diff
@@ -1096,6 +1097,7 @@ func TestHideHelpCommand_RunAsSubcommand_False(t *testing.T) {
 		t.Errorf("Run returned unexpected error: %v", err)
 	}
 }
+*/
```
meatballhat, line 1100: Are you planning to add (back) a compatible `RunAsSubcommand` method so that the API is retained?

command.go
dearchap: Sorted in same order as definition

app_test.go
dearchap: Yes. There seems to be a bug in the existing code(not this PR). Should we expect the help to have ? 

```
command custom bar - does bar things
```

help_test.go
dearchap: Added. Thanks for pointing this out.

app_test.go
dearchap: Updated the test to reflect actual usage

app.go
```diff
@@ -268,136 +296,19 @@ func (a *App) RunContext(ctx context.Context, arguments []string) (err error) {
...
-	err = a.Action(cCtx)
+	return a.rootCommand.Run(cCtx, arguments...)
+}
 
-	a.handleExitCoder(cCtx, err)
-	return err
+// This is a stub function to keep public API unchanged from old code
+// No one should really use this. Always use a.Run to execute app
+func (a *App) RunAsSubcommand(ctx *Context) (err error) {
+	return a.RunContext(ctx.Context, ctx.Args().Slice())
```
meatballhat, line 311:
```suggestion
// This is a stub function to keep public API unchanged from old code
// 
// Deprecated: use App.Run or App.RunContext
func (a *App) RunAsSubcommand(ctx *Context) (err error) {
	return a.RunContext(ctx.Context, ctx.Args().Slice())
```

meatballhat (approved):

PR #1508, opened 2022-09-30
Call FlagStringer in String() method of slice flags

fjl:
## What type of PR is this?

- bug/cleanup

The default help template relies on the `String()` method of Flag to render the flag. For most flag types, `String()` indirects through `FlagStringer`, so that is the best place to customize flag rendering.

`FlagStringer` was not called for slice flags because their help output differs from other flags in two ways: there can be multiple default values, and the flag name is shown two times to indicate that the flag can be specified multiple times.

To make multiple values work in the `FlagStringer`, I simply changed `GetValue()` to return all values. This should be OK since the purpose of GetValue() is rendering documentation.

Showing the flag more than once is achieved through a new interface, `DocGenerationSliceFlag`, which the `FlagStringer` uses to decide whether the flag is a slice flag type. I'm not entirely happy with adding a new interface, but
it seems like the only fully backwards-compatible solution.

### DIFF
--- a/flag.go
+++ b/flag.go
@@ -129,6 +129,14 @@ type DocGenerationFlag interface {
 	GetEnvVars() []string
 }
 
+// DocGenerationSliceFlag extends DocGenerationFlag for slice-based flags.
+type DocGenerationSliceFlag interface {
+	DocGenerationFlag
+
+	// IsSliceFlag returns true for flags that can be given multiple times.
+	IsSliceFlag() bool
+}
+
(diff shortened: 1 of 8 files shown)


fjl:
If the new interface type is not wanted, I'm also happy to port this patch to the v3 branch and extend the `DocGenerationFlag` interface there.

dearchap (approved):
@fjl Thank you so much for this.

PR #1526, opened 2022-10-09
Build and run `urfave-cli-genflags` via its `Makefile`

meatballhat:
## What type of PR is this?

- bug
- cleanup

Without this, running `make generate` can result in errors about missing `github.com/urfave/cli/v2` because the top-level `go.mod` context is in use rather than the `cmd/urfave-cli-genflags/go.mod`.

For example, this failure in the `v3-porting` branch: https://github.com/urfave/cli/actions/runs/3214832631/jobs/5255497140

NA 😬

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -22,4 +22,4 @@
 //	}
 package cli
 
-//go:generate go run cmd/urfave-cli-genflags/main.go
+//go:generate make -C cmd/urfave-cli-genflags run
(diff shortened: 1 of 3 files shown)


abitrolly:
Can this help avoid `make` dependency on `go build` https://github.com/golang/go/issues/46687 ?

meatballhat:
@abitrolly I don't think so (?) If anything, I'm adding _more_ dependence on `make` here 😅 , although it's only for when one wants to run `go generate`.

abitrolly:
I see. If that doesn't affect `cd cmd; go build` then that's probably ok.

Dokiys (approved):

Issue #1288, opened 2021-07-12
Opt-out of old-school windows ENV flags


codefromthecrypt:
## Checklist

* [x] Are you running the latest v2 release? The list of releases is [here](https://github.com/urfave/cli/releases).
* [x] Did you check the manual for your release? The v2 manual is [here](https://github.com/urfave/cli/blob/master/docs/v2/manual.md)
* [x] Did you perform a search about this feature? Here's the [Github guide](https://help.github.com/en/github/managing-your-work-on-github/using-search-to-filter-issues-and-pull-requests) about searching.

## What problem does this solve?
Env variables like `%FOO%` are old-school, like cmd.exe. Those using powershell will use normal env conventions like `$FOO`. Right now, the prefix/suffic logic is hard-coded based on `runtime.GOOS` and can't be easily overridden.

## Solution description

Assuming some want the `%FOO%`, a global flag to undo it without having to override the entire stringer function.

## Describe alternatives you've considered

I've tried string replacement to undo this with a wrapped stringer function. Another option is to let it slide, but I prefer the help/markdown to be in powershell syntax.

PR #1528, opened 2022-10-10
Fix:(issue_1206) Default value shouldnt depend on env variable or val…

### LINKED ISSUE #1206: Flag default value is overwritten by environment in help output.
## my urfave/cli version is

v2.3.0

## Describe the bug

Flag default value should never change in help output.

## To reproduce

Build the following program against v2.3.0.

```
package main

import (
	"github.com/urfave/cli/v2"
	"os"
)

func main() {
	app := cli.NewApp()
	app.Name = "myapp"
	app.Flags = []cli.Flag{
		&cli.StringFlag{
			Name:        "myflag",
			Usage:       "my flag is my flag",
			EnvVars:     []string{"MYFLAG"},
			Value:       "defval",
		},
  }
  app.Run(os.Args)
}
```

## Observed behavior

What did you do?

```
MYFLAG=foo ./app --help
```

What did you see happen immediately after the reproduction steps above?

```
NAME:
   myapp - A new cli application

USAGE:
   app [global options] command [command options] [arguments...]

COMMANDS:

GLOBAL OPTIONS:
   --myflag value  my flag is my flag (default: "foo") [$MYFLAG]
   --help, -h      show help (default: false)
```

What did you expect to see?

```
[...]
GLOBAL OPTIONS:
   --myflag value  my flag is my flag (default: "defval") [$MYFLAG]
   --help, -h      show help (default: false)
```

## Expected behavior

Default value should never change (in help output).

dearchap:
…ue set from cmdline

## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

_(REQUIRED)_

Fixes #1206 

## Testing

make test

## Release Notes

_(REQUIRED)_

```release-note

```

### DIFF
--- a/flag_test.go
+++ b/flag_test.go
@@ -3111,7 +3111,7 @@ func TestFlagDefaultValueWithEnv(t *testing.T) {
 			name:    "stringSlice",
 			flag:    &StringSliceFlag{Name: "flag", Value: NewStringSlice("default1", "default2"), EnvVars: []string{"ssflag"}},
 			toParse: []string{"--flag", "parsed"},
-			expect:  `--flag value [ --flag value ]	(default: "default1", "default2") [$ssflag]`,
+			expect:  `--flag value [ --flag value ]	(default: "default1", "default2")` + withEnvHint([]string{"ssflag"}, ""),
 			environ: map[string]string{
 				"ssflag": "some-other-env_value",
 			},
@@ -3120,7 +3120,7 @@ func TestFlagDefaultValueWithEnv(t *testing.T) {
 			name:    "float64Slice",
 			flag:    &Float64SliceFlag{Name: "flag", Value: NewFloat64Slice(1.1, 2.2), EnvVars: []string{"fsflag"}},
 			toParse: []string{"--flag", "13.3"},
-			expect:  `--flag value [ --flag value ]	(default: 1.1, 2.2) [$fsflag]`,
+			expect:  `--flag value [ --flag value ]	(default: 1.1, 2.2)` + withEnvHint([]string{"fsflag"}, ""),
 			environ: map[string]string{
 				"fsflag": "20304.222",
 			},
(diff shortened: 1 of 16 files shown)


flag_test.go
```diff
@@ -859,16 +847,21 @@ func TestInt64FlagWithEnvVarHelpOutput(t *testing.T) {
...
+		expectedSuffix := suffixForEnv("APP_BAR")
 		if !strings.HasSuffix(output, expectedSuffix) {
 			t.Errorf("%s does not end with"+expectedSuffix, output)
 		}
 	}
 }
 
+func suffixForEnv(s string) string {
+	expectedSuffix := fmt.Sprintf(" [$%s]", s)
+	if runtime.GOOS == "windows" {
```
julian7, line 859: can we do this with [constraints](https://pkg.go.dev/cmd/go#hdr-Build_constraints)? Like, having a file like `suffix_windows.go`:

```go
//go:build windows
package cli

import "fmt"

func suffixForEnv(s string) string {
        return fmt.Sprintf(" [%%%s%%]", s)
}
```

and another one called like `suffix_nonwindows.go`:

```go
//go:build !windows
package cli

import "fmt"

func suffixForEnv(s string) string {
        return fmt.Sprintf(" [$%s]", s)
}
```

julian7 (changes requested):

flag_test.go
dearchap: Interesting. Yes we could do that. However I found a function in the code which does the suffix stuff already

dearchap:
@julian7 What do you think of current implementation ?

dearchap:
Merging as all requirements are satisfied

flag_test.go
meatballhat: Worth considering how this relates to https://github.com/urfave/cli/issues/1288 ?

PR #1540, opened 2022-10-21
Feature:(issue_269) Allow external package flag definitions

### LINKED ISSUE #269: Combine flags from other package
Hi

Is there any solutaion for combining flags from other package. For example, glog package requires flag.Parse for using his flags but when i put the flag.Parse(), his flag overrided whole my cli flags. What should i do if i want to use both glog flags and my cli flags ?

dearchap:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

_(REQUIRED)_

Fixes #269 

## Testing
go test -run=TestApp_FlagsFromExtPackage

## Release Notes

_(REQUIRED)_

```release-note

```

### DIFF
--- a/app.go
+++ b/app.go
@@ -200,7 +200,7 @@ func (a *App) Setup() {
 	flag.VisitAll(func(f *flag.Flag) {
 		// skip test flags
 		if !strings.HasPrefix(f.Name, "test.") {
-			a.Flags = append(a.Flags, extFlag{f})
+			a.Flags = append(a.Flags, &extFlag{f})
 		}
 	})
 
--- a/app.go
+++ b/app.go
@@ -21,6 +21,7 @@ var (
 	errInvalidActionType    = NewExitError("ERROR invalid Action type. "+
 		fmt.Sprintf("Must be `func(*Context`)` or `func(*Context) error).  %s", contactSysadmin)+
 		fmt.Sprintf("See %s", appActionDeprecationURL), 2)
+	ignoreFlagPrefix = "test." // this is to ignore test flags when adding flags from other packages
 
 	SuggestFlag               SuggestFlagFunc    = suggestFlag
 	SuggestCommand            SuggestCommandFunc = suggestCommand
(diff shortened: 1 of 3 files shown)


app.go
```diff
@@ -195,6 +196,14 @@ func (a *App) Setup() {
 		a.ErrWriter = os.Stderr
 	}
 
+	// add global flags added by other packages
+	flag.VisitAll(func(f *flag.Flag) {
+		// skip test flags
+		if !strings.HasPrefix(f.Name, "test.") {
+			a.Flags = append(a.Flags, extFlag{f})
```
meatballhat, line 203: I'm good with specifically ignoring flags with `"test."` prefix in this implementation 👍🏼 I looked briefly and couldn't find any exported code that could be used to get this prefix from go libraries 🤷🏼 

What's the reasoning behind using a `extFlag{f}` instead of `&extFlag{f}` and having the `extFlag` methods use a pointer receiver?

dearchap: We need the test prefix check because otherwise when we run unit tests those flags get added !!!!

meatballhat: @dearchap I'm in favor of filtering out `"test." in general 👍🏼  My thinking was more around if we could read this value from somewhere instead of having it hard-coded here 🤷🏼 

As for `extFlag`, I meant that I wondered if you meant to use the struct and method receivers as structs rather than pointers to structs. In this implementation, there's going to be a lot more implicit copying of structs, IIUC.

dearchap: @meatballhat do you want it to be defined as an App field or as a library var ?

Dokiys: > @meatballhat do you want it to be defined as an App field or as a library var ?

@dearchap Maybe he means use pointer receivers(*extFlag) rather than value receivers(extFlag) to avoid unnecessary data duplication.

dearchap: @meatballhat I updated the extFlag to be pointer receiver.

PR #1544, opened 2022-10-23
Documentation:(issue_786) Add docs for flag validation

### LINKED ISSUE #786: Adding Flag Validators

dearchap:
## What type of PR is this?

_(REQUIRED)_

- documentation

_(REQUIRED)_

_(REQUIRED)_

Fixes #786 

## Testing

## Release Notes

_(REQUIRED)_

```release-note

```

### DIFF
--- a/docs/v2/examples/flags.md
+++ b/docs/v2/examples/flags.md
@@ -612,6 +612,7 @@ package main
 import (
 	"log"
 	"os"
+	"fmt"
 
 	"github.com/urfave/cli/v2"
 )
@@ -628,6 +629,7 @@ func main() {
 					if v >= 65536 {
 						return fmt.Errorf("Flag port value %v out of range[0-65535]", v)
 					}
+					return nil
 				},
 			},
 		},
--- a/docs/v2/examples/flags.md
+++ b/docs/v2/examples/flags.md
@@ -603,7 +603,7 @@ Handlers can be registered per flag which are triggered after a flag has been pr
 This can be used for a variety of purposes, one of which is flag validation
 
 <!-- {
-  "args": ["&#45;&#45;port 70000"],
+  "args": ["&#45;&#45;port","70000"],
   "output": "Flag port value 70000 out of range[0-65535]"
 } -->
 ```go
--- a/docs/v2/examples/flags.md
+++ b/docs/v2/examples/flags.md
@@ -596,3 +596,51 @@ The precedence for flag value sources is as follows (highest to lowest):
 0. Environment variable (if specified)
 0. Configuration file (if specified)
 0. Default defined on the flag
+
+#### Flag Actions
+
+Handlers can be registered per flag which are triggered after a flag has been processed. 
+This can be used for a variety of purposes, one of which is flag validation
+
+<!-- {
+  "args": ["&#45;&#45;port 70000"],
+  "output": "Flag port value 70000 out of range[0-65535]"
+} -->
+```go
+package main
+
+import (
+	"log"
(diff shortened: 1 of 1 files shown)


docs/v2/examples/flags.md
```diff
@@ -596,3 +596,53 @@ The precedence for flag value sources is as follows (highest to lowest):
...
 0. Default defined on the flag
+
+#### Flag Actions
+
+Handlers can be registered per flag which are triggered after a flag has been processed. 
+This can be used for a variety of purposes, one of which is flag validation
+
+<!-- {
+  "args": ["&#45;&#45;port","70000"],
+  "output": "Flag port value 70000 out of range[0-65535]"
```
meatballhat, line 607:
```suggestion
  "error": "Flag port value 70000 out of range[0-65535]"
```

PR #1568, opened 2022-11-06
Feature: Add support for persistent flags

### LINKED ISSUE #427: Allow global flags after command

### LINKED ISSUE #1113: Relax ordering requirements for flags/commands to improve upon cli ergonomics
## Checklist

* [X ] Are you running the latest v2 release? The list of releases is [here](https://github.com/urfave/cli/releases).
* [X ] Did you check the manual for your release? The v2 manual is [here](https://github.com/urfave/cli/blob/master/docs/v2/manual.md)
* [ X] Did you perform a search about this feature? Here's the [Github guide](https://help.github.com/en/github/managing-your-work-on-github/using-search-to-filter-issues-and-pull-requests) about searching.

CLI ergonomics when editing/adding flags to command are very bad.

Quick example, consider how many times you have to move cursor back in commandline to go from

    $ cmd query users

to

    $ cmd --out=csv query --filter "project=kittens" users --with-permission=Deploy

vs

    $ cmd query users --with-permission=Deploy  --filter  "project=kittens", --out=csv 

It also "reads" worse:
 "output CSV on query project kittens on users with permissions to deploy:
 vs
 " query users that can deploy in project kittens, output csv"

Without forced ordering you can add a flag at the end at any point so ad-hoc usage of commands is much more streamlined, it also allows writing CLI scripts with arguments grouped by what makes more sense, not by how program author structured it.

Now the long winded example, consider the following:
* App has global flags applicable to all subcommands like
  * change output type (for example switch between machine-parseable and user-frien
  * server address
  * config/credentials pass
  * environment
   * etc
* Application has subcommand with their own flags user might want to iterate
   
Let's say app has a subcommand to query systems state

    cmd query

Which has `--filter` string option, and this command have subcommands for various parts of system

    cmd query nodes
    cmd query services
    cmd query users
    
with various sub-subcommand specific switches.
Let's say user wants to only see stuff about a certain project. No worries, we have `--filter` flag for it

    $ cmd query --filter project=kittens

User now wants to see who has access to that project. But we have options just for that

    $ cmd query --filter project=kittens users 

So far so good. Let's see which of them can deploy the application

    $ cmd query --filter project=kittens users --with-permission=Deploy

Okay, now let's compare it with another project...

    $ cmd query --filter project=dogs users --with-permission=Deploy

Now we need to go back 4 words (4x `C-b`) and we can start changing it. But we have few more projects to check? How about just move filter at the end ?

    $ cmd query  users --with-permission=Deploy --filter project=dogs
    flag provided but not defined: -filter

Okay, we can't. Bit annoying but it is only few commands. 

    $ for project in kittens dogs otters ducks ; do cmd query --filter "project=$project" users --with-permission=Deploy ; done

Right, it does what we wanted. Now onto writing that report. There is a `csv` output format so it should be trivial.
 
    $ for project in kittens dogs otters ducks ; do cmd query --filter "project=$project" users --with-permission=Deploy --out=csv; done
    flag provided but not defined: -out

Oh, right, *go C-b NINE times* because that needs to be after a command but before subcommand.

    $ for project in kittens dogs otters ducks ; do cmd --out=csv query --filter "project=$project" users --with-permission=Deploy ; done

Want to bump it to verbose or enable debug ? Go back nearly to the start of the line. Want to set environment or target server ? Same. Only real option to have convenient ability to change the parameters used in whole app is duplicating them as command-local variables.

The migration manual mentions something that "it is more POSIX-like" but POSIX conventions never had a notion of subcommands in the first place, the "stuff after options" was not a subcommand but also command's parameters (usually file to operate on), and even now most of the say `coreutils` tools allow for having options in any place just for convenience (GNU's `ls -la /tmp` and `ls /tmt -la` have same effect for example)

## Solution description

Every `--option` in commandline should be considered for a flag, from closest "leaf" subcommand to the root

so

    $ cmd --env=dev query --filter type=sqldb
    $ cmd query --filter type=sqldb --env=dev

should be equivalent. Sometimes even having options before subcommand makes sense, like

    $ cmd --filter="type=sqldb" query users
    $ cmd --filter="type=sqldb" query nodes
    $ cmd --filter="type=sqldb" query services

Copying every flag to local "works" but it isn't exactly elegant, and messes up with generated help.

dearchap:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

Long asked for feature to have flags in flexible order

_(REQUIRED)_

Fixes #427 
Fixes #1113 

## Testing

go test -run=TestPersistentFlag

## Release Notes

_(REQUIRED)_

Flags can be marked as persistent in which case they can appear in subcommands. Flags with the same name in subcommand take precedence over commands defined higher in the hierarchy. Since a flag can have multiple names/aliases it is upto the user to ensure that none of the aliases of persistent flags are defined in subcommands. Even if one of the names is matched in a subcommand none of the aliases are applied to subcommand flag parsing

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -1461,6 +1461,10 @@ func TestApp_AfterFunc(t *testing.T) {
 		reset
 	*/
 	counts = &opCounts{}
+	// reset the flags since they are set previously
+	app.Flags = []Flag{
+		&StringFlag{Name: "opt"},
+	}
 
 	// run with none args
 	err = app.Run([]string{"command"})
@@ -3052,4 +3056,132 @@ func TestFlagAction(t *testing.T) {
 			}
 		})
 	}
+}
+
+func TestPersistentFlag(t *testing.T) {
+
+	var topInt, topPersistentInt, subCommandInt, appOverrideInt int
+	var appFlag string
+	var appOverrideCmdInt int64
+	var appSliceFloat64 []float64
+	var persistentAppSliceInt []int64
+
+	a := &App{
+		Flags: []Flag{
+			&StringFlag{
+				Name:        "persistentAppFlag",
+				Persistent:  true,
+				Destination: &appFlag,
+			},
+			&Int64SliceFlag{
+				Name:        "persistentAppSliceFlag",
+				Persistent:  true,
+				Destination: &persistentAppSliceInt,
+			},
+			&Float64SliceFlag{
+				Name:       "persistentAppFloatSliceFlag",
+				Persistent: true,
+				Value:      []float64{11.3, 12.5},
+			},
+			&IntFlag{
+				Name:        "persistentAppOverrideFlag",
+				Persistent:  true,
+				Destination: &appOverrideInt,
+			},
+		},
+		Commands: []*Command{
+			{
+				Name: "cmd",
+				Flags: []Flag{
+					&IntFlag{
+						Name:        "cmdFlag",
+						Destination: &topInt,
+					},
+					&IntFlag{
+						Name:        "cmdPersistentFlag",
+						Persistent:  true,
+						Destination: &topPersistentInt,
+					},
+					&Int64Flag{
+						Name:        "paof",
+						Aliases:     []string{"persistentAppOverrideFlag"},
+						Destination: &appOverrideCmdInt,
+					},
+				},
+				Subcommands: []*Command{
+					{
+						Name: "subcmd",
+						Flags: []Flag{
+							&IntFlag{
+								Name:        "cmdFlag",
+								Destination: &subCommandInt,
+							},
+						},
+						Action: func(ctx *Context) error {
+							appSliceFloat64 = ctx.Float64Slice("persistentAppFloatSliceFlag")
+							return nil
+						},
+					},
+				},
+			},
+		},
+	}
+
+	err := a.Run([]string{"app",
+		"--persistentAppFlag", "hello",
+		"--persistentAppSliceFlag", "100",
+		"--persistentAppOverrideFlag", "102",
+		"cmd",
+		"--cmdFlag", "12",
+		"--persistentAppSliceFlag", "102",
+		"--persistentAppFloatSliceFlag", "102.455",
+		"--paof", "105",
+		"subcmd",
+		"--cmdPersistentFlag", "20",
+		"--cmdFlag", "11",
+		"--persistentAppFlag", "bar",
+		"--persistentAppSliceFlag", "130",
+		"--persistentAppFloatSliceFlag", "3.1445",
+	})
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if appFlag != "bar" {
+		t.Errorf("Expected 'bar' got %s", appFlag)
+	}
+
+	if topInt != 12 {
+		t.Errorf("Expected 12 got %d", topInt)
+	}
+
+	if topPersistentInt != 20 {
+		t.Errorf("Expected 20 got %d", topPersistentInt)
+	}
+
+	// this should be changed from app since
+	// cmd overrides it
+	if appOverrideInt != 102 {
+		t.Errorf("Expected 102 got %d", appOverrideInt)
+	}
+
+	if subCommandInt != 11 {
+		t.Errorf("Expected 11 got %d", subCommandInt)
+	}
+
+	if appOverrideCmdInt != 105 {
+		t.Errorf("Expected 105 got %d", appOverrideCmdInt)
+	}
+
+	expectedInt := []int64{100, 102, 130}
+	if !reflect.DeepEqual(persistentAppSliceInt, expectedInt) {
+		t.Errorf("Expected %v got %d", expectedInt, persistentAppSliceInt)
+	}
+
+	expectedFloat := []float64{102.455, 3.1445}
+	if !reflect.DeepEqual(appSliceFloat64, expectedFloat) {
+		t.Errorf("Expected %f got %f", expectedFloat, appSliceFloat64)
+	}
+
 }
--- a/command.go
+++ b/command.go
@@ -146,7 +146,7 @@ func (c *Command) Run(cCtx *Context, arguments ...string) (err error) {
 	}
 
 	a := args(arguments)
-	set, err := c.parseFlags(&a, cCtx.shellComplete)
+	set, err := c.parseFlags(&a, cCtx)
 	cCtx.flagSet = set
 
 	if c.isRoot {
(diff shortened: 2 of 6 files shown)


abitrolly:
And what happens if local flags conflict with global?

dearchap:
Local wins over global

abitrolly:
With no warning during compilation?

dearchap:
This cant be detected during compile time. At runtime the local flags are applied to flagset first and then we walk higher up the chain . Only those flags marked as persistent up the chain will be applied to current flagset only if none of their names/aliases conflict with local . It is up to the user to make sure that the conflicts dont happen.

app_test.go
```diff
@@ -3077,3 +3077,64 @@ func TestFlagAction(t *testing.T) {
...
+					},
+					&IntFlag{
+						Name:        "flag2",
+						Persistent:  true,
+						Destination: &vflag2,
+					},
+				},
+				Subcommands: []*Command{
+					{
+						Name: "bar",
```
meatballhat, line 3102: I would greatly appreciate more descriptive flag and command names here, both for human comprehension and greppability 😁 Something along the lines of `var topInt, topPersistentInt, subCommandInt int` and commands named `"top-test-persistent-flag"` and `"sub-test-persistent-flag"`?

cmd/urfave-cli-genflags/main.go
```diff
@@ -313,6 +313,10 @@ func (ft *FlagType) GenerateVisibleFlagInterface() bool {
 	return ft.skipInterfaceNamed("VisibleFlag")
 }
 
+func (ft *FlagType) GeneratePersistentFlagInterface() bool {
+	return ft.skipInterfaceNamed("PersistentFlag")
```
meatballhat, line 317: I'm realizing now that this function is confusingly named 😅 TODO for later! renaming so that it reads `return ft.needsGeneratedInterfaceNamed(...)`

command.go
```diff
@@ -317,7 +317,29 @@ func (c *Command) parseFlags(args Args, shellComplete bool) (*flag.FlagSet, erro
...
+					})
+					if applyPersistentFlag {
+						if err := fl.Apply(set); err != nil {
+							return nil, err
+						}
+					}
+				}
+			}
+		}
+	}
```
meatballhat, line 340: Inverting some conditions and breaking early to reduce nesting

```suggestion
	for pCtx := ctx.parentContext; pCtx != nil; pCtx = pCtx.parentContext {
		if pCtx.Command == nil {
			continue
		}

		for _, fl := range pCtx.Command.Flags {
			pfl, ok := fl.(PersistentFlag)
			if !ok || !pfl.IsPersistent() {
				continue
			}

			applyPersistentFlag := true
			set.VisitAll(func(f *flag.Flag) {
				for _, name := range fl.Names() {
					if name == f.Name {
						applyPersistentFlag = false
					}
				}
			})

			if !applyPersistentFlag {
				continue
			}

			if err := fl.Apply(set); err != nil {
				return nil, err
			}
		}
	}
```

command.go
```diff
@@ -317,7 +317,29 @@ func (c *Command) parseFlags(args Args, shellComplete bool) (*flag.FlagSet, erro
...
+
+	err = parseIter(set, c, args.Tail(), ctx.shellComplete)
```
meatballhat, line 342: I realize this is pre-existing, but I'd prefer this to be combined with the following line as an `if err := ...; err != nil {`

meatballhat (changes requested):
This is very exciting! Thank you for doing this work 🎉

cmd/urfave-cli-genflags/main.go
dearchap: Yes it is !!!!!

command.go
dearchap: Yes good idea.

command.go
dearchap: Updated

app_test.go
dearchap: Yes nice idea. Updated

abitrolly:
@dearchap just to clarify. If `-v` is defined globally as `--verbose` and in subsommand as `--version`, will this
```
go -v subcommand -v
```
be equal to this
```
go --verbose subcommand --version
```
?

dearchap:
@abitrolly Yes correct.

flag.go
```diff
@@ -162,6 +162,12 @@ type Countable interface {
 	Count() int
 }
 
+// PersistentFlag is an interface to enable detection of flags which are persistent
+// through subcommands
+type PersistentFlag interface {
```
joeycumines, line 167: This interface needs to be handled by the `SliceFlag` type.

```go
func (x *SliceFlag[T, S, E]) IsPersistent() bool {
    v, ok := x.Target.(PersistentFlag)
    return ok && v.IsPersistent()
}
```

^ or something

We shouldn't change the `SliceFlagTarget` interface for v2 (breaking change). For v3, ya might want to embed `PersistentFlag` into `SliceFlagTarget` for v3, along with removing the type assertion. CC @meatballhat

command.go
```diff
@@ -317,13 +317,41 @@ func (c *Command) parseFlags(args Args, shellComplete bool) (*flag.FlagSet, erro
...
+				continue
+			}
+
+			if err := fl.Apply(set); err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	if err = parseIter(set, c, args.Tail(), ctx.shellComplete); err != nil {
```
meatballhat, line 350:
```suggestion
	if err := parseIter(set, c, args.Tail(), ctx.shellComplete); err != nil {
```

command.go
```diff
@@ -317,13 +317,41 @@ func (c *Command) parseFlags(args Args, shellComplete bool) (*flag.FlagSet, erro
...
 		return nil, err
 	}
 
-	err = normalizeFlags(c.Flags, set)
-	if err != nil {
+	if err = normalizeFlags(c.Flags, set); err != nil {
```
meatballhat, line 354:
```suggestion
	if err := normalizeFlags(c.Flags, set); err != nil {
```

flag.go
meatballhat: Thank you @joeycumines!

dearchap: Updated. Thanks @joeycumines

meatballhat (approved):
🎉🍕🎷🐻

PR #1584, opened 2022-11-11
Argh core

### LINKED ISSUE #1583: Use a core flag parser other than stdlib `flag` to unlock wider range of arg types
tl;dr - By switching away from stdlib `flag`, we will be able to support a wider range of capabilities with significantly less workaround code.

Much of the code around here is arguably _working around_ stdlib `flag`, for example:

- concept of commands
    - nesting of commands
    - commands with aliases
- persistent flags (in progress)
- flags with aliases
- flags of types not already supported in stdlib (via `FlagSet.Var`)
    - slice-specific (de)serialization hacks
    - other inconsistently-implemented (de)serialization hacks
- probably other stuff!

By building on top of stdlib `flag`, we are limited to two categories of string arguments:

- flags
    - must start with `-`
    - may accept zero or one argument
- non-flag string literals
    - must not start with `-`
    - no additional help from stdlib `flag` 🤷🏼 

There are [many other command line argument parsers available in the Go third-party package ecosystem](https://github.com/avelino/awesome-go#standard-cli) such as:

- [pflag](https://github.com/ogier/pflag), a lower-level library that provides a compatibility layer with stdlib `flag`
    - typically seen via [this fork](https://github.com/spf13/pflag) which isn't being actively maintained
- [cobra](https://github.com/spf13/cobra), a holistic command and flag building solution that encourages imperative construction of command contexts (depends on pflag)
- [kong](https://github.com/alecthomas/kong), a holistic command and flag building solution with strong support for deriving applications from nested struct variables
- [argparse](https://github.com/akamensky/argparse), an imperative solution which aims to provide functionality like the Python stdlib `argparse`

meatballhat:
## What type of PR is this?

- foundational
- scary
- experimental

Swap out stdlib `flag` with the "argh" parser

Closes #1583 

This is highly experimental ⚠️ I don't expect reviews at this time, which is why the PR is in draft 💖


meatballhat:
While working through this mess, I'm realizing that by moving away from stdlib `flag` we are not only changing out the flag parsing but also the source of truth for existence, values, everything (?) I think this is generally a good thing, though, since the mismatch between `urfave/cli`'s declarative shell and an imperative `flag` core has been a source of bugs and weirdness for a long time. I'm currently attempting to move the source of truth out to the `urfave/cli` declarative shell while using `internal/argh` only for parsing.

meatballhat:
reopening shortly...

PR #1590, opened 2022-11-18
Add StringMap flag

### LINKED ISSUE #1580: Add StringMap flag
## Checklist

There is no string mapping flag so I usually end up using a string slice and then splitting the elements into key/value pairs. pflag has the option to use the StringToString flag: https://github.com/spf13/pflag/blob/master/string_to_string.go.

## Solution description

Add a `StringMapFlag` and `StringMap` type that can be used like `StringSlice/Flag`, but is backed by a `map[string]string` and splits the elements accordingly with validation. Separator could be configurable, but I think `=` is a common default.

Create a custom flag, but that would be even more boilerplate than now.

avorima:
## What type of PR is this?

- feature

Add `StringMap` and `StringMapFlag` types that are backed by a `map[string]string` and parse comma-separated key=value pairs.

This is an analogue of pflag's [StringToString](https://github.com/spf13/pflag/blob/master/string_to_string.go).

Fixes #1580

I mostly repurposed code from StringSlice. The tests were especially confusing to me since it seems a lot of it is duplicated.

## Testing

Created a sample app that uses the flag.

```go
package main

import (
	"fmt"
	"os"

	cli "github.com/urfave/cli/v3"
)

func main() {
	app := cli.NewApp()
	app.Flags = []cli.Flag{
		&cli.StringMapFlag{Name: "kv", EnvVars: []string{"KV"}},
	}
		dict := c.StringMap("kv")
		fmt.Println(dict)
		return nil
	}

	app.Run(os.Args)
}
```

## Release Notes

```release-note
Add StringMapFlag that parses comma-separated key=value pairs
```

### DIFF
--- a/flag_impl.go
+++ b/flag_impl.go
@@ -193,8 +193,8 @@ func (f *FlagBase[T, C, V]) RunAction(ctx *Context) error {
 // values from cmd line. This is true for slice and map type flags
 func (f *FlagBase[T, C, VC]) IsMultiValueFlag() bool {
 	// TBD how to specify
-	return reflect.TypeOf(f.Value).Kind() == reflect.Slice ||
-		reflect.TypeOf(f.Value).Kind() == reflect.Map
+	kind := reflect.TypeOf(f.Value).Kind()
+	return kind == reflect.Slice || kind == reflect.Map
 }
 
 // IsPersistent returns true if flag needs to be persistent across subcommands
--- a/flag_map_impl.go
+++ b/flag_map_impl.go
@@ -55,7 +55,7 @@ func (i *MapBase[T, C, VC]) Set(value string) error {
 		if !ok {
 			return fmt.Errorf("item %q is missing separator %q", item, defaultMapFlagKeyValueSeparator)
 		}
-		if err := i.value.Set(strings.TrimSpace(value)); err != nil {
+		if err := i.value.Set(value); err != nil {
 			return err
 		}
 		tmp, ok := i.value.Get().(T)
--- a/flag_map_impl.go
+++ b/flag_map_impl.go
@@ -1,0 +1,116 @@
+package cli
+
+import (
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"sort"
+	"strings"
+)
+
+// MapBase wraps map[string]T to satisfy flag.Value
+type MapBase[T any, C any, VC ValueCreator[T, C]] struct {
+	dict       *map[string]T
+	hasBeenSet bool
+	value      Value
+}
+
+func (i MapBase[T, C, VC]) Create(val map[string]T, p *map[string]T, c C) Value {
+	*p = map[string]T{}
+	for k, v := range val {
+		(*p)[k] = v
+	}
+	var t T
+	np := new(T)
+	var vc VC
+	return &MapBase[T, C, VC]{
+		dict:  p,
+		value: vc.Create(t, np, c),
+	}
+}
+
+// NewMapBase makes a *MapBase with default values
+func NewMapBase[T any, C any, VC ValueCreator[T, C]](defaults map[string]T) *MapBase[T, C, VC] {
+	return &MapBase[T, C, VC]{
+		dict: &defaults,
+	}
+}
+
+// Set parses the value and appends it to the list of values
+func (i *MapBase[T, C, VC]) Set(value string) error {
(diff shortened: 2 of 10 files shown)


dearchap:
@avorima the changes look good . Can you rebase off of 3.x since we are not adding new flags in 2.x anymore

avorima:
@dearchap What branch is that, v3-dev-main?

dearchap:
main

avorima:
Code generation is somewhat broken by the changes to documentation in 1.19

avorima:
Yeah, so there definitely is a difference when generating code with 1.18 and 1.19. I went with 1.19 now.

flag_test.go
```diff
@@ -3125,3 +3164,112 @@ func TestFlagSplitMultiValues_Disabled(t *testing.T) {
...
+	_ = os.Setenv("MY_GOAT", "vincent van goat=scape goat")
+	var val map[string]string
+	fl := StringMapFlag{Name: "goat", EnvVars: []string{"MY_GOAT"}, Value: val}
+	set := flag.NewFlagSet("test", 0)
+	_ = fl.Apply(set)
+
+	err := set.Parse(nil)
+	expect(t, err, nil)
+	expect(t, val, map[string]string(nil))
+	expect(t, set.Lookup("goat").Value.(*StringMap).Value(), map[string]string{"vincent van goat": "scape goat"})
```
dearchap, line 3229: LookupStringMap ?

flag_map_impl.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// String returns a readable representation of this value (for usage defaults)
+func (i *MapBase[T, C, VC]) String() string {
+	v := i.Value()
+	var t T
+	if reflect.TypeOf(t).Kind() == reflect.String {
+		return fmt.Sprintf("%v", v)
+	}
+	return fmt.Sprintf("%T{%s}", v, i.ToString(v))
```
dearchap, line 78: Can you add a simple test case to test map[string]int to enable codecov of this line ?

flag_map_impl.go
```diff
@@ -0,0 +1,116 @@
...
+		i.hasBeenSet = true
+		return nil
+	}
+
+	for _, item := range flagSplitMultiValues(value) {
+		key, value, ok := strings.Cut(item, defaultMapFlagKeyValueSeparator)
+		if !ok {
+			return fmt.Errorf("item %q is missing separator %q", item, defaultMapFlagKeyValueSeparator)
+		}
+		if err := i.value.Set(strings.TrimSpace(value)); err != nil {
```
dearchap, line 58: would we ever need to trimspace for this kind of flags ?

flag_impl.go
```diff
@@ -192,7 +192,8 @@ func (f *FlagBase[T, C, V]) RunAction(ctx *Context) error {
 // IsSliceFlag returns true if the value type T is of kind slice
 func (f *FlagBase[T, C, VC]) IsSliceFlag() bool {
```
dearchap, line 193: I think we need to rename this to IsMultiValueFlag as we are overloading this function name now . I can make that change in a separate PR.

flag_map_impl.go
avorima: I was thinking about this, but shouldn't that be added together with `IntMap` or similar?

flag_map_impl.go
avorima: It's easy to do something like `./prog --myflag "foo=$BAR"` and BAR containing unwanted whitespace because it's a script output or whatever. On the other hand removing whitespace can also be delegated to the users of the library. How about I add  a `TrimSpace` flag field that controls this behavior?

flag_map_impl.go
dearchap: Yeah you could do that.

flag_test.go
dearchap: oh sorry I was thinking about ctx.StringMap.

flag_map_impl.go
dearchap: Sure that would be good. Thanks

avorima: Ok, I added it via the flag config. I had to add it to all string related flags to make the compiler happy

flag_impl.go
dearchap: @avorima  PR #1615

PR #1596, opened 2022-11-26
Run linter fixes + go mod tidy

marwan-at-work:
I ran the linter to find functions, methods, and variables not being used as well as ran go mod tidy. 

Plus, ioutil as of Go1.17 has been deprecated, so replaced all calls to os/io functions.

### DIFF
--- a/app.go
+++ b/app.go
@@ -432,30 +432,6 @@ func (a *App) handleExitCoder(cCtx *Context, err error) {
 	}
 }
 
-func (a *App) commandNames() []string {
-	var cmdNames []string
-
-	for _, cmd := range a.Commands {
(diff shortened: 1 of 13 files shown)


meatballhat (approved):
Thank you! I would love to see linting tooling fully integrated via CI, too, fwiw. Do you have strong opinions on what that should look like @marwan-at-work? I usually reach for [golangci-lint](https://golangci-lint.run/) but I admit it's been awhile since I looked around.

marwan-at-work:
@meatballhat no problem! golangci sounds good to me, but hopefully we keep the rules to a minimum so new comers don't find it difficult to contribute. 

Now tat I have access to the merge button, I'll go ahead and push it, I think the codecov/patch is a false positive here. 

Thanks!

PR #1597, opened 2022-11-26
Make *cli.Context implement context.Context by renaming Value

marwan-at-work:
This PR runs the `gopls` rename function on the `*cli.Context.Value` method to be renamed to `GetValue`. This makes the *cli.Context implement context.Context by default as the interface is already embedded into the struct, the only reason it didn't implement it before was because of that method name conflicting in the signature. 

See https://github.com/urfave/cli/issues/833#issuecomment-1328067910


context.go
```diff
@@ -20,6 +20,9 @@ type Context struct {
 	parentContext *Context
 }
 
+// ensures that Context always implements context.Context
+var _ context.Context = &Context{}
```
skelouse, line 24: Put this in context_test.go so that it does not effect built binary.

jalavosus (approved):

dearchap:
@marwan-at-work Can you rebase main ?

dolmen:
The description doesn't say *WHY* you want cli.Context to implement context.Context. What value does this bring?
I only see bugs and confusion coming...

bartekpacia:
This PR changes _a lot_ of code, there are tons of merge conflicts, and v2 is effectively in maintenance mode right now.

I'm leaning toward closing this PR, but would like to hear @meatballhat / @dearchap opinion first.

dearchap:
Yes I agree with @bartekpacia . There is no need for this anymore given that v3 has moved to a different architecture. v2 is in maint mode. I am going to close this.

PR #1606, opened 2022-11-30
Feature:(issue_351) Add support for generation of shell completion sc…

### LINKED ISSUE #351: Support for completion in other shells (fish, sh, etc.)
Hello,

I was going through the repo (I am using `cli` for my new project) and I found that it supports generation of bash and zsh completion files automatically. That is pretty amazing. Is there a way to support other shells (e.g. fish, sh)? I am an avid fish shell user and have been using it since a long time.

I would like to propose implementation of `EnableCompletion` flag instead of `EnableBashCompletion` flag to accept a string for supported shells (like bash, zsh, fish, sh, ...) and generate autocomplete file for the mentioned shell.

If it is a desirable feature, I would like to brainstorm the implementation with the author and community members and take the initial steps towards its development.

dearchap:
…ripts

## What type of PR is this?

_(REQUIRED)_

- cleanup
- feature

_(REQUIRED)_

_(REQUIRED)_

Fixes #351 

## Testing

 Add new tests 
```
go test -run=TestCompletion
```
## Release Notes

_(REQUIRED)_

```release-note
EnableBashCompletion has been renamed to EnableShellCompletion. When enabled completions can be generated on the fly and sourced in current shell environment

PROG=cli-app source <(cli-app generate-completion bash)

Supported shells are bash, fish, zsh, powershell
```

### DIFF
--- a/completion.go
+++ b/completion.go
@@ -40,6 +40,8 @@ var completionCommand = &Command{
 		for k := range shellCompletions {
 			shells = append(shells, k)
 		}
+		
+		sort.Strings(shells)
 
 		if ctx.Args().Len() == 0 {
 			return Exit(fmt.Sprintf("no shell provided for completion command. available shells are %+v", shells), 1)
--- a/completion.go
+++ b/completion.go
@@ -1,8 +1,9 @@
 package cli
 
 import (
-	_ "embed"
+	"embed"
 	"fmt"
+	"sort"
 )
 
 const (
@@ -40,7 +41,7 @@ var completionCommand = &Command{
 		for k := range shellCompletions {
 			shells = append(shells, k)
 		}
-		
+
 		sort.Strings(shells)
 
 		if ctx.Args().Len() == 0 {
--- a/completion.go
+++ b/completion.go
@@ -1,0 +1,61 @@
+package cli
+
+import (
+	_ "embed"
+	"fmt"
+)
+
+const (
+	completionCommandName = "generate-completion"
+)
+
+//go:embed autocomplete/bash_autocomplete
+var bash_autocomplete string
+
+//go:embed autocomplete/powershell_autocomplete.ps1
+var powershell_autocomplete string
+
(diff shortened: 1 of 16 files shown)


completion.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+var shellCompletions = map[string]renderCompletion{
+	"bash": getCompletion(bash_autocomplete),
+	"ps":   getCompletion(powershell_autocomplete),
+	"zsh":  getCompletion(zsh_autocomplete),
+	"fish": func(a *App) (string, error) {
+		return a.ToFishCompletion()
+	},
+}
```
meatballhat, line 36: If you want to keep the file names closely related to what's in the source code, instead of having multiple `snake_case` variable names, you could use an `embed.FS` and reference the files by full name (?)

```suggestion
var (
	//go:embed autocomplete
	autoCompleteFS embed.FS

	shellCompletions = map[string]renderCompletion{
		"bash": getCompletion("autocomplete/bash_autocomplete"),
		"ps":   getCompletion("autocomplete/powershell_autocomplete.ps1"),
		"zsh":  getCompletion("autocomplete/zsh_autocomplete"),
		"fish": func(a *App) (string, error) {
			return a.ToFishCompletion()
		},
	}
)

type renderCompletion func(a *App) (string, error)

func getCompletion(s string) renderCompletion {
	return func(a *App) (string, error) {
		b, err := autoCompleteFS.ReadFile(s)
		return string(b), err
	}
}
```

completion.go
```diff
@@ -0,0 +1,61 @@
...
+
+var completionCommand = &Command{
+	Name:   completionCommandName,
+	Hidden: true,
+	Action: func(ctx *Context) error {
+		var shells []string
+		for k := range shellCompletions {
+			shells = append(shells, k)
+		}
```
meatballhat, line 45: Sort the shells slice for deterministic behavior?

```suggestion
		}
		
		sort.Strings(shells)
```

PR #1622, opened 2022-12-08
Typo fix in example

octoshikari:


abitrolly:
Nope. That's intentional. Otherwise how do you know that the name changed from default?

Issue #1586, opened 2022-11-15
Consolidate all `App` code into `Command`


meatballhat:
The `App` struct and `Command` struct have very similar surface area and behaviors. Greatly reducing the duplication between them has already been done in the `v2` series, but I'd like to see if a single `Command` struct with `.Commands` field can be used everywhere and the `App` struct can be removed (or type alias?).

PR #1627, opened 2022-12-18
Remove `Author` type

meatballhat:
## What type of PR is this?

- cleanup

Removes the `Author` type and accepts `string`s for the `Authors` field instead, essentially leaving any formatting considerations to whomever is using the library.

Supports #1586 given that the `Author` type is part of the `app.go` file which is slated for collapse into `command.go` and this type is adding very little value &lt;/opinions&gt;

### DIFF
--- a/app.go
+++ b/app.go
@@ -79,7 +79,7 @@ type App struct {
 	// Execute this function when an invalid flag is accessed from the context
 	InvalidFlagAccessHandler InvalidFlagAccessFunc
 	// List of all authors who contributed
-	Authors []*Author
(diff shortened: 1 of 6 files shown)


dearchap:
I would suggest to keep the current format. No one has complained which means that either they like it or it isnt used much.

meatballhat:
@dearchap I'm good with it staying or going, so if you'd like to see it stay 👍🏼 😁

meatballhat:
Oops I lied! https://github.com/urfave/cli/pull/1630

PR #1630, opened 2022-12-19
Allow `any` instead of custom `*Author` type

meatballhat:
## What type of PR is this?

- cleanup

Drop the custom `*Author` type and allow `any` with the assumption that it either implements `fmt.Stringer` or can otherwise be safely `Sprintf`'d via `%s`.

Supports #1586 given the `*Author` type was defined in `app.go` which is slated for removal.

### DIFF
--- a/app.go
+++ b/app.go
@@ -5,7 +5,6 @@ import (
 	"flag"
 	"fmt"
 	"io"
-	"net/mail"
 	"os"
 	"path/filepath"
 	"sort"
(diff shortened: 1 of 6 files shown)


asahasrabuddhe (approved):
This looks good. Why are the tests failing though?

abitrolly:
@nathanleclaire looks like `net/mail` adds 500kB to the size of the package.

abitrolly:
@meatballhat I think like this issue needs to be addressed upstream somehow. Implementing interface for storing two strings should not result in 500kB increase. Why does this happen?

meatballhat:
> @meatballhat I think like this issue needs to be addressed upstream somehow. Implementing interface for storing two strings should not result in 500kB increase. Why does this happen?

@abitrolly wow, yeah, I hadn't thought of it that way!

asahasrabuddhe:
> 

This change does not make sense considering the increase in binary size. If we find a way like implementing an interface which is still compatible with the standard library and also does not affect the binary size, that's the way to go.

dearchap:
@meatballhat Just revert to old struct it was just 2 fields. Simple and direct or make an interface and add an example in docs of how to use mail.Address with it.

meatballhat:
@dearchap I think now that I'd like to use `Authors []any` given that the values are only ever passed to Sprintf-like functions with `%s` anyway. I'd like to keep reducing the surface area as much as possible, and making such a change would support that effort. WDYT?

dearchap:
@meatballhat That would be fine.

dearchap (approved):

PR #1631, opened 2022-12-23
Feature:(issue_1222) Add support for mutually exclusive groups

### LINKED ISSUE #1222: Make options mutually exclusive
## mutual exclusive options

Is it possible to make an option required if another option is not provided.

**e.g.**

If I provide `--app-id` `--app-secret` then `--token` is not required.
If I provide `--token`, then `--app-secret` and `--app-id` is not required.

dearchap:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

_(REQUIRED)_

Fixes #1222 

## Testing

Add new tests. 

## Release Notes

_(REQUIRED)_

```release-note

```

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -69,6 +69,39 @@ func (e *errRequiredFlags) getMissingFlags() []string {
 	return e.missingFlags
 }
 
+type mutuallyExclusiveGroup struct {
+	flag1Name string
+	flag2Name string
+}
+
+func (e *mutuallyExclusiveGroup) Error() string {
+	return fmt.Sprintf("option %s cannot be set along with option %s", e.flag1Name, e.flag2Name)
+}
+
+type mutuallyExclusiveGroupRequiredFlag struct {
+	flags *FlagExGroup
+}
+
+func (e *mutuallyExclusiveGroupRequiredFlag) Error() string {
+
+	var missingFlags []string
+	for _, grpf := range e.flags.Flags {
+		var grpString []string
+		for _, f := range grpf {
+			grpString = append(grpString, f.Names()...)
+		}
+		if len(e.flags.Flags) == 1 {
+			err := errRequiredFlags{
+				missingFlags: grpString,
+			}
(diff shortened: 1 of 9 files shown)


errors.go
```diff
@@ -69,6 +69,39 @@ func (e *errRequiredFlags) getMissingFlags() []string {
...
+		if len(e.flags.Flags) == 1 {
+			err := errRequiredFlags{
+				missingFlags: grpString,
+			}
+			return err.Error()
+		}
+		missingFlags = append(missingFlags, strings.Join(grpString, " "))
+	}
+
+	return fmt.Sprintf("one of these flags needs to be provided %s", strings.Join(missingFlags, "\n"))
```
meatballhat, line 102: so that the error text doesn't use multiple lines by default

```suggestion
	return fmt.Sprintf("one of these flags needs to be provided: %s", strings.Join(missingFlags, ", "))
```

flag_ex_group.go
```diff
@@ -0,0 +1,43 @@
...
+				}
+			}
+			if oneSet {
+				break
+			}
+		}
+	}
+
+	if !oneSet && grp.Required {
+		return &mutuallyExclusiveGroupRequiredFlag{&grp}
```
meatballhat, line 40: prefer explicit struct field names

```suggestion
		return &mutuallyExclusiveGroupRequiredFlag{flags: &grp}
```

meatballhat (commented):
Thank you so much for working on this! I have an API naming question/suggestion. Instead of `FlagExGroup`, what do you think of one of these?

- `MutuallyExclusiveFlags`
- `MutExFlags`
- `MutexFlags` (following stdlib `sync.Mutex` casing)

dearchap:
@meatballhat Renamed to MutuallyExclusiveFlags. Mutex/MutEx connotes a different meaning.

meatballhat (approved):

PR #1633, opened 2022-12-30
Fix:(issue_1277) Remove default text for version/help flags

### LINKED ISSUE #1277: can I remove default text when a flag's value is set

dearchap:
## What type of PR is this?

_(REQUIRED)_

- bug
- documentation
- feature

_(REQUIRED)_

Remove default text for help/version flags

_(REQUIRED)_

Fixes #1277 

## Testing

make test

## Release Notes

_(REQUIRED)_

```release-note

```

### DIFF
--- a/flag.go
+++ b/flag.go
@@ -340,7 +340,7 @@ func stringifyFlag(f Flag) string {
 	defaultValueString := ""
 
 	// set default text for all flags except bool flags
-	// for bool flags text is set only is DisableDefaultText is not
+	// for bool flags display default text if DisableDefaultText is not
 	// set
 	if bf, ok := f.(*BoolFlag); !ok || !bf.DisableDefaultText {
 		if s := df.GetDefaultText(); s != "" {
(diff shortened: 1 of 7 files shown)


flag.go
```diff
@@ -337,8 +339,13 @@ func stringifyFlag(f Flag) string {
 
 	defaultValueString := ""
 
-	if s := df.GetDefaultText(); s != "" {
-		defaultValueString = fmt.Sprintf(formatDefault("%s"), s)
+	// set default text for all flags except bool flags
+	// for bool flags text is set only is DisableDefaultText is not
```
lorensr, line 343:
```suggestion
	// for bool flags text is set only if DisableDefaultText is not
```

lorensr (approved):
Could potentially be useful if it worked on non-book fields? But not for my use case

flag.go
dearchap: Fixed.

dearchap:
> Could potentially be useful if it worked on non-book fields? But not for my use case

Yes it could but for now I think this is good. With other flags it becomes more involved/non-intuitive so I'm going to punt for now.

lorensr:
Sounds good!

meatballhat (approved):

PR #1636, opened 2023-01-08
Feature:(issue_1071) Add check for duplicated flags

### LINKED ISSUE #1071: Add optional check for duplicated string and int type flags.
## Checklist

* [y] Are you running the latest v2 release? The list of releases is [here](https://github.com/urfave/cli/releases).
* [y] Did you check the manual for your release? The v2 manual is [here](https://github.com/urfave/cli/blob/master/docs/v2/manual.md)
* [y] Did you perform a search about this feature? Here's the [Github guide](https://help.github.com/en/github/managing-your-work-on-github/using-search-to-filter-issues-and-pull-requests) about searching.

Sometimes users can accidently duplicate string type flags and then the latest value we be taken. It is not clear and bad for usability. 

## Solution description

Add an option called Duplicated true\false to flag options. Example:

```
flagFoo = &cli.StringFlag{
	Name: "foo",
	Duplicated: false,
}
```

A detailed description of what you want to happen.

if Duplicated set to false command should fail with error
"flag foo can not be duplicated"

For now, as a workaround, we're using StringSliceType for flags we don't want to be duplicated and check that slice have only one element in it.

https://github.com/Venafi/vcert/blob/8e1dc13c0afdd615d7f27b80190ddf3a1d07e848/cmd/vcert/commands.go#L105-L120

dearchap:
## What type of PR is this?

_(REQUIRED)_

- documentation
- feature

_(REQUIRED)_

_(REQUIRED)_

Fixes #1071 

## Testing

make test

## Release Notes

_(REQUIRED)_

```release-note

```

### DIFF
--- a/flag_impl.go
+++ b/flag_impl.go
@@ -16,9 +16,9 @@ type Value interface {
 // simple wrapper to intercept Value operations
 // to check for duplicates
 type valueWrapper struct {
-	value      Value
-	count      int
-	duplicated bool
+	value    Value
+	count    int
+	onlyOnce bool
 }
 
 func (v *valueWrapper) String() string {
@@ -26,7 +26,7 @@ func (v *valueWrapper) String() string {
 }
 
 func (v *valueWrapper) Set(s string) error {
-	if v.count == 1 && !v.duplicated {
+	if v.count == 1 && v.onlyOnce {
 		return fmt.Errorf("cant duplicate this flag")
 	}
 	v.count++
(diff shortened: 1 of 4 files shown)


flag_impl.go
```diff
@@ -57,6 +81,8 @@ type FlagBase[T any, C any, VC ValueCreator[T, C]] struct {
 
 	Config C // Additional/Custom configuration associated with this flag type
 
+	NonDuplicated bool // whether this flag can be duplicated on the command line
```
meatballhat, line 84: I feel weird about a bool type name that starts with a negation, but I understand we want the default behavior to be to allow multiple instances. What do you think about:

- `StrictOnce`
- `StrictlyOnce`
- `Once`
- `OnlyOnce`

?

abitrolly: * `Repeatable`

?

dearchap: I think OnlyOnce is good since by default it would be false.

meatballhat (approved):
Yay! :tada:

PR #1643, opened 2023-01-12
Create an inverseBoolFlag

### LINKED ISSUE #424: Add support for flags that also behave as switches
If there's a way to do this, it's not apparent to me or in the docs, but there doesn't seem to be an obvious way to distinguish between --someflag "" and no flag specified at all. Using a magic sentinel as the default value is _not_ a good approach for many cases.

/cc @ffrank

skelouse:
## What type of PR is this?

This is a new feature for handling IsSet of a BoolFlag by introducing an inverseFlag alongside it.

- feature

I made it here when trying to implement flags where the following commands are valid with a flag `--local`

`$ tau`  local is not set, thus we prompt the user

`$ tau --local`  local is set as true

`$ tau --local=false`  local is set as false

I decided on an alternative where I take a flag, say `local`

`$ tau --local`  local is set as true

`$ tau --no-local`  local is set as false

Somewhat Fixes #424 

## Testing

I simply tested the basic IsSet, value and an error with:
- TestBoolWithInverse<tests>
- Basic  for testing the basic functionality
- Action for testing the running of the provided BoolFlag's ActionFunc
- EnvVars for testing positive and negative environment variables being set
- WithPrefix for confirming the InversePrefix value is being used as expected
- An example: ExampleNewBoolWithInverse

## Release Notes

```release-note
BoolWithInverseFlag is a new flag struct to handle IsSet of a BoolFlag by introducing an inverseFlag alongside it
```

## TODO

Coverage

Options handling:
- Negative Description
- Negative suffix || ~prefix~
- Negative FilePath
~- Negative Destination && Count~
- Negative Usage

There are likely things I did not take into account which represent how a normal flag is expected to function.


flag_bool_with_inverse.go
```diff
@@ -0,0 +1,165 @@
+package cli
+
+import (
+	"flag"
+	"fmt"
+)
+
+type BoolWithInverseFlag interface {
```
dearchap, line 8: do we really need this interface ?

dearchap:
@skelouse Can you add more tests for to exercise new interface ?

flag_bool_with_inverse.go
skelouse: Not likely,  I'll remove it and add the provided BoolFlag as an exported value in the struct.  Then I can also remove the method to created, and have the flag defined just like any other flag.

skelouse:
> @skelouse Can you add more tests for to exercise new interface ?

Added more tests and removed the interface so that it acts, and can be used as any other flag type.

skelouse:
Need to fix flag.Names() being called before cli is instantiated.

This can be done be checking whether negative flag is still nil.  If so create the inverse names prior to instantiation and cache them.

flag_bool_with_inverse.go
```diff
@@ -0,0 +1,171 @@
...
+		for idx, envVar := range child.EnvVars {
+			parent.negativeFlag.EnvVars[idx] = strings.ToUpper(parent.InversePrefix) + envVar
+		}
+	}
+
+	return
+}
+
+func (s *BoolWithInverseFlag) Apply(set *flag.FlagSet) error {
+	s.initialize()
```
dearchap, line 138: what happens if the underyling bool flag is marked as persistent ? Will this make this flag be applied multiple times ?

skelouse: I'm not sure what you mean by persistent

dearchap: https://github.com/urfave/cli/blob/main/flag_impl.go#L92

dearchap:
@skelouse Also main is the default v3 branch now no need for v3-dev-main

flag_bool_with_inverse.go
skelouse: Discovered a new issue with persistence when attempting to handle this: https://github.com/urfave/cli/issues/1644

dearchap: @skelouse Can you just fix the failing tests for now ? We can focus on #1644 later after this PR is merged.

skelouse:
Seems the action CLI may be broken.

dearchap (approved):

skelouse:
Good to go.  Swapped in place for my original BoolWithInverse interface, and ran e2e tests on current project.

PR #1654, opened 2023-01-22
Update many v1 things

meatballhat:
## What type of PR is this?

- cleanup

including:
- testing on `windows-latest` in CI
- updating license year
- updating badges in README
- removing references to outdated v2 docs
- removing references to `GOPATH`
- removing appveyer config
- introducing stretchr/testify for comparison assertions
- replacing usage of `ioutil`
- updating all dependencies
- removing outdated version-specific test files


build.go
```diff
@@ -163,6 +165,11 @@ func GfmrunActionFunc(c *cli.Context) error {
 }
 
 func TocActionFunc(c *cli.Context) error {
+	if runtime.GOOS == "windows" {
```
dearchap, line 168: do you want to move this to build_windows.go ?

dearchap (approved):

build.go
meatballhat: I was trying to avoid adding a separate file 🤷🏼

Issue #1592, opened 2022-11-21
There are two problems when loading yaml (slice, Uint64 flags)


kjeom:
test.yaml
```yaml
scope: 
  test: 
    - test1
    - test2
test: 
  temp: 
    test: 1
```
testYaml.go
```go
func main() {
	flags := []cli.Flag{
		&cli.StringFlag{
			Name:  "config",
			Usage: "yaml config file name",
		},
		altsrc.NewUint64Flag(
			&cli.Uint64Flag{
				Name:        "test.temp.test",
				Aliases:     []string{},
				Value:       0,
				Destination: &test,
			},
		),
		altsrc.NewStringSliceFlag(
			&cli.StringSliceFlag{
				Name:    "scope.test",
				Aliases: []string{},
				EnvVars: []string{},
			},
		),
	}

	app := &cli.App{
		Flags:  flags,
		Before: altsrc.InitInputSourceWithContext(flags, altsrc.NewYamlSourceFromFlagFunc("config")),
		Action: func(ctx *cli.Context) error {
			if ctx.IsSet("scope.test") {
				fmt.Println("true", ctx.StringSlice("scope.test"))
			} else {
				fmt.Println("false", ctx.StringSlice("scope.test"))
			}
			if ctx.IsSet("test.temp.test") {
				fmt.Println("true", ctx.Uint64("test.temp.test") {
			} else {
				fmt.Println("false", ctx.Uint64("test.temp.test")
			}
			return nil
		},
	}
	fmt.Println(app.Run(os.Args))
}
```
- Expected Result
  - scope.test : true [test1 test2]
  - test.temp.test : true 1
- Result
  - scope.test : **false** [test1 test2]
  - test.temp.test : **false** **0**

- Reason
  - slice : not setting flagSet (https://github.com/urfave/cli/blob/v2-maint/altsrc/flag.go#L114)
  - Uint64: there is no receiver function for Uint64 (https://github.com/urfave/cli/blob/v2-maint/altsrc/flag.go#L66)

dearchap:
@kjeom The ctx.IsSet behaviour is expected since that function returns true if the flag has been set from the cmd line. Think of loading envvars, yaml files etc to set the default value of the flag which can then be modified from cmdline. yes the uint64 isnt implemented. Seems like a simple fix. Would you like to submit a PR ?

lobachpavel:
Hi

[...]

PR #1669, opened 2023-01-27
add Integer type casting when loading yaml file

kjeom:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

- Latest v2 release (2.24.1) resolved this https://github.com/urfave/cli/issues/1592
- But also we cannot load the Uint64 flag because the casting error
- The integer from yaml file is unmarshaled to the int about lower than `9223372036854775807` regardless of the flag type
- This PR can accept even the integer that can be casted to the flag type
- IsSet()
  - The IsSet() returns `true` after loaded from yaml about all the flags except the slice flags
  - This PR can return `true` about slice flag type also

_(REQUIRED)_

- #1592 

## Testing

## Release Notes

_(REQUIRED)_

```release-note

```

### DIFF
--- a/altsrc/map_input_source.go
+++ b/altsrc/map_input_source.go
@@ -128,7 +128,7 @@ func (fsm *MapInputSource) Float64(name string) (float64, error) {
 	return 0, nil
 }
 
-func CastToInt64(v interface{}) (int64, bool) {
+func castToInt64(v interface{}) (int64, bool) {
 	int64Value := int64(0)
 	isType := false
 	if v == nil {
@@ -152,7 +152,7 @@ func CastToInt64(v interface{}) (int64, bool) {
 func (fsm *MapInputSource) Int64(name string) (int64, error) {
 	otherGenericValue, exists := fsm.valueMap[name]
 	if exists {
-		otherValue, isType := CastToInt64(otherGenericValue)
+		otherValue, isType := castToInt64(otherGenericValue)
 		if !isType {
 			return 0, incorrectTypeForFlagError(name, "int64", otherGenericValue)
 		}
@@ -160,7 +160,7 @@ func (fsm *MapInputSource) Int64(name string) (int64, error) {
 	}
 	nestedGenericValue, exists := nestedVal(name, fsm.valueMap)
 	if exists {
-		otherValue, isType := CastToInt64(otherGenericValue)
+		otherValue, isType := castToInt64(otherGenericValue)
 		if !isType {
 			return 0, incorrectTypeForFlagError(name, "int64", nestedGenericValue)
 		}
@@ -170,7 +170,7 @@ func (fsm *MapInputSource) Int64(name string) (int64, error) {
 	return 0, nil
 }
 
-func CastToUint(v interface{}) (uint, bool) {
+func castToUint(v interface{}) (uint, bool) {
 	uintValue := uint(0)
 	isType := false
 	if v == nil {
@@ -209,7 +209,7 @@ func CastToUint(v interface{}) (uint, bool) {
 func (fsm *MapInputSource) Uint(name string) (uint, error) {
 	otherGenericValue, exists := fsm.valueMap[name]
 	if exists {
-		otherValue, isType := CastToUint(otherGenericValue)
+		otherValue, isType := castToUint(otherGenericValue)
 		if !isType {
 			return 0, incorrectTypeForFlagError(name, "uint", otherGenericValue)
 		}
@@ -217,7 +217,7 @@ func (fsm *MapInputSource) Uint(name string) (uint, error) {
 	}
 	nestedGenericValue, exists := nestedVal(name, fsm.valueMap)
 	if exists {
-		otherValue, isType := CastToUint(nestedGenericValue)
+		otherValue, isType := castToUint(nestedGenericValue)
 		if !isType {
 			return 0, incorrectTypeForFlagError(name, "uint", nestedGenericValue)
 		}
@@ -227,7 +227,7 @@ func (fsm *MapInputSource) Uint(name string) (uint, error) {
 	return 0, nil
 }
 
-func CastToUint64(v interface{}) (uint64, bool) {
+func castToUint64(v interface{}) (uint64, bool) {
 	uint64Value := uint64(0)
 	isType := false
 	if v == nil {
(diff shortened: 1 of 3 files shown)


abitrolly:
@urfave/cli I am not watching it closely, but is there an API to type cast and report errors for just any type? The specificity of int64 sound like something that should not come out of the box for generic command line handling library.

dearchap:
@kjeom Can you add some tests in altsrc for this condition ?

altsrc/map_input_source.go
```diff
@@ -127,19 +128,46 @@ func (fsm *MapInputSource) Float64(name string) (float64, error) {
...
+func castToInt64(v interface{}) (int64, bool) {
+	int64Value := int64(0)
+	isType := false
+	if v == nil {
+		return int64Value, true
+	}
+
+	// There are only four cases(int, int64, uint64, float64) when parsing the integer in yaml.v3 pkg
+	// But the last case, float64, is an error case so that ignored
+	vType := reflect.TypeOf(v).Kind()
```
dearchap, line 140: does reflect need to be used ? You cant just do 

```
switch v.(type) {
case uint64:
   ...
}
```
?

altsrc/map_input_source.go
```diff
@@ -127,19 +128,46 @@ func (fsm *MapInputSource) Float64(name string) (float64, error) {
...
+	case reflect.Int:
+		int64Value = int64(v.(int))
+		isType = true
+	case reflect.Int64:
+		int64Value = v.(int64)
+		isType = true
+	case reflect.Uint64:
+		uint64Value := v.(uint64)
+		if uint64Value <= math.MaxInt64 {
+			int64Value = int64(uint64Value)
```
dearchap, line 151: Test to update a test case to cover this

altsrc/map_input_source.go
```diff
@@ -149,19 +177,55 @@ func (fsm *MapInputSource) Int64(name string) (int64, error) {
...
+func castToUint(v interface{}) (uint, bool) {
+	uintValue := uint(0)
+	isType := false
+	if v == nil {
+		return uintValue, true
+	}
+
+	// There are only four cases(int, int64, uint64, float64) when parsing the integer in yaml.v3 pkg
+	// But the last case, float64, is an error case so that ignored
+	vType := reflect.TypeOf(v).Kind()
```
dearchap, line 189: please use 

```
switch v.(type) {


}
```

altsrc/map_input_source.go
```diff
@@ -149,19 +177,55 @@ func (fsm *MapInputSource) Int64(name string) (int64, error) {
...
+	switch vType {
+	case reflect.Int:
+		intValue := v.(int)
+		if intValue >= 0 {
+			uintValue = uint(intValue)
+			isType = true
+		}
+	case reflect.Int64:
+		int64Value := v.(int64)
+		if int64Value >= 0 && uint64(int64Value) <= math.MaxUint {
```
dearchap, line 199: Add/update test case to test this condition

altsrc/map_input_source.go
```diff
@@ -127,19 +128,46 @@ func (fsm *MapInputSource) Float64(name string) (float64, error) {
...
+	// But the last case, float64, is an error case so that ignored
+	vType := reflect.TypeOf(v).Kind()
+	switch vType {
+	case reflect.Int:
+		int64Value = int64(v.(int))
+		isType = true
+	case reflect.Int64:
+		int64Value = v.(int64)
+		isType = true
+	case reflect.Uint64:
```
dearchap, line 148: if user expects an int64 but a uint64 is provided should we still convert ? Discuss

altsrc/map_input_source.go
skelouse: Even simpler can use:

```go
// quick example
switch val := v.(type) {
    case int:
        int64Value = int64(val)
    case int64:
        int64Value = val
}
```

altsrc/map_input_source.go
kjeom: I agree. It is better to occur an error.

altsrc/map_input_source.go
kjeom: It needs the 32bits platform to cover this.

altsrc/map_input_source.go
kjeom: resolved it. fb6fcd2

altsrc/map_input_source.go
```diff
@@ -290,12 +378,10 @@ func (fsm *MapInputSource) Int64Slice(name string) ([]int64, error) {
 
 	var int64Slice = make([]int64, 0, len(otherValue))
 	for i, v := range otherValue {
-		int64Value, isType := v.(int64)
-
+		int64Value, isType := castToInt64(v)
 		if !isType {
-			return nil, incorrectTypeForFlagError(fmt.Sprintf("%s[%d]", name, i), "int", v)
+			return nil, incorrectTypeForFlagError(fmt.Sprintf("%s[%d]", name, i), "int64", v)
```
dearchap, line 383: Please add a test case for this

altsrc/map_input_source.go
```diff
@@ -171,19 +227,51 @@ func (fsm *MapInputSource) Uint(name string) (uint, error) {
...
+		}
+	case int64:
+		int64Value := int64(value)
+		if int64Value >= 0 {
+			uint64Value = uint64(int64Value)
+			isType = true
+		}
+	case uint:
+		uint64Value = uint64(value)
+		isType = true
```
dearchap, line 254: Please add a test case for this

altsrc/map_input_source.go
kjeom: covered it.

altsrc/map_input_source.go
kjeom: covered it.

altsrc/map_input_source.go
kjeom: All the patched code is covered.

altsrc/map_input_source.go
kjeom: Resolved it.

kjeom:
@dearchap please finish your review

dearchap (approved):

altsrc/yaml_file_loader_test.go
```diff
@@ -1,13 +1,17 @@
-package altsrc_test
+package altsrc
```
dearchap, line 1: Lets keep the original package name . we are verifying the uint64 cast types indirectly through tests. No need to verify them separately here.

kjeom: Okay, thanks. resolved it [f8fe5e8](https://github.com/urfave/cli/pull/1669/commits/f8fe5e8769bcf52dcca74897610983eaa56cfa29) and removed the separate uint64 test case.

dearchap (approved):

PR #1673, opened 2023-01-29
Show non categorized flags with categorized on help

### LINKED ISSUE #1672: fix categorized flags alongside non categorized
## My urfave/cli version is

latest v2

## Checklist

- [X] Did you check the manual for your release? The v2 manual is [here](https://github.com/urfave/cli/blob/main/docs/v2/manual.md)

## Describe the bug

Commit: https://github.com/urfave/cli/commit/8227be1fe9df91e3d69e17514df21b8e609d6cd0 
introduced a backward compatibility issue.  As before non categorized flags would show alongside categorized flags.  After this commit non categorized do not show at all.

## To reproduce

TestCommand_VisibleFlagCategories is a direct test of the breaking change, Just because no category is set, the flag should not be invisible.

## Expected behavior

Flags should show in help unless explicitly hidden

## Want to fix this yourself?

Sure, I'll introduce a PR shortly

```
go version go1.19.4 linux/amd64
```

skelouse:
## What type of PR is this?

- bug
- feature

Reverting the removal of visible flags from help if their sibling flags have categories defined.

Example:
```
App:
    Flags:
        Flag1: // non categorized
        Flag2: Category1
        
Help Before PR:
    Options:
      Category1
        Flag2
        
Help After:
    Options:
        Flag1
      
      Category1
        Flag2
```

fixes #1672 

## Release Notes

```release-note
Reverted removal of non categorized flags alongside categorized flags
```

### DIFF
--- a/category.go
+++ b/category.go
@@ -100,10 +100,23 @@ func newFlagCategories() FlagCategories {
 
 func newFlagCategoriesFromFlags(fs []Flag) FlagCategories {
 	fc := newFlagCategories()
+
+	var categorized bool
 	for _, fl := range fs {
 		if cf, ok := fl.(CategorizableFlag); ok {
-			if cf.GetCategory() != "" {
-				fc.AddFlag(cf.GetCategory(), cf)
+			if cat := cf.GetCategory(); cat != "" {
+				fc.AddFlag(cat, cf)
+				categorized = true
+			}
+		}
+	}
+
+	if categorized == true {
+		for _, fl := range fs {
+			if cf, ok := fl.(CategorizableFlag); ok {
+				if cf.GetCategory() == "" {
+					fc.AddFlag("", fl)
+				}
 			}
 		}
 	}
--- a/app.go
+++ b/app.go
@@ -250,14 +250,7 @@ func (a *App) Setup() {
 	}
 	sort.Sort(a.categories.(*commandCategories))
 
-	a.flagCategories = newFlagCategories()
-	for _, fl := range a.Flags {
-		if cf, ok := fl.(CategorizableFlag); ok {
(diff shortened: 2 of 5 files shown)


category.go
```diff
@@ -100,10 +100,23 @@ func newFlagCategories() FlagCategories {
 
 func newFlagCategoriesFromFlags(fs []Flag) FlagCategories {
 	fc := newFlagCategories()
+
+	var categorized bool
 	for _, fl := range fs {
 		if cf, ok := fl.(CategorizableFlag); ok {
-			if cf.GetCategory() != "" {
-				fc.AddFlag(cf.GetCategory(), cf)
+			if cat := cf.GetCategory(); cat != "" {
```
dearchap, line 107: isnt this equivalent to ```fc.AddFlag(cf.GetCategory(), cf)``` with out the ```cf.getCategory() != "" ``` condition ?

dearchap:
I think the main concern is whether empty string should be counted as a category. I am ok with either but it would be good if some stakeholders can chime in on this cc @urfave/cli

category.go
skelouse: Yes, but then we're always working with categorized flags

dearchap: Ah ok got it.

dearchap (approved):

skelouse:
v2.24.2?  @dearchap

tarampampam:
@dearchap Apologies for the necroposting, but it seems that `v3.0.0-alpha7` still faces the same issue. At least, flags without categories are not printed when at least one flag has it. Could you please confirm this, and if so, could you create an issue to address this?

Or will it be better if I will do that?

/cc @jetexe

PR #1674, opened 2023-01-29
Fix some issues in bash autocompletion

MrNaif2018:
## What type of PR is this?
- bug

With current version of autocomplete script, shell quotes gets expanded incorrectly

Consider the following example:
`my-cli --url http://localhost:8000 he` (pressing tab should auto-complete to help, but it doesn't)
Why doesn't it auto-complete?
Because it calls the following command:
`my-cli --url http : //localhost : 8000 --generate-bash-completion`
And if we put it in quotes my-cli --url "http://localhost:8000" he
It calls:
`my-cli --url '"http://localhost:8000"' --generate-bash-completion`

Both fail, so CLI fails to call correct URL and return a list of completions.
With this PR it sends correct data
Actually I am not sure if that's the best solution, I checked how cobra does this and converted to our use case
It looks like the `_init_completion` function does all the magic and makes it work, I don't understand why unfortunately

P.S. I hope it'll be backported to 2.x series too (:

## Testing

I tested multiple cases, i.e. sending one positional argument, 2 of them, or argument + flag, the CLI receives same arguments as before, but now more cases work

## Release Notes

```release-note
Fixed shell expansion in bash autocompletion scripts
```

### DIFF
--- a/autocomplete/bash_autocomplete
+++ b/autocomplete/bash_autocomplete
@@ -2,17 +2,31 @@
 
 : ${PROG:=$(basename ${BASH_SOURCE})}
 
+# Macs have bash3 for which the bash-completion package doesn't include
+# _init_completion. This is a minimal version of that function.
+_cli_init_completion() {
+  COMPREPLY=()
+  _get_comp_words_by_ref "$@" cur prev words cword
+}
+
 _cli_bash_autocomplete() {
   if [[ "${COMP_WORDS[0]}" != "source" ]]; then
-    local cur opts base
+    local cur opts base words
     COMPREPLY=()
     cur="${COMP_WORDS[COMP_CWORD]}"
-    if [[ "$cur" == "-"* ]]; then
-      opts=$( ${COMP_WORDS[@]:0:$COMP_CWORD} ${cur} --generate-shell-completion )
+    if declare -F _init_completion >/dev/null 2>&1; then
+      _init_completion -n "=:" || return
     else
-      opts=$( ${COMP_WORDS[@]:0:$COMP_CWORD} --generate-shell-completion )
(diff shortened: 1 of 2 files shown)


dearchap:
@MrNaif2018 Can you add a test case(s) for this ? I will attempt to backport to v2.x after that. Thanks

MrNaif2018:
Sure! Could you guide me where to check? Because bash behavior I tested was via tab, so not sure how to test it non-interactively

dearchap:
@MrNaif2018 You could use this test as a starting point. You can either add additional cases for this test or add a new test 

https://github.com/urfave/cli/blob/f9831417f33ad87e175dd44abbd91ce13840b2ee/help_test.go#L1177

MrNaif2018:
@dearchap I've added a simple test, but the issue is, it won't catch the bug anyway :D

No go-based code can catch the bug
The library is parsing and doing everything correctly, just that the bundled bash autocomplete script did something wrong, which caused arguments to be transformed wrong, hence by default bash parsing rules, go cli couldn't parse arguments correctly (because they were passed so)

dearchap:
@MrNaif2018 Just to confirm you want to merge this in v2 or v3 ?

MrNaif2018:
I am currently using v2 version, so I would like if it would land in v2 too. v3 probably needs this too if it uses same setup

dearchap (approved):

dearchap:
@MrNaif2018 Can you re-submit a PR for v2 ? I will merge this into main for now

MrNaif2018:
Is it v2-maint?

dearchap:
Yes

MrNaif2018:
#1676

PR #1677, opened 2023-02-01
Deprecation of package ioutil in Go 1.16

ArangoGutierrez:
ioutil.ReadFile(...) became -> os.ReadFile(...)

https://github.com/go-critic/go-critic/issues/1019


avorima:
I see a lot more uses of `ioutil`. Check the output of `grep -R ioutil` for example.

PR #1683, opened 2023-02-16
Adjust sub-command template to properly output sub-sub-command placeholder.

### LINKED ISSUE #1682: Sub-command help template shows `command [command options]` inappropriately
The sub-command help template checks [`Command.VisibleFlags()`](https://pkg.go.dev/github.com/urfave/cli/v3#Command.VisibleFlags) instead of [`Command.VisibleCommands()`](https://pkg.go.dev/github.com/urfave/cli/v3#Command.VisibleCommands) when deciding to render the trailing `command [command options]` part of the `USAGE` section.

https://github.com/urfave/cli/blob/87b48e2ddd3bd22a4df7942b572c76cd41c294c1/template.go#L84-L87

This looks to be unintentional and causes the help text to imply that there are sub-commands even when none are present.

aryounce:
## What type of PR is this?

- bug
- cleanup

- `template.go` includes an improper conditional in the sub-command help text template that relies on flag visibility instead command visibility for the `Command` instance.

Fixes #1682

## Release Notes

```release-note
Fix sub-command help output to exclude sub-sub-command placeholder when command has no sub-commands.
```

### DIFF
--- a/godoc-current.txt
+++ b/godoc-current.txt
@@ -146,7 +146,7 @@ var SubcommandHelpTemplate = `NAME:
    {{template "helpNameTemplate" .}}
 
 USAGE:
-   {{if .UsageText}}{{wrap .UsageText 3}}{{else}}{{.HelpName}} {{if .VisibleFlags}}command [command options]{{end}} {{if .ArgsUsage}}{{.ArgsUsage}}{{else}}[arguments...]{{end}}{{end}}{{if .Description}}
+   {{if .UsageText}}{{wrap .UsageText 3}}{{else}}{{.HelpName}} {{if .VisibleCommands}}command [command options] {{end}}{{if .ArgsUsage}}{{.ArgsUsage}}{{else}}[arguments...]{{end}}{{end}}{{if .Description}}
 
 DESCRIPTION:
    {{template "descriptionTemplate" .}}{{end}}{{if .VisibleCommands}}
--- a/template.go
+++ b/template.go
@@ -83,7 +83,7 @@ var SubcommandHelpTemplate = `NAME:
    {{template "helpNameTemplate" .}}
 
 USAGE:
-   {{if .UsageText}}{{wrap .UsageText 3}}{{else}}{{.HelpName}} {{if .VisibleFlags}}command [command options]{{end}} {{if .ArgsUsage}}{{.ArgsUsage}}{{else}}[arguments...]{{end}}{{end}}{{if .Description}}
+   {{if .UsageText}}{{wrap .UsageText 3}}{{else}}{{.HelpName}} {{if .VisibleCommands}}command [command options] {{end}}{{if .ArgsUsage}}{{.ArgsUsage}}{{else}}[arguments...]{{end}}{{end}}{{if .Description}}
 
 DESCRIPTION:
    {{template "descriptionTemplate" .}}{{end}}{{if .VisibleCommands}}
--- a/testdata/godoc-v3.x.txt
+++ b/testdata/godoc-v3.x.txt
@@ -146,7 +146,7 @@ var SubcommandHelpTemplate = `NAME:
    {{template "helpNameTemplate" .}}
 
 USAGE:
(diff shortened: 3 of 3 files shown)


avorima:
I played around with this a bit and did not get it to produce different output than current main. Could you provide an example?

dearchap (dismissed):

dearchap (commented):
This may not be sufficient since Help Command is added by default. So it really wont change any behaviour

aryounce:
@dearchap This is for when the help command is disabled. Then there are zero sub-commands (which is the use case where I encountered the problem).

dearchap:
That makes sense. Can you see why the checks are failing ? You might have to run make v2approve or make v3approve

aryounce:
@dearchap Running `make diffcheck` (which is failing in the tests) locally succeeds. Running `make v2approve` and `make v3approve` don't appear to have any affect. I don't have the ability to re-run the Github actions and thought that perhaps the encountered error is transient. Advice on how to proceed would be greatly appreciated.

avorima:
@aryounce I checked the output with the following program and there was no change from main to your PR. Maybe my example is incorrect?

```go
package main

import "github.com/urfave/cli/v3"

func main() {
	app := &cli.App{
		Name: "test",
		Commands: []*cli.Command{
			{
				Name: "sub",
				Commands: []*cli.Command{
					{Name: "sub-sub", Hidden: true},
				},
				HideHelp: true,
			},
		},
	}
	_ = app.Run([]string{"test", "sub"})
}
```

avorima:
You need to run `make generate`

aryounce:
@avorima I've modified your example to demonstrate the issue I'm attempting to address. Be sure to run this outside of my branch (or off of `main`) to prevent my changes having an effect. My version of your code is:

```golang
package main

import (
	"fmt"
	"os"
	"github.com/urfave/cli/v3"
)

func main() {
	app := &cli.App{
		Name: "test",
		Commands: []*cli.Command{
			{
				Name: "default",
				Action: dumpCmdInfo,
			},
			{
				Name: "hide-help",
				Action: dumpCmdInfo,
				HideHelp: true,
			},
			{
				Name: "hide-help-cmd",
				Action: dumpCmdInfo,
				HideHelpCommand: true,
			},
			{
				Name: "hidden",
				Action: dumpCmdInfo,
				Hidden: true,
			},
		},
	}
	app.Run(os.Args)
}

func dumpCmdInfo(c *cli.Context) error {
	fmt.Println(c.Command.Name)
	fmt.Printf("  VisibleCommands() = %+v\n", c.Command.VisibleCommands())
	fmt.Printf("  VisibleFlags() = %+v\n", c.Command.VisibleFlags())
	return nil
}
```

## default

When running the `default` sub-command (which has no sub-commands of its own) the output is as follows. Notice that the help text only shows `[command options]` under `USAGE`. This appears to use the `CommandHelpTemplate`: https://github.com/urfave/cli/blob/main/template.go#L59-L67

### Help Output
```
NAME:
   test default

USAGE:
   test default [command options] [arguments...]

OPTIONS:
```

### Command Info Dump
```
default
  VisibleCommands() = [0x105159b40]
  VisibleFlags() = [--help, -h	show help (default: false)]
```

## hidden

This uses `CommandHelpTemplate` to form its `USAGE` section, like `default`.

### Help Output
```
NAME:
   test hidden

USAGE:
   test hidden [command options] [arguments...]

OPTIONS:
```
### Command Info Dump
```
hidden
  VisibleCommands() = [0x104471b40]
  VisibleFlags() = [--help, -h	show help (default: false)]
```

## hide-help

Has no help output because the `HideHelp` property on `Command` is set to `false`.

### Command Info Dump
```
hide-help
  VisibleCommands() = []
  VisibleFlags() = []
```

## hide-help-cmd

**This is the use case my PR is attempting to address.** Although it has no sub-commands the help text output contains `command [command options]` which implies that it does have one or more sub-commands. This is because it uses `SubcommandHelpTemplate`, https://github.com/urfave/cli/blob/main/template.go#L79-L87, which relies upon the whether or not `VisibleFlags()` returns a zero-value.

### Help Output
```
NAME:
   test hide-help-cmd

USAGE:
   test hide-help-cmd command [command options] [arguments...]

OPTIONS:
```

### Command Info Dump
```
hide-help-cmd
  VisibleCommands() = []
  VisibleFlags() = [--help, -h	show help (default: false)]
```

avorima:
Thank you, I can see the difference now.

avorima (approved):

PR #1691, opened 2023-03-05
Shift tested Go versions

meatballhat:
## What type of PR is this?

- maintenance

## What this PR does / why we need it:

Bump the tested go versions since the first patch of 1.20.x is available now.

## Which issue(s) this PR fixes:

N/A 😬

PR #1692, opened 2023-03-05
Shift tested Go versions in v2-maint

meatballhat:
as with #1691


meatballhat:
string substitution ordering error 😂

skelouse (approved):

PR #1694, opened 2023-03-05
Fix linting issues

### LINKED ISSUE #1663: Add integration with golangci-lint
The [golangci-lint](https://golangci-lint.run/) meta-linting tool is widely used to ensure various classes of bugs and maintenance hindrances stay out of the code. Introducing this tool will almost certainly result in a lot of warnings and errors, so I'm happy to consider a _gradual adoption_ via config file tweaking.

skelouse:
## What type of PR is this?

- cleanup

Fixes 100% of `golangci-lint` linting issues

Partially fixes #1663 

## Testing

https://asciinema.org/a/ucDYTAfRZzASFpkmpfdrTxQHJ

ran make

## Release Notes
```release-note
Update to pass `golangci-lint`
```

### DIFF
--- a/completion.go
+++ b/completion.go
@@ -54,7 +54,10 @@ var completionCommand = &Command{
 		} else if c, err := rc(ctx.App); err != nil {
 			return Exit(err, 1)
 		} else {
-			ctx.App.Writer.Write([]byte(c))
+			_, err = ctx.App.Writer.Write([]byte(c))
+			if err != nil {
+				return Exit(err, 1)
+			}
 		}
 		return nil
 	},
--- a/completion.go
+++ b/completion.go
@@ -54,8 +54,7 @@ var completionCommand = &Command{
 		} else if c, err := rc(ctx.App); err != nil {
 			return Exit(err, 1)
 		} else {
-			_, err = ctx.App.Writer.Write([]byte(c))
-			if err != nil {
+			if _, err = ctx.App.Writer.Write([]byte(c)); err != nil {
 				return Exit(err, 1)
 			}
 		}
--- a/context_test.go
+++ b/context_test.go
@@ -282,7 +282,7 @@ func TestContext_Set_InvalidFlagAccessHandler(t *testing.T) {
 	}
 
 	c := NewContext(app, set, nil)
-	expect(t, c.Set("missing", "") == nil, true)
+	expect(t, c.Set("missing", "") != nil, true)
 	expect(t, flagName, "missing")
 }
 
--- a/context_test.go
+++ b/context_test.go
@@ -280,9 +280,9 @@ func TestContext_Set_InvalidFlagAccessHandler(t *testing.T) {
 			flagName = name
 		},
 	}
+
 	c := NewContext(app, set, nil)
-	err := c.Set("missing", "")
-	if err == nil {
+	if err := c.Set("missing", ""); err == nil {
 		t.Error("Expected set error")
 		return
 	}
--- a/context_test.go
+++ b/context_test.go
@@ -282,10 +282,7 @@ func TestContext_Set_InvalidFlagAccessHandler(t *testing.T) {
 	}
 
 	c := NewContext(app, set, nil)
-	if err := c.Set("missing", ""); err == nil {
-		t.Error("Expected set error")
-		return
-	}
+	expect(t, c.Set("missing", "") == nil, true)
 	expect(t, flagName, "missing")
 }
 
--- a/context_test.go
+++ b/context_test.go
@@ -281,7 +281,11 @@ func TestContext_Set_InvalidFlagAccessHandler(t *testing.T) {
 		},
 	}
 	c := NewContext(app, set, nil)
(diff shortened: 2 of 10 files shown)


skelouse:
Can we override the patch diff, or should I add test cases for the error checks?

flag_test.go
```diff
@@ -426,7 +426,11 @@ func TestStringFlagHelpOutput(t *testing.T) {
 		fl := &StringFlag{Name: test.name, Aliases: test.aliases, Usage: test.usage, Value: test.value}
 		// create a tmp flagset
 		tfs := flag.NewFlagSet("test", 0)
-		fl.Apply(tfs)
+		err := fl.Apply(tfs)
+		if err != nil {
+			t.Error(err)
+			return
+		}
```
meatballhat, line 433: I'm happy to start using `stretchr/testify` more around here, e.g.:

```suggestion
		require.Nil(t, fl.Apply(tfs))
```

help.go
```diff
@@ -352,27 +363,32 @@ func printHelpCustom(out io.Writer, templ string, data interface{}, customFuncs
...
+	handleTemplateError(err)
+	_, err = t.New("visibleGlobalFlagCategoryTemplate").Parse(strings.Replace(visibleFlagCategoryTemplate, "OPTIONS", "GLOBAL OPTIONS", -1))
+	handleTemplateError(err)
+	_, err = t.New("authorsTemplate").Parse(authorsTemplate)
+	handleTemplateError(err)
+	_, err = t.New("visibleCommandCategoryTemplate").Parse(visibleCommandCategoryTemplate)
+	handleTemplateError(err)
+
+	err = t.Execute(w, data)
+	handleTemplateError(err)
```
meatballhat, line 390: imho this is a great place to use the "iferr" form so that the `err` var isn't getting overwritten and adding to the potential for unexpected debug output, and the last case can be folded up without assigning `err` at all 👍🏼

meatballhat (dismissed):
Overall I'm very excited about this change! Thank you @skelouse 🎉 

- the notes here don't need to be addressed in this PR, but I'd like to get them done soonish
- ignoring the patch-level coverage check is fine in this case, imho
- can either this PR or a near future follow-up PR add `golangci-lint` integration in CI to assert that new linting errors are not introduced?

completion.go
```diff
@@ -54,7 +54,10 @@ var completionCommand = &Command{
 		} else if c, err := rc(ctx.App); err != nil {
 			return Exit(err, 1)
 		} else {
-			ctx.App.Writer.Write([]byte(c))
+			_, err = ctx.App.Writer.Write([]byte(c))
+			if err != nil {
+				return Exit(err, 1)
+			}
```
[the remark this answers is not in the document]
meatballhat: 🤦🏼 this is targeting the v3 alpha, so never mind what I said! Let's exit non-zero on failure to write completion 👍🏼

context_test.go
```diff
@@ -281,7 +281,11 @@ func TestContext_Set_InvalidFlagAccessHandler(t *testing.T) {
 		},
 	}
 	c := NewContext(app, set, nil)
-	c.Set("missing", "")
+	err := c.Set("missing", "")
+	if err == nil {
```
dearchap, line 285: can we add a expectNotNil helper function for this kind of stuff ?

help.go
```diff
@@ -352,27 +363,32 @@ func printHelpCustom(out io.Writer, templ string, data interface{}, customFuncs
...
-	err := t.Execute(w, data)
-	if err != nil {
-		// If the writer is closed, t.Execute will fail, and there's nothing
-		// we can do to recover.
-		if os.Getenv("CLI_TEMPLATE_ERROR_DEBUG") != "" {
-			_, _ = fmt.Fprintf(ErrWriter, "CLI TEMPLATE ERROR: %#v\n", err)
-		}
-		return
-	}
+	_, err := t.New("helpNameTemplate").Parse(helpNameTemplate)
```
dearchap, line 366: better to do 

`
if _, err := t.New(...); err != nil {
 handleTemplateError(err)
}
`

completion.go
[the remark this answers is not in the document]
skelouse: Can easily do _, _ for v2 and support linting just the same, good note.

skelouse:
@meatballhat  
> * can either this PR or a near future follow-up PR add `golangci-lint` integration in CI to assert that new linting errors are not introduced?

I'd love to make this change, but i'm not sure where to start with CI integrations.

flag_test.go
skelouse: Do we want to introduce another import in go.mod?  From my perspective it seems that this package has very few external imports.  Maybe in another PR as there are many use cases for using testify.

skelouse: https://github.com/urfave/cli/issues/1695

completion.go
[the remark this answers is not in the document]
skelouse: https://github.com/urfave/cli/pull/1696

meatballhat (approved):

PR #1697, opened 2023-03-08
v3 Add integration with golangci-lint

### LINKED ISSUE #1663: Add integration with golangci-lint

skelouse:
## What type of PR is this?

- github action

adds an action for `golangci-lint run`

Fixes #1663 

## Testing

https://github.com/skelouse/cli/actions/runs/4361565058

## Release Notes

```release-note
Integrate golangci-lint action
```

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,0 +1,25 @@
+name: golangci-lint
+on:
+  push:
+    tags:
+      - v3.*
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+permissions:
+  contents: read
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-go@v3
+        with:
+          go-version: "1.20"
+      - uses: actions/checkout@v3
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3
+        with:
+          version: latest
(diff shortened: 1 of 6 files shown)


avorima:
Is there a plan to also add a .golanci.yml file? I find the default config that it uses somewhat lacking. I think at least some linters from the `bugs` preset category would be good to have, for example the `makezero` linter would've spotted the following: `errors.go:45:10: append to slice `errs` with non-zero initialized length`.
https://golangci-lint.run/usage/linters/ has a table with linters and their presets.

skelouse:
> Is there a plan to also add a .golanci.yml file? I find the default config that it uses somewhat lacking. I think at least some linters from the `bugs` preset category would be good to have, for example the `makezero` linter would've spotted the following: `errors.go:45:10: append to slice `errs` with non-zero initialized length`. https://golangci-lint.run/usage/linters/ has a table with linters and their presets.

Adding more linters will be a future iteration, as that will entail cleaning up those bugs as well as integrating them.  I will look into adding the golangci.yml.

avorima (dismissed):

skelouse:
@avorima I've added the .golangci.yaml along with the misspell package.  If anything else should be added we can make an issue for it along with the current output of running the additional linter.

avorima (approved):

PR #1706, opened 2023-03-18
Better Generic Support For ISCs

danhunsaker:
## What type of PR is this?

- feature

- Allow generics to be set as their Set() string in the config
- Allow converting internal ISC (InputSourceContext) types to Generics when:
  - The data used to define the Generic can be expressed in JSON
  - The target Generic implements `FromJson([]byte) error`
  - The source ISC implements `Json(name string) ([]byte, error)`
- Update existing ISC implementations to also implement `Json(name string) ([]byte, error)`

And of course, there are tests for the new features.

As to why, the internal ISCs don't actually support Generics, as they don't know about these types to be able to deserialize into them directly. Proper support requires the ability to coerce the intermediate types into a Generic type. This can be done one of two ways - by using a string equivalent as you would on the CLI, or by implementing methods that can handle a JSON intermediate version on the Generic itself. 

Given:

```go
	app.Flags = []cli.Flag{
		altsrc.NewGenericFlag(&cli.GenericFlag{
			Name:   "location",
			Usage:  "where to connect",
			Value:  HostPort{},
		}),
		// ...
	}
```

The following examples are equivalent for the sample `HostPort` type added in this PR to the `altsrc` tests:

```yaml
location: localhost:33
```

```yaml
location:
  host: localhost
  port: 33
```

Third-party code that implements its own InputSourceContext may already support storing Generics in its internal structure, in which case the new code paths won't be hit, rendering full BC.

## Testing

Tests were updated in the code to cover the new scenarios.

## Release Notes

```release-note
Better Generic support when using an InputSourceContext
```

### DIFF
--- a/altsrc/flag.go
+++ b/altsrc/flag.go
@@ -73,7 +73,32 @@ func (f *GenericFlag) ApplyInputSourceValue(cCtx *cli.Context, isc InputSourceCo
 		}
 		value, err := isc.Generic(name)
 		if err != nil {
-			return err
+			stringValue, stringErr := isc.String(name)
+			if stringErr != nil {
+				isce, isType := isc.(InputSourceContextWithExport)
+				if isType {
+					flagValue, isType := f.Value.(GenericWithImport)
+					if !isType {
+						return err
+					}
+					jsonValue, err := isce.Json(name)
+					if err != nil {
+						return err
+					}
+					if len(jsonValue) > 0 {
+						err := flagValue.FromJson(jsonValue)
+						if err != nil {
+							return err
+						}
+						value = flagValue
+					}
+				}
+			} else {
+				value = &stringGeneric{stringValue}
+			}
+			if value == nil {
+				return err
+			}
 		}
 		if value == nil {
 			continue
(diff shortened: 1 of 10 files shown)


dearchap:
@danhunsaker v2 is only in maintenance mode. You do realise that v3/main branch has support for generics already ?

danhunsaker:
v3 isn't released/usable/documented yet. I imagine this will need to be forward-ported to v3 as well at some point, but until it's actually _released_, I can't use it in my own apps. Thus targeting v2 for the moment.

meatballhat:
@danhunsaker Thank you again for your patience! I've got much the same feedback as in #1702. How would you like to handle next steps?

bartekpacia:
Hey, @danhunsaker. Thank you for this PR.

Some merge conflicts showed up, and as Dan explained above, we're very reluctant to accept new features into v2 (much smaller PRs were getting rejected because of this policy).

Now, I'm a new maintainer and new to this codebase, so I'm not yet sure if it makes sense to re-target this PR against https://github.com/urfave/cli-altsrc.

I'm just trying to see what's the state that this PR is in. I don't like if they live for long, because they rot.

Again, thanks for all the work.

abitrolly:
Extending code surface to  +416 new lines for a maintenance project is also a security risk.

dearchap:
Wont merge into v2 as it is in maint mode. v3 has better support for multiple value sources and is amenable to be extended easily.

PR #1707, opened 2023-03-19
v2 - Fix `incorrectTypeForFlagError` for pointers

danhunsaker:
## What type of PR is this?

- bug

When configs parsed into `MapInputSource` are looked up, they report type mismatch errors, as expected. However, if the type is mismatched because it somehow ended up becoming a pointer, the error message doesn't actually indicate this. Instead, it reports an "actual type" of `''` - the empty string. This is a known limitation of the way `reflect` works, and is used elsewhere in the code to dereference the pointer to get the actual type. So I'm not sure why it was overlooked in this case.

## Testing

A test was added to ensure this change works as expected. 

## Release Notes

```release-note
NONE
```

### DIFF
--- a/altsrc/map_input_source.go
+++ b/altsrc/map_input_source.go
@@ -475,6 +475,9 @@ func incorrectTypeForFlagError(name, expectedTypeName string, value interface{})
 	valueTypeName := ""
 	if valueType != nil {
 		valueTypeName = valueType.Name()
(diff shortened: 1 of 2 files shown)


danhunsaker:
I'll do a v3 version of this if needed once v3 supports configuration files.

danhunsaker:
Further exploration reveals this is inaccurate. I need to figure out what type I actually get and provide a fix that way. Closing until then.

PR #1709, opened 2023-03-20
v2 - Fix: missing required flag error uses flag name and not alias

### LINKED ISSUE #1701: Missing required flag error uses alias
## My urfave/cli version is

v2.25.0

## Checklist

- [ ] Did you check the manual for your release? The v2 manual is [here](https://github.com/urfave/cli/blob/main/docs/v2/manual.md) <-- The link is broken, but I read the documentation.

## Dependency Management

## Describe the bug

The error message for a missing required flag uses the alias instead of the main flag name, which is not untrue, but confusing.

## To reproduce

```go
package main

import (
	"fmt"
	"log"

	"github.com/urfave/cli/v2"
)

func main() {
	app := &cli.App{
		Flags: []cli.Flag{
			&cli.StringFlag{
				Name:     "collection",
				Aliases:  []string{"c"},
				Required: true,
			},
		},
			fmt.Println(cCtx.String("collection"))
			return nil
		},
	}

	if err := app.Run([]string{""}); err != nil {
		log.Fatal(err)
	}
}
```

## Observed behavior

```shell
Required flag "c" not set
```

## Expected behavior

```shell
Required flag "collection" not set
```

## Additional context

It seems like [checkRequiredFlags](https://github.com/urfave/cli/blob/v2-maint/context.go#LL200C22-L200C40) is iterating over the the flag names (which is the append(name, aliases...)) and sets `flagName = key`. After the loop is over, `flagName` is the last alias.

## Want to fix this yourself?

I can submit a PR if you want, but according to the [guidelines](https://github.com/urfave/cli/blob/main/docs/CONTRIBUTING.md) you are not accepting v2 fixes. Let me know if this is ok, and I'll move forward with a fix.

A fix could be:

```go
func (cCtx *Context) checkRequiredFlags(flags []Flag) requiredFlagsErr {
	var missingFlags []string
	for _, f := range flags {
		if rf, ok := f.(RequiredFlag); ok && rf.IsRequired() {
			var flagPresent bool
			var flagName string

			flagNames := f.Names()
			flagName = flagNames[0]

			for _, key := range flagNames {
				if cCtx.IsSet(strings.TrimSpace(key)) {
					flagPresent = true
				}
			}

			if !flagPresent && flagName != "" {
				missingFlags = append(missingFlags, flagName)
			}
		}
	}

	if len(missingFlags) != 0 {
		return &errRequiredFlags{missingFlags: missingFlags}
	}

	return nil
}
```

```
go version go1.20.2 darwin/amd64
```

```
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/nhaas/Library/Caches/go-build"
GOENV="/Users/nhaas/Library/Application Support/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/nhaas/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/nhaas/go"
GOPRIVATE=""
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOVCS=""
GOVERSION="go1.20.2"
GCCGO="gccgo"
GOAMD64="v1"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/dev/null"
GOWORK="/Users/something"
CGO_CFLAGS="-O2 -g"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-O2 -g"
CGO_FFLAGS="-O2 -g"
CGO_LDFLAGS="-O2 -g"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -arch x86_64 -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/t8/83vffqv11_bbfxbbdc3fsfx40000gn/T/go-build1340975921=/tmp/go-build -gno-record-gcc-switches -fno-common"
```

nirhaas:
## What type of PR is this?

- bug

Up until now, the missing required flag error showed the last alias of a flag, instead of the actual flag name. This results in a relatively unclear error messages:
```
Required flag "c" not set
```

Instead of 
```
Required flag "collection" not set
```

Fixes #1701 

## Testing

Added test case to ensure this change works as expected.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/context.go
+++ b/context.go
@@ -204,9 +204,10 @@ func (cCtx *Context) checkRequiredFlags(flags []Flag) requiredFlagsErr {
 			var flagPresent bool
 			var flagName string
 
-			for _, key := range f.Names() {
-				flagName = key
+			flagNames := f.Names()
+			flagName = flagNames[0]
 
+			for _, key := range flagNames {
 				if cCtx.IsSet(strings.TrimSpace(key)) {
 					flagPresent = true
 				}
--- a/context_test.go
+++ b/context_test.go
@@ -602,6 +602,14 @@ func TestCheckRequiredFlags(t *testing.T) {
 				&StringFlag{Name: "n", Required: true},
 			},
 		},
+		{
+			testCase:              "required_flag_with_alias_errors_with_actual_name",
+			expectedAnError:       true,
+			expectedErrorContents: []string{"Required flag \"collection\" not set"},
+			flags: []Flag{
+				&StringFlag{Name: "collection", Required: true, Aliases: []string{"c"}},
+			},
+		},
 	}
 
 	for _, test := range tdata {


skelouse (approved):

context.go
```diff
@@ -204,9 +204,10 @@ func (cCtx *Context) checkRequiredFlags(flags []Flag) requiredFlagsErr {
 			var flagPresent bool
 			var flagName string
 
-			for _, key := range f.Names() {
-				flagName = key
+			flagNames := f.Names()
+			flagName = flagNames[0]
```
dearchap, line 208: what happens if someone passes in a flag with no name or aliases defined ? Will the slice be nil ?

skelouse: Good catch 🦾

nirhaas: It returns an empty slice (length 0), not a nil. I can make a test case for it to prove that it doesn't panic if you want, even though the code doesn't handle this kind of flag nicely (showing as a newline in the help).

What would a flag without a name or alias means to the user? Shouldn't the code have a different check and error for this case?

skelouse: if len(flagNames) == 0 ln 208 will panic:

https://go.dev/play/p/aN-gKini_p9

nirhaas: @skelouse it's length 1 with an empty string. Sorry for the confusion. https://go.dev/play/p/tsnBea9RKk4 - No panic

dearchap: Both examples provided are illustrative but its better to add a simple test case for this. Just have a flag without a name and see what happens. If everything works well and good.

nirhaas: Sure. Added another test case

dearchap (approved):

PR #1728, opened 2023-04-30
Bug/fix issue 1703

### LINKED ISSUE #1703: CustomHelpTemplate is ignored with --help flag
## My urfave/cli version is

_**2.25.0**_

## Checklist

## Dependency Management

## Describe the bug

While `cli top-level-command help` respects `CustomHelpTemplate` when printing help, `cli top-level-command --help` ignores it and prints `SubcommandHelpTemplate` instead.

Is this intentional? What is the purpose of `SubcommandHelpTemplate`?

If this is not intentional i can send a PR switching to  `CustomHelpTemplate`

## To reproduce

```
app := cli.NewApp()
	app.Name = "testapp"
		        {
			        Name:   "top-level-command",
			        Action: func(c *cli.Context) error { return nil },
			        CustomHelpTemplate: "TESTING",
			 }
}
```

```
go run ./cmd/app top-level-command --help
```

## Observed behavior

It prints `SubcommandHelpTemplate`

## Expected behavior

`go run ./cmd/app top-level-command --help` should print `CustomHelpTemplate`

## Additional context

https://github.com/temporalio/cli/pull/153

## Want to fix this yourself?

**Yes**

go version

jojje:
## What type of PR is this?

- bug

_(REQUIRED)_

fixes #1703

Let me know if you want the solution somewhere else in the code-base.

I noticed that you try to contain corner case handling to the `helpCommand.Action` callback. However I'm not familiar enough with the code-base to fully understand what the right set of conditions would be by comparing various disparate but seemingly interconnected objects and fields/attributes.

Instead I went for the most intuitive solution of simply ensuring that whenever the help command is about to be rendered, instead of assuming that the caller has populated the cCtx correctly, I used _defensive programming_ to ensure honoring the intent of using a custom template for given command, at the closest location to the template use, if the command happens to have one defined.

## Testing

Two unit tests added. One for the happy path to ensure the behavior didn't change in that regard. The other for the error path which triggers the bug and which this fix addresses.

tests were performed by running `make test`

## Release Notes

_(REQUIRED)_

```release-note
* Custom help templates are now used when -h[elp] flag is given, even when no sub-commands are defined. 
```

### DIFF
--- a/help.go
+++ b/help.go
@@ -84,19 +84,6 @@ var helpCommand = &Command{
 	},
 }
 
-func anyFlag(candidates []string, flags []Flag) bool {
-	for _, flag := range flags {
-		for _, alias := range flag.Names() {
-			for _, c := range candidates {
-				if c == alias {
-					return true
-				}
-			}
-		}
-	}
-	return false
-}
-
 // Prints help for the App or Command
 type helpPrinter func(w io.Writer, templ string, data interface{})
 
--- a/help.go
+++ b/help.go
@@ -42,6 +42,7 @@ var helpCommand = &Command{
 		// 3 $ app foo
 		// 4 $ app help foo
 		// 5 $ app foo help
+		// 6 $ app foo -h (with no other sub-commands nor flags defined)
 
 		// Case 4. when executing a help command set the context to parent
 		// to allow resolution of subsequent args. This will transform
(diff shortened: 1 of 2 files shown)


help_test.go
```diff
@@ -1126,6 +1126,61 @@ func TestHideHelpCommand_WithSubcommands(t *testing.T) {
...
+func TestHideHelpCommand_RunAsSubcommand_False_CustomTemplate(t *testing.T) {
+	var buf bytes.Buffer
+
+	app := &App{
+		Writer: &buf,
+		Commands: []*Command{
+			{
+				Name:               "dummy",
+				CustomHelpTemplate: "TEMPLATE",
+				HideHelpCommand:    false,
```
dearchap, line 1138: This is redundant since default is false.

dearchap (approved):

dearchap:
@jojje Thanks for your PR. I've changed the merge from main to v2-maint.

help.go
```diff
@@ -77,10 +78,25 @@ var helpCommand = &Command{
...
+			for _, c := range candidates {
+				if c == alias {
+					return true
+				}
+			}
+		}
+	}
+	return false
+}
+
```
meatballhat, line 99: I can't find where this function is used 🤔

jojje: Oh, that should never have made it into the PR.
It was a different approach to solving the problem. Thanks for spotting this Dan!

Pushed a corrective commit that removes this function, and also the redundant test which DC pointed out.
Once you're happy, I'll squash the PR.

dearchap (approved):

PR #1736, opened 2023-05-21
Fix:(issue_1610). Keep RunAsSubcommand behaviour as before

### LINKED ISSUE #1610: RunAsSubcommand has been broken
## My urfave/cli version is

_**v2.25.0**_

## Checklist

## Dependency Management

## Describe the bug

On https://github.com/urfave/cli/commit/d5947d781442f834bf417ecd020ec9568206844b the `RunAsSubcommand` method was removed. It was re-added on https://github.com/urfave/cli/commit/00afca42c863a8ab6b551bc41817a1558f626fa7 as a deprecated stub.

However, `a.RunContext(ctx.Context, ctx.Args().Slice())` behaves differently than what `RunAsSubcommand` did. For example, all the global flags in the original `cli.Context` are lost, where RunAsSubcommand correctly transferred them.

I have not found a way to run dynamic subcommands (i.e. an app which generates different subcommands at runtime by creating a subApp and calling `subApp.RunAsSubcommand(origCtx)`. The subApp no longer has access to global flags defined in the parent app as before.

For me this is a breaking and silent behavior change, moreover since it arrived with absolutely no heads up: https://github.com/urfave/cli/releases/tag/v2.20.0. It seems unintended, therefore a bug.

## To reproduce

See above.

## Observed behavior

Sub-apps cannot get global flags anymore.

## Expected behavior

There should be a way to pass global flags to sub-apps.

## Want to fix this yourself?

I think removals were done for a reason that is in conflict with keeping this functionality.

```
go version go1.19.3 linux/amd64
```

dearchap:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

_(REQUIRED)_

Fixes #1610 

## Testing

## Release Notes

_(REQUIRED)_

```release-note

```

### DIFF
--- a/app.go
+++ b/app.go
@@ -336,7 +336,15 @@ func (a *App) RunContext(ctx context.Context, arguments []string) (err error) {
 //
 // Deprecated: use App.Run or App.RunContext
 func (a *App) RunAsSubcommand(ctx *Context) (err error) {
-	return a.RunContext(ctx.Context, ctx.Args().Slice())
+	a.Setup()
+
+	cCtx := NewContext(a, nil, ctx)
+	cCtx.shellComplete = ctx.shellComplete
+
+	a.rootCommand = a.newRootCommand()
+	cCtx.Command = a.rootCommand
+
+	return a.rootCommand.Run(cCtx, ctx.Args().Slice()...)
 }
 
 func (a *App) suggestFlagFromError(err error, command string) (string, error) {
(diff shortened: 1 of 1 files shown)


hsanjuan:
This works for me (RunAsSubcommand seems to behave as before).

But also this forces me to keep using the method marked as deprecated. The new RunAsSubcommand implementation uses internal methods/fields so I cannot just do the things directly on my code.

Would we un-deprecate the method?

dearchap:
@hsanjuan Removed the deprecate.

PR #1742, opened 2023-05-28
Fix:(issue_1644) Run persistent flag action

### LINKED ISSUE #1644: A Persistent flag's action is not ran.
## Checklist

* [❌ ] Are you running the latest v3 release? Running the main branch
* [✔️] Did you check the manual for your release?
* [✔️] Did you perform a search about this feature?

When setting a global flag to persist and running a command within, the action of the flag is not ran.

```go
package cli_test

import (
	"testing"

	"github.com/urfave/cli/v3"
)

func TestFlagActionFail(t *testing.T) {

	var flagActionRan bool
	testFlag := &cli.BoolFlag{
		Name:       "env",
		Persistent: true,
		Action: func(ctx *cli.Context, b bool) error {
			flagActionRan = true
			return nil
		},
	}

	app := &cli.App{
		Flags: []cli.Flag{testFlag},
		Commands: []*cli.Command{
			{
				Name: "run",
					return nil
				},
			},
		},
	}

	err := app.Run([]string{"prog", "run", "--env"})
	if err != nil {
		t.Error(err)
		return
	}

	if !flagActionRan {
		t.Error("Flag action did not run")
	}
}

```

## Solution description

In `command.go` we are calling the following, but c.Flags does not include persistent flags.
```go
if err = runFlagActions(cCtx, c.Flags); err != nil {
	return err
}
```

dearchap:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

_(REQUIRED)_

Fixes #1644 

## Testing

Updated test cases

## Release Notes

_(REQUIRED)_

```release-note

```

### DIFF
--- a/app.go
+++ b/app.go
@@ -227,7 +227,7 @@ func (a *App) Setup() {
 	for _, fl := range a.Flags {
 		if cf, ok := fl.(CategorizableFlag); ok {
 			if cf.GetCategory() != "" {
-				a.flagCategories.AddFlag(cf.GetCategory(), cf)
+				a.flagCategories.AddFlag(cf.GetCategory(), fl)
 			}
 		}
 	}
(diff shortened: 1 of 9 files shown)


skelouse:
Nice!  Does that test fail without your changes?

dearchap:
> Nice! Does that test fail without your changes?

Yes it does.

PR #1745, opened 2023-05-30
Feat: Add value source

dearchap:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

This places Envs and Filepaths on same footing instead of having one take precedence over the other. In fact the user can choose which to prioritize but placing the sources in the given order. The new interface also allows plugin implementations i.e for reading values from networks, custom format files etc etc. 

_(REQUIRED)_

## Testing

Updates all test case 

## Release Notes

_(REQUIRED)_

```release-note

```

### DIFF
--- a/flag_value_source.go
+++ b/flag_value_source.go
@@ -43,6 +43,16 @@ func (e EnvSource) Identifier() string {
 	return fmt.Sprintf("environment variable %q", string(e))
 }
 
+// EnvSources is a helper function to encapsulate a bunch
+// of envs together as ValueSources
+func EnvSources(envs ...string) ValueSources {
+	vs := []FlagValueSource{}
+	for _, env := range envs {
+		vs = append(vs, EnvSource(env))
+	}
+	return vs
+}
+
 // FileSource encapsulates an file source
 type FileSource string
 
@@ -53,4 +63,14 @@ func (f FileSource) Get() (string, bool) {
 
 func (f FileSource) Identifier() string {
 	return "File"
+}
+
+// FileSources is a helper function to encapsulate a bunch
+// of file paths together as ValueSources
+func FileSources(paths ...string) ValueSources {
+	vs := []FlagValueSource{}
+	for _, path := range paths {
+		vs = append(vs, FileSource(path))
+	}
+	return vs
 }
--- a/flag_value_source.go
+++ b/flag_value_source.go
@@ -43,9 +43,9 @@ func (e EnvSource) Identifier() string {
 	return fmt.Sprintf("environment variable %q", string(e))
 }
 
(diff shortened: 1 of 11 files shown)


flag_value_source.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func (e EnvSource) Identifier() string {
+	return fmt.Sprintf("environment variable %q", string(e))
+}
+
+// FileSource encapsulates an file source
+type FileSource string
+
+func (f FileSource) Get() (string, bool) {
```
avorima, line 49: Not returning an error could hide important information. For example wrong permissions would cause this source to silently be ignored.
I see that this behavior is consistent with `flagFromEnvOrFile`, but maybe it can be improved.

avorima (commented):
It would be nice to have a convenience constructor for simple cases, e.g.

```go
func EnvVars(...string) []FlagValueSource
func FilePaths(...string) []FlagValueSource

_ = &cli.StringFlag{
        Name: "foo",
        Sources: cli.EnvVars("FOO"),
}
```

There are a lot of changes in v3 and I think helpers like this would definitely make the transition easier.

flag_value_source.go
dearchap: So far no one has asked for it but yes I think it might be worthwhile.

flag_value_source.go
```diff
@@ -0,0 +1,56 @@
...
+// EnvSource encapsulates an env
+type EnvSource string
+
+func (e EnvSource) Get() (string, bool) {
+	envVar := strings.TrimSpace(string(e))
+	return syscall.Getenv(envVar)
+}
+
+func (e EnvSource) Identifier() string {
+	return fmt.Sprintf("environment variable %q", string(e))
```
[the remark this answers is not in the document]
dearchap: Updated.

flag_value_source.go
dearchap: The problem with err is should it be ignored or not. Some may want to ignore and continue to next source, others may want to return an error to the cli user. So that means we have to setup a policy on whether to ignore those errors or not. Add in flag validation as well and it starts getting complicated. So for now lets stick to current behaviour and then re-evaulate when the feature request actually comes up

flag_value_source_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"fmt"
+	"os"
+	"testing"
+)
+
+func TestEnvSource(t *testing.T) {
+
+	os.Setenv("foo", "bar")
```
avorima, line 11: `t.Setenv` unsets the env after the test ends

avorima (approved):
:+1:

PR #1746, opened 2023-06-02
Consolidate all `App` code into `Command`

### LINKED ISSUE #1586: Consolidate all `App` code into `Command`

meatballhat:
## What type of PR is this?

- cleanup
- feature

In addition to what is described in #1586, this change:

- adds `Command.Root()` method for returning the "root command" (previously what was `App`) which is generally used to get the "root Writer". I'd like to make this access more sugary than `cCtx.Command.Root().Writer` once `Context` and `Command` are collapsed 🤞🏼 
- consolidates "running" to a single `Run` method that accepts only stdlib `context.Context` and a string slice like `os.Args`
- removes the `Commands` type alias for `[]*Command`
- adds and uses `tracef` func for runtime tracing to aid with development, testing, and debugging
- modifies many (but not all!) tests to use `testify/require`
- moves `expectFileContent` from `fish_test.go` to `cli_test.go` and modifies it to use `testify/require`
- replaces mutated `HelpName` string with `FullName` method, which also removes the ability to customize the name used for a command in help text
- replaces global variables for some commands and flags with private functions for building copies when needed
- removes (I assume) accidentally added binary file in `./cmd/`
- probably other notable things 😬 I could have been more careful. Critical feedback wanted! 🙇🏼 

Closes #1586


cmd/urfave-cli-genflags/urfave-cli-genflags
dearchap, line 1: Is this supposed to be checked into git ?

command.go
```diff
@@ -1,253 +1,456 @@
...
+			}
+		})
+	}
+
+	for _, subCmd := range cmd.Commands {
+		tracef("setting sub-command parent as self")
+		subCmd.parent = cmd
+
+		tracef("assigning subCmd.flagCategories from subCmd.Flags")
+		subCmd.flagCategories = newFlagCategoriesFromFlags(subCmd.Flags)
```
dearchap, line 226: Does this need to be done here and not in cmd.Setup ?

context.go
```diff
@@ -7,37 +7,50 @@ import (
...
 	flagSet       *flag.FlagSet
-	parentContext *Context
+	parent        *Context
 }
 
-// NewContext creates a new context. For use in when invoking an App or Command action.
-func NewContext(app *App, set *flag.FlagSet, parentCtx *Context) *Context {
-	c := &Context{App: app, flagSet: set, parentContext: parentCtx}
+// NewContext creates a new context. For use in when invoking a Command action.
+func NewContext(cmd *Command, set *flag.FlagSet, parentCtx *Context) *Context {
```
dearchap, line 29: rename parentCtx to parent as well ?

help.go
```diff
@@ -119,22 +126,24 @@ func ShowAppHelpAndExit(c *Context, exitCode int) {
...
+	tmpl := cCtx.Command.CustomRootCommandHelpTemplate
+	if tmpl == "" {
+		tracef("using RootCommandHelpTemplate")
+		tmpl = RootCommandHelpTemplate
 	}
 
-	if cCtx.App.ExtraInfo == nil {
-		HelpPrinter(cCtx.App.writer(), tpl, cCtx.App)
+	if cCtx.Command.ExtraInfo == nil {
+		HelpPrinter(cCtx.Command.Root().Writer, tmpl, cCtx.Command)
```
dearchap, line 136: should be be 

```
HelpPrinter(cCtx.Command.Root().Writer, tmpl, cCtx.Command.Root())
```

help.go
```diff
@@ -119,22 +126,24 @@ func ShowAppHelpAndExit(c *Context, exitCode int) {
...
-		return map[string]interface{}{
-			"ExtraInfo": cCtx.App.ExtraInfo,
+	tracef("setting ExtraInfo in customAppData")
+	customAppData := func() map[string]any {
+		return map[string]any{
+			"ExtraInfo": cCtx.Command.ExtraInfo,
 		}
 	}
-	HelpPrinterCustom(cCtx.App.writer(), tpl, cCtx.App, customAppData())
+	HelpPrinterCustom(cCtx.Command.Root().Writer, tmpl, cCtx.Command, customAppData())
```
dearchap, line 146: same comment as above about Root()

dearchap (dismissed):

command.go
meatballhat: oooh yeah this looks like a duplicate

thank you!

context.go
meatballhat: Yes please!

help.go
meatballhat: Ah! Thank you 😁

dearchap (approved):

PR #1750, opened 2023-06-12
Document slice flags as part of examples (v3)

### LINKED ISSUE #1005: v2 documentation: add a StringSliceFlag example to the manual

This would solve people creating issues about unexpected behavior from the `StringSliceFlag`, like these ones:

- https://github.com/urfave/cli/issues/910
- https://github.com/urfave/cli/issues/1003

## Solution description

Add an example, or maybe multiple examples, of how to input `StringSliceFlag`

I've considered possibly changing `StringSliceFlag` to work in the way that both those users were expecting, but that sounds dangerous.

carhartl:
This PR adds an example for how to use slice flags to `flags.md` in the v2 docs example directory, using a `StringSliceFlag` in the specific example. As instructed in #1005, which seems to indicate there's some demand for it.

## What type of PR is this?

- documentation

Make the documentation more complete, so that users can get started without looking at source/tests/discussions, like me :)

Fixes #1005

## Release Notes

```release-note
Add example for using slice flags.
```

### DIFF
--- a/docs/v2/examples/flags.md
+++ b/docs/v2/examples/flags.md
@@ -230,6 +230,52 @@ That flag can then be set with `--lang spanish` or `-l spanish`. Note that
 giving two different forms of the same flag in the same command invocation is an
 error.
 
+#### Multiple Values per Single Flag
+
+Using a slice flag allows you to pass multiple values for a single flag; the values will be provided as a slice:
+
+- `Int64SliceFlag`
+- `IntSliceFlag`
+- `StringSliceFlag`
+
+<!-- {
+  "args": ["&#45;&#45;greeting Hello", "&#45;&#45;greeting Hola"],
+  "output": "Hello, Hola"
+} -->
+```go
+package main
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"strings"
+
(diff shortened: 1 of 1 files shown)


docs/v2/examples/flags.md
```diff
@@ -230,6 +230,52 @@ That flag can then be set with `--lang spanish` or `-l spanish`. Note that
...
+
+```go
+package main
+
+import (
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/urfave/cli/v2"
```
dearchap, line 254: This needs to be v3 if we need to merge this PR to main. If you want this change in v2 can you rebase your changes off of v2-maint branch ?

carhartl: Ah, totally overlooked all of that, makes sense!

carhartl: So it'd be useful to have for both v2 and v3?

dearchap: Yes. But we would need separate PRs for each

dearchap (approved):
Thank you @carhartl

PR #1752, opened 2023-06-12
Some docs housekeeping and minor cleanups

meatballhat:
## What type of PR is this?

- cleanup
- documentation

- the `.flake8` file is no longer used
- I felt that the first section of the top-level `README.md` didn't help explain why someone might choose to use this project and the `docs/index.md` intro should probably be the same (opinions wanted)
- the `LICENSE` year was outdated
 
N/A 😬


dearchap (approved):

flag.go
```diff
@@ -373,7 +372,7 @@ func hasFlag(flags []Flag, fl Flag) bool {
 func flagFromEnvOrFile(envVars []string, filePaths []string) (value string, fromWhere string, found bool) {
```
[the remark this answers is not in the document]
meatballhat: doesn't look like it! 🚮

README.md
```diff
@@ -1,18 +1,32 @@
...
 [![Go Report Card](https://goreportcard.com/badge/github.com/urfave/cli/v3)](https://goreportcard.com/report/github.com/urfave/cli/v3)
 [![codecov](https://codecov.io/gh/urfave/cli/branch/main/graph/badge.svg?token=t9YGWLh05g)](https://codecov.io/gh/urfave/cli)
 
-cli is a simple, fast, and fun package for building command line apps in Go. The
-goal is to enable developers to write fast and distributable command line
-applications in an expressive way.
+urfave/cli is a **declarative**, simple, fast, and fun package for building command line tools in Go featuring:
+
+- nestable commands with alias and prefix match support
+- flexible and permissive help system
```
abitrolly, line 11: Permissive help? What does it permit?

README.md
```diff
@@ -1,18 +1,32 @@
...
+- dynamic shell completion for `bash`, `zsh`, `fish`, and `powershell`
+- `man` and markdown format documentation generation
+- input flag types for primitives, slices of primitives, time, duration, and others
```
abitrolly, line 14: The terminology "primitives" and "slices of primitives" seems unique here. Simple types is better. Why would it be a feature?

README.md
```diff
@@ -1,18 +1,32 @@
...
+- compound short flag support (`-a` `-b` `-c` :arrow_right: `-abc`)
+- input value sources including:
```
abitrolly, line 16:
```suggestion
- options lookup from:
```

README.md
meatballhat: I was thinking of how the default behavior allows `help` before and after positional commands, and has a default `h` alias, and also the `--help` `-h` flags, although these behave in a more standard way. I picked "permissive" because I wanted a single word that describes how the help system is trying to be compatible with various usage styles.

README.md
meatballhat: I like "simple types" :+1: How would you like to see it worded?

README.md
```diff
@@ -1,18 +1,32 @@
...
+    - environment variables
+    - plain text files
+    - [structured file formats supported via the `urfave/cli-altsrc` package](https://github.com/urfave/cli-altsrc)
 
 ## Documentation
 
-More documentation is available in [`./docs`](./docs) or the hosted
-documentation site at <https://cli.urfave.org>.
+More documentation is available in [`./docs`](./docs) or the hosted documentation site published from the latest release
+at <https://cli.urfave.org>.
```
abitrolly, line 24: What about https://pkg.go.dev/github.com/urfave/cli#section-documentation ?

meatballhat: I'm not sure how best to handle the pkg.go.dev docs in relation to the `./docs` tree that's published via mkdocs and the `gh-pages` branch. Do you think mentioning it here is good? More than the link button alone, that is?

README.md
abitrolly: Depends on what purpose does this feature serve. Why I would need type autoconversion at all? It may bloat the library.

README.md
abitrolly: I would skip that, but if there us a dedicated documentation section, the autogenerated godoc API should be mentioned there. There is no replacement for it in `./docs`.

README.md
meatballhat: The ordering is true for the flag form, but the behavior is a little different for the command form which treats adjacent command name and `help` as equivalent no matter the order, meaning these are equivalent:

```
prog cmd help
```

```
prog help cmd
```

README.md
meatballhat: I don't follow 😅 Maybe it would help for me to define the audience more clearly (?) I'm attempting to speak to a software engineer who has worked with Go and knows about the standard library `flag` package.

dearchap (dismissed):

README.md
meatballhat: @abitrolly I'm in favor of replacing "primitive" with "simple" 👍🏼 I wasn't following what you meant by:

> Depends on what purpose does this feature serve. Why I would need type autoconversion at all? It may bloat the library.

abitrolly: @meatballhat my first language is Python, so the comment was that the library exposes too much types. Uint, Float64, which are too specific for a command line API. And if I understand it correctly, if these types are not used, they will still take space in importing app.

Issue #1531, opened 2022-10-13
Minimize top-level API surface area for both usability clarity and maintenance reduction


Dokiys:
- [x] #1586 
- [x] #1600
- [x] #1587
- [x] #1759 
- [x] #1780
- [x] #1791

meatballhat:
(Continuing discussion from #833)

> > context.Context is explicitly not meant to be extended
> 
> @meatballhat any documentation or links to where that explicit convention maybe written?


@marwan-at-work I should have explained myself better, sorry! The interpretation of the go docs I'm leaning on here comes from [the overview]()

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx:
> 
> ```
> func DoSomething(ctx context.Context, arg Arg) error {
> 	// ... use ctx ...
> }
> ```

meaning that we're not supposed to be wrapping `context.Context` with our own type because that goes against the convention of accepting `context.Context` as first argument for the purpose of keeping the semantics/expectations of using `context.Context` as simple as possible.

Aside from that particular interpretation issue, I'd like to make sure we are moving towards the design detailed in the series of issues tracked in #1531 which would result in an action func signature of:

```go
type ActionFunc func(ctx context.Context, cmd *Command) error
```

meatballhat:
@marwan-at-work can you move [your comment](https://github.com/urfave/cli/issues/833#issuecomment-1328265820) over here? 🙇🏼

marwan-at-work:
@meatballhat ah, yes storing context vs passing it. In most cases, you want a function to accept a context. But FWIW the Go team themselves break that rule in multiple places: 
1. http.Request stores its context as a [private field](https://github.com/golang/go/blob/d5de62df152baf4de6e9fe81933319b86fd95ae4/src/net/http/request.go#L326) and exposes `Context()` and `WithContext` methods to get/set a request's context. 
2. the sql package also stores its context for queries: https://github.com/golang/go/blob/d5de62df152baf4de6e9fe81933319b86fd95ae4/src/database/sql/sql.go#L2159
3. The `os` package actually does exactly what we're doing here and embedding the context interface: https://github.com/golang/go/blob/d5de62df152baf4de6e9fe81933319b86fd95ae4/src/os/signal/signal.go#L299

With all of that said, I'm definitely happy with any solution as long as we have a full context accessible from the outside :) 

[...]

PR #1760, opened 2023-06-17
Reworking value source API

meatballhat:
## What type of PR is this?

- cleanup
- feature

Reworking the "flag value source" API in prep for ensuring [urfave/cli-altsrc](https://github.com/urfave/cli-altsrc) can be easily used with `urfave/cli/v3` by:
- reducing the surface area by keeping the environment variable and file implementations private
- generalizing to "value source" in name
- using the stdlib `fmt.Stringer` and `fmt.GoStringer` interfaces
- switching the value source interface to use `Lookup` to mimic stdlib semantics such as `os.Lookup`

Connected to #1531


value_source_test.go
```diff
@@ -0,0 +1,67 @@
...
+	f := &fileValueSource{Path: "junk_file_name"}
+	_, ok := f.Lookup()
+	require.False(t, ok)
+}
+
+func TestFilePaths(t *testing.T) {
+	r := require.New(t)
+
+	r.Nil(os.Chdir(t.TempDir()))
+	r.Nil(os.WriteFile("some_file_name_1", []byte("Hello"), 0644))
```
dearchap, line 51: defer os.Delete("some_file_name_1")

value_source.go
```diff
@@ -0,0 +1,108 @@
...
+	fmt.GoStringer
+
+	// Lookup returns the value from the source and if it was found
+	// or returns an empty string and false
+	Lookup() (string, bool)
+}
+
+// ValueSourceChain is a slice of ValueSource that allows for
+// lookup where the first ValueSource to resolve is returned
+type ValueSourceChain struct {
```
dearchap, line 22: I would prefer 

```
type ValueSources []ValueSource
```

instead of defining a struct. Of course if we need to store state about sources then struct is needed. Also no need to name it ValueSourceChain

value_source.go
```diff
@@ -0,0 +1,108 @@
...
+		if value, found := src.Lookup(); found {
+			return value, src, true
+		}
+	}
+
+	return "", nil, false
+}
+
+// envVarValueSource encapsulates a ValueSource from an environment variable
+type envVarValueSource struct {
```
dearchap, line 62:
```
type envVarValueSource string
```

value_source.go
```diff
@@ -0,0 +1,108 @@
...
+
+	for _, key := range keys {
+		vsc.Chain = append(vsc.Chain, &envVarValueSource{Key: key})
+	}
+
+	return vsc
+}
+
+// fileValueSource encapsulates a ValueSource from a file
+type fileValueSource struct {
```
dearchap, line 86:
```
type fileValueSource string
```

dearchap:
@meatballhat When you say "reintegration with altsrc" do you mean just have the API compatible with altsrc or actually pulling the alstrc codebase back into this repo ?

meatballhat:
> @meatballhat When you say "reintegration with altsrc" do you mean just have the API compatible with altsrc or actually pulling the alstrc codebase back into this repo ?

@dearchap I definitely **do not** mean bringing the altsrc codebase back here 😅 I'll update the description 👍🏼

value_source.go
meatballhat: I intentionally moved away from a slice type because:
- imho the semantics and behavior of literals and funcs with slice type receivers is **weird**, meaning it's different enough from base types and structs that I'd like to avoid it whenever possible
- wrapping the slice in a struct means that the struct can grow and change without working around the above point

I intentionally changed the name to `ValueSourceChain` because `ValueSources` is only one letter off from `ValueSource` and I wanted there to be more "distance" between these types *at least* for visual differentiation.

WDYT?

value_source_test.go
meatballhat: This is not needed anymore because of the use of `t.TempDir()`

value_source.go
meatballhat: Roughly the same comments about literals and func receivers for simple types as with slice types, and also future-proofing. I did this on purpose ❤️

value_source.go
meatballhat: Same here, too

value_source_test.go
meatballhat: well gosh, seems that using `testing.T.TempDir` on windows is asking for trouble 😂 

working on it...

value_source.go
dearchap: I think its fine as it is.

dearchap (approved):

PR #1776, opened 2023-06-23
Chore: Rename mkdocs requirements file name

dearchap:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

Due to the file mkdocs-requirements.txt FrogBot detects this repo as a mixed golang & python module and tries to scan python vulnerabilites which results in a failure since there are no python files in this repo. So we just rename the file to avoid its py scan

_(REQUIRED)_

## Testing

## Release Notes

_(REQUIRED)_

```release-note

```


mkdocs-r.txt
meatballhat, line 1: I know this is such a nit pick, but I work with `R` every day and this name caught me off guard 😅  Can it be `mkdocs-reqs.txt`?

dearchap: Shouldn't be a problem. I can make the change

meatballhat (approved):
Thank you!

PR #1784, opened 2023-06-25
Collapse `Context` into `Command`

### LINKED ISSUE #1587: Consolidate all `Context` code into `Command`
(wild generalizations warning) The `Context` struct exists for reasons mostly related to the use of stdlib `flag` and has a concept of "lineage" that forms a tree with other `Context` nodes. I would like to see this lineage relationship move to `Command` given the existence of `.Commands` for traversing outward, and for `Command` to be the source of truth for values that are fetched via the type-specific flag accessor functions like `.String(string) string`, `.Int(string) int`, `.Bool(string) bool`.

### LINKED ISSUE #1780: Pass `(context.Context, *Command)` to action funcs instead of `(*Context)`

meatballhat:
## What type of PR is this?

_(REQUIRED)_

- bug
- cleanup
- documentation
- feature

_(REQUIRED)_

Closes #1587 
Closes #1780 

## Testing

## Release Notes

_(REQUIRED)_

```release-note

```


skelouse:
Quick thought: What if rather than passing (ctx.Context, cmd) to Action, Before, and After, pass a context.WithValue?  Then for example rather than calling `ctx.String("<flag>")` you'd use `<import>.String(ctx, "<flag>")`.  Although this would simply be flipping storing context within the `urfave/cli`'s context to storing package context within context.Context and may complicate unit testing.

Also, will command be passed to flag Action?  `Action func(context.Context, cmd, T) error`

meatballhat:
> Quick thought: What if rather than passing (ctx.Context, cmd) to Action, Before, and After, pass a context.WithValue? Then for example rather than calling `ctx.String("<flag>")` you'd use `<import>.String(ctx, "<flag>")`. Although this would simply be flipping storing context within the `urfave/cli`'s context to storing package context within context.Context and may complicate unit testing.

@skelouse I'm pretty nervous about API signatures that intentionally pass values through `context.Context`'s `Value` func. Is that what you meant?

skelouse:
> > @skelouse I'm pretty nervous about API signatures that intentionally pass values through `context.Context`'s `Value` func. Is that what you meant?

Yeah that's what I meant.

meatballhat:
> > @skelouse I'm pretty nervous about API signatures that intentionally pass values through `context.Context`'s `Value` func. Is that what you meant?
> 
> Yeah that's what I meant.

Yeeeeahh I'd like to avoid that if possible 😅 I feel nervous enough about using a `context.Context` to track the _internal_ relationship between a `*Command` and it's potential parent `*Command`.

flag_duration.go
```diff
@@ -36,8 +36,8 @@ func (d *durationValue) Get() any { return time.Duration(*d) }
 
 func (d *durationValue) String() string { return (*time.Duration)(d).String() }
 
-func (cCtx *Context) Duration(name string) time.Duration {
-	if v, ok := cCtx.Value(name).(time.Duration); ok {
+func (cmd *Command) Duration(name string) time.Duration {
+	if v, ok := cmd.Value(name).(time.Duration); ok {
```
dearchap, line 40: no tracef here like for cmd.Bool(..)

flag_impl.go
```diff
@@ -263,18 +264,18 @@ func (f *FlagBase[T, C, V]) GetDefaultText() string {
 }
 
 // Get returns the flag’s value in the given Context.
```
dearchap, line 266: update comment from Context - > Command

dearchap (approved):

PR #1785, opened 2023-06-26
Feature: Add support for validation functions

dearchap:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

This PR adds a framework for allowing users to be able to define custom validators on flags. Some very basic rudimentary ones have been provided for demonstration purposes to be able to gather user feedback on easy of use. Other validation functions may be provided in the future if users requests common ones. Note that the validation happens in the same context as when the value is *set* from the command line so slice flag validations for length may not work as expected.

_(REQUIRED)_

## Testing

Add new validation test cases, 

## Release Notes

_(REQUIRED)_

```release-note

```


command_test.go
```diff
@@ -3068,26 +3068,29 @@ func TestPersistentFlag(t *testing.T) {
...
-				OnlyOnce: true,
-			},
-			&IntSliceFlag{
-				Name: "isflag",
-			},
-			&FloatSliceFlag{
-				Name:     "fsflag",
-				OnlyOnce: true,
+
+	cmd_get := func() *Command {
```
meatballhat, line 3072: If you need a fresh one for each test case, maybe define inside the `t.Run(...)` scope?

flag_impl.go
```diff
@@ -13,47 +13,35 @@ type Value interface {
...
+	isBool bool
+	v      Value
 }
 
-func (v *valueWrapper) Set(s string) error {
-	if v.count == 1 && v.onlyOnce {
-		return fmt.Errorf("cant duplicate this flag")
+func (f fnValue) Get() any           { return f.v.Get() }
+func (f fnValue) Set(s string) error { return f.fn(s) }
+func (f fnValue) String() string {
```
meatballhat, line 28: Are these function receivers intentionally not `*fnValue`?

flag_impl.go
```diff
@@ -162,13 +153,29 @@ func (f *FlagBase[T, C, V]) Apply(set *flag.FlagSet) error {
...
-		value:    f.value,
-		onlyOnce: f.OnlyOnce,
+	isBool := false
+	if b, ok := f.value.(boolFlag); ok && b.IsBoolFlag() {
+		isBool = true
 	}
 
 	for _, name := range f.Names() {
-		set.Var(vw, name, f.Usage)
+		set.Var(fnValue{
```
meatballhat, line 162: Not a `&fnValue{`?

go.mod
```diff
@@ -10,5 +10,7 @@ require (
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
+	github.com/russross/blackfriday/v2 v2.1.0 // indirect
+	golang.org/x/exp v0.0.0-20230522175609-2e198f4a06a1 // indirect
```
meatballhat, line 14: This seems odd that `github.com/russross/blackfriday/v2` is re-introduced 🤔 It doesn't appear to be used.

I'm _more than a little bit_ nervous about including `golang.org/x/exp` given how volatile it is and having a pre-`v1` version that's more likely to differ with downstream `go.mod` files. Are the capabilities we're using something that can be copy-vendor'd into this codebase until it graduates out of `golang.org/x/exp`?

flag_impl.go
```diff
@@ -162,13 +153,29 @@ func (f *FlagBase[T, C, V]) Apply(set *flag.FlagSet) error {
...
+			fn: func(val string) error {
+				if f.count == 1 && f.OnlyOnce {
+					return fmt.Errorf("cant duplicate this flag")
+				}
+				f.count++
+				if err := f.value.Set(val); err != nil {
+					return err
+				}
+				if f.Validator != nil {
+					return f.Validator(f.value.Get().(T))
```
meatballhat, line 172: I know this isn't the first place we have an unchecked type coercion, but for some reason this one makes me really nervous 😅 Is there an alternative that would allow us to bubble up an error instead of panicking?

flag_validation.go
```diff
@@ -0,0 +1,69 @@
+package cli
+
+import (
+	"fmt"
+
+	"golang.org/x/exp/constraints"
+)
+
+// ConditionOrError ir a helper function to make writing
```
meatballhat, line 9:
```suggestion
// ConditionOrError is a helper function to make writing
```

flag_validation.go
```diff
@@ -0,0 +1,69 @@
...
+	return func(v T) error {
+		var errs []error
+		for _, fn := range fns {
+			if err := fn(v); err == nil {
+				return nil
+			} else {
+				errs = append(errs, err)
+			}
+		}
+		return fmt.Errorf("%+v", errs)
```
meatballhat, line 45: I thought we had a multi-error type already somewhere 🤔

meatballhat (commented):
- I really like this capability 
- I'm super nervous about depending on golang.org/x/exp
- I don't like the alternatives to depending on golang.org/x/exp
- I'm nervous about how much this expands the API surface area

What do you think about making some/most of this an "addon" library like with urfave/cli-docs and urfave/cli-altsrc?

command_test.go
dearchap: Yeah makes sense. Done

flag_impl.go
dearchap: Not really. I tend to avoid pointer receivers if there is nothing to change in the struct. But yeah I'll change it

flag_impl.go
dearchap: Updated

flag_impl.go
dearchap: I understand. For now lets just roll with this. Maybe we can enforce a stronger type check in all the other ctx.Type(str) functions as well along with this. It is mostly safe if you use internal libraries. For user defined types yes it is problematic. Lets revisit later

go.mod
dearchap: Actually that is a very good point. Lets create a separate urfave/cli-validation library which has all this. That way the core repo is unaffected by all this. So the only addition in our repo is the Validator field and nothing else.

dearchap: ran go mod tidy and removed blackfriday

flag_impl.go
```diff
@@ -160,15 +151,40 @@ func (f *FlagBase[T, C, V]) Apply(set *flag.FlagSet) error {
...
+					return fmt.Errorf("cant duplicate this flag")
+				}
+				f.count++
+				if err := f.value.Set(val); err != nil {
+					return err
+				}
+				if f.Validator != nil {
+					if err := f.Validator(f.value.Get().(T)); err != nil {
+						return err
+					}
```
meatballhat, line 181: I think this needs a type check around `f.value.Get().(T)`

flag_validation_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestFlagDefaultValidation(t *testing.T) {
+
```
meatballhat, line 11:
```suggestion
```

flag_validation_test.go
```diff
@@ -0,0 +1,114 @@
...
+	r := require.New(t)
+
+	// Default value of flag is 2 which should fail validation
+	err := cmd.Run(buildTestContext(t), []string{"foo", "--if", "5"})
+	r.Error(err)
+}
+
+func TestFlagValidation(t *testing.T) {
+
```
meatballhat, line 36:
```suggestion
```

meatballhat (approved):
Aww, I liked the validation helper bits! 😭  WDYT about renaming `urfave/cli-altsrc` to `urfave/cli-ext` and including them over there?

flag_impl.go
dearchap: Added.

flag_validation_test.go
dearchap: We need this to pass codecov

flag_validation_test.go
dearchap: We need this to pass codecov

dearchap:
> Aww, I liked the validation helper bits! 😭 WDYT about renaming `urfave/cli-altsrc` to `urfave/cli-ext` and including them over there?

Hmm thats not a bad idea at all. I'm ok with that. cli-docs would go in there as well ? Would you like to take point on that ?

PR #1789, opened 2023-06-28
Feat: Add minimal support for i18n

### LINKED ISSUE #980: i18n support for urfave/cli

We have many users of this CLI from all over the world, so it may be useful to provide translations for our error messages? Possibly ones that default to the language set by the user's environment variables? That would likely make the CLI easier to use for them.

## Solution description

This is more-so a discussion and idea generation issue, rather than an explicit feature spec.

dearchap:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

Enable basic internalization support for urfave/cli

_(REQUIRED)_

Fixes #980 

## Testing

## Release Notes

_(REQUIRED)_

```release-note

This adds basic framework support for internationalization for urfave/cli. Right now the languages built in are en-US/en-GB which zero changes between the messages. As people request more languages the capability to add this is present with no changes to core code base
```


meatballhat:
Ack sorry about the conflicts! 😭

meatballhat (commented):
- This is very exciting! 🎉 
- It looks like the generated catalog code sets the `message.DefaultCatalog` package-level variable 🤔 Does this mean that this library is intended for use at the _application_ level and not from a dependency like `github.com/urfave/cli/v3`?
- Is there supposed to be an import of `_ "github.com/urfave/cli/v3/internal/translations"` somewhere in the top level so that the catalog is initialized?
- I had to manually install `gotext` to get this working locally. Can a `gotext`-ensuring command be added to `internal/build/build.go` like the one that ensures `goimports` exists?

dearchap:
@meatballhat

- Addressed all the conflicts
- Added import of _ "github.com/urfave/cli/v3/internal/translations"
- Added ensure-gotext to internal build 
- Added ensure-gotext and go-generation to the build system

meatballhat:
@dearchap Thank you! I sincerely appreciate your work on this 🤘🏼 

I'm still confused about the intended usage given [this line in the generated code](
https://github.com/dearchap/cli/blob/d43b64899ad6275591450859aac28458995d87bf/internal/translations/catalog.go#L38). How is this supposed to work if `urfave/cli/v3` is being used by an application that _also_ uses code generated by `gotext`?

Given that [the template file for the generated code](https://github.com/golang/text/blob/e50348080f29449bcd6808c11400b3d45f08b09d/message/pipeline/generate.go#L321) hasn't been touched in ~4 years, I'm wondering if using `golang.org/x/text/message/pipeline` (via `gotext`) is the right move 🤔

dearchap:
@meatballhat you have a keen eye. I tested cli with these changes along with a sample app that uses its own catalog and basically the translation doesn't work as one catalog overrides the other due to the message.DefaultCatalog in init. I'm going to redo this PR.

dearchap:
@meatballhat I've hacked around the package global Catalog initialization. I dont like it but this is what we have until support for multiple Catalogs is rolled into the text/message package itself. 
https://cs.opensource.google/go/x/text/+/refs/tags/v0.10.0:message/catalog/catalog.go;l=281

For now this is the simplest route without adding dependencies on other i18n libraries which look to me more heavyweight

dearchap:
@meatballhat Also I'm not sure why only the window tests are failing. Can you take a look ?

meatballhat:
> @meatballhat Also I'm not sure why only the window tests are failing. Can you take a look ?

Offhand I don't see why that failure is happening 🤔 Can the `make go-generate` step be limited to most recent Go on ubuntu like with some of the nearby steps?

internal/translations/dcatalog.go
meatballhat, line 1: Is the intent here that the load order of the files is by name and that using `init` functions to stash the `message.DefaultCatalog` as `oldCatalog` will allow us to keep a reference to the catalog that is "ours"?

dearchap: Yes correct.

meatballhat: OK. hmm. Is this going to work reliably? Is load order and how it relates to `init` func running even guaranteed? (my quick search isn't encouraging 😅)

For me, this all tells a story of `golang.org/x/text/message` *not really being ready* for production usage, and yet there's no specific warning against using it in production. I'm confused!

dearchap: https://medium.com/@markbates/go-init-order-dafa89fcef22

dearchap: @meatballhat i cannot guarantee the load order. So I think we should just scrap this approach.

PR #1799, opened 2023-07-22
Always get 0 for a nested int64 value in v2.25.7

stephenfire:
## bug fix

- always get 0 for a nested int64 value in v2.25.7

### DIFF
--- a/altsrc/map_input_source.go
+++ b/altsrc/map_input_source.go
@@ -159,7 +159,7 @@ func (fsm *MapInputSource) Int64(name string) (int64, error) {
 	}
 	nestedGenericValue, exists := nestedVal(name, fsm.valueMap)
 	if exists {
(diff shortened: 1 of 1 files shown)


abitrolly (approved):
I can not test it ATM, because `make`, as described here https://cli.urfave.org/CONTRIBUTING/#development-workflow fails on my machine ("no goimports").

And it looks like we need to give an explicit approval to run CI after installing Frogbot in https://github.com/urfave/cli/pull/1778

abitrolly:
@urfave/cli frogbot looks broken. CI for PRs is not being executed.

stephenfire:
> @urfave/cli frogbot looks broken. CI for PRs is not being executed.

bypassing this problem with using uint64 by now :)

Issue #785, opened 2018-12-13
Creating two cli.Command with the same Name results in magic


johnwyles:
If you name a `cli.Command` the same as another `cli.Command` only *one* of them will execute with no error or warning that there is also a duplicate command by the same name and with an exit code of `0`. Consider the following:

```go
import (
	"fmt"
	"log"
	"os"

	"github.com/urfave/cli"
)

func main() {
	app := cli.NewApp()
	app.Commands = []cli.Command{
		// This is the only command that WILL execute since it is the first "foo" command found
		cli.Command{
			Action: func(c *cli.Context) { fmt.Println("foo1") }, // Print "foo1"
			Name:   "foo", // Name is "foo"
		},
		// This command will NEVER execute since it is the second "foo" match in the array
		cli.Command{
			Action: func(c *cli.Context) { fmt.Println("foo2") }, // Print "foo2"
			Name:   "foo", // Name is ALSO "foo"
		},
	}
	app.Name = "test"

	if err := app.Run(os.Args); err != nil {
		log.Fatalf("An error occurred: %s", err)
	}
}
```

Running this code yields the following output:
```shell
$ go run test.go foo
foo1
$ echo $?
0
```

Suggest checking all `cli.Command`s for any duplicates and exiting with a message and non-zero exit code *or* running all `cli.Command`s that match the `cli.Command.Name` field.

gist link: https://gist.github.com/johnwyles/3129a05391e1f749c661b1efc98abc67

johnwyles:
I should also point out this is especially more cryptic to  debug and discover (as it was for me 😄) when you have moved your `cli.Command` functions to separate files as you would probably do in most modest sized projects ([see here](https://github.com/johnwyles/vrddt/tree/master/cmd/cli)).

michaeljs1990:
It seems reasonable that this would just blow up when this happens. It shouldn't be a hard check to add internally.

johnwyles:
I split my commands per-file and certainly when you have several if you're not careful about what you name them you can screw this up and the debugging was atrocious - iirc 2 hours of killed

johnwyles:
Another thought I just had is that another seemingly expected behaviour would be that *both* would execute in some order (presumably as code reads) - I am unsure if you want to blow up or execute both because them ordering (however one *expect* ordering to go?) would determine their serial execution. Let me know if I need to clarify that point...

johnwyles:
[...]

PR #1805, opened 2023-08-19
fix: check duplicated sub command name and alias

linrl3:
## What type of PR is this?
- bug

Fix https://github.com/urfave/cli/issues/785.

Fix https://github.com/urfave/cli/issues/785

This fix is only for v2. Do we need to fix it in other versions?

## Testing

- I test it in unit test.
- Also, I write a dummy program and see the output:

```golang
package main

import (
	"fmt"

	"github.com/urfave/cli/v2"
)

func main() {
	app := &cli.App{
		Name: "linrl3",
		Commands: []*cli.Command{
			{Name: "sub1"},
			{Name: "sub1"},
		},
	}
	if err := app.Run([]string{"linrl3", "sub1"}); err != nil {
		fmt.Println(err)
	}
}
```
result: 
```
result:
➜  cli git:(fix/v2_duplicate_subcommand) ✗ go run cmd/dummyrun/main.go
parent command [linrl3] has duplicated subcommand name or alias: sub1
```

## Release Notes

_(REQUIRED)_

```release-note
- fix: if sub-commands in the same level have same name or alias, an error would return
```

### DIFF
--- a/command.go
+++ b/command.go
@@ -149,6 +149,9 @@ func (c *Command) Run(cCtx *Context, arguments ...string) (err error) {
 
 	if !c.isRoot {
 		c.setup(cCtx)
+		if err := checkDuplicatedCmds(c); err != nil {
+			return err
+		}
 	}
 
 	a := args(arguments)
@@ -403,4 +406,17 @@ func hasCommand(commands []*Command, command *Command) bool {
 	}
 
 	return false
+}
+
+func checkDuplicatedCmds(parent *Command) error {
+	seen := make(map[string]struct{})
+	for _, c := range parent.Subcommands {
+		for _, name := range c.Names() {
+			if _, exists := seen[name]; exists {
+				return fmt.Errorf("parent command [%s] has duplicated subcommand name or alias: %s", parent.Name, name)
+			}
+			seen[name] = struct{}{}
+		}
+	}
+	return nil
 }
--- a/app.go
+++ b/app.go
@@ -329,6 +329,9 @@ func (a *App) RunContext(ctx context.Context, arguments []string) (err error) {
 	a.rootCommand = a.newRootCommand()
 	cCtx.Command = a.rootCommand
 
+	if err := checkDuplicatedCmds(a.rootCommand); err != nil {
+		return err
+	}
 	return a.rootCommand.Run(cCtx, arguments...)
 }
 
--- a/app_test.go
+++ b/app_test.go
@@ -3087,3 +3087,92 @@ func TestFlagAction(t *testing.T) {
 		})
 	}
 }
+
+func TestDuplicateSubcommand(t *testing.T) {
+	var testdata = []struct {
+		app           *App
+		expectNoError bool
+	}{
+		{&App{
+			Name: "p1",
+		}, true},
+		{&App{
+			Name:     "p2",
+			Commands: []*Command{},
+		}, true},
+		{&App{
+			Name:     "p3",
+			Commands: []*Command{{Name: "sub1"}},
+		}, true},
+		{&App{
+			Name:     "p4",
+			Commands: []*Command{{Name: "sub1"}, {Name: "sub1"}},
+		}, false},
+		{&App{
+			Name:     "p5",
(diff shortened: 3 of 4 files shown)


command.go
```diff
@@ -404,3 +408,29 @@ func hasCommand(commands []*Command, command *Command) bool {
...
+	if parent == nil {
+		return nil
+	}
+	if len(parent.Subcommands) == 0 {
+		return nil
+	}
+
+	var (
+		seen = make(map[string]struct{})
+	)
```
abitrolly, line 422: Why not?
```suggestion
	seen := make(map[string]struct{})
```

command.go
```diff
@@ -404,3 +408,29 @@ func hasCommand(commands []*Command, command *Command) bool {
...
+	for _, c := range parent.Subcommands {
+		if c == nil {
+			continue
+		}
```
abitrolly, line 426: What is the purpose of `nil` subcommands here?

command.go
linrl3: sure, I will change it

command.go
linrl3: I'm not sure if we will have nil subcommand situation. I make it more defensive.

abitrolly: If `nil` is an allowed subcommand, it should be documented here. Otherwise the code is dead.

https://github.com/urfave/cli/blob/b686f4fd3343d353d0f079f6ce620850e23d329d/command.go#L39-L40

linrl3:
```golang
func Test_checkDuplicatedCmds_nilInput(t *testing.T) {
	expect(t, checkDuplicatedCmds(nil), nil)
	cmd := Command{
		Name:        "",
		Subcommands: []*Command{nil},
	}
	expect(t, checkDuplicatedCmds(&cmd), nil)
}
```
In this test, it will panic if we don't check nil pointer.

command.go
```diff
@@ -404,3 +408,24 @@ func hasCommand(commands []*Command, command *Command) bool {
...
 	return false
 }
+
+func checkDuplicatedCmds(parent *Command) error {
+	if parent == nil {
+		return nil
+	}
+	if len(parent.Subcommands) == 0 {
+		return nil
+	}
```
abitrolly, line 418: No need for this check either. `range` will skip empty list.

linrl3: sure, will delete it soon

linrl3: @abitrolly Please check again

command.go
```diff
@@ -404,3 +408,20 @@ func hasCommand(commands []*Command, command *Command) bool {
 
 	return false
 }
+
+func checkDuplicatedCmds(parent *Command) error {
+	if parent == nil {
+		return nil
+	}
```
abitrolly, line 415: Why there is a need to call the function with `nil` argument?

linrl3: @abitrolly  It's a defensive way too. Already deleted.

abitrolly:
On the surface LGTM. Need somebody to get into deeper logic, like using new `error` module for v2 branch.

linrl3:
> On the surface LGTM. Need somebody to get into deeper logic, like using new `error` module for v2 branch.

@abitrolly Who do you think should review the code?

command.go
```diff
@@ -404,3 +408,17 @@ func hasCommand(commands []*Command, command *Command) bool {
 
 	return false
 }
+
+func checkDuplicatedCmds(parent *Command) error {
+	seen := make(map[string]struct{})
+	for _, c := range parent.Subcommands {
+		for _, name := range c.Names() {
+			if _, exists := seen[name]; exists {
+				errMsg := fmt.Sprintf("parent command [%s] has duplicated subcommand name or alias: %s", parent.Name, name)
```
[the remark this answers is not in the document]
linrl3: @avorima Sure, done. Please check again.

avorima (approved):

linrl3:
> @linrl3 Thank you so much for your PR.

@dearchap Thanks! What is the next step? Is there a routine merging and release?

PR #1833, opened 2023-12-06
Feat:(issue_1074) Add basic support for cmd args

### LINKED ISSUE #1074: Args (an alternative to Flags)
## Checklist

Undocumented arguments can be a burden and forces developers to generate write their own `ArgsUsage`

## Solution description

Add a new field to `App` and `Command` called `Args` of type `[]ArgSpec` which is an interface like `Flag`. There would many structs that implement this interface, like those for parsing flags.

```go
type ArgSpec interface {
    // Name returns the name that this arg will be accessed by
    Name() string
    // Required returns true if this arg is required 
    Required() bool
    // Priority means the order (highest to lowest) which are parsed after the required ones
    Priority() uint
    // Min returns the minimum args to be parsed
    Min() uint
    // Max returns the maximum args to be parsed, if 0 then any number can be parsed 
    Max() uint
}
```

Along with this there are several other rules that will result in a panic (always no matter what the provided args to be parsed are)

- All Optional fields must either come before all required fields or after all required fields 
- At most one `Max() == 0` arg per field

Args never mean anything for sub-comands. 

Args will be accessed in the same manner as `Flags` currently, they are parsed after flags so will override them (that is how one can have a EnvVar be the default to an arg)

### Implementation 

If there are any `[]ArgSpec` then all provided args must be used, otherwise a error is printed as if a required flag was not supplied. 

This also means that the `Arg(n)` and `Args()` won't return anything after parsing args.

- Do nothing, this a hard problem to solve and this solution is complicated to use.

dearchap:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

_(REQUIRED)_

Fixes #1074 

## Testing

Added 2 new test cases in args_test.go

## Release Notes

_(REQUIRED)_

```release-note

```

### DIFF
--- a/args.go
+++ b/args.go
@@ -1,5 +1,10 @@
 package cli
 
+import (
+	"fmt"
+	"time"
+)
+
 type Args interface {
 	// Get returns the nth argument, or else a blank string
 	Get(n int) string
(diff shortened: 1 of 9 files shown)


meatballhat (approved):
Whoa! 🤯🤩

PR #1836, opened 2023-12-08
Fix:(issue_1720) Add support for reading args from stdin

### LINKED ISSUE #1720: How do you set up cliv2 for command line inputs  $ cli_command < input.txt > output.txt
I'm re-writing an **old** fortran 77 piece of code into Go. The scientists are used to using it like: 

```bash
water < input_file.txt 
```
and then having it walk them through the command parameters it needs (druid style). 

I googled and looked and can't seem to find any documentation on how to accomplush that (though assume if I make the input file the first expected argument it might work? I thought I should ask the question since I haven't seen it mentioned anywhere in the docs (or missed it totally) and I imagine other people might want to find this for the google juice it gives. 

thanks!
PS> I've already used cliv2 to write one of my long standing little cli apps, [harsh](https://github.com/wakatara/harsh) and it's been great, so just wanted to say thanks for writing a simple, easy to use framework that doesn't overcomplicate. I prefer cliv2 over all the alternatives I evaluated.

dearchap:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

_(REQUIRED)_

Fixes #1720 

## Testing

go test -run=TestCommandReadArgsFromStdIn

## Release Notes

_(REQUIRED)_

A new field 'ReadArgsFromStdin' has been added to Command. This flag will be applicable to root command only and will allow the root command to process arguments from cmdline as well as from standard input. 

```release-note

```

### DIFF
--- a/command.go
+++ b/command.go
@@ -408,9 +408,10 @@ outer:
 				if token != "" {
 					args = append(args, token)
 					token = ""
-				} else {
-					//TODO. Should we pass in empty strings ?
 				}
+				/*else {
+					//TODO. Should we pass in empty strings ?
+				}*/
 				st = STATE_SEARCH_FOR_TOKEN
 			}
 		}
--- a/command.go
+++ b/command.go
@@ -343,6 +343,61 @@ func (cmd *Command) ensureHelp() {
 	}
 }
 
+func (cmd *Command) parseArgsFromStdin() ([]string, error) {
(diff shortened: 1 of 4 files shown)


meatballhat (approved):
What do you think about supporting `--` as the "end" of reading from stdin instead of consuming all of it?

Instead of referencing `os.Stdin` directly, could it use the `Reader` field?

PR #1848, opened 2023-12-21
v2: GitHub Actions: upgrade Go, upgrade actions

dolmen:
## What type of PR is this?

- cleanup

Upgrade CI:
* Upgrade actions/checkout, actions/setup-go.
* Move setup-go after checkout as recommended in doc.
* Use symbolic names 'stable', 'oldstable' for Go versions.

## Release Notes

NONE

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -15,19 +15,19 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        go: [1.19.x, 1.20.x]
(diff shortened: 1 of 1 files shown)


dolmen:
Cc: @skelouse for review

skelouse (approved):
Neat, would be good on main as well.

https://github.com/actions/setup-go?tab=readme-ov-file#using-stableoldstable-aliases

dearchap:
Does this mean that we would be automatically getting the latest stable go version instead of pinning them down ?

PR #1864, opened 2024-02-11
Missing arg can cause a panic

joshfrench:
## What type of PR is this?

- bug

- In the special case where an arg has `max: 1`, `args.Parse` would panic when arg was omitted.
- Also ensure omitted args retain their `Value`, if provided.

```
--- FAIL: TestSingleOptionalArg (0.00s)
panic: runtime error: index out of range [0] with length 0 [recovered]
        panic: runtime error: index out of range [0] with length 0

goroutine 15 [running]:
testing.tRunner.func1.2({0x163bc20, 0xc00001c660})
        /usr/local/Cellar/go/1.21.4/libexec/src/testing/testing.go:1545 +0x366
testing.tRunner.func1()
        /usr/local/Cellar/go/1.21.4/libexec/src/testing/testing.go:1548 +0x630
panic({0x163bc20?, 0xc00001c660?})
        /usr/local/Cellar/go/1.21.4/libexec/src/runtime/panic.go:920 +0x270
github.com/urfave/cli/v3.(*ArgumentBase[...]).Parse(0x1729d40, {0xc0003381a0, 0x0?, 0x0})
        /Users/josh/src/cli/args.go:136 +0xc88
```

## Testing

Tests added for an omitted arg with `max: 1`, with and without a default `Value`.

## Release Notes

```release-note
Optional arguments with arity 1 no longer panic when omitted.
```

### DIFF
--- a/args_test.go
+++ b/args_test.go
@@ -170,4 +170,7 @@ func TestSingleOptionalArg(t *testing.T) {
 	arg.Value = "bar"
 	require.NoError(t, cmd.Run(context.Background(), []string{"foo"}))
 	require.Equal(t, "bar", s1)
+
+	require.NoError(t, cmd.Run(context.Background(), []string{"foo", "zbar"}))
+	require.Equal(t, "zbar", s1)
 }
--- a/args_test.go
+++ b/args_test.go
@@ -168,6 +168,6 @@ func TestSingleOptionalArg(t *testing.T) {
 	require.Equal(t, "", s1)
 
 	arg.Value = "bar"
-	require.NoError(t, cmd.Run(context.Background(), []string{"foo", "bar"}))
+	require.NoError(t, cmd.Run(context.Background(), []string{"foo"}))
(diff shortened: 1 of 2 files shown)


args_test.go
```diff
@@ -151,3 +151,23 @@ func TestArgsUsage(t *testing.T) {
...
+	}
+	cmd.Arguments = []Argument{
+		arg,
+	}
+
+	require.NoError(t, cmd.Run(context.Background(), []string{"foo"}))
+	require.Equal(t, "", s1)
+
+	arg.Value = "bar"
+	require.NoError(t, cmd.Run(context.Background(), []string{"foo"}))
```
dearchap, line 171: could you also add a test for 

```
[]string{"foo", "zbar"}
```

to make sure that value is picked up from args

joshfrench: all set!

dearchap (approved):
Thanks @joshfrench

PR #1865, opened 2024-02-13
Fix variadic args

joshfrench:
## What type of PR is this?
- bug

Fixes the following:
- If variadic args are passed, only the first is captured because `Parse` exits early when comparing `count >= -1`.
- If no args are passed, default values are clobbered by the empty set.

If more arguments are defined after an unbounded argument, the end user may experience an error:
```
Arguments: []cli.Argument{
    &cli.StringArg{
        Name: "First",
        Max: -1,
    },
    &cli.StringArg{
        Name: "Second,
    },
}
```
After `First` is parsed, there will be no remaining args. If `Second.Min` == 0, it will just use its default/zero value. If `Second.Min` > 0, this will result in an insufficient arg count error. Both of those behaviors seem intuitive to me, but we could also add a runtime check and proactively error if an unbounded arg is defined as anything other than the last argument.

## Testing

Unit tests updated with cases for variadic args.

## Release Notes

```release-note
Variadic args are now parsed correctly.
```

### DIFF
--- a/args.go
+++ b/args.go
@@ -128,7 +128,7 @@ func (a *ArgumentBase[T, C, VC]) Parse(s []string) ([]string, error) {
 
 	if a.Values == nil {
 		a.Values = &values
-	} else {
+	} else if count > 0 {
 		*a.Values = values
 	}
 
@@ -139,6 +139,7 @@ func (a *ArgumentBase[T, C, VC]) Parse(s []string) ([]string, error) {
 			*a.Destination = t
 		}
 	}
+
 	return s[count:], nil
 }
 
--- a/args.go
+++ b/args.go
@@ -118,7 +118,7 @@ func (a *ArgumentBase[T, C, VC]) Parse(s []string) ([]string, error) {
 		}
 		values = append(values, value.Get().(T))
 		count++
-		if count >= a.Max && a.Max != -1 {
+		if count >= a.Max && a.Max > -1 {
 			break
 		}
 	}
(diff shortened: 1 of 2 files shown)


args.go
```diff
@@ -128,7 +128,7 @@ func (a *ArgumentBase[T, C, VC]) Parse(s []string) ([]string, error) {
 
 	if a.Values == nil {
 		a.Values = &values
-	} else {
+	} else if count > 0 {
```
dearchap, line 131: shouldnt the if count be outside the if a.Values ...?

args_test.go
```diff
@@ -174,3 +174,49 @@ func TestSingleOptionalArg(t *testing.T) {
...
+		},
+	}
+
+	cmd := buildMinimalTestCommand()
+	cmd.Arguments = []Argument{arg}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			arg.Values = &test.values
+			require.NoError(t, cmd.Run(context.Background(), test.args))
+			require.Equal(t, test.expected, test.values)
```
dearchap, line 219: `require.Equal(t, test.expected, arg.Values`

dearchap (commented):
Added comments

args.go
joshfrench: My thinking was that there's only one out of four cases where we _don't_ want to mutate `a.Values`: if `a.Values` has defaults && no args were supplied to override those. So the second branch of this conditional just needs to check if any args were collected before mutating `a.Values`; otherwise we should leave the defaults unmodified.

I don't feel strongly about how the logic is implemented though, if you think there's a cleaner way to phrase this.

dearchap (approved):

args.go
dearchap: Thats fine @joshfrench . Lets leave it as it is

PR #1872, opened 2024-02-22
Fix repeated [arguments...] in usage template in v2

edeustua:
## What type of PR is this?

- bug

This PR fixes a small bug in the `usageTemplate` template, where `[arguments ...]` would
be printed even if `command.Args == false` and would be printed twice otherwise.

None yet. I can create one.

Very simple thing. No need for much.

## Testing

Visually. It is only a rendering issue.

## Release Notes`

```release-note
Fixes a repeating [arguments...] string in the usageTemplate
```

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -167,6 +167,7 @@ func ExampleApp_Run_commandHelp() {
 				Aliases:     []string{"d"},
 				Usage:       "use it to see a description",
 				Description: "This is how we describe describeit the function",
+				Args:        true,
 				Action: func(c *Context) error {
 					fmt.Printf("i like to describe things")
 					return nil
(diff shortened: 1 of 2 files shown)


Juneezee (commented):
I think this has been addressed in commit https://github.com/urfave/cli/commit/bda5fdd70f625b4b53f88a40c0b733237e2ca70a on the `main` branch already.

edeustua:
> I think this has been addressed in commit [bda5fdd](https://github.com/urfave/cli/commit/bda5fdd70f625b4b53f88a40c0b733237e2ca70a) on the `main` branch already.

This is true. However, it hasn't been fixed in v2. That's why this PR is targeting v2-maint.

edeustua:
Sure. I'll push tonight.

dearchap (approved):

edeustua:
> > So one needs to add `.local/bin` to the `PATH`. That would be a nice thing to add to `CONTRIBUTING.md` along with the actual requirements, i.e. `gfmrun` and `goimports`.
> 
> Actually a simple `make` would have sufficed. It does all the heavy lifting for you.

That actually didn't work for me. Should I investigate further?

PR #1903, opened 2024-05-01
Include flag descriptions when using zsh completion

Toalaah:
## What type of PR is this?

- feature

This PR ensures that flags' `Usage` fields are included when using zsh completion (if their description is non-empty). This functionality is in line with how this is handled for command completion, see https://github.com/urfave/cli/blob/043b774d0f2c64d2ed4e22e87c2f6ec54bfcb30b/help.go#L163

### Without Changes
without-desc

### With changes 
with-desc

## Testing

I tested these changes locally against a CLI tool that I am migrating to urfave/cli/v3. I also ran `make test` and encountered no errors, but then again I also did not add any test cases (would be open to doing so if this feature is desired).

## Release Notes

```release-note
Flag usage fields are now included when using zsh completion
```

### DIFF
--- a/help.go
+++ b/help.go
@@ -192,6 +192,11 @@ func printFlagSuggestions(lastArg string, flags []Flag, writer io.Writer) {
 			continue
 		}
 
+		usage := ""
+		if docFlag, ok := flag.(DocGenerationFlag); ok {
+			usage = docFlag.GetUsage()
+		}
+
 		name := strings.TrimSpace(flag.Names()[0])
 		// this will get total count utf8 letters in flag name
 		count := utf8.RuneCountInString(name)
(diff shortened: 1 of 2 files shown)


bartekpacia:
Hi @Toalaah! Thank you very much for the contribution :)

bartekpacia (approved):
LGTM.

bartekpacia:
@meatballhat @Juneezee What should be done when codecov fails, like in this PR?

Juneezee:
> @meatballhat @Juneezee What should be done when codecov fails, like in this PR?

@bartekpacia I think the simplest way would be adding a test case for `zsh` in `TestDefaultCompleteWithFlags`

https://github.com/urfave/cli/blob/043b774d0f2c64d2ed4e22e87c2f6ec54bfcb30b/help_test.go#L1079-L1083

Toalaah:
Thanks for the hint, added some zsh-specific test cases. Looks like codecov is no longer complaining.

bartekpacia:
Thanks a lot :-)

Issue #1902, opened 2024-05-01
Doc generation default value set by environment variable


sj14:
## Checklist

- [x] Are you running the latest v2 release? The list of releases is [here](https://github.com/urfave/cli/releases).
- [x] Did you check the manual for your release? The v2 manual is [here](https://github.com/urfave/cli/blob/main/docs/v2/manual.md)
- [x] Did you perform a search about this problem? Here's the [GitHub guide](https://help.github.com/en/github/managing-your-work-on-github/using-search-to-filter-issues-and-pull-requests) about searching.
  - Seems like a related issue: https://github.com/urfave/cli/issues/1206

## Dependency Management

- My project is using go modules.

## Describe the bug

When you have a flag which can also be set by an env var, and the flag is empty, the value from the env will be used in the output docs markdown.

## To reproduce

```go
package main

import (
	"log"
	"os"

	"github.com/urfave/cli/v2"
)

func main() {
	app := &cli.App{
		Flags: []cli.Flag{
			&cli.StringFlag{
				Name:    "url",
				Value:   "http://127.0.0.1",
				EnvVars: []string{"URL"},
			},
			&cli.StringFlag{
				Name:    "token",
				Value:   "",
				EnvVars: []string{"TOKEN"},
			},
		},
		Commands: []*cli.Command{
			{
				Name:   "docs",
				Hidden: true,
				Action: func(ctx *cli.Context) error {
					s, err := ctx.App.ToMan()
					if err != nil {
						return err
					}
					return os.WriteFile("DOCS.man", []byte(s), os.ModePerm)
				},
			},
		},
	}

	if err := app.Run(os.Args); err != nil {
		log.Fatal(err)
	}
}
```

## Observed behavior

When running the example like this 

```
TOKEN=my-token URL=my-url go run test/main.go docs
```

a file called DOCS.man is created with the follwing content:

```
.nh
.TH main 8

.SH NAME
.PP
main - A new cli application


.SH SYNOPSIS
.PP
main

.EX
[--help|-h]
[--token]=[value]
[--url]=[value]
.EE

.PP
\fBUsage\fP:

.EX
main [GLOBAL OPTIONS] command [COMMAND OPTIONS] [ARGUMENTS...]
.EE


.SH GLOBAL OPTIONS
.PP
\fB--help, -h\fP: show help

.PP
\fB--token\fP="":  (default: my-token)

.PP
\fB--url\fP="":  (default: "http://127.0.0.1")


.SH COMMANDS
[...]

PR #1910, opened 2024-05-09
v2 Docs: Mention value from env as default value

sj14:
## What type of PR is this?

_(REQUIRED)_

- documentation

_(REQUIRED)_

Mention that values from the environment can be used as default values in the generated docs when no `Value` was set for a flag. Without knowing this, it can easily lead to exposed secrets.

_(REQUIRED)_

Related to #1902 

## Release Notes

_(REQUIRED)_

```release-note
Docs: Mention that values from the environment can be used as default values in the generated docs when no Value was set for a flag.
```

### DIFF
--- a/docs/v2/examples/flags.md
+++ b/docs/v2/examples/flags.md
@@ -470,6 +470,8 @@ func main() {
 }
 ```
 
+When `Value` is not set for the flag, but a matching environment variable is found, the value from the environment will be used in the generated docs as the default value.
+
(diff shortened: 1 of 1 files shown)


Juneezee (approved):
Thanks!

ArangoGutierrez (approved):
Thanks

bartekpacia (approved):
LGTM!

PR #1919, opened 2024-06-02
Fix:(issue_1916) Add additional check for --

### LINKED ISSUE #1916: Should I expect completion of nested command's flags to work?
### Discussed in https://github.com/urfave/cli/discussions/1915

<div type='discussions-op-text'>

<sup>Originally posted by **ldelossa** May 25, 2024</sup>
Hello, 

I have a root command called 'cmd'. 
Now I have a sub command hanging off this called "sub-command".
The cli.Command for "sub-command" defines 2 flags "--a, --b".
All commands have the "EnableShellCompletion" boolean set to true. 

I'm using `zsh`. Should I expect to see completion for `cmd subcommand --<Tab>`. 
My assumption was to see completion for `--a, --b` but instead the only completion I get is for `--h, -help` essentially.</div>

dearchap:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

_(REQUIRED)_

Fixes #1916 

## Testing

## Release Notes

_(REQUIRED)_

```release-note

```

### DIFF
--- a/help.go
+++ b/help.go
@@ -237,6 +237,11 @@ func DefaultCompleteWithFlags(ctx context.Context, cmd *Command) {
 		lastArg = args[argsLen-1]
 	}
 
+	if len(cmd.runArgs) > 1 && cmd.runArgs[len(cmd.runArgs)-1] == "-" {
+		tracef("using last arg from cmd.runArgs [%v]", cmd.runArgs)
+		lastArg = "-"
+	}
+
 	if strings.HasPrefix(lastArg, "-") {
 		tracef("printing flag suggestion for flag[%v] on command %[1]q", lastArg, cmd.Name)
 		printFlagSuggestions(lastArg, cmd.Flags, cmd.Root().Writer)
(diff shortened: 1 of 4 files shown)


help.go
```diff
@@ -237,6 +237,11 @@ func DefaultCompleteWithFlags(ctx context.Context, cmd *Command) {
 		lastArg = args[argsLen-1]
 	}
 
+	if len(cmd.runArgs) > 1 && cmd.runArgs[len(cmd.runArgs)-1] == "-" {
```
meatballhat, line 240: This looks like an equality check for `"-"` but then `lastArg` is later checked for a _prefix_ of `"-"`. Is that right?

dearchap: I dont think so. This is mainly for tab completion after user hits tab after entering a '-'. Let me double check. I might need to add more tests

Juneezee (approved):
:+1:

Issue #1733, opened 2023-05-18
Ordering of arguments and flags impact one another


mntlty:
## My urfave/cli version is

_** v2.25.3**_

## Checklist

- [x] Are you running the latest v2 release? The list of releases is [here](https://github.com/urfave/cli/releases).
- [x] Did you check the manual for your release? The v2 manual is [here](https://github.com/urfave/cli/blob/main/docs/v2/manual.md)
- [x] Did you perform a search about this problem? Here's the [GitHub guide](https://help.github.com/en/github/managing-your-work-on-github/using-search-to-filter-issues-and-pull-requests) about searching.

## Dependency Management
- My project is using go modules.

## Describe the bug

Ordering of arguments and flags impact one another, causing a flag value to be set or unset depending on whether it comes before or after an argument. This can result in a silent error where the flag name is recognized but the value isn't set.

## To reproduce

Using the flag example from https://cli.urfave.org/v2/examples/flags/ (copying below for ease of reference)

```go
package main

import (
	"fmt"
	"log"
	"os"

	"github.com/urfave/cli/v2"
)

func main() {
	app := &cli.App{
		Flags: []cli.Flag{
			&cli.StringFlag{
				Name:  "lang",
				Value: "english",
				Usage: "language for the greeting",
			},
		},
		Action: func(cCtx *cli.Context) error {
			name := "Nefertiti"
			if cCtx.NArg() > 0 {
				name = cCtx.Args().Get(0)
			}
			if cCtx.String("lang") == "spanish" {
				fmt.Println("Hola", name)
			} else {
				fmt.Println("Hello", name)
			}
			return nil
		},
	}

	if err := app.Run(os.Args); err != nil {
		log.Fatal(err)
	}
}
```

## Observed behavior

```bash
# argument
$ ./example my-name
Hello my-name
# flag
$ ./example --lang spanish        
Hola Nefertiti
# flag before argument
$ ./example --lang spanish my-name 
Hola my-name
# argument before flag
$ ./example my-name --lang spanish
Hello my-name
```

## Expected behavior

`$ ./example my-name --lang spanish` and `./example --lang spanish my-name ` should both set the `--lang` flag to the value `spanish`

## Additional context

N/A

## Want to fix this yourself?

We'd love to have more contributors on this project! If the fix for
this bug is easily explained and very small, feel free to create a
pull request for it.

## Run `go version` and paste its output here

```
go version go1.20.4 darwin/arm64
```

## Run `go env` and paste its output here

```
[...]

PR #1928, opened 2024-06-26
reorder flags so they can come after arguments

### LINKED ISSUE #976: v2 feature: allow flags after arguments
## Checklist

## Solution description

* `ctx.Args() --> {"a", "b"}`

fiatjaf:
## What type of PR is this?

- feature

This brings back (and improves) the `reorderFlags` function from v1 that allows flags to be specified after arguments and still work. It's a very much requested feature, see https://github.com/urfave/cli/issues/1733 and https://github.com/urfave/cli/issues/976.

Fixes #976

## Testing

I've done some manual testing.

### DIFF
--- a/command.go
+++ b/command.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"slices"
 	"sort"
 	"strings"
 	"unicode"
@@ -1282,6 +1283,7 @@ func reorderArgs(cmd *Command, args Args) Args {
 		}
 
 		isFlag, isBooleanFlag := cmd.argIsFlag(arg)
+
 		// if we're expecting an option-value, check if this arg is a value, in
 		// which case it should be re-ordered next to its associated flag
 		if nextIndexMayContainValue && !isFlag {
@@ -1336,33 +1338,40 @@ func (cmd *Command) argIsFlag(arg string) (isFlag bool, isBooleanFlag bool) {
 		shortFlags := strings.Split(arg[1:], "")
 		for s, shortFlag := range shortFlags {
 			// look through all the flags, to see if the `arg` is one of our flags
-			for _, flag := range cmd.Flags {
-				for _, key := range flag.Names() {
-					if key == shortFlag {
-						_, isBooleanFlag = flag.(boolFlag)
-						if isBooleanFlag {
-							// this is the default, expected case
-							continue
-						} else {
-							// only the last flag may not be a boolean
-							if s == len(shortFlag)-1 {
-								// this is an arg and not a boolean arg (it expects a value right after it)
-								return true, false
-							} else {
-								// this is not a valid flag bundle for this command
-								return false, false
-							}
-						}
-					} else {
-						// this short mode doesn't correspond to any flags for this Command
-						return false, false
-					}
+			var flag Flag
+			idx := slices.IndexFunc(cmd.Flags, func(f Flag) bool { return slices.Contains(f.Names(), shortFlag) })
+			if idx != -1 {
+				flag = cmd.Flags[idx]
+			} else {
+				vpf := cmd.VisiblePersistentFlags()
+				idx := slices.IndexFunc(vpf, func(f Flag) bool { return slices.Contains(f.Names(), shortFlag) })
+				if idx != -1 {
+					flag = vpf[idx]
+				} else {
+					// this short flag doesn't correspond to any flags for this Command
+					return false, false
 				}
 			}
 
-			// in this case we didn't exit, which means all the short flags are booleans
-			return true, true
+			// got a flag match
+			_, isBooleanFlag = flag.(boolFlag)
+			if isBooleanFlag {
+				// this is the default, most common case
+				continue
+			} else {
+				// only the last flag may not be a boolean
+				if s == len(shortFlag)-1 {
+					// this is an arg and not a boolean arg (it expects a value right after it)
+					return true, false
+				} else {
+					// this is not a valid flag bundle for this command
+					return false, false
+				}
+			}
 		}
+
+		// in this case we didn't exit, which means all the short flags are booleans
+		return true, true
 	} else {
 		// this line turns `-flag` into `flag`
 		arg, _ = strings.CutPrefix(arg, "-")
--- a/command.go
+++ b/command.go
@@ -110,8 +110,9 @@ type Command struct {
 	// i.e. foobar -o -v -> foobar -ov
 	UseShortOptionHandling bool `json:"useShortOptionHandling"`
 	// Boolean to enable reordering flags before passing them to the parser
-	// such that `cli -f <val> <arg>` behaves the same as `cli <arg> -f <val>`
-	AllowFlagsAfterArguments bool `json:"AllowFlagsAfterArguments"`
+	// such that `cli -f <val> <arg>` behaves the same as `cli <arg> -f <val>`.
+	// This only has effect when set at the top-level command.
+	AllowFlagsAfterArguments bool `json:"allowFlagsAfterArguments"`
 	// Enable suggestions for commands and flags
 	Suggest bool `json:"suggest"`
 	// Allows global flags set by libraries which use flag.XXXVar(...) directly
@@ -1273,7 +1274,6 @@ func reorderArgs(cmd *Command, args Args) Args {
 
 	nextIndexMayContainValue := false
 	for i, arg := range tail {
-
 		subCmd := cmd.Command(arg)
 		if subCmd != nil {
 			cmd = subCmd
@@ -1281,9 +1281,10 @@ func reorderArgs(cmd *Command, args Args) Args {
 			continue
 		}
 
+		isFlag, isBooleanFlag := cmd.argIsFlag(arg)
 		// if we're expecting an option-value, check if this arg is a value, in
 		// which case it should be re-ordered next to its associated flag
-		if isFlag, _ := argIsFlag(cmd.Flags, arg); nextIndexMayContainValue && !isFlag {
+		if nextIndexMayContainValue && !isFlag {
 			nextIndexMayContainValue = false
 			reorderedArgs = append(reorderedArgs, arg)
 		} else if arg == "--" {
@@ -1299,15 +1300,14 @@ func reorderArgs(cmd *Command, args Args) Args {
 			remainingArgs = append(remainingArgs, tail[i+1:]...)
 			break
 			// checks if this is an arg that should be re-ordered
-		} else if isFlag, isBooleanFlag := argIsFlag(cmd.Flags, arg); isFlag {
+		} else if isFlag {
 			// we have determined that this is a flag that we should re-order
 			reorderedArgs = append(reorderedArgs, arg)
 
 			// if this arg does not contain a "=", then the next index may contain the value for this flag
 			nextIndexMayContainValue = !strings.Contains(arg, "=") && !isBooleanFlag
-
-			// simply append any remaining args
 		} else {
+			// simply append any remaining args
 			remainingArgs = append(remainingArgs, arg)
 		}
 	}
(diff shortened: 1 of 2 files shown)


skelouse:
I did something similar in another repo if you need this: https://github.com/urfave/cli/issues/1583#issuecomment-1385733549

linrl3:
Some Run tests are failed, please check @fiatjaf

fiatjaf:
I don't understand what is failing. Looks like it's test coverage only? I can fix that, but can you tell me first if this change is wanted and would be merged if tests were ok?

dearchap:
@fiatjaf No one has specifically asked for this feature, so it depends.

fiatjaf:
I've linked to two issues above with many people asking for it.

joshfrench:
This trips up my team all the time — I’d definitely appreciate the feature!

dearchap:
@fiatjaf This starts getting complicated. Can you provide more examples of the behaviour ? if you have say

```
cli subcmd1 subcmd2 -f -s -t
```

that would be equivalent to 

```
cli -f -s -t subcmd1 subcmd2
```

even if say '-s -t' are defined for subcmd1 and not root command.

fiatjaf:
Subcommands should not be involved here. So `cli subcmd -s -t -m file.txt otherfile.txt` should be equivalent to `cli subcmd file.txt otherfile.txt -s -t -m`.

dearchap:
I understand the appeal of this but you are interpreting what the user want to achieve. Its is entirely possible that the subcmd wants to handle all the args itself in the command action. 

```
cli subcmd file.txt s.txt -f -s
```

So you need a new boolean field in Command, something like ReorderArgs, which is off by default and can be enabled if a particular user of this library wishes it. Also you have to ensure that the reorder can be applied only to a leaf command and not to root/branch commands which have subcommands.

decentral1se:
> No one has specifically asked for this feature

*checks notes* one of the most critically requested features and reason for several projects migrating back to v1 and/or dropping `urfave/cli` altogether in favour of `cobra` 😆

abitrolly:
> _checks notes_ one of the most critically requested features and reason for several projects migrating back to v1 and/or dropping `urfave/cli` altogether in favour of `cobra` 😆

`[reference needed]`. I don't see "reorder" in the critical list of requested features.

https://github.com/urfave/cli/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

Alzo the v1 migration dataset is very interesting to see.

decentral1se:
@abitrolly you don't have to look far https://github.com/urfave/cli/issues/1113 There are *several* folks (including myself) who had to remigrate to v1. Additionally, several folks there *and in other issue* migrating to cobra because of this. I don't know who curates [this list](https://github.com/urfave/cli/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc) but this ordering issue should be on it imho

fiatjaf:
I've opened a feature request here: https://github.com/urfave/cli/issues/1950

Let's go there an upvote it!

XANi:
Can confirm, can't move to v2 in any of our project for ergonomic issues. "It's more POSIX compliant" is a very weak arguments, humans are not POSIX compliant.

We have a lot of pattern that can be described as 

`<cmd> --global --config <subcmd> --subcmd-config`

because for CLI user it is easier to just copy front part (`<cmd> --global --config `) or outright alias it for shorter cmdnline and then change subcommand and its options when using it.

Forcing flags to be at certain position just means unnecesary cursor movement

decentral1se:
I've done manual some testing @fiatjaf and it doesn't seem to be working for flags on sub-commands?

```
./<cli> <command> --help # (1)
./<cli> <command> <sub-command> --help # (2)
```

Where both `(1)`/`(2)` show help output for `<command>` *only*. I can't seem to trigger help output for `<sub-command>` at all.

fiatjaf:
That's true. It's broken indeed. I'll try to fix it soon.

dearchap:
How many utilities you see provide 3-4 level subcmds ? This library has to support that use case so we cannot make choices assuming a certain depth. urfave/cli is more akin to a platform than a cli library. It allows lots of configurations so that people can achieve what they want from a cli library. Just because something is good for one user doesnt mean that others want it. Any changes have to be backward compatible(within major versions) and if it is an optional feature it needs to turned off by default. Interested users should be able to enable it as needed. I have thought about this in the past and felt there needs to be a "personality" setting for the library which configures a set of defaults automatically to provide a given behavior. As an example we would provide say 'ls' personality or 'git' personality and the cli would behave similar to what those utilities do in terms of flags/subcmds in terms of how options are parsed(reorder args) and help output. Additionally users would be able to plugin their own personality engine which would define behaviors expected of the platform.

fiatjaf:
I made it work for subcommands -- now it seems to be working fine in all my manual tests -- and made the behavior optional.

But some test in the suite and I'm having trouble understanding which. When I run `go test` I get a thousand lines of output from command calls. Please help me!

command.go
```diff
@@ -1255,3 +1263,88 @@ func makeFlagNameVisitor(names *[]string) func(*flag.Flag) {
...
+		} else {
+			remainingArgs = append(remainingArgs, arg)
+		}
+	}
+
+	return &stringSliceArgs{append([]string{args.First()}, append(reorderedArgs, remainingArgs...)...)}
+}
+
+// argIsFlag checks if an arg is one of our command flags
+func argIsFlag(commandFlags []Flag, arg string) (isFlag bool, isBooleanFlag bool) {
```
dearchap, line 1319:
```suggestion
func *c *Command) argIsFlag(arg string) (isFlag bool, isBooleanFlag bool) {
```

command.go
```diff
@@ -1255,3 +1263,88 @@ func makeFlagNameVisitor(names *[]string) func(*flag.Flag) {
...
+		// `-` is never a flag
+		// `--` is an option-value when following a flag, and a delimiter indicating the end of options in other cases.
+		return false, false
+	}
+	// flags always start with a -
+	if !strings.HasPrefix(arg, "-") {
+		return false, false
+	}
+	// this line turns `--flag` into `flag`
+	if strings.HasPrefix(arg, "--") {
```
dearchap, line 1330:
```suggestion
	arg, _ = strings.CutPrefix(arg, "--") 
```

command.go
```diff
@@ -1255,3 +1263,88 @@ func makeFlagNameVisitor(names *[]string) func(*flag.Flag) {
...
+		arg = strings.Replace(arg, "-", "", 2)
+	}
+	// this line turns `-flag` into `flag`
+	if strings.HasPrefix(arg, "-") {
```
dearchap, line 1334:
```suggestion
	arg, _ = strings.CutPrefix(arg, "-")
```

fiatjaf: I'm curious why, is this more efficient?

command.go
```diff
@@ -1255,3 +1263,88 @@ func makeFlagNameVisitor(names *[]string) func(*flag.Flag) {
...
+		arg = strings.Replace(arg, "-", "", 1)
+	}
+	// this line turns `flag=value` into `flag`
+	arg = strings.Split(arg, "=")[0]
+	// look through all the flags, to see if the `arg` is one of our flags
+	for _, flag := range commandFlags {
```
dearchap, line 1340: You have to also think about short options handling here. What happens if the user gives multiple options like -fhsk where each of the f, h, s & k are bool flags

command.go
```diff
@@ -109,6 +109,9 @@ type Command struct {
 	// single-character bool arguments into one
 	// i.e. foobar -o -v -> foobar -ov
 	UseShortOptionHandling bool `json:"useShortOptionHandling"`
+	// Boolean to enable reordering flags before passing them to the parser
+	// such that `cli -f <val> <arg>` behaves the same as `cli <arg> -f <val>`
+	AllowFlagsAfterArguments bool `json:"AllowFlagsAfterArguments"`
```
dearchap, line 114:
```suggestion
	AllowFlagsAfterArguments bool `json:"allowFlagsAfterArguments"`
```

command.go
```diff
@@ -1255,3 +1263,88 @@ func makeFlagNameVisitor(names *[]string) func(*flag.Flag) {
...
+			// >   The first -- argument that is not an option-argument should be accepted
+			// >   as a delimiter indicating the end of options. Any following arguments
+			// >   should be treated as operands, even if they begin with the '-' character.
+
+			// make sure the "--" delimiter itself is at the start
+			remainingArgs = append([]string{"--"}, remainingArgs...)
+			remainingArgs = append(remainingArgs, tail[i+1:]...)
+			break
+			// checks if this is an arg that should be re-ordered
+		} else if isFlag, isBooleanFlag := argIsFlag(cmd.Flags, arg); isFlag {
```
dearchap, line 1302: move this call out to the top and use the return values if all the if/else instead of calling argIsFlag again

command.go
```diff
@@ -1255,3 +1263,88 @@ func makeFlagNameVisitor(names *[]string) func(*flag.Flag) {
...
+			// we have determined that this is a flag that we should re-order
+			reorderedArgs = append(reorderedArgs, arg)
+
+			// if this arg does not contain a "=", then the next index may contain the value for this flag
+			nextIndexMayContainValue = !strings.Contains(arg, "=") && !isBooleanFlag
```
dearchap, line 1307: if for some reason the value matches the name of a command/subcommand then the logic breaks down right ?

command.go
dearchap: you dont need the if statement then. if string doesnt start with "-" it becomes a no-op

dearchap:
@fiatjaf Can you add some test cases for this first ? If you want to test locally you can do 

```
go test -v -failfast
```

that should give you test which failed immediately

command.go
```diff
@@ -1255,3 +1263,88 @@ func makeFlagNameVisitor(names *[]string) func(*flag.Flag) {
...
+	if strings.HasPrefix(arg, "-") {
+		arg = strings.Replace(arg, "-", "", 1)
+	}
+	// this line turns `flag=value` into `flag`
+	arg = strings.Split(arg, "=")[0]
+	// look through all the flags, to see if the `arg` is one of our flags
+	for _, flag := range commandFlags {
+		for _, key := range flag.Names() {
+			if key == arg {
+				_, isBooleanFlag = flag.(*BoolFlag)
```
dearchap, line 1343: use the internal boolFlag interface to check if flag is a bool flag. 

https://github.com/urfave/cli/blob/main/flag_impl.go#L17C6-L17C14

command.go
fiatjaf: I don't think so, if `git` has a subcommand `remote` but someone calls it with `git -C remote` then the correct interpretation should be that `remote` is the `<path>` value given to `-C`, it's a user error.

fiatjaf:
Working on tests now.

dearchap:
@fiatjaf One more thing is to check the behavior when ShellCompletion is enabled. Pressing TAB and having the reorder, now sure how that might work.

decentral1se:
One thing I ran into on manual testing latest changes. One example to compare from another tool. You can do `hugo -h` and `hugo -h serve` and still get the same main command help output. In the current changes, the `-h` would get reordered to `hugo serve -h` which I think will cause confusion. I think this might relate to the discussion in https://github.com/urfave/cli/issues/1950 where reordering should be "local first"?

fiatjaf:
OK, I'm giving up.

dearchap:
@fiatjaf if you'd like us to add this feature please keep the branch open. We can try picking it up later . As you can see it's not that trivial to implement . For a particular use case it's easy but to have it work for all users for different use cases is extremely difficult. That's something we've always struggled with

abitrolly:
We can add "Unsolvable challenges" chapter with examples and clarification to documentation. That will be useful for people from all CLI frameworks to tackle this problem without spending extra too much time on researching these tangled bits.

fiatjaf:
This "unsolvable challenge" is implemented in every other CLI framework, it's the number 1 requested feature according to your own link above, and this PR implements it perfectly as far as I know (except for the autocomplete thing which I have no idea of how it works, never managed to get it working on my apps).

abitrolly:
@fiatjaf hey, I am not dismissive. Now that you've created the issue, the link https://github.com/urfave/cli/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc correctly lists it as the most upvoted feature (with my upvote included), but before that there was no data to back up you claim. )

I can confirm that in Python `argparse` flags can come after arguments by default.

```py
$ cat postarg.py
import argparse

parser = argparse.ArgumentParser()

parser.add_argument('filename')           # positional argument
parser.add_argument('-c', '--count')      # option that takes a value
parser.add_argument('-v', '--verbose',
                    action='store_true')  # on/off flag

args = parser.parse_args()
print(args.filename, args.count, args.verbose)
```

```sh
$ python postarg.py -v filename -c 4
filename 4 True
$ python postarg.py -c4 -v filename 
filename 4 True 
```

I personally ~~hate~~ don't like that `go` binary itself has very awful command help interface.

```go
$ go mod -h
go mod: unknown command
Run 'go help mod' for usage. 
```

It could be much better if `go` supported `-h` option, which could be stuffed at the end.

ellistarn:
👍 for this feature!

I decided to try out urfave/cli, as it had some nice idiomatic patterns compared to cobra. The lack of this feature is a deciding factor that will keep me using cobra. I will always favor the ergonomics of my customers rather than my own ergonomics as a developer.

dearchap:
@fiatjaf I've implemented this feature in #1977 so as to keep backward compatibility and work with autocomplete. I am closing this. Can you review that PR ? Much appreciated

dearchap:
@fiatjaf I have a potential fix at github.com:dearchap/cli/pos_args_flags_mix . Want to try it out  ?

fiatjaf:
Seems to work pretty well for me now!

PR #1935, opened 2024-06-28
Fix:(issue_1925) Dont print default text for required flags

### LINKED ISSUE #1925: Allow omission of default text in help text
## Checklist

* [x] Did you check the manual for your release? The v3 manual is [here](https://github.com/urfave/cli/blob/main/docs/v3/manual.md). -- This link is a 404

In 3.0.0-alpha9, default text is always printed, irrespective of whether the field is required, etc.

This means that required numeric flags have help text that prints a default value of 0, for example.  The current API supports bringing your own DefaultText, but if you set it to "", it doesn't get applied. See relevant code [here](https://github.com/urfave/cli/blob/main/flag.go#L348-L354) and [here](https://github.com/urfave/cli/blob/1210bdf785b2bddda680a262acc84d5e7b583460/flag_impl.go#L245-L251).

Arguably this is a bug for implicit zero-valued default values. Certainly this is a bug for required flags with an implicit zero-valued default. Regardless of its status as a bug or expected behavior, it would be nice to be able to opt out of showing default text altogether on a flag-by-flag basis.

## Solution description

I'd like for an `OmitDefaultText` boolean field to be added to flag base, or a similarly turn-key solution.

I think you could also make the `DefaultText` field a `*string`, and use the empty string to indicate that default text shouldn't be printed, and I think that would also be fine. I could also imagine at the flag level being able to specify the help template for that flag only, which would be slightly more cumbersome.

dearchap:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

_(REQUIRED)_

Fixes #1925 

## Testing

go test -run=ExampleCommand_Run_appHelp

## Release Notes

_(REQUIRED)_

```release-note
Required flags will not have default text printed. 
```

### DIFF
--- a/flag.go
+++ b/flag.go
@@ -347,8 +347,11 @@ func stringifyFlag(f Flag) string {
 
 	defaultValueString := ""
 
-	if s := df.GetDefaultText(); s != "" {
-		defaultValueString = fmt.Sprintf(formatDefault("%s"), s)
+	// dont print default text for required flags
+	if rf, ok := f.(RequiredFlag); !ok || !rf.IsRequired() {
+		if s := df.GetDefaultText(); s != "" {
+			defaultValueString = fmt.Sprintf(formatDefault("%s"), s)
+		}
 	}
 
 	usageWithDefault := strings.TrimSpace(usage + defaultValueString)
--- a/flag.go
+++ b/flag.go
@@ -347,7 +347,7 @@ func stringifyFlag(f Flag) string {
 
 	defaultValueString := ""
 
-	// dont print default text for required flags
+	// don't print default text for required flags
 	if rf, ok := f.(RequiredFlag); !ok || !rf.IsRequired() {
 		if s := df.GetDefaultText(); s != "" {
 			defaultValueString = fmt.Sprintf(formatDefault("%s"), s)
--- a/examples_test.go
+++ b/examples_test.go
@@ -97,7 +97,6 @@ func ExampleCommand_Run_appHelp() {
 		},
 		Flags: []cli.Flag{
 			&cli.StringFlag{Name: "name", Value: "bob", Usage: "a name to say"},
-			&cli.StringFlag{Name: "id", Value: "abcd-dndndnd", Usage: "an id", Required: true},
 		},
 		Commands: []*cli.Command{
 			{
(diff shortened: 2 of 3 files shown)


flag.go
```diff
@@ -347,8 +347,11 @@ func stringifyFlag(f Flag) string {
 
 	defaultValueString := ""
 
-	if s := df.GetDefaultText(); s != "" {
-		defaultValueString = fmt.Sprintf(formatDefault("%s"), s)
+	// dont print default text for required flags
```
ccoVeille, line 350:
```suggestion
	// don't print default text for required flags
```

or


```suggestion
	// do not print default text for required flags
```

flag.go
```diff
@@ -347,8 +347,11 @@ func stringifyFlag(f Flag) string {
...
+	if rf, ok := f.(RequiredFlag); !ok || !rf.IsRequired() {
```
ccoVeille, line 351: Maybe this test could be moved in a function of `f`

```suggestion
	if rf, ok := f.(RequiredFlag); !ok || !rf.IsRequired() {
```

flag.go
```diff
@@ -347,8 +347,11 @@ func stringifyFlag(f Flag) string {
...
+		if s := df.GetDefaultText(); s != "" {
```
ccoVeille, line 352: Maybe this block could be moved in a method of `df`

abitrolly:
The requied value for  `required` flag doesn't make sense. Why mask it?

dearchap:
> The default value for `required` flag doesn't make sense. Why mask it?

Thats exactly why we need to mask it right ?

flag.go
dearchap: I dont understand. Can you elaborate ?

meatballhat (approved):

abitrolly:
> > The default value for `required` flag doesn't make sense. Why mask it?
> 
> Thats exactly why we need to mask it right ?

Maybe fail with an error then?

dearchap:
> > > The default value for `required` flag doesn't make sense. Why mask it?
> > 
> > 
> > Thats exactly why we need to mask it right ?
> 
> Maybe fail with an error then?



> > > The default value for `required` flag doesn't make sense. Why mask it?
> > 
> > 
> > Thats exactly why we need to mask it right ?
> 
> Maybe fail with an error then?

This is to do with display and not the behavior of the flag itself.

PR #1941, opened 2024-06-30
Fix:(issue_1930) Fix for invalid bool counts

### LINKED ISSUE #1929: What is the purpose of `normalizeFlags`?
I read the code but my understanding of it doesn't match what happens when I remove that function entirely.

It looks like the only things it does in practice is to disallow using the long form and the short form alias of the same flag in a same call.

Is there a reason why that should be disallowed? Or is that a necessity due to some constraint I am not seeing?

### LINKED ISSUE #1930: Bug in v3: Count is off by 1
## Checklist

* [ x] Are you running the latest v3 release? The list of releases is [here](https://github.com/urfave/cli/releases).
* [ x] Did you check the manual for your release? The v3 manual is [here](https://github.com/urfave/cli/blob/main/docs/v3/manual.md).
* [ x] Did you perform a search about this feature? Here's the [GitHub guide](https://help.github.com/en/github/managing-your-work-on-github/using-search-to-filter-issues-and-pull-requests) about searching.

## What problem did you observe?

When using cmd.Count("verbose") on my bool flag, it's value is always one too high

## Test Program

```go
package main

import (
	"context"
	"fmt"
	"os"

	"github.com/urfave/cli/v3"
)

func main() {
	cmd := &cli.Command{
		Name:                   "test",
		Usage:                  "Tool for testing",
		HideHelp:               true,
		UseShortOptionHandling: true,
		Flags: []cli.Flag{
			&cli.BoolFlag{
				Name:    "verbose",
				Aliases: []string{"v"},
				Usage:   "Be verbose",
			},
		},
		Action: tester,
	}

	if err := cmd.Run(context.Background(), []string{"prog", "-v"}); err != nil {
		fmt.Fprintln(os.Stderr, err)
		os.Exit(1)
	}
}


func tester(_ context.Context, cmd *cli.Command) (err error) {
	fmt.Printf(`verbosecount: %d`, cmd.Count("verbose"))
	return
}
```

Also on playground: https://play.golang.com/p/om6L1TD_6s5

dearchap:
## What type of PR is this?

_(REQUIRED)_

- bug
- cleanup

_(REQUIRED)_

_(REQUIRED)_

Fixes #1930 
Fixes #1929 

## Testing

go test -run=TestCommand_Value
go test -run=TestBoolFlagCountFromCommand

## Release Notes

_(REQUIRED)_

```release-note
The normalizeFlags function is redundant and causes the value.Set to be triggered for every alias which in turn increments count for flag unnecessarily. This function is a holdover from very very old code and has been removed. This fixes the count issue and also allows multiple aliases of same flag to be used without any errors. 
```

### DIFF
--- a/command_test.go
+++ b/command_test.go
@@ -3204,16 +3204,55 @@ func TestCommand_Bool(t *testing.T) {
 }
 
 func TestCommand_Value(t *testing.T) {
-	set := flag.NewFlagSet("test", 0)
-	set.Int("myflag", 12, "doc")
-	parentSet := flag.NewFlagSet("test", 0)
-	parentSet.Int("top-flag", 13, "doc")
-	pCmd := &Command{flagSet: parentSet}
-	cmd := &Command{flagSet: set, parent: pCmd}
+	subCmd := &Command{
+		Name: "test",
+		Flags: []Flag{
+			&IntFlag{
+				Name:    "myflag",
+				Usage:   "doc",
+				Aliases: []string{"m", "mf"},
+			},
+		},
+		Action: func(ctx context.Context, c *Command) error {
+			return nil
+		},
+	}
+
+	cmd := &Command{
+		Flags: []Flag{
+			&IntFlag{
+				Name:    "top-flag",
+				Usage:   "doc",
+				Aliases: []string{"t", "tf"},
+			},
+		},
+		Commands: []*Command{
+			subCmd,
+		},
+	}
+	err := cmd.Run(buildTestContext(t), []string{"main", "--top-flag", "13", "test", "--myflag", "14"})
+	assert.NoError(t, err)
 
 	r := require.New(t)
-	r.Equal(12, cmd.Value("myflag"))
-	r.Equal(13, cmd.Value("top-flag"))
+	r.Equal(int64(13), cmd.Value("top-flag"))
+	r.Equal(int64(13), cmd.Value("t"))
+	r.Equal(int64(13), cmd.Value("tf"))
+
+	r.Equal(int64(14), subCmd.Value("myflag"))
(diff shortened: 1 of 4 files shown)


Skeeve:
Are there any plans as to when alpha10 will be released?

command_test.go
```diff
@@ -3204,16 +3204,55 @@ func TestCommand_Bool(t *testing.T) {
...
+	err = cmd.Run(buildTestContext(t), []string{"main", "-tf", "15", "test", "-m", "16"})
+	assert.NoError(t, err)
+
+	r.Equal(int64(15), cmd.Value("top-flag"))
+	r.Equal(int64(15), cmd.Value("t"))
+	r.Equal(int64(15), cmd.Value("tf"))
+
+	r.Equal(int64(16), subCmd.Value("myflag"))
+	r.Equal(int64(16), subCmd.Value("m"))
+	r.Equal(int64(16), subCmd.Value("mf"))
```
Juneezee, line 3255: Just a small suggestion. I think we can make the test more self-documenting by using subtests.

```go
    t.Run("flag name", func(t *testing.T) {
	err := cmd.Run(buildTestContext(t), []string{"main", "--top-flag", "13", "test", "--myflag", "14"})
	assert.NoError(t, err)
	r := require.New(t)
	...
    })

    t.Run("flag alias", func(t *testing.T) {
	err := cmd.Run(buildTestContext(t), []string{"main", "-tf", "15", "test", "-m", "16"})
	assert.NoError(t, err)
	r := require.New(t)
	...
    })
```

dearchap: @Juneezee Thats an excellent idea. I've made the change

Juneezee (approved):

Issue #1942, opened 2024-07-01
BoolFlag Count and/or compound not working 


native0090:
## My urfave/cli version is

_**v2.27.2**_

## Dependency Management

- My project is using go modules.

## Describe the bug

The v2 documentation states:

> For bool flags you can specify the flag multiple times to get a count(e.g -v -v -v or -vvv)

It seems that you can specify the flag multiple times with `-v -v -v` but **not** with `-vvv`.

## To reproduce

Minimal example to reproduce:
```
package main

import (
	"fmt"
	"log"
	"os"

	"github.com/urfave/cli/v2"
)

func main() {
	var count int

	app := &cli.App{
		Flags: []cli.Flag{
			&cli.BoolFlag{
				Name:    "verbosity",
				Aliases: []string{"v"},
				Count:   &count,
			},
		},
		Action: func(cCtx *cli.Context) error {
			fmt.Println("count", count)
			return nil
		},
	}

	if err := app.Run(os.Args); err != nil {
		log.Fatal(err)
	}
}

``` 

## Observed behavior

Running this program without parameters works as expected.
Running with `-v` works as expected.
Running with `-v -v` works as expected.
Running with `-vv` does **not** work as expected and prints following message:

```
2024/07/01 12:17:32 flag provided but not defined: -vv
```

## Run `go version` and paste its output here

```
go version go1.22.4 linux/amd64
```

dearchap:
@native0090 You need to set App.UseShortOptionHandling = true for this to work

https://go.dev/play/p/8o9KPIu8a8x

native0090:
Thank you, that works!
Maybe a hint to this variable in the [Flags section](https://cli.urfave.org/v2/examples/flags/) would be useful.

dearchap:
@native0090 can you submit a PR with doc changes ?

PR #1953, opened 2024-07-11
docs(flag): add `UseShortOptionHandling` description

BlackHole1:
## What type of PR is this?

- documentation

Add a description of `UseShortOptionHandling` in the Flag section so that users can clearly understand that to support Flags like `-vvv`, it is necessary to set `UseShortOptionHandling`.

Close: #1942

nil

## Testing

nil

## Release Notes

```release-note
NONE
```

### DIFF
--- a/docs/v2/examples/flags.md
+++ b/docs/v2/examples/flags.md
@@ -107,8 +107,8 @@ For bool flags you can specify the flag multiple times to get a count(e.g -v -v 
 > If you want to support the `-vvv` flag, you need to set `App.UseShortOptionHandling`.
 
 <!-- {
-  "args": ["&#45;&#45;foo", "&#45;&#45;foo"],
-  "output": "count 2"
+  "args": ["&#45;&#45;foo", "&#45;&#45;foo", "&#45;fff",  "&#45;f"],
+  "output": "count 6"  
 } -->
 ```go
 package main
--- a/docs/v2/examples/flags.md
+++ b/docs/v2/examples/flags.md
@@ -104,6 +104,8 @@ See full list of flags at https://pkg.go.dev/github.com/urfave/cli/v2
 
 For bool flags you can specify the flag multiple times to get a count(e.g -v -v -v or -vvv)
 
+> If you want to support the `-vvv` flag, you need to set `App.UseShortOptionHandling`.
+
 <!-- {
   "args": ["&#45;&#45;foo", "&#45;&#45;foo"],
   "output": "count 2"
@@ -123,10 +125,12 @@ func main() {
 	var count int
 
 	app := &cli.App{
+		UseShortOptionHandling: true,
 		Flags: []cli.Flag{
 			&cli.BoolFlag{
 				Name:        "foo",
 				Usage:       "foo greeting",
+				Aliases:     []string{"f"},
 				Count: &count,
 			},
 		},
--- a/docs/v3/examples/flags.md
+++ b/docs/v3/examples/flags.md
@@ -106,6 +106,8 @@ See full list of flags at https://pkg.go.dev/github.com/urfave/cli/v3
 
 For bool flags you can specify the flag multiple times to get a count(e.g -v -v -v or -vvv)
 
+> If you want to support the `-vvv` flag, you need to set `Command.UseShortOptionHandling`.
+
 <!-- {
   "args": ["&#45;&#45;foo", "&#45;&#45;foo"],
   "output": "count 2"
@@ -126,10 +128,12 @@ func main() {
 	var count int
 
 	cmd := &cli.Command{
+		UseShortOptionHandling: true,
 		Flags: []cli.Flag{
 			&cli.BoolFlag{
 				Name:        "foo",
 				Usage:       "foo greeting",
+				Aliases:     []string{"f"},
 				Config: cli.BoolConfig{
 					Count: &count,
 				},
--- a/docs/v3/examples/flags.md
+++ b/docs/v3/examples/flags.md
@@ -109,8 +109,8 @@ For bool flags you can specify the flag multiple times to get a count(e.g -v -v 
 > If you want to support the `-vvv` flag, you need to set `Command.UseShortOptionHandling`.
 
 <!-- {
-  "args": ["&#45;&#45;foo", "&#45;&#45;foo"],
-  "output": "count 2"
+  "args": ["&#45;&#45;foo", "&#45;&#45;foo", "&#45;fff"],
+  "output": "count 5"
 } -->
 ```go
 package main


Juneezee (commented):
LGTM. However, the documentation website https://cli.urfave.org/ is still built using the `docs` folder of `v2-maint` branch so this change won't be reflected on the website. 

@dearchap @meatballhat Should we consider moving the `publish` job from `v2-maint` branch to `main` branch?

https://github.com/urfave/cli/blob/a4832fd31def973ae9fb5aedba571db952463ceb/.github/workflows/cli.yml#L71-L88

abitrolly:
@junezee you mean this one?

https://github.com/urfave/cli/blob/937cfe918cb10bd15aeeab4519ef2d6058a296ea/.github/workflows/test.yml#L89-L96

abitrolly:
I would say, docs publishing needs a separate job, which will pull markup files from all branches. This way `v2` docs will be maintained in `v2-maint` branch.

BlackHole1:
> Should we consider moving the publish job from v2-maint branch to main branch?

This may not be a good idea because the current main branch contains the documentation for v3, which is still in the alpha stage.

I think we can merge this PR first, as the current changes include `docs/v3`. I can submit another PR and point it to the `v2-maint` branch.

docs/v3/examples/flags.md
```diff
@@ -106,6 +106,8 @@ See full list of flags at https://pkg.go.dev/github.com/urfave/cli/v3
 
 For bool flags you can specify the flag multiple times to get a count(e.g -v -v -v or -vvv)
 
+> If you want to support the `-vvv` flag, you need to set `Command.UseShortOptionHandling`.
```
bartekpacia, line 109: I would try to rephrase it. Currently it sounds like only _the_ `-vvv` flag will be supported if you set `UseShortOptionHandling` to true.

yassinebenaid: Suggestion:

To enable the parsing of combined flags (e.g., `-vvv` as `-v -v -v` or `-xfc` as `-x -f -c` ...etc.), you need to set `Command.UseShortOptionHandling`.

abitrolly: Is it too late to rename it to `Command.ShortOptionsJoined`

bartekpacia: I think not, it's still alpha, it's the right time to make such changes.

But I'd rename it to `JoinShortOption`. Even shorter:P

bartekpacia (approved):
Thanks for contributing! 🎉

yassinebenaid (approved):

docs/v3/examples/flags.md
```diff
@@ -126,6 +128,7 @@ func main() {
 	var count int
 
 	cmd := &cli.Command{
+		UseShortOptionHandling: true,
```
abitrolly, line 131: There are no short options here, so this option looks useless.

dearchap (approved):

docs/v3/examples/flags.md
BlackHole1: Hi @abitrolly. I didn't understand what you meant, are you saying that `UseShortOptionHandling` does not exist in v3? I just checked, it does exist. Ref: https://github.com/urfave/cli/blob/937cfe918cb10bd15aeeab4519ef2d6058a296ea/command.go#L108-L111

abitrolly: @BlackHole1 I mean there are no short options in this example.

BlackHole1: I re-examined the modified context, and I confirm that this example does involve the `-vvv` scenario.

21-12 11 21@2x

abitrolly: So how to use the short option to increase the count?

BlackHole1: I still don't understand the specific issue you're trying to address. Could you please describe it again?

The current changes in the PR are all in the `flags.md` of `v2` and `v3`, and the contextual descriptions have hardly changed. The only difference is that one uses `Flag.Count` and the other uses `Config.Count`, but the textual description is almost identical.
From your description, I get the impression that you mean: "The changes in v2 are fine, but there is a problem with the changes in v3."

The reason this PR changes here is that someone submitted an issue: https://github.com/urfave/cli/issues/1942, in which they pointed out that the current description of Flags in the documentation (which is where this PR is making changes) is misleading, causing them to be unable to run it.

Juneezee: @BlackHole1 I think @abitrolly was saying that the `cli.BoolFlag` in the example doesn't have e.g. `Aliases: []string{"f"}` that gives the `foo` flag a short option.

https://github.com/urfave/cli/blob/937cfe918cb10bd15aeeab4519ef2d6058a296ea/docs/v2/examples/flags.md?plain=1#L127-L131

https://github.com/urfave/cli/blob/937cfe918cb10bd15aeeab4519ef2d6058a296ea/docs/v3/examples/flags.md?plain=1#L130-L136

I created an example here: https://go.dev/play/p/8nVATQqJz2F

Without `Aliases: []string{"f"}`, we would see
```
GLOBAL OPTIONS:
   --foo       foo greeting (default: false)
```

With `Aliases: []string{"f"}`,
```
GLOBAL OPTIONS:
   --foo, -f   foo greeting (default: false)
```

abitrolly: > @BlackHole1 I think @abitrolly was saying that the `cli.BoolFlag` in the example doesn't have e.g. `Aliases: []string{"f"}` that gives the `foo` flag a short option.

This. @Juneezee thanks.

BlackHole1: Thanks! I understand.

BlackHole1: Fixed

docs/v3/examples/flags.md
```diff
@@ -106,6 +106,8 @@ See full list of flags at https://pkg.go.dev/github.com/urfave/cli/v3
 
 For bool flags you can specify the flag multiple times to get a count(e.g -v -v -v or -vvv)
 
+> If you want to support the `-vvv` flag, you need to set `Command.UseShortOptionHandling`.
+
 <!-- {
   "args": ["&#45;&#45;foo", "&#45;&#45;foo"],
   "output": "count 2"
```
abitrolly, line 113:
```suggestion
  "args": ["&#45;&#45;foo", "&#45;&#45;foo", "&#45;fff"],
  "output": "count 5"
```

docs/v2/examples/flags.md
```diff
@@ -104,6 +104,8 @@ See full list of flags at https://pkg.go.dev/github.com/urfave/cli/v2
 
 For bool flags you can specify the flag multiple times to get a count(e.g -v -v -v or -vvv)
 
+> If you want to support the `-vvv` flag, you need to set `App.UseShortOptionHandling`.
+
 <!-- {
   "args": ["&#45;&#45;foo", "&#45;&#45;foo"],
   "output": "count 2"
```
abitrolly, line 111:
```suggestion
  "output": "count 2"
```

abitrolly, line 111:
```suggestion
  "args": ["&#45;&#45;foo", "&#45;&#45;foo", "&#45;fff",  "&#45;f"],
  "output": "count 6"  
```

docs/v3/examples/flags.md
abitrolly: I added two suggestions that add short flags to parameters, so that the code is actually being used.

Juneezee (approved):

abitrolly (approved):

PR #1961, opened 2024-08-10
Feat:(issue_1960) make defaults validation optional

### LINKED ISSUE #1960: Strange `.Validator` behavior
Hi there! I'm wondering if I'm the only one who finds the behavior of this package a bit strange:

```go
package main

import (
  "context"
  "errors"
  "os"
  "strings"

  "github.com/urfave/cli/v3"
)

func main() {
  if err := (&cli.Command{
    Action: func(ctx context.Context, command *cli.Command) error {
      return nil
    },
    Flags: []cli.Flag{
      &cli.StringFlag{
        Name: "api-key",
        Validator: func(s string) error {
          if !strings.HasPrefix(s, "API-") {
            return errors.New("wrong API key format")
          }

          return nil
        },
      },
    },
  }).Run(context.Background(), os.Args); err != nil {
    println(err.Error())

    os.Exit(1)
  }
}
```

It seems that the default value validation is failing before the actual flag value is parsed. Since the default value will always be invalid, this validation function renders the app unusable, even when the correct flag value is provided:

```shell
❯ go run . -h

Incorrect Usage: wrong API key format

...

wrong API key format
exit status 1
```

```
❯ go run . --api-key API-123 -h

Incorrect Usage: wrong API key format

...

wrong API key format
exit status 1
```

Currently, the only option I have is to set a default value that appears to be "valid-like":

```diff
&cli.StringFlag{
	Name: "api-key",
+	Value: "API-<set-your-key>",
```

Could you please confirm if this is intentional or if it might be a bug?

> [!NOTE]
> ```
> github.com/urfave/cli/v3 v3.0.0-alpha9
> go version go1.22.5 linux/amd64
> ```

dearchap:
## What type of PR is this?

_(REQUIRED)_

- documentation
- feature

_(REQUIRED)_

The feature allows default validation to be optional.

_(REQUIRED)_

Fixes #1960 

## Testing

make test

## Release Notes

_(REQUIRED)_

```release-note
Defaults validation is optional(false). Validation of defaults, values from env can be enabled by setting ValidateDefaults to true
```

### DIFF
--- a/command_test.go
+++ b/command_test.go
@@ -3763,7 +3763,7 @@ func TestCommandReadArgsFromStdIn(t *testing.T) {
 		{
 			name: "empty2",
 			input: `
-			
+
 			`,
 			args:          []string{"foo"},
 			expectedInt:   0,
@@ -3781,7 +3781,7 @@ func TestCommandReadArgsFromStdIn(t *testing.T) {
 		{
 			name: "intflag-from-input2",
 			input: `
-			--if 
+			--if
 
 			100`,
 			args:          []string{"foo"},
@@ -3800,14 +3800,14 @@ func TestCommandReadArgsFromStdIn(t *testing.T) {
 			--ssf hello
 			--ssf
 
-			"hello	
+			"hello
   123
 44"
 			`,
 			args:          []string{"foo"},
 			expectedInt:   100,
 			expectedFloat: 100.1,
-			expectedSlice: []string{"hello", "hello\t\n  123\n44"},
+			expectedSlice: []string{"hello", "hello\n  123\n44"},
 		},
 		{
 			name: "end-args",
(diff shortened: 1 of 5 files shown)


yassinebenaid (approved):

dearchap:
@urfave/cli I'm wondering if we need to validate values from ValueSources(like Env, File etc) but not the default that the user sets explicitly via Value . Thoughts ?

Juneezee:
> I'm wondering if we need to validate values from ValueSources(like Env, File etc) but not the default that the user sets explicitly via Value . Thoughts ?

I believe this would be a valuable feature. We could potentially create a `Validator` struct to accommodate future extensions, such as:

```go
type Validator[T any] struct {
	ValidateFunc     func(T) error `json:"-"`
	ValidateDefaults bool          `json:"validateDefaults"`
	ValidateSources  bool          `json:"validateSources"`
}
```

PR #1968, opened 2024-08-19
feat: added HideHelpCommands to docs

### LINKED ISSUE #1965: Update docs for HideHelpCommands
## Checklist

* [ ] Are you running the latest v3 release? The list of releases is [here](https://github.com/urfave/cli/releases).
* [ ] Did you check the manual for your release? The v3 manual is [here](https://github.com/urfave/cli/blob/main/docs/v3/index.md).

Sometimes the user of `cli` would like their app to not have any subcommands. The `help` subcommand is only useful when the app actually has other subcommands. Users of the app may find the usage of the `help` subcommand misleading that other subcommands exist. Even though the `help` subcommand does not list any subcommands, the docs / help message even lead you to believe in the usage section there are subcommands to be used beyond a simple `help` command.

## Solution description

There should be an option of some sort to disable subcommands. When disabled, the `help` subcommand should be removed. Also, the docs and help messages should be adjusted accordingly. This means the "commands" section should be removed entirely. And the usage section will show something along the lines of:
```
USAGE:
   <name> [options]
```
...instead of:
```
USAGE:
   <name> [global options] command [command options]
 ```

Not only should `command` be removed along with its corresponding `[command options]`, but `[global options]` should also be renamed to `[options]` since there is no other options to distinguish global options from. In other words, all options are already global when there are no subcommands. This also means the "global options" section in the docs / help message should be renamed to simply "options" for consistency.

It is possible to create custom help messages. However, in this case, the `help` subcommand cannot be removed.

### PR DESCRIPTION
**Pull request by midirhee12**
## What type of PR is this?

- documentation

Adds `HideHelpCommands` to docs

Fixes #1965 

## Release Notes

```release-note
- Adds `HideHelpCommands` to docs
```

### DIFF
--- a/docs/v2/examples/full-api-example.md
+++ b/docs/v2/examples/full-api-example.md
@@ -107,10 +107,11 @@ func main() {
 						Action: wopAction,
 					},
 				},
-				SkipFlagParsing: false,
-				HideHelp:        false,
-				Hidden:          false,
-				HelpName:        "doo!",
+				SkipFlagParsing:    false,
+				HideHelp:           false,
+				HideHelpCommands:   false,
+				Hidden:             false,
+				HelpName:           "doo!",
 				BashComplete: func(cCtx *cli.Context) {
 					fmt.Fprintf(cCtx.App.Writer, "--better\n")
 				},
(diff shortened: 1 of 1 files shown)


dearchap:
@midirhee12 this is for v2 ? Do you want to change the base to urfave:v2_maint ?

dearchap (approved):

Issue #1884, opened 2024-04-08
Shell completions are broken when login shell is different than currently active shell


bartekpacia:
## My urfave/cli version is

2.27.1

## Checklist

- [x] Are you running the latest v2 release? The list of releases is [here](https://github.com/urfave/cli/releases).
- [x] Did you check the manual for your release? The v2 manual is [here](https://github.com/urfave/cli/blob/main/docs/v2/manual.md)
- [x] Did you perform a search about this problem? Here's the [GitHub guide](https://help.github.com/en/github/managing-your-work-on-github/using-search-to-filter-issues-and-pull-requests) about searching.

## Dependency Management

- [x] My project is using go modules.

## Describe the bug

Shell completions to show are determined by looking at contents of the `$SHELL` environment variable, which isn't a reliable way to do this because [the current login shell isn't always the current active shell](https://askubuntu.com/questions/590899/how-do-i-check-which-shell-i-am-using#comment821399_590903).

https://github.com/urfave/cli/blob/7656c5fb838ca8a6febca43100147d317b544fd3/help.go#L153-L161

## To reproduce

Here's [my simple CLI app](https://github.com/bartekpacia/fhome) with shell completions.

## Observed behavior

My shell is Zsh, and when I run:

```
$ fhome [tab]
config     -- Manage system configuration
event      -- Manage events
help    h  -- Shows a list of commands or help for one command
object  o  -- Manage objects
```

but when I run Bash first and then trigger completions, my CLI tools still outputs Zsh-style completions:

```
$ bash
bash-5.2$ fhome
a              commands       configuration  events         h:Shows        help:Shows     o:Manage       objects        one            system
command        config:Manage  event:Manage   for            help           list           object:Manage  of             or
```

## Expected behavior

I expect shell completion to work for the current active shell, so this bug won't happen.

## Additional context

- Maybe we can take a look at how Cobra does it.
- This also occurs in v3 alpha, please add appropriate labels

## Want to fix this yourself?

Sure!

## Run `go version` and paste its output here

```
$ go version
go version go1.22.2 darwin/arm64
```

## Run `go env` and paste its output here

```
$ go env
GO111MODULE=''
GOARCH='arm64'
GOBIN=''
GOCACHE='/Users/bartek/Library/Caches/go-build'
GOENV='/Users/bartek/Library/Application Support/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFLAGS=''
GOHOSTARCH='arm64'
[...]

PR #1969, opened 2024-08-21
Fix: Use $0 env variable to retrieve current active shell

imransolanki:
## What type of PR is this?

- bug

This is a fix to [bug- Shell completions are broken when login shell is different than currently active shell #1884
](https://github.com/urfave/cli/issues/1884)
- Using `$0` as an env variable instead of `SHELL` as `$0` gives the currently active shell

[Fixes the bug- Shell completions are broken when a login shell is different than currently active shell #1884
](https://github.com/urfave/cli/issues/1884)

## Testing
Updated the existing test case.

## Release Notes
```release-note
Fixed https://github.com/urfave/cli/issues/1884
```

### DIFF
--- a/examples_test.go
+++ b/examples_test.go
@@ -405,7 +405,7 @@ func ExampleCommand_Run_shellComplete_zsh() {
 
 	// Simulate a zsh environment and command line arguments
 	os.Args = []string{"greet", "--generate-shell-completion"}
-	os.Setenv("SHELL", "/usr/bin/zsh")
+	os.Setenv("0", "/usr/bin/zsh")
 
 	_ = cmd.Run(context.Background(), os.Args)
 	// Output:
--- a/help.go
+++ b/help.go
@@ -159,7 +159,7 @@ func printCommandSuggestions(commands []*Command, writer io.Writer) {
 		if command.Hidden {
 			continue
 		}
-		if strings.HasSuffix(os.Getenv("SHELL"), "zsh") {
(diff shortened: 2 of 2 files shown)


bartekpacia:
Hey, can you tell us more about how it works?

imransolanki:
> Hey, can you tell us more about how it works?

The code retrieves env var `$0` which gives us the active shell. 
Earlier the code was retrieving the current shell using `$SHELL` which denotes the login shell.
In the best-case scenario $SHELL and $0 are equal

Example of `$0` in action
image

bartekpacia (approved):
I see. Thank you, this change LGTM. I'm holding off with merge for other maintainers to take a look.

asahasrabuddhe (approved):
Looks good to me!

bartekpacia:
I think this is worth backporting, yeah

dearchap (approved):
@imransolanki Thanks !!!!

bartekpacia:
This actually doesn't fix anything, but breaks printing descriptions in `zsh`. This commit must be reverted.

I think this is also a lesson for us, maintainers of this module, to not trust that contributors did actually test their changes.

### Before this PR

https://github.com/user-attachments/assets/35856f1e-22df-4d2d-b397-846e9259ac7b

### After this PR

https://github.com/user-attachments/assets/b7b332f1-0bef-44b9-b468-ca1ae1127913

PR #1971, opened 2024-08-22
Limit Dependencies to the stdlib

### LINKED ISSUE #1890: Strictly limit default dependencies to stdlib
The runtime dependencies for core library usage should be strictly limited to the Go standard library ("stdlib"). This does not apply to the test dependencies.

- [x] #1892

asahasrabuddhe:
## What type of PR is this?

- cleanup

This PR removes the last remaining dependency `testify`.  In line with the famous Go Proverb: [A little copying is better than a little dependency](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=9m28s). I have decided to copy over the necessary code to make our tests work.

Fixes #1890 

## Testing

## Release Notes

_(REQUIRED)_

```release-note

```


internal/testing/testing.go
```diff
@@ -0,0 +1,765 @@
...
+
+func Empty(t *testing.T, object interface{}, msgAndArgs ...interface{}) bool {
+	pass := isEmpty(object)
+	if !pass {
+		t.Helper()
+		return fail(t, fmt.Sprintf("Should be empty, but was %v", object), msgAndArgs...)
+	}
+
+	return pass
+}
```
yassinebenaid, line 293: I suggest to shorten this:

```go
func Empty(t *testing.T, object interface{}, msgAndArgs ...interface{}) bool {
	if ! isEmpty(object) {
		t.Helper()
		return fail(t, fmt.Sprintf("Should be empty, but was %v", object), msgAndArgs...)
	}

	return true
}
```

Changes:
```diff
func Empty(t *testing.T, object interface{}, msgAndArgs ...interface{}) bool {
- 	pass := isEmpty(object)
- 	if !pass {
+	if ! isEmpty(object) {
		t.Helper()
		return fail(t, fmt.Sprintf("Should be empty, but was %v", object), msgAndArgs...)
	}

- 	return pass
+	return true
}
```

internal/testing/testing.go
```diff
@@ -0,0 +1,765 @@
...
+
+func NotEmpty(t *testing.T, object interface{}, msgAndArgs ...interface{}) bool {
+	pass := !isEmpty(object)
+	if !pass {
+		t.Helper()
+		return fail(t, fmt.Sprintf("Should NOT be empty, but was %v", object), msgAndArgs...)
+	}
+
+	return pass
+}
```
yassinebenaid, line 303: I suggest to shorten this too.

internal/testing/testing.go
yassinebenaid, line 1: Exported functions are missing go doc comments.

I believe that all exported types/functions must be well documented using go doc comments even those that are self documented or that look obvious.

yassinebenaid (changes requested):

bartekpacia:
As much as I like stdlib and desest testing frameworks like testify, I'm not sure it's the right course of action here.

internal/testing/testing.go
asahasrabuddhe: Noted. I will make the necessary changes.

internal/testing/testing.go
asahasrabuddhe: Noted. I will make the necessary changes.

internal/testing/testing.go
asahasrabuddhe: I had originally left them out, as the new `testing` package is our internal package. However, I think it would be a good idea to have them. I have just pushed a commit that includes the comments.

asahasrabuddhe:
> As much as I like stdlib and desest testing frameworks like testify, I'm not sure it's the right course of action here.

How do you think we should approach this then?

bartekpacia:
> How do you think we should approach this then?

I don't know, honestly. I'll pass on reviewing this PR and just defer to other maintainers here.

That said, thanks a lot for contributing to this project :) It's great! I'm just not versed well-enough in this codebase.

internal/testing/testing.go
```diff
@@ -0,0 +1,814 @@
...
+
+	if Containsf(t, s, contains, msg, args...) {
+		return
+	}
+
+	t.FailNow()
+}
+
+// NotContains asserts that the specified string does NOT contain the specified substring.
+func NotContains(t *testing.T, s, contains string, msgAndArgs ...interface{}) bool {
```
yassinebenaid, line 236: I believe that grammar is important even for us developers 👨‍🏫 
```diff
- func NotContains(t *testing.T, s, contains string, msgAndArgs ...interface{}) bool {
+ func DoesNotContain(t *testing.T, s, contains string, msgAndArgs ...interface{}) bool {
```

internal/testing/testing.go
```diff
@@ -0,0 +1,814 @@
...
+	if strings.Contains(s, contains) {
+		return fail(t, fmt.Sprintf("%#v should not contain %#v", s, contains), msgAndArgs...)
+	}
+
+	return true
+}
+
+// RequireNotContains asserts that the specified string does NOT contain the specified substring. If it does, the test
+// will stop.
+func RequireNotContains(t *testing.T, s, contains string, msgAndArgs ...interface{}) {
```
yassinebenaid, line 248:
```diff
- func RequireNotContains(t *testing.T, s, contains string, msgAndArgs ...interface{}) {
+ func RequireDoesNotContain(t *testing.T, s, contains string, msgAndArgs ...interface{}) {
```

yassinebenaid (changes requested):
I'm fan of perfectionism LOL

yassinebenaid:
I know this might sound weird. But, I think we should have a test for those testing functions too using the standard library only.

such tests would test that those testing function are actually doing the job of testing (term `test` is widely used here LOL).

for example:

```go
func TestThatNilActuallyWork(t *testing.T){
    if !Nil(nil){
        t.Fail("Nil doesn't actually work, our tests are risky then.")
    }
}
```

abitrolly:
You've forgot to include license.

dearchap:
Do we really want to maintain a testing framework ourselves ? We have to add tests for the test framework itself now. Is testify required by the users of this library to run their code ?

bartekpacia:
> Is testify required by the users of this library to run their code ?

It's not, it doesn't appear in my app's go.mod and go.sum.

dearchap:
Then what's the point of us struggling so much for this one library

bartekpacia:
Yeah, I am with @dearchap here. Let's stay with testify.

asahasrabuddhe:
Agreed

PR #1977, opened 2024-10-06
Fix: Make flags persistent by default

### LINKED ISSUE #976: v2 feature: allow flags after arguments

dearchap:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

_(REQUIRED)_

Fixes #976 

## Testing

Updated existing tests to work with new features

## Release Notes

_(REQUIRED)_

```release-note
All flags defined by a command flow to subcommands by default unless the Local field is set for the flag in which case it will apply to that command only. This is better than having to make flags persistent when needed as this is an oft requested feature
```

### DIFF
--- a/command.go
+++ b/command.go
@@ -636,12 +636,6 @@ func (cmd *Command) Run(ctx context.Context, osArgs []string) (deferErr error) {
 			return err
 		}
 
-		if err := cmd.checkPersistentRequiredFlags(); err != nil {
-			cmd.isInError = true
-			_ = ShowSubcommandHelp(cmd)
-			return err
-		}
-
 		if len(cmd.Arguments) > 0 {
 			rargs := cmd.Args().Slice()
 			tracef("calling argparse with %[1]v", rargs)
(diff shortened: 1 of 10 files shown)


fiatjaf:
It doesn't work.

I'm on af97b3fe9d60f8fb1be327058fb0cfcd6f948111. I created a directory `example/` and put the following `main.go` file in it:

```go
package main

import (
	"context"
	"fmt"
	"os"

	"github.com/urfave/cli/v3"
)

var app = &cli.Command{
	Name:                   "example",
	Suggest:                true,
	UseShortOptionHandling: true,
	Commands: []*cli.Command{
		blergh,
	},
}

var blergh = &cli.Command{
	Name: "blergh",
	Flags: []cli.Flag{
		&cli.BoolFlag{
			Name: "cool",
		},
		&cli.UintFlag{
			Name: "number",
		},
	},
	ArgsUsage: "[relay...]",
	Action: func(ctx context.Context, c *cli.Command) error {
		fmt.Println("args:", c.Args())
		fmt.Println("cool?", c.Bool("cool"))
		fmt.Println("number", c.Uint("number"))

		return nil
	},
}

func main() {
	if err := app.Run(context.Background(), os.Args); err != nil {
		os.Exit(1)
	}
}
```

Then I run it with `go run . blergh ok --cool --number 12`, expecing the output to be

```
args: &{[ok]}
cool? true
number 12
```

But instead I get

```
args: &{[ok --cool --number 12]}
cool? false
number 0
```

abitrolly:
I feel like this needs a proper blog post explaining logic without Go code.

dearchap:
Okay I see the problem. I misinterpreted what you are looking for @fiatjaf . I'm going to reopen your PR and make a fix .

PR #1980, opened 2024-10-13
Update dependencies in v2-maint

meatballhat:
## What type of PR is this?

- maintenance

Update to latest dependency versions and introduce diffing to the file comparison test func, but without bringing in a dependency like the `v1-maint` branch.

N/A

### DIFF
--- a/fish_test.go
+++ b/fish_test.go
@@ -1,9 +1,13 @@
 package cli
 
 import (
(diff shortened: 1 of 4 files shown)


Juneezee (approved):
:+1:

Scutua (approved):

PR #1987, opened 2024-10-14
Fix:(issue_976) All positional args and flags to be mixed in any order

### LINKED ISSUE #976: v2 feature: allow flags after arguments

### LINKED ISSUE #1733: Ordering of arguments and flags impact one another

dearchap:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

Working towards making positions arguments and flags interchangeable

_(REQUIRED)_

Fixes #976 
Fixes #1928 
Fixes #1733 

Focus on command.parseFlags which has the logic

## Testing

All tests have been updated to ensure they work 

## Release Notes

_(REQUIRED)_

```release-note
Allows positions args and flags to be placed in any order
```


command.go
```diff
@@ -802,14 +804,60 @@ func (cmd *Command) parseFlags(args Args) (Args, error) {
...
+			tracef("returning-1 (cmd=%[1]q) args %[2]q", cmd.Name, posArgs)
+			cmd.parsedArgs = &stringSliceArgs{posArgs}
+			return cmd.parsedArgs, err
+		}
+		tracef("rearrange-4 (cmd=%[1]q) check %[2]q", cmd.Name, cmd.flagSet.Args())
+		rargs = cmd.flagSet.Args()
+		if len(rargs) == 0 || len(rargs[0]) == 0 || rargs[0] == "-" {
+			break
+		}
+	}
```
BlackHole1, line 855: Optimize nesting depth.

```go
for {
		tracef("rearrange:1 (cmd=%[1]q) %[2]q", cmd.Name, rargs)
		for {
			tracef("rearrange:2 (cmd=%[1]q) %[2]q %[3]q", cmd.Name, posArgs, rargs)

			if len(rargs) == 0 || len(rargs[0]) == 0 {
				break
			}

			if rargs[0] == "--" {
				posArgs = append(posArgs, rargs...)
				cmd.parsedArgs = &stringSliceArgs{posArgs}
				return cmd.parsedArgs, nil
			}

			if rargs[0][0] == '-' {
				break
			}

			tracef("rearrange-3 (cmd=%[1]q) check %[2]q", cmd.Name, rargs[0])

			if cmd.Command(rargs[0]) != nil {
				posArgs = append(posArgs, rargs...)
				cmd.parsedArgs = &stringSliceArgs{posArgs}
				return cmd.parsedArgs, nil
			}

			posArgs = append(posArgs, rargs[0])
			if len(rargs) == 0 {
				rargs = []string{}
				break
			}

			rargs = rargs[1:]
		}

		if err := parseIter(cmd.flagSet, cmd, rargs, cmd.Root().shellCompletion); err != nil {
			posArgs = append(posArgs, cmd.flagSet.Args()...)
			tracef("returning-1 (cmd=%[1]q) args %[2]q", cmd.Name, posArgs)
			cmd.parsedArgs = &stringSliceArgs{posArgs}
			return cmd.parsedArgs, err
		}
}
```

BlackHole1 (changes requested):

command_test.go
```diff
@@ -163,7 +163,7 @@ func TestCommandFlagParsing(t *testing.T) {
 		{testArgs: []string{"test-cmd", "blah", "blah"}, skipFlagParsing: true, useShortOptionHandling: false},   // Test SkipFlagParsing without any args that look like flags
 		{testArgs: []string{"test-cmd", "blah", "-break"}, skipFlagParsing: true, useShortOptionHandling: false}, // Test SkipFlagParsing with random flag arg
 		{testArgs: []string{"test-cmd", "blah", "-help"}, skipFlagParsing: true, useShortOptionHandling: false},  // Test SkipFlagParsing with "special" help flag arg
-		{testArgs: []string{"test-cmd", "blah", "-h"}, skipFlagParsing: false, useShortOptionHandling: true},     // Test UseShortOptionHandling
+		//{testArgs: []string{"test-cmd", "blah", "-h"}, skipFlagParsing: false, useShortOptionHandling: true},     // Test UseShortOptionHandling
```
BlackHole1, line 166: Did you forget to delete this?

command_test.go
```diff
@@ -693,16 +693,16 @@ var defaultCommandSubCommandTests = []struct {
...
-	{" ", "j", "", false},
-	{"bat", "", "batbaz", false},
-	{"nothing", "", "batbaz", false},
+	//{" ", "", "", true},
+	//{" ", "j", "", true},
+	{"bat", "", "batbaz", true},
+	{"nothing", "", "batbaz", true},
 	{"nothing", "", "", false},
-	{"nothing", "j", "batbaz", false},
+	//{"nothing", "j", "batbaz", true},
```
BlackHole1, line 705: Above

command_test.go
dearchap: no. for some reason the test fails and I couldnt figure it out

command_test.go
dearchap: no. for some reason the test fails and I couldnt figure it out

command_test.go
```diff
@@ -693,14 +693,14 @@ var defaultCommandSubCommandTests = []struct {
...
 	{"", "jimmers", "", true},
-	{" ", "jimmers", "foobar", false},
+	//{" ", "jimmers", "foobar", false},
 	{"", "", "", true},
-	{" ", "", "", false},
-	{" ", "j", "", false},
-	{"bat", "", "batbaz", false},
-	{"nothing", "", "batbaz", false},
+	//{" ", "", "", true},
+	//{" ", "j", "", true},
```
BlackHole1, line 701: Did you forget to remove the comments?
If there are issues, maybe we can add a TODO here to inform future readers why this part is commented out.

BlackHole1 (changes requested):

command_test.go
dearchap: I'm not even sure these are valid test cases. Why would an argument be just a string of spaces ?

danhunsaker: Some field separators are encoded as successive whitespace. Specifying `'    '` is simpler to use than `--space=4` for some users in some use cases, even if that's not strictly "better" technically. So I've seen it done.

dearchap: But these are the args specified directly to the application . So on command line if I type 

```
Foo    Bar
```

The application would get it as an array of strings ["Foo","Bar"] so I don't see when we would be passed empty spaces at all

danhunsaker:
```sh
foo      bar
```

won't give any whitespace, correct. But

```sh
foo '    ' bar
```

will. Shell quotation doesn't get passed along, obviously, but the contents do.

danhunsaker (approved):

PR #1994, opened 2024-10-26
Feature:(issue_1927) Add ability for before handler to return new context

### LINKED ISSUE #1927: Ability to add context values
## Checklist

* [x] Did you check the manual for your release? The v3 manual is [here](https://github.com/urfave/cli/blob/main/docs/v3/manual.md).

In old versions we could add context fields in Before function like this
```
  Before: func(c *cli.Context) error {
      db := "example"
      c.Context = context.WithValue(c.Context, "db", db)
      return nil
  },
```
It's useful for example for passing logger through all commands.

## Solution description
Add some middleware layer to command with signature
```
func(ctx context.Context, command *cli.Command) func(ctx context.Context, command *cli.Command)
```
where we could modify passing ctx

Return back Context field to command

dearchap:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

_(REQUIRED)_

Fixes #1927 

## Testing

go test -run=TestCommand_Run_BeforeReturnNewContext

## Release Notes

_(REQUIRED)_

```release-note
The Before handler for a command can optionally return a new context.Context which will then be propagated to the Action and After handlers. 
```

### DIFF
--- a/command.go
+++ b/command.go
@@ -562,9 +562,11 @@ func (cmd *Command) Run(ctx context.Context, osArgs []string) (deferErr error) {
 	}
 
 	if cmd.Before != nil && !cmd.Root().shellCompletion {
-		if err := cmd.Before(ctx, cmd); err != nil {
+		if bctx, err := cmd.Before(ctx, cmd); err != nil {
 			deferErr = cmd.handleExitCoder(ctx, err)
 			return deferErr
+		} else if bctx != nil {
+			ctx = bctx
 		}
 	}
 
--- a/command_test.go
+++ b/command_test.go
@@ -268,8 +268,8 @@ func TestParseAndRunShortOpts(t *testing.T) {
 func TestCommand_Run_DoesNotOverwriteErrorFromBefore(t *testing.T) {
 	cmd := &Command{
 		Name: "bar",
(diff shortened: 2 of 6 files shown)


command.go
```diff
@@ -562,9 +562,11 @@ func (cmd *Command) Run(ctx context.Context, osArgs []string) (deferErr error) {
 	}
 
 	if cmd.Before != nil && !cmd.Root().shellCompletion {
-		if err := cmd.Before(ctx, cmd); err != nil {
+		if bctx, err := cmd.Before(ctx, cmd); err != nil {
 			deferErr = cmd.handleExitCoder(ctx, err)
 			return deferErr
+		} else if bctx != nil {
+			ctx = bctx
```
meatballhat, line 569: Is it not a legitimate usage for one's `Before` to return a new context *and* an error?

meatballhat (approved):
I like it! I had the one question and it looks like the docs tests failures are related to the change in func signature.

command.go
dearchap: what would be the purpose of that since the app is going to return an error immediately anyway .

dearchap: Lets implement it if and when someone asks for it.

Issue #1904, opened 2024-05-02
Expose shell completions scripts in by default in the `completion` command


bartekpacia:
## Checklist

* [x] Are you running the latest v3 release? The list of releases is [here](https://github.com/urfave/cli/releases).
* [x] Did you check the manual for your release? The v3 manual is [here](https://github.com/urfave/cli/blob/main/docs/v3/manual.md).
* [x] Did you perform a search about this feature? Here's the [GitHub guide](https://help.github.com/en/github/managing-your-work-on-github/using-search-to-filter-issues-and-pull-requests) about searching.

## What problem does this solve?

Makes it easier for developers to add shell completion to the CLI tools they're building with `urfave/cli/v3`. This would make completion installation the same as the popular Cobra library, that is:

`myapp completion zsh` prints the completion script for `zsh`. The user of the CLI gets to decide what to do with it:
- put it into completions directory: `myapp completion zsh > /opt/homebrew/share/zsh/site-functions/_myapp` (on macOS), or
- source it in-place: `. <(myapp completion zsh)`

We already have [completion scripts in this repository](https://github.com/urfave/cli/tree/main/autocomplete) for all major popular shells (`bash`, `zsh`, `pwsh`), and furthermore, these scripts are already bundled with the module:

https://github.com/urfave/cli/blob/204d34fd57943cab87f9228e19b9b0230c65976a/completion.go#L14-L16

## Solution description

By default, each CLI app made with `urfave/cli/v3` should have the `completion` command (hidden by default)

I don't think this could be brought into `v2`, because it would break if the developer already defined the `completion` command.

## Describe alternatives you've considered

The alternative is the current situation – developers have to copy , which is unnecessary work, given that the completion files are embedded in `urfave/cli` anyway.

## Resources used

- [Cobra's zsh completion script generation function](https://github.com/spf13/cobra/blob/3a5efaede9d389703a792e2f7bfe3a64bc82ced9/zsh_completions.go#L87)

dearchap:
@bartekpacia I dont understand. The "generate-completion" is a hidden command in urfave/cli/v3. So if user invokes

```
myapp generate-completion zsh
```

It will output the zsh completion script which can be copied by user and sourced.

bartekpacia:
@dearchap Thank you for pointing it out – I wasn't aware of it.

It doesn't look like this feature is documented anywhere yet – but this may be actually a good thing, because v3 is at alpha, so we could:

[...]

PR #1998, opened 2024-10-28
Improve the command for printing completion scripts

bartekpacia:
## Which issue(s) this PR fixes

This is a feature PR that resolves issue #1904.

## Special notes for your reviewer

I would appreciate it if you check out the code locally, build it, and test the shell completions yourself.

I did that myself, but the more testing, the better. Also, if you have any ideas on how to test this better, I'd love to hear them.

## Release Notes

- Rename the `generate-completion` flag to simply `completion`. This makes our behavior in this matter consistent with the other very popular CLI library for Go – spf13/cobra.

- The completion scripts now include your CLI app name and are ready to be used right away.
 
To enable completion only for the current shell Zsh session:

```
. <(./my-awesome-app completion zsh
```

To enable completions permanently by placing the completion scripts into [the standard completions directory](https://docs.brew.sh/Shell-Completion):

```
my-awesome-app completion zsh > /opt/homebrew/share/zsh/site-functions/_my-awesome-app
```

Please note that for completion to work, your top-level `cli.Command` name and your binary name must be the same.


completion.go
```diff
@@ -8,43 +8,80 @@ import (
...
+func getCompletion(s string) renderCompletionFunc {
+	return func(cmd *Command, appName string) (string, error) {
 		b, err := autoCompleteFS.ReadFile(s)
 		return string(b), err
 	}
 }
 
-func buildCompletionCommand() *Command {
+func genZshCompletion(appName string) string {
+	return fmt.Sprintf(`#compdef %[1]s
```
dearchap, line 44: you could still load this content from autocompleteFS as before and then do a search/replace on a template string with appName. That way if other shells need the same logic we can reuse.

command.go
```diff
@@ -257,7 +257,7 @@ func (cmd *Command) setupDefaults(osArgs []string) {
 	}
 
 	if cmd.EnableShellCompletion || cmd.Root().shellCompletion {
-		completionCommand := buildCompletionCommand()
+		completionCommand := buildCompletionCommand(cmd.Name)
```
dearchap, line 260:
```suggestion
		completionCommand := buildCompletionCommand(osArgs[0])
```

That way you dont have to worry about binary name and cmd.Name ?

bartekpacia: Thanks. Is `os.Args[0]` always the binary name though? Will it wokr when the binary is called with a shell alias, not directly?

I'll have to check...

completion.go
bartekpacia: Thanks, I prefer to keep this inside of the Go code though. imho makes it easier to see what is passed inside

bartekpacia:
I'm having trouble understanding why it fails.

Running `FLAGS="--walk docs/v3/" make gfmrun` works without problem for me.

Will appreciate help here @dearchap @meatballhat @abitrolly

examples/simpletask/main.go
```diff
@@ -0,0 +1,70 @@
+package main
```
dearchap, line 1: I dont see the value of this example. Its not really doing anything. If you want to really test this move it into examples_test.go or call it func ExampleCompletion(...) in completion_test.go

completion.go
```diff
@@ -57,14 +54,103 @@ func completionCommandAction(ctx context.Context, cmd *Command) error {
...
-	} else {
-		if _, err = cmd.Writer.Write([]byte(c)); err != nil {
-			return Exit(err, 1)
-		}
 	}
+
 	return nil
 }
+
+func genBashCompletion(appName string) string {
```
dearchap, line 75: can you provide the reason why you moved away from embedFs to having the completion inline with the code ? In the previous approach it was very easy to lookup where the autcompletes were and what they were doing

abitrolly:
> I'm having trouble understanding why it fails.
> 
> Running `FLAGS="--walk docs/v3/" make gfmrun` works without problem for me.

I am not familiar with this piece of code. And don't see any failures. I guess it is fixed.

What concerns me is how much to the binary size is added by the completion feature and its templates? I guess it is non-optional, right?

dearchap:
@abitrolly All the template code is already in current binary. So @bartekpacia 's work doesnt really change.

examples/simpletask/main.go
bartekpacia: Hi, I do see its value, it's a simple yet quite realistic CLI app. It's quite useful for testing shell completions

completion.go
bartekpacia: This is to make it easier to use `%[1]s` since then declaration and usage is in the same place.

Juneezee: I'm not sure about this either. We get nice syntax highlighting from the code editor/IDE when the autocomplete is in a separate file. But when it's inlined, it's treated as a string, which makes it less readable.

bartekpacia: I wouldn't say it's a big problem. Characters like `%[1]` aren't valid shell function names anyway and will probably be highlighted wrongly.

I also think keeping everything related to completion in a single file is quote convenient.

abitrolly:
> @abitrolly All the template code is already in current binary. So @bartekpacia 's work doesnt really change.

It is still interesting what is the overhead and how to remove it?

examples/simpletask/main.go
abitrolly: I find complete examples very useful. Sometimes you just need a bit of working code to debug something that doesn't work anymore. If completions break when migrating from v2 to v3, then using this v3 code as a reference, I could find the cause much faster.

Complete working examples are also useful for training AI.

It needs some better organization, though. Maybe even numbers to sort contents in the order people usually learn the library. By most frequent use cases.

https://github.com/urfave/cli/tree/ef45965eeb9c1122885fafa4a391b6be6a674f3d/examples

completion.go
abitrolly: > I also think keeping everything related to completion in a single file is quote convenient.

Scrolling back and forth between template and its code doesn't seem convenient to me. It is better to have two panes - one with template and another with content.

Highlighting of template variables as errors could actually be a  feature.

abitrolly:
@bartekpacia I like the idea of introducing complete examples for specific features. It helps me to see how `urfave/cli` implements some decisions. But mixing up Go code with Bash code for completion scripts looks like a wrong way ahead. )

examples/simpletask/main.go
dearchap: Thats where the https://github.com/urfave/cli/blob/ef45965eeb9c1122885fafa4a391b6be6a674f3d/examples_test.go comes into play.

dearchap:
@abitrolly what I meant is that removing embedFS and templating it in code doesnt change the size of the overall cli binary. I am ok with either approach unless there is a good reason one way or the other

examples/simpletask/main.go
abitrolly: @dearchap https://github.com/urfave/cli/blob/ef45965eeb9c1122885fafa4a391b6be6a674f3d/examples_test.go is not discoverable, and also they do not work if copy-pasted.

abitrolly:
@bartekpacia I am not okay with moving code sections without really good reasons, because comparing historical changes in these pieces becomes really hard.

bartekpacia:
Thanks for the review and discussion.

I don't agree with your opinion  @abitrolly about mixing Go+ShellC but it's not really an important thing so let's not bikeshed. I'll move completions back to their respective files.

I plan to spend some time soon on improving the completions, so if this proves to be annoying we can revisit it down the road.

bartekpacia:
@Juneezee @dearchap i moved completion script strings back into individual files. I think this PR is good to be merged now.

help.go
```diff
@@ -214,7 +215,7 @@ func printFlagSuggestions(lastArg string, flags []Flag, writer io.Writer) {
 			continue
 		}
 		// match if last argument matches this flag and it is not repeated
-		if strings.HasPrefix(name, cur) && cur != name && !cliArgContains(name, os.Args) {
+		if strings.HasPrefix(name, currentArg) && currentArg != name && !cliArgContains(name, os.Args) {
```
abitrolly, line 218: After the prefix is stripped, it is not arg anymore, but a search string for a match. I would name it `match` or `lookup` or something line that.

help.go
```diff
@@ -190,8 +190,9 @@ func cliArgContains(flagName string, args []string) bool {
 }
 
 func printFlagSuggestions(lastArg string, flags []Flag, writer io.Writer) {
-	cur := strings.TrimPrefix(lastArg, "-")
-	cur = strings.TrimPrefix(cur, "-")
+	// Trim the prefix twice to handle both "-short" and "--long" flags.
+	currentArg := strings.TrimPrefix(lastArg, "-")
+	currentArg = strings.TrimPrefix(currentArg, "-")
```
abitrolly, line 195: https://pkg.go.dev/strings#TrimLeft

examples/simpletask/main.go
abitrolly: `example/simpletask.go` doesn't tell anything to the person who browses examples. For the sake of bikeshedding `examples/commands.go` may be a better name.

I understand that this example can be used to test command completions, but because it is not used actually used in tests, I would suggest to submit it in a separate PR, where we could also discuss how to integrate it with what we have in docs.

autocomplete/bash_autocomplete
```diff
@@ -1,23 +1,23 @@
-#! /bin/bash
+#!/usr/env/bin bash
```
abitrolly, line 1: Why? Maybe it is obvious, but I am not sure. In Python we use `#!/usr/env/bin python`, so that scripts can be run from `vitrualenv` where `python` points to different binary with different set of libs. What is the point here with `bash`?

abitrolly (commented):
I left some minor comments for the things I could review for the time being. It would be realistic to merge them if submitted in a separate PRs, because the rest of the diff is beyond my ability to review right now.

autocomplete/bash_autocomplete
bartekpacia: I was brought up in the practice of using `/usr/bin/env bash,` and I believe it's a good shebang, because it doesn't hardcode the path. [See this answer](https://stackoverflow.com/a/16365367/7009800).

help.go
Juneezee:
```suggestion
	currentArg := strings.TrimLeft(lastArg, "-")
```

Just in case it isn't obvious for other readers, we only need to trim once with `strings.TrimLeft`.

```go
package main

import (
	"fmt"
	"strings"
)

func main() {
	shortFlag := "-short"
	longFlag := "--long"

	fmt.Println(strings.TrimLeft(shortFlag, "-")) // short
	fmt.Println(strings.TrimLeft(longFlag, "-"))  // long

	fmt.Println(strings.TrimPrefix(shortFlag, "-")) // short
	fmt.Println(strings.TrimPrefix(longFlag, "-"))  // -long
}
```

examples/simpletask/main.go
bartekpacia: Good idea, I agree. I will add the `simpletask` example in a separate PR.

help.go
bartekpacia: thanks @Juneezee, I didn't understand it before:)

help.go
bartekpacia: I just reverted back to `cur`, don't want to spend time on this.

autocomplete/bash_autocomplete
abitrolly: There is a mistake in new shebang, and that makes me think that you haven't tested that it actually works. :D

Using `env` has drawbacks. I still don't see why it is so good. Where do you expect hardcoded `/bin/bash` to fail?

bartekpacia: mistake fixed, thanks.

abitrolly:
> Rename the `generate-completion` flag to simply `completion`.

I am not aware how this feature works.

Is it enabled by default?
Does it show up in the list of user configured commands by default?
Can users choose the name they are prefer for this completion command?

autocomplete/bash_autocomplete
bartekpacia: also, what kind of drawbacks?

bartekpacia:
This PR does 2 things:
- renames an existing `generate-completion` to be `completion` (in-line with Cobra and other popular CLI libraries)
- the completion script printed by this command now contains the app name, and is ready to be pasted in the completion directory without any kind of manual editing.

completion.go
bartekpacia: GoLand highlights when too many or too few arguments are used in the `fmt.Sprintf` call. I think this is already a nice benefit.

I may move scripts into Go code in separate PR.

autocomplete/bash_autocomplete
abitrolly: > also, what kind of drawbacks?

`bash` then can point to any binary. Any user or malicious script can override it. `/bin/bash` is guaranteed to be present if `bash` is installed. So it is just less moving parts, less attack surface, and less things to care about.

If you can ask your mentor about the reasons why he/she prefers `env` calls, maybe he can actually remember the case where `env` with `bash` was really necessary.

abitrolly:
> This PR does 2 things:

Maybe make 2 PRs making 1 thing, so that we can finally merge at least something already. :D

completion.go
```diff
@@ -8,43 +8,57 @@ import (
...
 	//go:embed autocomplete
 	autoCompleteFS embed.FS
 
 	shellCompletions = map[string]renderCompletion{
-		"bash": getCompletion("autocomplete/bash_autocomplete"),
-		"ps":   getCompletion("autocomplete/powershell_autocomplete.ps1"),
-		"zsh":  getCompletion("autocomplete/zsh_autocomplete"),
-		"fish": func(c *Command) (string, error) {
+		"bash": func(c *Command, appName string) (string, error) {
+			b, err := autoCompleteFS.ReadFile("autocomplete/bash_autocomplete")
```
dearchap, line 25: i wonder if we can ignore the error since we are loading from embedFS. Saves us from writing unit tests as well.

bartekpacia: nice catch! I think it's not worth worrying about the error here

dearchap (approved):
Please try to add additional tests or remove err in the lookup.

completion.go
bartekpacia: I just decided to not additionaly wrap the errors. I think it's fine now.

autocomplete/bash_autocomplete
bartekpacia: I see. I reverted back to `/bin/bash` (minus the additional whitespace).

examples/simpletask/main.go
bartekpacia: Oh, actually why I like `/bin/bash`, is that it will refer to whatever is first in 

For example on macOS, the `/bin/bash` is a 20-years old bash3, and almost everyone using bash installs a modern bash5 with Homebrew.

Output on my machine:

```
$ which -a bash
/opt/homebrew/bin/bash
/bin/bash
```

So I think hardcoding `/bin/bash` in this case will always default to the pre-installed bash3, which I think is not so nice.

wdyt?

abitrolly (approved):
I haven't tested Go code, and can not give it a proper review, but all questions that I had had been addressed, so no reason for me to hold this back. Except, maybe to squash the history, or rebase to a minimal set of commits to avoid stumbling into reverts while blaming the history.

bartekpacia:
thanks for the review everyone :)


I will of course "squash and merge" this PR to not butcher history

completion.go
dearchap: Yes but the code cov decreased, I've worked so much over the last month to get the code coverage from 80 something percent to 98.4% and now it decreases again.

bartekpacia: Oh, I didn't think about it. I'm sorry!

What command can I run locally to see coverage percentage?

dearchap:
```make``` will give you the coverage percentage

PR #2005, opened 2024-11-06
Re-add generic flag back

dearchap:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

Re add Generic flag as it is still useful in generated code by 3rd party tools. 

_(REQUIRED)_

## Testing

Added Generic tests from v2-maint.

## Release Notes

_(REQUIRED)_

```release-note
GenericFlag has been added back after initial removal from v3 as it seems to be useful for 3rd party plug-in code
```

### DIFF
--- a/flag_generic.go
+++ b/flag_generic.go
@@ -46,6 +46,14 @@ func (f *genericValue) String() string {
 	return ""
 }
 
+func (f *genericValue) IsBoolFlag() bool {
+	if f.val == nil {
+		return false
+	}
+	bf, ok := f.val.(boolFlag)
+	return ok && bf.IsBoolFlag()
+}
+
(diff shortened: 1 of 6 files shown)


Juneezee (approved):

flag_generic.go
```diff
@@ -0,0 +1,67 @@
+package cli
+
+type GenericFlag = FlagBase[Value, NoConfig, genericValue]
```
meatballhat, line 3: I think `genericValue` should be made public as mixed type access in a public API can be no fun.

dearchap: Not really since the public API would use only ```cli.Value``` . Any other customizations is upto the user to redefine FlagBase with other template types

PR #2009, opened 2024-11-20
Add map value source

dearchap:
## What type of PR is this?

_(REQUIRED)_

- documentation
- feature

_(REQUIRED)_

In order to make the cli-altsrc more modular in terms on its provider it seems to be better if map value source can be incorporated directly into cli. 

_(REQUIRED)_

## Testing

go test -run=TestMapValueSource

## Release Notes

_(REQUIRED)_

```release-note
New value source for maps has been included. This allows plugin providers for other formats(json, yaml, etc) to utilize base functionality instead of each source do custom nested lookups
```

### DIFF
--- a/value_source.go
+++ b/value_source.go
@@ -17,6 +17,11 @@ type ValueSource interface {
 	Lookup() (string, bool)
 }
 
+type EnvValueSource interface {
+	IsFromEnv() bool
+	Key() string
+}
+
 // ValueSourceChain contains an ordered series of ValueSource that
 // allows for lookup where the first ValueSource to resolve is
 // returned
@@ -38,8 +43,8 @@ func (vsc *ValueSourceChain) EnvKeys() []string {
 	vals := []string{}
 
 	for _, src := range vsc.Chain {
-		if v, ok := src.(*envVarValueSource); ok {
-			vals = append(vals, v.Key)
+		if v, ok := src.(EnvValueSource); ok {
+			vals = append(vals, v.Key())
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


value_source.go
```diff
@@ -17,6 +17,11 @@ type ValueSource interface {
 	Lookup() (string, bool)
 }
 
+type EnvValueSource interface {
+	IsFromEnv() bool
```
Juneezee, line 21: LGTM. Only 1 comment.

I'm wondering about the use cases for `IsFromEnv`. Doesn't it always return true?

abitrolly:
Is the the "map value" or the map of value sources?

For debugging and bug reports it might be useful to have a structure and output that shows where option sets and overrides. Like:

```
$ CLI_PATH=/eee cli --path=/ccc --debug
urfave/cli: --path (default: none, env: /eee, cmdline: /ccc, final: /ccc)
```

value_source.go
dearchap: Yes does in our case as it enables us to iterate over value sources and detect which ones are from env (for help print purposes). It is conceivable for us to have an envSource which masks IsFromEnv so that it doesnt show up in help. Think of boolFlag pattern. 

https://cs.opensource.google/go/go/+/refs/tags/go1.23.3:src/flag/flag.go;l=150

dearchap:
> Is it the "map value" or the map of value sources?
> 
> For debugging and bug reports it might be useful to have a structure and output that shows where option is sets, and which value is selected finally. Like:
> 
> ```
> $ CLI_PATH=/eee cli --path=/ccc --debug
> urfave/cli: --path (default: none, env: /eee, cmdline: /ccc, final: /ccc)
> ```

It is a map-value source i.e a source which can lookup values based on key.

Juneezee (approved):

meatballhat (approved):
I agree with @abitrolly that `mapValueSource` is a bit confusing, but I follow your reasoning 😅👍🏻

PR #2012, opened 2024-11-24
Use `cmd.Name` when generating completion script

bartekpacia:
This PR fixes a bug in completion scripts when the binary is called in any other way than by simply typing its name. Examples:

- When installing my program `emu` with Homebrew, `os.Args[0]` is `/opt/homebrew/Cellar/emu/0.2.16/bin/emu`.

- When running my program `emu` like this: `./emu`, then `os.Args[0]` is `./emu`.

This breaks the completion script.
I think using root command's `cmd.Name` is a good way to fix it.

[Also discussed here](https://github.com/urfave/cli/pull/1998#discussion_r1819174294)

### DIFF
--- a/command.go
+++ b/command.go
@@ -257,7 +257,7 @@ func (cmd *Command) setupDefaults(osArgs []string) {
 	}
 
 	if cmd.EnableShellCompletion || cmd.Root().shellCompletion {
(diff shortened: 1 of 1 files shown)


bartekpacia:
I ask for a quick approve so can quickly release new version and test if it works fine :-)

PR #2025, opened 2024-12-03
Add dependabot

mrueg:
## What type of PR is this?

- feature

Adds a dependabot configuration for automated dependency and github workflow updates.

None, let me know if I need to create one first.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


yassinebenaid (approved):
LGTM

bartekpacia:
Honestly, I don't see a need for it. I always find dependabots warnings to be annoying and not useful.

Also, this module has no dependencies (aside from some utilities for testing), and we aim to keep it this way.

I'm against.

abitrolly:
-1 on dependabot too. It pollutes search results with irrelevant hits, and mailbox with too much noise.

bartekpacia:
We don't have many GitHub Actions workflow files, and we only use first-party actions from GitHub, and as of now, they're all on the latest version.

Thank you for this PR and I appreciate enthusiasm, but I'm still against (just my opinion. I'm curious what other maintainers think).

PR #2028, opened 2024-12-06
Run Before actions after setting up subcommand

### LINKED ISSUE #2027: Before function for persistent flags
This is with v3.0.0-beta1.01.
I came across this issue while porting our commands in [go-ethereum](https://github.com/ethereum/go-ethereum) from v2 to v3.

In many of our tools, we use a `Before` hook on the root command to perform common actions such as setting up logging.
As an example, let's define a flag like this to set the loglevel:

```go
var logLevelFlag = &cli.IntFlag{
    Name: "loglevel",
    Value: 3,
}
```

The root command includes this flag and a `Before` hook to enable logging.

```go
var app = &cli.Command{
    Flags: []cli.Flag{
        logLevelFlag,
    },

    Commands: []*cli.Command{
        subCommand,
    },
    
    Before: func (ctx context.Context, cmd *cli.Command) (context.Context, error) {
        // here we would set the loglevel on the default logger
        fmt.Println("level:", cmd.Int(logLevelFlag.Name))
        return ctx, nil
    },
}

var subCommand = &cli.Command{
    Name: "sub",
}
```

This works as expected when invoking the root command (`app`) directly:

```shell
$ ./app --loglevel=5
level: 5
```

It also works for subcommands when putting the flag before the command:

```shell
$ ./app --loglevel=5 sub
level: 5
```

But it doesn't work when the flag is at the end:

```shell
$ ./app sub --loglevel=5
level: 3  # should be 5
```

It's unexpected behavior for me, because in v3, flags are supposed to be persistent by default, i.e. they should be applicable in any context. I think of the `app.Before` hook as a place where I can act on flags, regardless of which subcommand is in effect. But perhaps that's not the case. 

A workaround could be to set a `Before` hook on all subcommands. But it's a messy solution for us, since we have a lot of subcommands.

You can find the full example code here: https://gist.github.com/fjl/06ccd9ae685d91ee98dd05128637a8a5

fjl:
I'm posting this to explore a possible resolution for #2027. 

Let me explain the change to `TestFlagAction/mixture`. The test shows a behavior difference introduced by this PR: since flag actions will now run after all flags have been set, the test does not see the 'shadowed' flag `--f_string=app`. I personally think the new behavior is more correct: the flag is shadowed after all.

I can't really think of a scenario in a real app where you would want to set the same flag two times with different values for the root command and the subcommand. So I propose to change the test, locking in the new behavior.

I have also added a test for issue #2027 here.

## What type of PR is this?

- bug
- cleanup

This fixes an issue where the Before hook of a parent command would not see flag values set in the subcommand context.

Fixes #2027

### DIFF
--- a/command.go
+++ b/command.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"slices"
 	"sort"
 	"strings"
 	"unicode"
@@ -561,23 +562,7 @@ func (cmd *Command) Run(ctx context.Context, osArgs []string) (deferErr error) {
 		}
 	}
 
-	if cmd.Before != nil && !cmd.Root().shellCompletion {
-		if bctx, err := cmd.Before(ctx, cmd); err != nil {
-			deferErr = cmd.handleExitCoder(ctx, err)
-			return deferErr
-		} else if bctx != nil {
-			ctx = bctx
-		}
-	}
-
-	tracef("running flag actions (cmd=%[1]q)", cmd.Name)
-
-	if err := cmd.runFlagActions(ctx); err != nil {
-		return err
-	}
-
 	var subCmd *Command
-
 	if args.Present() {
 		tracef("checking positional args %[1]q (cmd=%[2]q)", args, cmd.Name)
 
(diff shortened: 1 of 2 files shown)


fjl:
Actually, after thinking a bit longer, the behavior change to flag actions may not be acceptable. The existing behavior of flag actions was basically to invoke the action for every instance of the flag, with the value of that instance. It can make sense for some applications. The problem is the context parameter.

If flag actions didn't take the `context.Context` parameter, they could be run before `Before` actions. This would make sense for applications where the flag action is used to set a global or something.

However, if the context is desired in flag actions, the only option would be collecting flag instances in another slice in `Command` so their local values can be accessed later when running the actions.

fjl:
Having spent quite a bit of time trying to restore the behavior of flag actions while keeping the new Before logic, I can now say it's impossible. But working on that has convinced me flag actions are handled kind of incorrectly right now.

At the moment, flag actions for persistent flags are invoked multiple times, basically once per 'command-level', with the value at that level. That doesn't make a lot of sense, given that the ultimate goal of flag parsing, setup etc. is to execute the innermost command action, and the flag values at that level are the ones that should be applied to the program.

So I propose that flag actions of persistent flags should be changed to run exactly once, with the value that will actually be set in the command action.

fjl:
@dearchap would be nice to get your input on this PR

dearchap:
@fjl i'm AFK and on travel. Will get to it in a day or two. Thank you so much for your patience. I have been following all the comments in this PR closely

dearchap:
@fjl sorry for the late approval. The changes look good.

Scutua (approved):

Scutua (approved):

PR #2036, opened 2025-01-01
Update LICENSE, fix copyright license year

JasonnnW3000:
Yearly copyright bump. Padding my changed LOC! 😛

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2023 urfave/cli maintainers
+Copyright (c) 2025 urfave/cli maintainers
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal


yassinebenaid (approved):
You waited a long time to do this 🤣

suzuki-shunsuke (commented):
Maybe this is my misunderstanding, but I don't think we should update this because this should be the year when this was published first. 🤔

https://www.copyrightlaws.com/copyright-symbol-notice-year/#:~:text=The%20general%20rule%20is%20that,posting%20content%20on%20your%20blog.

> Which Year Should You Use in Your Copyright Notice?
> The general rule is that the year to include in a copyright notice is the year of first publication of the work.
> First publication is when the work is made available to the public without restriction.
> This includes selling a book or leaving free copies of it in a public place or posting content on your blog.
> Publication does not include, for example, circulating copies internally to co-workers or sending copies of a book to a publisher.

suzuki-shunsuke:
In my understanding, the first publication year is required.
We can also include the last updated year, but in that case we should write it like `<first publication year> - <last updated year>` (e.g. `2021 - 2024`).

Juneezee (commented):
I'm not a copyright law expert, but a quick Google search seems to align with @suzuki-shunsuke (https://github.com/urfave/cli/pull/2036#pullrequestreview-2526832847).

> #### Which Year Should You Use in Your Copyright Notice?
>
> The general rule is that the year to include in a copyright notice is the year of first publication of the work. First publication is when the work is made available to the public without restriction.
>
> Source: https://www.copyrightlaws.com/copyright-symbol-notice-year/

I’m aware that some open-source repositories update the copyright year annually, while others don’t. For example, the Go repository still lists the copyright year as 2009: https://github.com/golang/go/blob/go1.24rc1/LICENSE#L1.

suzuki-shunsuke:
📝 LICENSE was added at 2013.

https://github.com/urfave/cli/commits/main/LICENSE
https://github.com/urfave/cli/commit/0bd64eace1a00a708503db149adcf87b9117d577

PR #2047, opened 2025-01-29
Support root level map keys in map sources

### LINKED ISSUE #2037: Support root level map keys in map sources
## Checklist

This issue is related to this one I opened here: https://github.com/urfave/cli-altsrc/issues/14

I noticed the same issue of root keys not being considered also occurs in `urfave/cli` itself:

Example:

```go
mapSource := cli.NewMapSource("config", map[any]any{
	"host": "localhost",
	"database": map[any]any{
		"url": "example.com",
	},
})

cmd := &cli.Command{
	Name: "Example",
		fmt.Println("Host:")
		fmt.Println(cmd.String("host"))
		fmt.Println("Database URL:")
		fmt.Println(cmd.String("database-url"))
		return nil
	},
	EnableShellCompletion: true,
	Flags: []cli.Flag{
		&cli.StringFlag{
			Name:    "host",
			Sources: cli.NewValueSourceChain(cli.NewMapValueSource("host", mapSource)),
		},
		&cli.StringFlag{
			Name:    "database-url",
			Sources: cli.NewValueSourceChain(cli.NewMapValueSource("database.url", mapSource)),
		},
	},
}

cmd.Run(context.Background(), os.Args)
```

will output (host is empty):
```
Host:

Database URL:
example.com
```

## Solution description

I want the above snipped to set host to `localhost` and `database-url` to `example.com`, instead of only setting the database url.

The issue is located here - because it always returns `nil` if there is no dot in a key:
https://github.com/urfave/cli/blob/main/value_source.go#L196

lukasbindreiter:
## What type of PR is this?

- bug

It fixes a bug in `MapValueSource` where lookup doesn't work for keys directly in the first level (root) of the map.

Fixes #2037 

The issue is related to this one I raised in `cli-altsrc`: https://github.com/urfave/cli-altsrc/issues/14
I'll create a similar PR there as well.

## Testing

I added a test case for this exact issue as well. It will fail without changes from the PR.
I also added another test case for the `map[string]any` special case that was in place in the code, to make sure that works to.

### DIFF
--- a/value_source.go
+++ b/value_source.go
@@ -189,12 +189,19 @@ func (ms *mapSource) GoString() string {
 	return fmt.Sprintf("&mapSource{name:%[1]q}", ms.name)
 }
 
+// Lookup returns a value from the map source. The lookup name may be a dot-separated path into the map.
+// If that is the case, it will recursively traverse the map based on the '.' delimited sections to find
+// a nested value for the key.
 func (ms *mapSource) Lookup(name string) (any, bool) {
-	// nestedVal checks if the name has '.' delimiters.
-	// If so, it tries to traverse the tree by the '.' delimited sections to find
-	// a nested value for the key.
-	if sections := strings.Split(name, "."); len(sections) > 1 {
-		node := ms.m
+	sections := strings.Split(name, ".")
+	if name == "" || len(sections) == 0 {
+		return nil, false
+	}
+
+	node := ms.m
+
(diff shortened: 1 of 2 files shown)


lukasbindreiter:
Related PR in `cli-altsrc`: https://github.com/urfave/cli-altsrc/pull/16

Juneezee (approved):
LGTM. Thanks! :blush:

PR #2055, opened 2025-02-10
Add TextFlag

### LINKED ISSUE #2051: Text flags
## Checklist

The standard library's `flag` package has had `flag.TextVar` since 1.19 which allows any type that satisfies `encoding.TextMarshaller` and `encoding.TextUnmarshaller` to be set using flags. A good use for this is to set log level using strings such as `slog.LevelVar`.

Other logging packages such as `go.uber.org/zap` satisfies these interfaces for its atomic level type. You can have a logger before the log level has been set. It'll be ready before arguments have been parsed and output structured logs in case of a parse error.

## Solution description

Implement `TextValue` and `TextFlag`.

Here's what I had to implement to get the functionality. It would be nice to see this being supported by this package.

```go
package clix

import (
	"encoding"
	"strings"

	"github.com/urfave/cli/v3"
)

type TextMarshalUnMarshaller interface {
	encoding.TextMarshaler
	encoding.TextUnmarshaler
}

type TextFlag = cli.FlagBase[TextMarshalUnMarshaller, cli.StringConfig, TextValue]

type TextValue struct {
	Value  TextMarshalUnMarshaller
	Config cli.StringConfig
}

func (v TextValue) String() string {
	text, err := v.Value.MarshalText()
	if err != nil {
		return ""
	}

	return string(text)
}

func (v TextValue) Set(s string) error {
	if v.Config.TrimSpace {
		return v.Value.UnmarshalText([]byte(strings.TrimSpace(s)))
	}

	return v.Value.UnmarshalText([]byte(s))
}

func (v TextValue) Get() any {
	return v.Value
}

func (v TextValue) Create(t TextMarshalUnMarshaller, _ *TextMarshalUnMarshaller, c cli.StringConfig) cli.Value {
	return &TextValue{
		Value:  t,
		Config: c,
	}
}

func (v TextValue) ToString(t TextMarshalUnMarshaller) string {
	text, err := t.MarshalText()
	if err != nil {
		return ""
	}

	return string(text)
}
```

I can contribute a solution based on this if so desired.

somebadcode:
* Added `TextFlag` which supports setting values for types that satisfies both `encoding.TextMarshaller` and `encoding.TextUnmarshaller` which is very handy when you want to set log levels or string-like types that satisfies the interfaces.

Fixes: #2051

## What type of PR is this?

- feature

Making it easier to set types that satisfies `encoding.TextMarshaller` and `encoding.TextUnmarshaller` such as `log/slog.LogLevelVar`, using flags as see in the standard library's  `flag`.

Fixes #2051

## Release Notes

```release-note
Added `TextFlag` to support setting values for types that satisfy both  `encoding.TextMarshaller` and `encoding.TextUnmarshaller`.
```


flag_text_test.go
```diff
@@ -0,0 +1,107 @@
...
+			flag: TextFlag{
+				Name:  "log-level",
+				Value: &slog.LevelVar{},
+				Validator: func(v TextMarshalUnMarshaller) error {
+					text, err := v.MarshalText()
+					if err != nil {
+						return err
+					}
+
+					if slices.Compare(text, []byte("INFO")) != 0 {
```
Juneezee, line 39:
```suggestion
					if !slices.Equal(text, []byte("INFO")) {
```

flag_text_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+				return
+			} else if (err != nil) == tt.wantErr {
+				// Expected error.
+				return
+			}
+
+			if got := tt.flag.GetValue(); got != tt.want {
+				t.Errorf("Value = %v, want %v", got, tt.want)
+			}
```
Juneezee, line 104: Use the [testify](https://github.com/stretchr/testify) package for assertions

Juneezee (commented):
Thanks for your contribution! :blush: 

Just a few comments regarding the tests. If possible, please add a few more test cases to cover edge cases and improve coverage.

flag_text_test.go
somebadcode: I don't use testify and I don't see it as idiomatic Go.
But I've added it and I hope that I'm using it the correct way.

flag_text.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	return v.Value.UnmarshalText([]byte(s))
+}
+
+func (v TextValue) Get() any {
+	return v.Value
+}
+
+func (v TextValue) Create(t TextMarshalUnMarshaller, _ *TextMarshalUnMarshaller, c StringConfig) Value {
```
dearchap, line 42:
```suggestion
func (v TextValue) Create(t TextMarshalUnMarshaller, p *TextMarshalUnMarshaller, c StringConfig) Value {
      b, _ := p.MarshalText()
      t.UnmarshalText(b)
```

flag_text.go
```diff
@@ -0,0 +1,56 @@
...
+type TextMarshalUnMarshaller interface {
+	encoding.TextMarshaler
+	encoding.TextUnmarshaler
+}
+
+// TextFlag enables you to set types that satisfies [TextMarshalUnMarshaller] using flags such as log levels.
+type TextFlag = FlagBase[TextMarshalUnMarshaller, StringConfig, TextValue]
+
+type TextValue struct {
+	Value  TextMarshalUnMarshaller
```
dearchap, line 17:
```suggestion
	Value  *TextMarshalUnMarshaller
```

flag_text.go
```diff
@@ -0,0 +1,56 @@
...
+	return v.Value.UnmarshalText([]byte(s))
+}
+
+func (v TextValue) Get() any {
+	return v.Value
+}
+
+func (v TextValue) Create(t TextMarshalUnMarshaller, _ *TextMarshalUnMarshaller, c StringConfig) Value {
+	return &TextValue{
+		Value:  t,
```
dearchap, line 44:
```suggestion
		Value:  p,
```

flag_text_test.go
```diff
@@ -0,0 +1,125 @@
...
+		flag    TextFlag
+		args    []string
+		want    string
+		wantErr bool
+	}{
+		{
+			name: "empty",
+			flag: TextFlag{
+				Name:  "log-level",
+				Value: &slog.LevelVar{},
```
dearchap, line 37: Note that Value is used to initialize the flag. The actual value of field should be in Destination

flag_text.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return ""
+	}
+
+	return string(text)
+}
+
+func (v TextValue) Set(s string) error {
+	if v.Config.TrimSpace {
+		return v.Value.UnmarshalText([]byte(strings.TrimSpace(s)))
```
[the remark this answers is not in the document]
somebadcode: Can you elaborate on what you mean?

flag_text_test.go
dearchap: See the flag_generic.go as an example.

flag_text.go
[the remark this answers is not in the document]
somebadcode: I've made the necessary changes.

flag_text.go
somebadcode: I've made the necessary changes.

flag_text.go
somebadcode: I've made the necessary changes.

flag_text.go
somebadcode: I've made the necessary changes.

flag_text_test.go
somebadcode: The tests should now have full coverage.

flag_text_test.go
somebadcode: I've address this in the latest version of the PR.

Juneezee (approved):

somebadcode:
I've discovered a limitation of the way the `FlagBase` forces you to have `T` and `*T`. That's not suitable in this case when when it would be better for it to be `encoding.TextMarshaler` and `encoding.TextUnmarshaler`.

So the interface that I created does not equate to those two to the extent that it allows the user of the package to just put their implementation in like you can with `flag.TextVar`.

```go
	text := &MockMarshaller{}

	cmd := &Command{
		Name: "foo",
		Flags: []Flag{
			&TextFlag{
				Name:        "mytext",
				Value:       text,
				Destination: &text,
			},
		},
	}
```

This is not valid. Because `&text` is of type `**MockMarshaller` and not `*TextMarshaler`. Type assertion for a simple case like this is unpleasant.

So I would either have to not use the `FlagBase` and re-implement everything that it gives or `FlagBase` must be refactored to allow for two disparate types to be used for value and destination like `flag.TextVar` does.

dearchap:
@somebadcode can you send me the test code you are using for this ? I think it should be like this. 

```
	text := MockMarshaller{}

	cmd := &Command{
		Name: "foo",
		Flags: []Flag{
			&TextFlag{
				Name:        "mytext",
				Value:       text,
				Destination: &text,
			},
		},
	}
```

somebadcode:
> @somebadcode can you send me the test code you are using for this ? I think it should be like this.
> 
> ```
> 	text := MockMarshaller{}
> 
> 	cmd := &Command{
> 		Name: "foo",
> 		Flags: []Flag{
> 			&TextFlag{
> 				Name:        "mytext",
> 				Value:       text,
> 				Destination: &text,
> 			},
> 		},
> 	}
> ```

Test code:
```go
type MockMarshaller struct {
	text []byte
}

var _ encoding.TextMarshaler = (*MockMarshaller)(nil)
var _ encoding.TextUnmarshaler = (*MockMarshaller)(nil)

func (m *MockMarshaller) MarshalText() ([]byte, error) {
	return m.text, nil
}

func (m *MockMarshaller) UnmarshalText(text []byte) error {
	m.text = text
	return nil
}

func TestTextFlagValueFromCommand(t *testing.T) {
	text := &MockMarshaller{}

	cmd := &Command{
		Name: "foo",
		Flags: []Flag{
			&TextFlag{
				Name:        "mytext",
				Value:       text,
				Destination: &text,
			},
		},
	}
}

```

Code that I'm trying to test:
```go
func (cmd *Command) Text(name string) TextMarshalUnmarshaler {
	if v, ok := cmd.Value(name).(TextMarshalUnmarshaler); ok {
		tracef("text available for flag name %[1]q with value=%[2]v (cmd=%[3]q)", name, v, cmd.Name)
		return v
	}

	tracef("text NOT available for flag name %[1]q (cmd=%[2]q)", name, cmd.Name)
	return nil
}
```
This is what I've added but not committed yet.

dearchap:
@somebadcode I think the problem is you are trying to use the same instance for both Value and Destination. Generally this distinction is present when you have "simple" types

```
Value: 10,
Destination: &fooInt
```

You should test your code like this

```
Value: text1,
Destination: &text2,
```
where text1 is a pointer to marshaller/unmarshaller and text2 is a simple reference

somebadcode:
> @somebadcode I think the problem is you are trying to use the same instance for both Value and Destination. Generally this distinction is present when you have "simple" types
> 
> ```
> Value: 10,
> Destination: &fooInt
> ```
> 
> You should test your code like this
> 
> ```
> Value: text1,
> Destination: &text2,
> ```
> 
> where text1 is a pointer to marshaller/unmarshaller and text2 is a simple reference

Why would pointing back to the same one matter?
The whole issue is that it's a pointer to an interface and an interface is already a pointer to the implementation. In this case, we want to accept two distinct types for `Value`  and `Destination`, `encoding.TextMarshaler` and `encoding.TextUnmarshaler` respectively, like `flag.TextVar` does.

In your example `text1` would be a pointer to the implementation, which does satisfy the interface. We don't need to have a pointer to an interface in  `Destination`, all we care about is that it satisfies a different interface. (see example further down)

Both `Value` and `Destination` should be plain interfaces and not pointers to interfaces.
To satisfy an interface, you must give it a pointer to the implementation after which the interface contains the pointer. So passing that interface around would be the same as passing a pointer to the implementation around. So `Destination` doesn't need to be a pointer and having a pointer to an interface is pointless in this case and gives us issues like this.

The type `FlagBase` is too strict in so far that it only works with simple types when it should work with any type with ease.

Because of the constraints of `FlagBase`, in `Destination` we must have a pointer to an interface that we're passing in as `Value` and to be able to do that we need to have our own interface that consists of the two interfaces that we would otherwise use for `Value` and `Destination` separately.

So `FlagBase[encoding.TextMarshaler, encoding.TextUnmarshaler, TextValue]` is essentially what we want.
So forcing `T` and `*T` is a huge hurdle in this case. It would be easier if we could have them completely separate and the need for `TextMarshalUnmarshaler` goes away and the flag would use the interfaces provided by the standard library.

Separating the two would make this implementation less complicated and those who want to add their own custom implementations for their special flags would have an easier time if it wouldn't be this constrained.

So if `FlagBase` is changed to this:
```go
type FlagBase[T any, T2 any, C any, VC ValueCreator[T, T2, C]] struct {
	// ...
	Value		T	`json:"defaultValue"`
	Destination	T2	`json:"-"
	// ...
}
```

This would enable us to do:
```
type TextFlag = FlagBase[encoding.TextMarshaler, encoding.TextUnmarshaler, StringConfig, TextValue]
```

This would make the implementation cleaner, more explicit and easier to use.

Simple types such as `StringFlag` would look like this:
```go
type StringFlag = FlagBase[string, *string, StringConfig, StringValue]
```

More explicit typing reduces the constraints and increases the versatility at minimal cost (no increased runtime cost).

Look at the standard library's `flag.TextVar` for how it should be done when it comes to accepting interfaces for flags. The default value and the destination are two distinct interfaces, neither of them are pointers to interfaces because interfaces are already pointing to the implementations, which can be incompatible implementations.

dearchap:
@somebadcode I like that idea. I dont want to increase the scope of this PR to include that as well. Do you want to do a separate PR for the FlagBase changes ?

somebadcode:
> @somebadcode I like that idea. Do you want include the FlagBase changes in this PR ? or should I make a separate PR for that ?

That would be a big change not directly related to this PR but an improvement on `FlagBase` itself. To make the commit history easier to follow, it should be a separate PR.

So this PR can be rebased on top of that once the changes are merged.

Also, what would the generic type be? `T` and `T2` seems weird, but maybe `T` and `DT`  or something.

dearchap:
@somebadcode I started working on adding a new Destination Type and ran into some issues. I dont think its that trivial to add. I dont have cycles to spare this week so I'm not sure when I will get around to it. For now I would suggest if you use GenericValue as your base or do this particular flag as a custom flag satisfying all the necessary interfaces. When we get around to making the Destination Type change I can look into compacting TextFlag into this.

somebadcode:
> @somebadcode I started working on adding a new Destination Type and ran into some issues. I dont think its that trivial to add. I dont have cycles to spare this week so I'm not sure when I will get around to it. For now I would suggest if you use GenericValue as your base or do this particular flag as a custom flag satisfying all the necessary interfaces. When we get around to making the Destination Type change I can look into compacting TextFlag into this.

I tried to do the same but ran into issues with `ArgumentBase`. So I haven't even managed to run tests yet.
So it looks like more refactoring is needed than just changing the types. The flags that cover the simple types and slices of such types seems to work but the `ArgumentBase` is much more tricky.

I'll look into changing the `TextFlag` implementation to not use the generic structs with these limitations.

flag_text_test.go
```diff
@@ -0,0 +1,169 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			cmd := &cli.Command{
+				Name:      tt.name,
+				Flags:     []cli.Flag{tt.flag},
+				Writer:    io.Discard,
+				ErrWriter: io.Discard,
+			}
+
+			err := cmd.Run(context.TODO(), append([]string{"mock"}, tt.args...))
```
dearchap, line 147: Use buildTestContext(t)

dearchap:
@somebadcode Thinking about this a bit more. Cant you achieve the same behavior using a GenericValueFlag ? Say you used

```
   GenericFlag{
       Name: "test",
       Destination: textValue(&slog.Logger{})
  }
```

Would that fit the bill itself ? I am wary of adding more code into the cli library

dearchap:
@somebadcode Here is the test code I posted in the issue https://github.com/urfave/cli/issues/2051#issuecomment-2816482242

dearchap:
If you want to remove the GenericFlag templating of FlagBase and have both GenericFlag.Value and GenericFlag.Destination be set to cli.Value type it might be better suited for your use case ?

dearchap:
Use GenericFlag instead of creating a custom flag type for this

PR #2074, opened 2025-03-10
Fix:(issue_2066) Remove dependency on golang flag library

### LINKED ISSUE #2066: **Terrible CLI Library** Both v2/v3 have serious bugs
The v3 version fails to properly recognize the `--` terminator, making it completely unusable in my use case.  
Related: #1114 (unresolved) ,#1562 (useless, insane patch) ...

The v2 version can handle the terminator correctly, but it cannot disable the comma-splitting behavior of `StringSliceFlag`, and `cli.separatorSpec` is not exported, making all features related to `separatorSpec` (like `WithSeparatorSpec`) unavailable.  
Related: #1856 ,#1552 (should never happen) ,  #1878, #1541 ...

The above issues are just what I found, there are definitely more than these
# These are all known issues, but they were closed without being fixed.

Additionally, there are weird functions like `cli.Args().Slice()`, `Args().Tail()`, etc., and I can only guess their functionality based on their output.

Since I couldn't resolve these issues, I switched to Cobra. I'm just posting this to vent.  
Simply changing the command parsing part from `urfave/cli` to `cobra`, my binary size dropped from 12MB to 5MB, which is incredible.  
Although Cobra’s API isn’t as pleasant to write as `urfave/cli`, at least it doesn’t have serious bugs and can handle basic command parsing.

Sorry for raising this issue, please feel free to close

### LINKED ISSUE #2078: Support flag sources in a required group of MutuallyExclusiveFlags
## Checklist

* [x] Did you check the manual for your release? The v3 manual is [here](https://github.com/urfave/cli/blob/main/docs/v3/index.md).

A required group of MutuallyExclusiveFlags with flags that use sources, such as environment variables, should work as expected.

Here is an example program that demonstrates the problem. Set the `ONE` environment variable and then run the program. You get the error `one of these flags needs to be provided: one, two`. If you set required to false, then the MutuallyExclusiveFlags work as expected.

```
package main

import (
	"context"
	"fmt"
	"log"
	"os"

	"github.com/urfave/cli/v3"
)

func main() {
	cmd := &cli.Command{
		Name: "test",
		MutuallyExclusiveFlags: []cli.MutuallyExclusiveFlags{
			{
				Flags: [][]cli.Flag{
					{
						&cli.StringFlag{
							Name:    "one",
							Sources: cli.NewValueSourceChain(cli.EnvVar("ONE")),
						},
					},
					{
						&cli.StringFlag{
							Name:    "two",
							Sources: cli.NewValueSourceChain(cli.EnvVar("TWO")),
						},
					},
				},
				Required: true,
			},
		},
		Action: func(_ context.Context, cmd *cli.Command) error {
			if cmd.String("one") != "" {
				fmt.Println("var one set")
			}
			return nil
		},
	}

	if err := cmd.Run(context.Background(), os.Args); err != nil {
		log.Fatal(err)
	}
}
```

## Solution description

The [mutuallyExclusiveGroupRequiredFlag Error() func](https://github.com/urfave/cli/blob/main/errors.go#L77-L88) should be updated to check sources.

A clear and concise description of any alternative solutions or
features you've considered.

dearchap:
## What type of PR is this?

_(REQUIRED)_

- cleanup
- feature

_(REQUIRED)_

Remove dependence of this library on golang's flag library. We are being constrained by parsing capabilities of golang's flag library and have to create workarounds. This PR does away with all workarounds and does the parsing of the arguments from scratch. The command code has been split into multiple files 
* command_run.go   : the start point of command.Run
* command_setup.go : everything related to setting up defaults for a Command instance
* command_parse.go : the heart of the parsing algorithm

All prior workaround from parse.go have been folded into command_parse.go. The BoolWithInverseFlag has been rewritten from scratch to do away with hacky behaviour of using a BoolFlag underneath. All tests have been updated to remove flag.FlagSet requirement. The public interface for Flags has been kept the same with the exception of addition of lifecycle functions for the Flag interface. (TBD. should those be moved into a separate Lifecycle interface)

_(REQUIRED)_

Fixes #2066 
Fixes #2078 

## Testing

go test -run=. 

## Release Notes

_(REQUIRED)_

```release-note

```


abitrolly:
> We are being constrained by parsing capabilities of golang's flag library and have to create workarounds. This PR does away with all workarounds and does the parsing of the arguments from scratch.

Hi @dearchap. It would be useful to document all the bad sides, because eventually it may improve upstream state of things for everybody. Like the fresh story here https://www.dolthub.com/blog/2025-03-07-archiving-the-dolthub-swiss-github-repository/ We just need one page to do that.

TBH I am puzzled what's wrong with Go `flag` and I don't see this PR removing the import.

dearchap:
> > We are being constrained by parsing capabilities of golang's flag library and have to create workarounds. This PR does away with all workarounds and does the parsing of the arguments from scratch.
> 
> Hi @dearchap. It would be useful to document all the bad sides, because eventually it may improve upstream state of things for everybody. Like the fresh story here https://www.dolthub.com/blog/2025-03-07-archiving-the-dolthub-swiss-github-repository/ We just need one page to do that.
> 
> TBH I am puzzled what's wrong with Go `flag` and I don't see this PR removing the import.

@abitrolly There is no issue itself with golang flag. However its parsing is not exposed so we have had to workaround to support short options, skipping flag parsing, positional arguments, detect underlying errors based on error string and so on. This gives us an opportunity to get rid of it altogether. The flag imports are still needed since we support ExtFlags i.e flags defined in the golang flag libary can co-exist with cli flag.

dearchap:
@System233 Thank you for the exhaustive test cases and the test code. Much appreciated. I have added it to the test suite and they all pass with this PR changes with 2 exceptions, both very tricky cases

### `TestStringFlagTerminator`

| Name| Input| Expected `--flag` value| Expected args | error? | Expected error message |
|----------|---------|------------------|--------------|---------|--------------|
| missing flag value due to terminator | `["test", "--flag", "--", "x"]` | - | - | Yes | `"flag needs an argument"` |

---

How do we know that the user doesnt want the "--" to be the value for the string flag ? 

### `TestBoolFlagTerminator`

| Name| Input| Expected `--flag` value| Expected args | error? | Expected error message |
|----------|---------|------------------|--------------|---------|--------------|
| bool flag with invalid non-bool value | `["test", "--flag", "x", "--", "test", "a1", "a2", "a3"]` | - | - | Yes | `"invalid syntax"` |

---
How do we know the 'x' is actually an arg that the user wants ? 

This issue here has been raised before for example bool flags are not expected to have args and sometime users do type 
```--boolFlag true``` and get a cryptic error due to true being parsed as an arg subsequently. 

Would love your thought on both these cases .

System233:
I agree with your view. In the first case, treating `"--"` as a normal string flag value is more logical and easier to implement.  

Regarding the `boolFlag` issue, the main concern is compatibility. Using `--bool-flag=true|false` or other designs like `--no-bool-flag` / `--disable-bool-flag` would be more reliable. However, if `--bool-flag true|false` was already supported, changing it now requires careful consideration of its impact. Personally, I also prefer that `boolFlag` should not accept arguments.

dearchap:
This library does not support ``` --bool-flag true|false``` so I think we can remove these two tests. Thanks a lot for the feedback. I'm going to merge this PR.

PR #2083, opened 2025-03-31
chore: Bump golangci-lint to v2

mrueg:
## What type of PR is this?

- cleanup

Bumps to golangci-lint v2.

None 

## Testing

## Release Notes

_(REQUIRED)_

```release-note
NONE
```


.github/workflows/lint.yml
```diff
@@ -34,23 +34,6 @@ jobs:
...
-
-      - name: Clone repository
-        uses: actions/checkout@v4
-
-      - name: Set up staticcheck
-        run: go install honnef.co/go/tools/cmd/staticcheck@latest
-
-      - name: Run staticcheck
-        run: staticcheck ./...
-
```
bartekpacia, line 53: Why remove it? Does new golang-ci-lint v2 include staticcheck?

mrueg: Yes it's included. https://golangci-lint.run/usage/linters/#enabled-by-default

dearchap (approved):

.github/workflows/lint.yml
Juneezee: Just FYI, the `staticcheck` linter in Golangci-lint V1 is not the same as the standalone [Staticcheck linter](https://github.com/dominikh/go-tools). The [V1 docs](https://github.com/golangci/golangci-lint/blob/v1.64.8/assets/linters-info.json#L1131) included an additional statement that says (https://github.com/golangci/golangci-lint/pull/2894):

> It's not the same thing as the staticcheck binary. The author of staticcheck doesn't support or approve the use of staticcheck as a library inside golangci-lint.

This statement has been removed in V2.

In V1, `stylecheck`, `gosimple,` and `staticcheck` cover the [4 options](https://staticcheck.dev/docs/configuration/options/) of the standalone Staticcheck linter. They have been merged in https://github.com/golangci/golangci-lint/pull/5487, that means the `staticcheck` in Golangci-lint V2 now actually works like the standalone Staticcheck linter.

.golangci.yaml
```diff
@@ -1,5 +1,23 @@
...
+      - third_party$
+      - builtin$
+      - examples$
+formatters:
+  exclusions:
+    generated: lax
+    paths:
+      - third_party$
+      - builtin$
+      - examples$
```
Juneezee, line 23: Did you copy this config from somewhere else? Because we only have the `examples` directory https://github.com/urfave/cli/tree/main/examples

Since our examples code are minimal, I think we should be able to include them in the linter too and not cause any errors. What do you think?

mrueg: This was auto-generated by golangci-lint migrate

.github/workflows/lint.yml
mrueg: thanks for the context! 

I think with v2 we should be safe unless I'm missing something?

.github/workflows/lint.yml
```diff
@@ -17,14 +17,14 @@ jobs:
...
+        uses: actions/checkout@v4
+
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
           go-version: stable
 
-      - name: Clone repository
-        uses: actions/checkout@v4
-
```
Juneezee, line 27: This fixes that the CI cache isn't getting used because of the order of our clone and Go setup.

Juneezee (approved):

PR #2088, opened 2025-04-02
Fix:(issue_2056) Add cmd.XXXArgs() functions for retrieving args

### LINKED ISSUE #2056: Clarification regarding Command.Arguments property
Hi
Guys please could you clarify the idea behind `Command.Arguments` property.

Actually my use case is a command with one required positional argument
```
app command <required-position-argument>
```
So without `Command.Arguments`  declaration `cmd.Args().Get(0)` works well.  Yet in this case there's no validation  performed for argument was provided

With adding
```golang
Arguments: []cli.Argument{
  &cli.StringArgument{
       Name: "required-position-argument",
  }
}
```
it works the very same way (without args number validation)

While trying to extend arg declaration with
```golang
  &cli.StringArgument{
       Name: "required-position-argument",
       Min: 1,
       Max: 1,
  }
```
it actually starts validating against arguments number. Yet the argument value is not accessible neither through `cmd.Args().Get(0)` nor `cmd.String("required-position-argument")`. Instead it could be reached over something like `cmd.Arguments[0].Values[0]`

So as the examples seem not covered such a deep cases it is not clear (for `v3` specifically)
 * How to make in general the case of one Required argument with a single scalar string value
 * What is the purpose of `Argument.Name` though. Are such argument values could be accessed by name somehow or is it just for help referring matter only
 * What is the general case for `Command.Arguments` declaration

Thank you.

dearchap:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

Add functions to allow Args access. Two classes of functions are available on the command
one for single value arguments and one for multi value/slice

_(REQUIRED)_

Fixes #2056 

## Testing

go test -run=Test*Args

## Release Notes

_(REQUIRED)_

```release-note
New classes of accessor functions for different argument types has been added. One for single value arguments and one for multiple values/slices

- StringArg
- StringArgs
- FloatArg
- FloatArgs
- IntArg
- IntArgs
- UintArg
- UintArgs
- TimestampArg
- TimestampArgs

Only the single value arguments can have a default value. Parsing of arguments will stop at the first argument that fails either parsing or constraint mismatch
```

### DIFF
--- a/args_test.go
+++ b/args_test.go
@@ -2,7 +2,6 @@ package cli
 
 import (
 	"context"
-	"errors"
 	"testing"
 	"time"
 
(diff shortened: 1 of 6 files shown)


args_test.go
```diff
@@ -179,43 +303,67 @@ func TestArgsUsage(t *testing.T) {
...
+		{
+			name: "no args",
+			args: []string{"foo"},
+			exp:  []string{},
+		},
+		/*{
+			name: "no arg with def value",
+			args: []string{"foo"},
+			exp:  []string{"bar"},
+		},*/
```
Juneezee, line 321: Should this case be handled? Currently the test fails when it is included

meatballhat (approved):
I fully admit that I mostly skimmed this! I'd love to know the answer to @Juneezee's question, too, but I don't have any blocking feedback.

Other notes:
- I'm reminded that I'd like to make sure argument and flag types from external packages are possible.
- The use of private members in the public `ArgumentBase` declarations still looks funny to me 🤔 e.g. the `timestampValue` in `ArgumentsBase[time.Time, TimestampConfig, timestampValue]`

args_test.go
dearchap: Added

Juneezee (approved):

PR #2090, opened 2025-04-03
Fix docs for shell completions

antimatter96:
## What type of PR is this?

- documentation

The documentation at [shell-completions](https://cli.urfave.org/v3/examples/completions/shell-completions/) seems to be broken
This PR fixes it

*Before*
Before

## Testing

Tested locally using `mkdocs serve`

*After*
After

## Release Notes

```release-note
docs: fix broken code block


antimatter96:
Also verified all other places where this might occur

grep '^\s+```'

PR #2094, opened 2025-04-13
feat!: add more integers and unsigned integers

### LINKED ISSUE #2050: Removal of integer flags causes issues.
## Checklist

The new opinionated `IntFlag` does not document on how to work with other types than `int64`. If I have a struct which I want to configure using flags, I do not want to restrict myself to using `int64` since the struct is not defined by my project. I also do not want to create custom flags for builtin types.

I've tried to use the `Action` function on the flag to set the value for a field of type `int` but it doesn't work with default values, making any workaround too convoluted.

The issue of `intValue` not being exported cause issues if I want to do any form of customization of the flag and this applies to all other flag types. Allowing reuse in custom flags would be nice.

## Solution description

Use generics to implement `IntFlag`, `Int8Flag`, `Int16Flag`, `Int32Flag` and `Int64Flag`.

One way would be to recreate the types that you removed but use generics to avoid code duplication.

```go
package clix

import (
	"strconv"
	"unsafe"

	"github.com/urfave/cli/v3"
)

type IntFlag = cli.FlagBase[int, IntegerConfig, *IntValue[int]]
type Int8Flag = cli.FlagBase[int8, IntegerConfig, *IntValue[int8]]
type Int16Flag = cli.FlagBase[int16, IntegerConfig, *IntValue[int16]]
type Int32Flag = cli.FlagBase[int32, IntegerConfig, *IntValue[int32]]
type Int64Flag = cli.FlagBase[int64, IntegerConfig, *IntValue[int64]]

// IntegerConfig is the configuration for all integer type flags.
type IntegerConfig struct {
	Base int
}

// Validate will ensure that the base is set. If base is 0 then it set to 10. This function is called by [IntValue.Create].
func (cfg *IntegerConfig) Validate() {
	if cfg.Base == 0 {
		cfg.Base = 10
	}
}

type IntValue[T int | int8 | int16 | int32 | int64] struct {
	Val  *T
	Base int
}

func (i *IntValue[T]) Create(val T, p *T, c IntegerConfig) cli.Value {
	c.Validate()

	*p = val

	return &IntValue[T]{
		Val:  p,
		Base: c.Base,
	}
}

func (i *IntValue[T]) ToString(b T) string {
	return strconv.FormatInt(int64(b), i.Base)
}

func (i *IntValue[T]) Set(s string) error {
	v, err := strconv.ParseInt(s, i.Base, int(unsafe.Sizeof(T(0))))
	if err != nil {
		return err
	}

	*i.Val = T(v)

	return err
}

func (i *IntValue[T]) Get() any { return *i.Val }

func (i *IntValue[T]) String() string { return strconv.FormatInt(int64(*i.Val), i.Base) }
```

All methods use pointer receivers as recommended by the Go documentation. https://go.dev/doc/faq#methods_on_values_or_pointers
Making all the receivers either values or pointers avoids mistakes. Linters and IDEs will pick up on this and warn you about mixing values and pointers.

I can contribute a solution based on this if so desired.

somebadcode:
* Reintroduced all variants of integers and unsigned integers.

Fixes: #2050

## What type of PR is this?

- feature

Only having flags for `int64` and `uint64` when generics can be used to make it support all of them is very limiting and forces the users of the package to create wrappers if the variable that they want to configure using a flag is not part of their project or adds unnecessary bloat by having less appropriate integer types and conversions of those integers.

Fixes #2050

## Release Notes

_(REQUIRED)_

```release-note
Added support for all integer types, signed and unsigned.
```

### DIFF
--- a/args.go
+++ b/args.go
@@ -152,9 +152,17 @@ func (a *ArgumentBase[T, C, VC]) Parse(s []string) ([]string, error) {
 
 type (
 	FloatArg     = ArgumentBase[float64, NoConfig, floatValue]
-	IntArg       = ArgumentBase[int64, IntegerConfig, intValue]
+	IntArg       = ArgumentBase[int, IntegerConfig, intValue[int]]
+	Int8Arg      = ArgumentBase[int8, IntegerConfig, intValue[int8]]
+	Int16Arg     = ArgumentBase[int16, IntegerConfig, intValue[int16]]
+	Int32Arg     = ArgumentBase[int32, IntegerConfig, intValue[int32]]
+	Int64Arg     = ArgumentBase[int64, IntegerConfig, intValue[int64]]
 	StringArg    = ArgumentBase[string, StringConfig, stringValue]
 	StringMapArg = ArgumentBase[map[string]string, StringConfig, StringMap]
 	TimestampArg = ArgumentBase[time.Time, TimestampConfig, timestampValue]
-	UintArg      = ArgumentBase[uint64, IntegerConfig, uintValue]
+	UintArg      = ArgumentBase[uint, IntegerConfig, uintValue[uint]]
+	Uint8Arg     = ArgumentBase[uint8, IntegerConfig, uintValue[uint8]]
+	Uint16Arg    = ArgumentBase[uint16, IntegerConfig, uintValue[uint16]]
+	Uint32Arg    = ArgumentBase[uint32, IntegerConfig, uintValue[uint32]]
+	Uint64Arg    = ArgumentBase[uint64, IntegerConfig, uintValue[uint64]]
 )
--- a/command_test.go
+++ b/command_test.go
@@ -355,7 +355,7 @@ func TestCommand_OnUsageError_hasCommandContext(t *testing.T) {
 	cmd := &Command{
 		Name: "bar",
 		Flags: []Flag{
-			&IntFlag{Name: "flag"},
+			&Int64Flag{Name: "flag"},
 		},
 		OnUsageError: func(_ context.Context, cmd *Command, err error, _ bool) error {
 			return fmt.Errorf("intercepted in %s: %s", cmd.Name, err.Error())
@@ -370,7 +370,7 @@ func TestCommand_OnUsageError_WithWrongFlagValue(t *testing.T) {
 	cmd := &Command{
 		Name: "bar",
 		Flags: []Flag{
-			&IntFlag{Name: "flag"},
+			&Int64Flag{Name: "flag"},
 		},
 		OnUsageError: func(_ context.Context, _ *Command, err error, _ bool) error {
 			assert.ErrorContains(t, err, "strconv.ParseInt: parsing \"wrong\"")
@@ -391,7 +391,7 @@ func TestCommand_OnUsageError_WithSubcommand(t *testing.T) {
 			},
 		},
 		Flags: []Flag{
-			&IntFlag{Name: "flag"},
+			&Int64Flag{Name: "flag"},
 		},
 		OnUsageError: func(_ context.Context, _ *Command, err error, _ bool) error {
 			assert.ErrorContains(t, err, "parsing \"wrong\": invalid syntax")
@@ -478,7 +478,7 @@ func TestCommand_Run_CustomShellCompleteAcceptsMalformedFlags(t *testing.T) {
 				Name:                  "bar",
 				Usage:                 "this is for testing",
 				Flags: []Flag{
-					&IntFlag{
+					&Int64Flag{
 						Name:  "number",
 						Usage: "A number to parse",
 					},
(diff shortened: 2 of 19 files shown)


args.go
```diff
@@ -152,9 +152,9 @@ func (a *ArgumentBase[T, C, VC]) Parse(s []string) ([]string, error) {
 
 type (
 	FloatArg     = ArgumentBase[float64, NoConfig, floatValue]
-	IntArg       = ArgumentBase[int64, IntegerConfig, intValue]
+	IntArg       = ArgumentBase[int, IntegerConfig, intValue[int]]
```
dearchap, line 155: you want to add int8. int16 etc args here as well ?

command_test.go
```diff
@@ -1245,15 +1245,15 @@ func TestCommand_Float64Flag(t *testing.T) {
...
-	var parsedIntSlice []int64
+	var parsedIntSlice []int
 	var parsedStringSlice []string
 
 	cmd := &Command{
 		Commands: []*Command{
 			{
 				Name: "cmd",
 				Flags: []Flag{
-					&IntSliceFlag{Name: "p", Value: []int64{}, Usage: "set one or more ip addr"},
```
dearchap, line 1256: Nitpick you could have changed this to Int64SliceFlag and avoided all the 
int64->int changes. Same for other places.

dearchap:
@somebadcode Can you run ```make v3approve``` and push your changes. That should fix the failing checks.

somebadcode:
@dearchap I've addressed the things you pointed out and I've added basic tests for all the new flag types.

xoxys:
This change is breaking again, right? I'm wondering about how this will be released. Will we see v4 soon?

somebadcode:
> This change is breaking again, right? I'm wondering about how this will be released. Will we see v4 soon?

Yes, it's breaking since it changes `Int` from `int64` to `int`.
So that's what I'm wondering. The same thing has to be done with floats. Not supporting all the builtin types was a huge mistake. The `flag` package should be the minimum requirement. Any package that try to be an alternative should have the same base feature-set.

I was a bit too slow to make it to the v3 release. I initially hoped that they'd held back the release of v3 because of these mistakes since the use of generics makes is really easy to cover all the different integers and floats.

So maybe correcting these short-comings will cause a v4 in a not too distant future?

dearchap:
Yes breaking change. Can't help it.

flag_uint.go
```diff
@@ -2,53 +2,89 @@ package cli
...
-	*i.val = v
+	*i.val = T(v)
 	return err
 }
 
-func (i *uintValue) Get() any { return uint64(*i.val) }
+func (i *uintValue[T]) Get() any { return T(*i.val) }
 
-func (i *uintValue) String() string { return strconv.FormatUint(uint64(*i.val), 10) }
+func (i *uintValue[T]) String() string { return strconv.FormatUint(uint64(*i.val), 10) }
```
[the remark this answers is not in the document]
somebadcode: I fixed that and uncovered more problems. I had to make sure that base is not 0, defaulting it to 10. Also, a test that tested invalid octals caused a panic because integer config had base 0 instead of 8.

flag_int_slice_test.go
```diff
@@ -0,0 +1,232 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			cmd := &cli.Command{
+				Name:      "mock",
+				Flags:     tt.flags,
+				Writer:    io.Discard,
+				ErrWriter: io.Discard,
+			}
+
+			err := cmd.Run(context.TODO(), append([]string{"mock"}, tt.arguments...))
```
dearchap, line 43:
```suggestion
			err := cmd.Run(buildTestContext(t), append([]string{"mock"}, tt.arguments...))
```

dearchap: Same for other place below and in other new test files

flag_uint_slice_test.go
```diff
@@ -0,0 +1,232 @@
...
+		expectErr bool
+	}{
+		{
+			flags: []cli.Flag{
+				&cli.UintSliceFlag{
+					Name: "numbers",
+				},
+			},
+			flagName:  "numbers",
+			arguments: []string{"--numbers", "1,2,3,4"},
```
dearchap, line 30: Also add a test like 

```
{"--numbers", "1", "--numbers" ,"2", "--numbers", 3"}
```

{"--numbers", "1,2", "--numbers" ,"3,4"}

dearchap:
@somebadcode Final request. There are 2 spots not covered by any tests. Not sure why 

flag_uint.go#L55-L59 and L61 .

Basically the String() function of uintValue[T]. Not sure why its not covered since we have Uint.String() tests in flag_test.go(see TestUintFlagXXX functions). Can you fix that ?

dearchap (approved):

PR #2096, opened 2025-04-17
fix: make fish completions work

yjp20:
## What type of PR is this?

- bug

This PR fixes a bug where `./cli completion fish` always outputs 

```
# completion fish shell completion

function __fish_completion_no_subcommand --description 'Test if there has been any subcommand yet'
    for i in (commandline -opc)
        if contains -- $i
            return 1
        end
    end
    return 0
end

complete -c completion -n '__fish_completion_no_subcommand' -f -l help -s h -d 'show help'
complete -c completion -n '__fish_completion_no_subcommand' -f -l help -s h -d 'show help'
```

This is because the command used to generate the completions is _always_ the one corresponding to the `completion` command, and it needs to know the root command.

issue #2105 

## Testing

After making these changes, completions for my CLI generated successfully.

## Release Notes

```release-note
Fixes issue with fish completions which showed completions only for the `complete` command
```

### DIFF
--- a/command_setup.go
+++ b/command_setup.go
@@ -89,7 +89,7 @@ func (cmd *Command) setupDefaults(osArgs []string) {
 	}
 
 	if cmd.EnableShellCompletion || cmd.Root().shellCompletion {
-		completionCommand := buildCompletionCommand(cmd.Name)
+		completionCommand := buildCompletionCommand(cmd)
 
 		if cmd.ShellCompletionCommandName != "" {
 			tracef(
--- a/completion.go
+++ b/completion.go
@@ -14,42 +14,42 @@ const (
 	completionFlag = "--generate-shell-completion"
 )
 
-type renderCompletion func(cmd *Command, appName string) (string, error)
+type renderCompletion func(cmd *Command) (string, error)
 
 var (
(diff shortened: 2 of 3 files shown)


suzuki-shunsuke (approved):

dearchap:
@yjp20 You dont need to make all these changes you could have done the following 

https://github.com/urfave/cli/blob/main/command_setup.go#L92

```
completionCommand := buildCompletionCommand(cmd.Root().Name)
```

yjp20:
> You dont need to make all these changes you could have done the following
> `completionCommand := buildCompletionCommand(cmd.Root().Name)`


I'm pretty sure this isn't sufficient, because you have to do `c.ToFishCompletion()` as a part of fish handler, so `c` here definitely has to be the rootCmd. You can check this out yourself if you'd like.

If you're averse to code changes, I could also change this such that the fish handler does `c.Root().ToFishCompletion()`, but frankly this feels a bit weirder to me, though I'm happy to make this change.

dearchap:
Will check and get back to you

dearchap:
Fixed by #2106

PR #2103, opened 2025-04-20
Enable to customize completion commands

suzuki-shunsuke:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

To customize completion command's fields such as `Hidden`, `Usage`, and `Description`.

_(REQUIRED)_

- https://github.com/urfave/cli/discussions/2102

## Testing

```go
package main

import (
	"context"
	"fmt"
	"log"
	"os"

	"github.com/urfave/cli/v3"
)

func main() {
	cmd := &cli.Command{
		Name: "greet",
		// EnableShellCompletion is unnecessary
		ConfigureShellCompletionCommand: func(cmd *cli.Command) {
			cmd.Hidden = false
			// cmd.Usage = "shell comletion"
			// cmd.Description = "Generate shell completion scripts"
		},
		Commands: []*cli.Command{
			{
				Name:  "hello",
				Usage: "Say hello",
					fmt.Println("Hello")
					return nil
				},
			},
		},
	}

	if err := cmd.Run(context.Background(), os.Args); err != nil {
		log.Fatal(err)
	}
}
```

```console
$ go run main.go --help           
NAME:
   greet - A new cli application

USAGE:
   greet [global options] [command [command options]]

COMMANDS:
   hello       Say hello
   help, h     Shows a list of commands or help for one command
   completion  Output shell completion script for bash, zsh, fish, or Powershell

GLOBAL OPTIONS:
   --help, -h  show help

$ go run main.go completion --help
NAME:
   greet completion - Output shell completion script for bash, zsh, fish, or Powershell

USAGE:
   greet completion [command [command options]]

DESCRIPTION:
   Output shell completion script for bash, zsh, fish, or Powershell.
   Source the output to enable completion.

   # .bashrc
   source <(greet completion bash)

   # .zshrc
   source <(greet completion zsh)

   # fish
   greet completion fish > ~/.config/fish/completions/greet.fish

   # Powershell
   Output the script to path/to/autocomplete/greet.ps1 an run it.


OPTIONS:
   --help, -h  show help
```

## Release Notes

_(REQUIRED)_

```release-note

```

### DIFF
--- a/completion.go
+++ b/completion.go
@@ -39,10 +39,28 @@ var (
 	}
 )
 
+const completionDescription = `Output shell completion script for bash, zsh, fish, or Powershell.
+Source the output to enable completion.
+
+# .bashrc
+source <($COMMAND completion bash)
+
+# .zshrc
+source <($COMMAND completion zsh)
+
+# fish
+$COMMAND completion fish > ~/.config/fish/completions/$COMMAND.fish
+
+# Powershell
+Output the script to path/to/autocomplete/$COMMAND.ps1 an run it.
+`
+
 func BuildCompletionCommand(appName string) *Command {
 	return &Command{
-		Name:   completionCommandName,
-		Hidden: true,
+		Name:        completionCommandName,
+		Hidden:      true,
+		Usage:       "Output shell completion script for bash, zsh, fish, or Powershell",
+		Description: completionDescription,
 		Action: func(ctx context.Context, cmd *Command) error {
 			return printShellCompletion(ctx, cmd, appName)
 		},
--- a/completion.go
+++ b/completion.go
@@ -56,7 +56,7 @@ $COMMAND completion fish > ~/.config/fish/completions/$COMMAND.fish
 Output the script to path/to/autocomplete/$COMMAND.ps1 an run it.
 `
 
-func BuildCompletionCommand(appName string) *Command {
+func buildCompletionCommand(appName string) *Command {
 	return &Command{
 		Name:        completionCommandName,
 		Hidden:      true,
--- a/completion.go
+++ b/completion.go
@@ -5,6 +5,7 @@ import (
 	"embed"
 	"fmt"
 	"sort"
+	"strings"
 )
 
 const (
@@ -60,7 +61,7 @@ func BuildCompletionCommand(appName string) *Command {
 		Name:        completionCommandName,
 		Hidden:      true,
 		Usage:       "Output shell completion script for bash, zsh, fish, or Powershell",
-		Description: completionDescription,
+		Description: strings.ReplaceAll(completionDescription, "$COMMAND", appName),
 		Action: func(ctx context.Context, cmd *Command) error {
 			return printShellCompletion(ctx, cmd, appName)
 		},
--- a/completion.go
+++ b/completion.go
@@ -39,7 +39,7 @@ var (
 	}
 )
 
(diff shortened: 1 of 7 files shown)


completion.go
```diff
@@ -39,10 +40,28 @@ var (
...
+source <($COMMAND completion zsh)
+
+# fish
+$COMMAND completion fish > ~/.config/fish/completions/$COMMAND.fish
+
+# Powershell
+Output the script to path/to/autocomplete/$COMMAND.ps1 an run it.
+`
+
+func BuildCompletionCommand(appName string) *Command {
```
dearchap, line 59: Why the name change to make it public ?

dearchap (commented):
@suzuki-shunsuke we have the docs to show how to invoke this https://cli.urfave.org/v3/examples/completions/shell-completions/

We generally want to keep the cli binary to a minimum so we avoid unnecessary descriptions/etc. I am not against adding the command description as part of the command but I would like others to chime in here

completion.go
suzuki-shunsuke: - https://github.com/urfave/cli/discussions/2102

To allow to change the completion command's filelds such as `Hidden`.

suzuki-shunsuke:
> suzuki-shunsuke we have the docs to show how to invoke this https://cli.urfave.org/v3/examples/completions/shell-completions/
> 
> We generally want to keep the cli binary to a minimum so we avoid unnecessary descriptions/etc. I am not against adding the command description as part of the command but I would like others to chime in here

I'm aware the document.
But it doesn't describe how to customize fields such as `Hidden`, `Usage`, and `Description` of `completion` command.
Now `buildCompletionCommand` is private, so there is no way to customize them unless overriding the command itself.

- https://github.com/urfave/cli/discussions/2102

I set the default usage and description to make it user friendly, but if you are against, it's okay to remove them.
Users can set them explicitly.

dearchap:
@suzuki-shunsuke do you want to update the docs under docs/v3/completions to include this as well ? That way I can push docs out to server once we merge this in

suzuki-shunsuke:
> suzuki-shunsuke do you want to update the docs under docs/v3/completions to include this as well ? That way I can push docs out to server once we merge this in

Sure. Updated. 23f8f563b32adb2f6000b8d17f91883ff3aa01c2

dearchap:
@suzuki-shunsuke urfave/cli is declarative in nature. So instead of exporting the completion command we should add a flag saying whether the completion command should be hidden or not. So usage would be 

```
  cmd := &Command{
      Name: "foo",
      ShowCompletionCommand: true
....
```

Advantage of this method is that we dont need to export the completion command. Disadvantage is that we can change only the hidden value for command but not usage or description.

suzuki-shunsuke:
How about this?
https://github.com/urfave/cli/commit/98bb3899329c95b3862a4b98aa2fc70d4700af7e

dearchap (approved):

dearchap:
@suzuki-shunsuke You have to run "make v3approve" since the public API has changed

suzuki-shunsuke:
I updated this pull request and added a field `ConfigureShellCompletionCommand` to *cli.Command.

```go
// ConfigureShellCompletionCommand is a function to configure a shell completion command
type ConfigureShellCompletionCommand func(*Command)
```

PR #2107, opened 2025-04-21
add test for MutuallyExclusiveFlags with After

bystones:
## What type of PR is this?

- additonal test

New test to lock in the current interaction of MututallyExclusiveFlags and the After function.

None.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/command_test.go
+++ b/command_test.go
@@ -5011,3 +5011,38 @@ func TestJSONExportCommand(t *testing.T) {
 `
 	assert.JSONEq(t, expected, string(out))
 }
+
+func TestCommand_ExclusiveFlagsWithAfter(t *testing.T) {
+	var called bool
+	cmd := &Command{
+		Name: "bar",
+		MutuallyExclusiveFlags: []MutuallyExclusiveFlags{
+			{
+				Category: "cat1",
+				Flags: [][]Flag{
+					{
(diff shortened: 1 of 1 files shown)


command_test.go
```diff
@@ -5011,3 +5011,38 @@ func TestJSONExportCommand(t *testing.T) {
...
+			return nil
+		},
+	}
+
+	require.Error(t, cmd.Run(buildTestContext(t), []string{
+		"bar",
+		"--foo", "v1",
+		"--foo2", "v2",
+	}))
+	require.True(t, called)
```
dearchap, line 5047: This is a positive test. What about if there is an error in Mutex flags say we pass in ```--foo 1 --foo2 v2``` by mistake. Would after be called still ? Do we expect it to be called ?

bystones: I am not sure whether I understand your question. 

> What about if there is an error in Mutex flags say we pass in --foo 1 --foo2 v2 by mistake.

This is exactly what the test is supposed to test. The current version of the code runs After when you pass an invalid flag combination.

Whether this is good or bad I am unsure, but either way that behavior should not change and that's why I want to add a test for it.

dearchap (approved):

PR #2108, opened 2025-04-21
use correct context in After function with subcommand

### LINKED ISSUE #2098: v3: context value added in Before missing in After when subcommands are involved
When adding values to a context in a Before function that value is not present in the After function when a subcommand was run. Without the subcommand the value is present as expected.

The following test demonstrates that behavior and should pass:

```go
func TestCommand_Run_BeforeReturnNewContextSubcommand(t *testing.T) {
        var receivedValFromAction, receivedValFromAfter string
        type key string

        bkey := key("bkey")

        cmd := &Command{
                Name: "bar",
                Before: func(ctx context.Context, cmd *Command) (context.Context, error) {
                        return context.WithValue(ctx, bkey, "bval"), nil
                },
                After: func(ctx context.Context, cmd *Command) error {
                        if val := ctx.Value(bkey); val == nil {
                                return errors.New("bkey value not found")
                        } else {
                                receivedValFromAfter = val.(string)
                        }
                        return nil
                },
                Commands: []*Command{
                        {
                                Name: "baz",
                                Action: func(ctx context.Context, cmd *Command) error {
                                        if val := ctx.Value(bkey); val == nil {
                                                return errors.New("bkey value not found")
                                        } else {
                                                receivedValFromAction = val.(string)
                                        }
                                        return nil
                                },
                        },
                },
        }

        require.NoError(t, cmd.Run(buildTestContext(t), []string{"bar", "baz"}))
        require.Equal(t, "bval", receivedValFromAfter)
        require.Equal(t, "bval", receivedValFromAction)
}
```

This test is a slight variation of an already existing test: https://github.com/urfave/cli/blob/efab65a84b56dd9c80562dbc4e9b7a52b1f97a87/command_test.go#L320-L352

bystones:
## What type of PR is this?

bugfix

Fixes #2098 

This still passes with the new test from #2107.
Instead of returning the context we could also use a pointer to a context, but that looks very strange.

## Testing

Added unit test.

## Release Notes

```release-note
Command.After now uses the correct context (as set by Command.Before) even when subcommands are involved
```

### DIFF
--- a/command_run.go
+++ b/command_run.go
@@ -91,6 +91,11 @@ outer:
 // arguments are parsed according to the Flag and Command
 // definitions and the matching Action functions are run.
 func (cmd *Command) Run(ctx context.Context, osArgs []string) (deferErr error) {
+	_, deferErr = cmd.run(ctx, osArgs)
+	return
+}
+
+func (cmd *Command) run(ctx context.Context, osArgs []string) (_ context.Context, deferErr error) {
 	tracef("running with arguments %[1]q (cmd=%[2]q)", osArgs, cmd.Name)
 	cmd.setupDefaults(osArgs)
 
@@ -102,7 +107,7 @@ func (cmd *Command) Run(ctx context.Context, osArgs []string) (deferErr error) {
 	if cmd.parent == nil {
 		if cmd.ReadArgsFromStdin {
 			if args, err := cmd.parseArgsFromStdin(); err != nil {
-				return err
+				return ctx, err
 			} else {
 				osArgs = append(osArgs, args...)
 			}
@@ -132,7 +137,7 @@ func (cmd *Command) Run(ctx context.Context, osArgs []string) (deferErr error) {
 	var rargs Args = &stringSliceArgs{v: osArgs}
 	for _, f := range cmd.allFlags() {
 		if err := f.PreParse(); err != nil {
-			return err
+			return ctx, err
 		}
 	}
 
@@ -149,7 +154,7 @@ func (cmd *Command) Run(ctx context.Context, osArgs []string) (deferErr error) {
 	tracef("using post-parse arguments %[1]q (cmd=%[2]q)", args, cmd.Name)
 
 	if checkCompletions(ctx, cmd) {
-		return nil
+		return ctx, nil
 	}
 
 	if err != nil {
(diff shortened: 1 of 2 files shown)


dearchap:
@bystones This is a really elegant solution. Thanks. Here is a sequence I want to test

1. Main command -> run
2. Main command -> Before
3. subcommand -> run
4. subcommand -> Before adds key
5. subcommnd -> Action has key
6. subcommand -> After has key
7. Main command -> After : does this have key ? Do we want to have it ?

bystones:
@dearchap I changed the test to check the values in the context at more places, hopefully that covers what you have in mind.

> Main command -> After : does this have key ? Do we want to have it ?

Generally speaking my answer would be that the context values added in a child should not go upwards to the parent. But this is already happening with Flag.Actions, see the newly added test TestCommand_Run_FlagActionContext (which passes with and without the bugfix). So we should either declare that as a bug as well or allow the same thing with Command.After.

command_test.go
```diff
@@ -351,6 +351,121 @@ func TestCommand_Run_BeforeReturnNewContext(t *testing.T) {
 	require.Equal(t, "bval", receivedValFromAction)
 }
 
+func TestCommand_Run_BeforeReturnNewContextSubcommand(t *testing.T) {
+	type key string
+
+	bkey := key("bkey")
+	bkey2 := key("bkey2")
+
+	contextValues := make(map[string]string)
```
dearchap, line 360:
```suggestion
	contextValues := make(map[string]map[string]string)
```

Make it a map from string(function name) to dict

bystones:
Added a new commit with your suggested changes.

command_run.go
```diff
@@ -262,7 +267,12 @@ func (cmd *Command) Run(ctx context.Context, osArgs []string) (deferErr error) {
 	// If a subcommand has been resolved, let it handle the remaining execution.
 	if subCmd != nil {
 		tracef("running sub-command %[1]q with arguments %[2]q (cmd=%[3]q)", subCmd.Name, cmd.Args(), cmd.Name)
-		return subCmd.Run(ctx, cmd.Args().Slice())
+
+		// It is important that we overwrite the ctx variable in the current
+		// function so any defer'd functions use the new context returned
+		// from the sub command.
+		ctx, err = subCmd.run(ctx, cmd.Args().Slice())
```
dearchap, line 274:
```suggestion
		return subCmd.run(ctx, cmd.Args().Slice())
```

dearchap (approved):

PR #2143, opened 2025-05-21
optimize: use self properties instead of declared new variables

huiyifyj:
## What type of PR is this?

- cleanup

Reduce declared variables and use self properties directly in `TakesValue` and `GetDefaultText` methods.

### DIFF
--- a/flag_impl.go
+++ b/flag_impl.go
@@ -247,8 +247,7 @@ func (f *FlagBase[T, C, V]) GetEnvVars() []string {
 
 // TakesValue returns true if the flag takes a value, otherwise false
 func (f *FlagBase[T, C, V]) TakesValue() bool {
-	var t T
-	return reflect.TypeOf(t) == nil || reflect.TypeOf(t).Kind() != reflect.Bool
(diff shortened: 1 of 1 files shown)


Juneezee (approved):

flag_impl.go
```diff
@@ -247,17 +247,15 @@ func (f *FlagBase[T, C, V]) GetEnvVars() []string {
...
 }
 
 // GetDefaultText returns the default text for this flag
 func (f *FlagBase[T, C, V]) GetDefaultText() string {
 	if f.DefaultText != "" {
 		return f.DefaultText
 	}
-	var v V
-	return v.ToString(f.Value)
+	return f.creator.ToString(f.Value)
```
dearchap, line 258: what is f.creator hasnt been initialized yet ? Define a flag(with empty DefaultText) and call GetDefaultTtext without doing a cmd.Run and see what happens. It is likely the code will crash

dearchap (commented):
While this change is well intentioned I dont think we have sufficient tests to verify

Issue #1607, opened 2022-11-30
`ToJSON` or JSON format for import/export of CLI options


abitrolly:
I wonder if it is possible to create a generic options definition serialization/exchange format in JSON?

Then the stuff like Markdown help, man pages and command line completion (https://github.com/urfave/cli/pull/1606/files) could be generated from that JSON format by external tools without bloating the size of `urfave/cli` binaries (https://github.com/urfave/cli/discussions/1594).

I already tried to play with that a bit with Python here - https://github.com/abitrolly/podman-compose/commit/0cd4197a90453dbb299aab8e8b6c3d0191e59d16 - so it is possible to extract that bit to see if `urfave/cli` export and Python `argparse` export can be compatible.

dearchap:
@abitrolly yes I'm sure its possible. The completion here in cli is minimal except for the fish completion code. Do you have any serialization formats and tools in mind for this ?

abitrolly:
I have no particular format in mind. It should probably [be designed](https://modern-json-schema.com/json-schema-is-a-constraint-system) from scratch.  The use case is to generate everything that is currently possible (man, markdown, completions) given just the JSON.

abitrolly:
Another way this could be useful - generate markdown parameter tables to `README.md` https://github.com/Zxilly/animated-timeline

PR #2150, opened 2025-06-05
Export help display functions as variables to allow custom help display logic

almas-x:
## What type of PR is this?

- feature

- Exported `ShowAppHelp`, `ShowCommandHelp`, and `ShowSubcommandHelp` as variables assigned to private functions.

- This allows users to override the help display logic for the app and commands, improving extensibility and flexibility for custom help behaviors.

## Release Notes

_(REQUIRED)_

```release-note
Exported `ShowAppHelp`, `ShowCommandHelp`, and `ShowSubcommandHelp` as variables assigned to private functions, allowing users to override help display logic for custom behaviors.
```

### DIFF
--- a/godoc-current.txt
+++ b/godoc-current.txt
@@ -131,6 +131,15 @@ COPYRIGHT:
     cli.go uses text/template to render templates. You can render custom help
     text by setting this variable.
 
+var ShowAppHelp = showAppHelp
+    ShowAppHelp is an action that displays the help
+
+var ShowCommandHelp = showCommandHelp
+    ShowCommandHelp prints help for the given command
+
+var ShowSubcommandHelp = showSubcommandHelp
+    ShowSubcommandHelp prints help for the given subcommand
+
 var SubcommandHelpTemplate = `NAME:
    {{template "helpNameTemplate" .}}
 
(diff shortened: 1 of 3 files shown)


almas-x:
> Is there something that is impossible with the current API/templates?

The readability of Go templates is quite poor, which makes debugging difficult. Additionally, they are not flexible enough to support more advanced customization, such as using colors to distinguish between commands, flags, and so on. While it is possible to add custom functions to templates by defining a custom printer function, this approach is hard to maintain. For examples of more personalized requirements, please refer to @goravel/framework/pull/766.

abitrolly:
How people are going to discover what is available in `cmd` object to produce help?

It would be nice to have some page listing all variables/props used. Maybe render them from JSON https://github.com/urfave/cli/issues/1607 so that it won't be specific to functional cli API.

dearchap:
I'm still not convinced having this would have helped @almas-x . Maybe I'm missing something .

almas-x:
> I'm still not convinced having this would have helped @almas-x . Maybe I'm missing something .

Perhaps I didn’t express myself clearly. My point is to export these `ShowXXXHelp` methods so that when printing help information, we can directly use something like fmt.Sprintf within these methods for output, instead of relying on template rendering. This way, it allows for i18n handling or adding color formatting before fmt.Sprintf, reducing the limitations compared to templates. If you still feel this doesn’t add much value, please go ahead and help close this PR.

PR #2164, opened 2025-06-25
Fix/custom help for subcommand

adrian-thurston:
## What type of PR is this?

- bug

The custom help template was not getting used in subcommand help. This is called when using `HideHelpCommand`. Updated `ShowSubcommand` to use the custom template.

It's possible a minor refactor in the help display cases is needed here, but this change fixes our immediate problem.

No known issue, just discovered locally.

None.

## Testing

Test case for the bug added.

## Release Notes

```release-note
fix: use the custom help template in a subcommand when in HideHelpCommand is active
```

### DIFF
--- a/help.go
+++ b/help.go
@@ -324,7 +324,11 @@ func ShowSubcommandHelpAndExit(cmd *Command, exitCode int) {
 
 // ShowSubcommandHelp prints help for the given subcommand
 func ShowSubcommandHelp(cmd *Command) error {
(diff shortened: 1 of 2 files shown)


dearchap:
@adrian-thurston the subcommand help is supposed to use the subcommand template. I think you have your logic backwards. If subcommand template is empty then it can use the main command template

dearchap:
@adrian-thurston When you say customization for subcommand what do you mean ?

PR #2172, opened 2025-07-26
fix: zsh autocompletions

mahyarmirrashed:
Actually auto-complete.

For some reason, the auto-completion wasn't working for me. Running `<prog> -- --generate-shell-completion` just ran the command, and did not provide the correct results. The reason that was happening (if I had to guess) is because `--` is being treated as an end of flags. As such, the literal string `--generate-shell-completion` was being passed to the underlying program.

## What type of PR is this?

- bug

Current completions do not work for ZSH. I spent 3 hours trying to figure out what it was.

N/A

## Release Notes

```release-note
fix: zsh autocompletions
```

### DIFF
--- a/autocomplete/zsh_autocomplete
+++ b/autocomplete/zsh_autocomplete
@@ -9,7 +9,7 @@ _%[1]s() {
 	current=${words[-1]} # -1 means "the last element"
 	if [[ "$current" == "-"* ]]; then
 		# Current word starts with a hyphen, so complete flags/options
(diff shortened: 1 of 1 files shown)


mahyarmirrashed:
Admittedly, this is just a patch over the old problem. There probably needs to be actual source code changes to make it work properly.

e.g.

```
mahyar@tiger tmp % ./greet add - --generate-shell-completion
--config:path to config file
--help:show help
mahyar@tiger tmp % ./greet add --conf --generate-shell-completion
help:Shows a list of commands or help for one command
```

PR #2177, opened 2025-08-09
Add installation instructions for gfmrun

jllovet:
## What type of PR is this?

- Documentation

Contributor docs did not include that `gfmrun` is a dependency for running `make all`. The PR adds a few lines with instructions on doing that.

This is my first PR for this repo. Any suggestions - or a correction if I'm off base here - would be helpful.

## Testing

On a machine that does not have`gfmrun`, run:

```
git clone https://github.com/urfave/cli.git
```

Run

```
make
```

Final steps should fail.

Now run installation described in PR:

```
go install github.com/urfave/gfmrun/cmd/gfmrun@latest
```

Re-run 

```
make
```

Final steps should pass.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -66,7 +66,7 @@ verify one's changes are harmonious in nature. The same steps are also run durin
 [continuous integration
 phase](https://github.com/urfave/cli/blob/main/.github/workflows/test.yml).
 
-`gfmrun` is required to run the examples, and without it `make all` may fail.
+`gfmrun` is required to run the examples, and without it `make all` will fail.
 
 You can find `gfmrun` here:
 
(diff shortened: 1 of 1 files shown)


docs/CONTRIBUTING.md
```diff
@@ -66,6 +66,18 @@ verify one's changes are harmonious in nature. The same steps are also run durin
 [continuous integration
 phase](https://github.com/urfave/cli/blob/main/.github/workflows/test.yml).
 
+`gfmrun` is required to run the examples, and without it `make all` may fail.
```
dearchap, line 69:
```suggestion
`gfmrun` is required to run the examples, and without it `make all` will fail.
```

dearchap (approved):

jllovet:
@meatballhat's suggestion seems reasonable to me, based on my understanding of `build.go`. `gfmrun` will get installed by  `EnsureGfmrunActionFunc` if it's missing. Should we refashion this portion of the docs to suggest that a user run `make ensure-gfmrun` directly? While we're at it, are there other things that would be good to check or run for someone trying to contribute?

PR #2196, opened 2025-09-02
feat: add ability to override usage text of default help command

Maks1mS:
## What type of PR is this?

- feature

This PR introduces the ability to override the usage text of the default help command.

---

In v2, I implemented in my app translations for the help command by overriding the usage text after `app.Setup()`:
```go
func Localize(app *cli.App) {
	app.Setup()
	// ...
	for _, cmd := range app.Commands {
		if cmd.Name == "help" {
			cmd.Usage = gotext.Get("Shows a list of commands or help for one command")
		}
	}
}
```

However, v3 removed the `cli.App` and `Setup` method, making this approach impossible. 
This PR adds a way to customize help command text, enabling use cases like localization.

N/A

## Release Notes

```release-note
Feature: add ability to override usage text of default help command
```


dearchap (approved):

dearchap:
@Maks1mS can you fix the gofmt error check ?

PR #2214, opened 2025-10-21
Fix: Make DefaultText behaviour consistent

dearchap:
## What type of PR is this?

_(REQUIRED)_

- cleanup

_(REQUIRED)_

The behaviour of GetDefaultText was too unpredictable, depending on whether the DefaultText is set or whether Value is set and what not. This change takes the guesswork out of flag_impl.go and puts the logic into docs.go where it should be

_(REQUIRED)_

## Testing

make test

## Release Notes

_(REQUIRED)_

```release-note
flag.GetDefaultText returns the DefaultText set by the user without any modification
```

### DIFF
--- a/docs.go
+++ b/docs.go
@@ -107,9 +107,12 @@ func stringifyFlag(f Flag) string {
 
 	// don't print default text for required flags
 	if rf, ok := f.(RequiredFlag); !ok || !rf.IsRequired() {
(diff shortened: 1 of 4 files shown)


oprudkyi:
@dearchap  yes, sure. Sorry

PR #2227, opened 2025-11-21
Fix:(issue_2225) Make external flags more robust.

### LINKED ISSUE #2225: Improve `AllowExtFlags` robustness
## Checklist

* [X] Are you running the latest v3 release?
* [X] Did you check the manual for your release?
* [X] Did you perform a search about this feature?

- cryptic error messages like `"syntax error: expected flag.go:234"`
- errors requiring external boolean flags have a value when used
- avoidable astonishment when `AllowExtFlags` flags behave differently from identical flags defined using urfave/cli

Recently, while addressing CVE-2025-52881, I found urfave/cli/v3 became a transitive dependency of numerous programs with a large number of dependencies, many of which define flags using the standard golang "flags" package.

The support for external flags in urfave/cli/v3, `AllowExtFlags` [added code to set the value of each external flag to its default value.](https://github.com/urfave/cli/commit/9bb00e2e9d78514163e08bc45ff95017171e559b) While it seems reasonable that every flag would accept its default value, that is not always the case.

It's not clear to me that urfave/cli should ever fail in this case. That said, it is possible to identify a tiny subset of flags that are most likely to reject their default value, namely variable flags that use a pointer to a structure as the variable and lazily use the string representation of the zero-value of the structure instead of an empty string as the default. In any case, where urfave/cli does fail, the user would benefit from additional context.

All of the urfave/cli flag types follow a Value+Flag paradigm except `extFlag`. As a result, an identical flag brought in via `AllowExtFlags` behaves differently in perhaps subtle ways.

## Solution description

Reimplement extFlag using delegation to implement the Value+Flag paradigm, add context to any errors reported when setting the value to the default value, and suppress those errors some or all of the time.

One could insist that all external flags accept the default value, but: 1) the `"flags"` package explicitly describes the only purpose for the default value as the usage message, and 2) a small change can make a big difference in the transitive dependency footprint introducing many new dependencies some of which might be externally sourced.

bbadour:
Use delegation to implement the Value+Flag paradigm used for other flag types.

Implement all of the Flag interfaces except Count.

Borrowed heavily from flag_bool.go and flag_generic.go

Add context to the error message when setting an external flag's value to its default value to avoid cryptic error messages like: `"syntax error: expected file.go:234"`

And suppress the error for odd-ball external flags that report the string representation of the zero-value of some structure as the default, but do not accept that string as input. Detect such odd-balls by their `Get()` method returning `nil`.

## What type of PR is this?

- feature

Solve:

   - cryptic error messages like "syntax error: expected flag.go:234"
   - errors requiring external boolean flags have a value when used
   - avoidable astonishment when AllowExtFlags flags behave differently from identical flags defined using urfave/cli

Fixes #2225

## Testing

Installed gfmrun and make passes. Also the same code passes the test cases for a large code base that uses the feature.

## Release Notes

```release-note
Improve support for `AllowExtFlags`
```

### DIFF
--- a/flag_ext.go
+++ b/flag_ext.go
@@ -1,14 +1,74 @@
 package cli
 
-import "flag"
+import (
+	"context"
+	"flag"
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+var _ Value = (*externalValue)(nil)
+
+var (
+	_ Flag                        = (*extFlag)(nil)
+	_ ActionableFlag              = (*extFlag)(nil)
+	_ CategorizableFlag           = (*extFlag)(nil)
+	_ DocGenerationFlag           = (*extFlag)(nil)
+	_ DocGenerationMultiValueFlag = (*extFlag)(nil)
+	_ LocalFlag                   = (*extFlag)(nil)
+	_ RequiredFlag                = (*extFlag)(nil)
+	_ VisibleFlag                 = (*extFlag)(nil)
+)
+
+// -- Value Value
+type externalValue struct {
+	e *extFlag
+}
+
+// Below functions are to satisfy the flag.Value interface
+
+func (ev *externalValue) Set(s string) error {
+	if ev != nil && ev.e.f != nil {
+		return ev.e.f.Value.Set(s)
+	}
+	return nil
+}
+
+func (ev *externalValue) Get() any {
+	if ev != nil && ev.e.f != nil {
+		return ev.e.f.Value.(flag.Getter).Get()
+	}
+	return nil
+}
+
+func (ev *externalValue) String() string {
+	if ev != nil && ev.e.f != nil {
+		return ev.e.String()
+	}
+	return ""
+}
+
+func (ev *externalValue) IsBoolFlag() bool {
+	if ev == nil || ev.e.f == nil {
+		return false
+	}
+	bf, ok := ev.e.f.Value.(boolFlag)
+	return ok && bf.IsBoolFlag()
+}
 
 type extFlag struct {
-	f *flag.Flag
+	f        *flag.Flag
+	category string
 }
 
 func (e *extFlag) PreParse() error {
 	if e.f.DefValue != "" {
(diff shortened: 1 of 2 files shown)


flag_ext.go
```diff
@@ -30,6 +90,43 @@ func (e *extFlag) Names() []string {
...
+	return kind == reflect.Slice || kind == reflect.Map
+}
+
+// IsRequired returns whether or not the flag is required
+func (e *extFlag) IsRequired() bool {
+	return false
+}
+
 func (e *extFlag) IsSet() bool {
 	return false
```
bbadour: Migrating some code from v1, I found this could be less astonishing if, instead, it were:

```
        if e == nil || e.f == nil || e.f.Value == nil {
                return false
        }
        return e.f.Value.String() != e.f.DefValue
```

flag_ext.go
```diff
@@ -1,14 +1,74 @@
...
-	f *flag.Flag
+	f        *flag.Flag
+	category string
 }
 
 func (e *extFlag) PreParse() error {
 	if e.f.DefValue != "" {
-		return e.Set("", e.f.DefValue)
+		// suppress errors for write-only external flags that always return nil
+		if err := e.Set("", e.f.DefValue); err != nil && e.f.Value.(flag.Getter).Get() != nil {
```
dearchap:
```suggestion
		if err := e.Set("", e.f.DefValue); err != nil  {
```

do you need the Get() != nil check ?

flag_ext.go
```diff
@@ -61,3 +158,61 @@ func (e *extFlag) GetDefaultText() string {
 func (e *extFlag) GetEnvVars() []string {
 	return nil
 }
+
+// RunAction executes flag action if set
+func (e *extFlag) RunAction(ctx context.Context, cmd *Command) error {
```
dearchap: you dont need to define this method. interface casting will check if Flag is an ActionableFlag and call accordingly

PR #2258, opened 2026-02-05
Fix issue: 2257 Next arg flag name, no default cmd

bbadour:
If the next argument matches a flag name, and there is no default command, a spurious empty string gets inserted before the flag name.

Existing code that expects the flag name in element 0 fails because it is now in element 1 with an empty string in element 0.

When there is no default command and an empty string is detected in element 0 (i.e. `cmd.parsedArgs.First()`), change `cmd.parsedArgs` to omit the first empty element.

## What type of PR is this?

- bug

Removes spurious empty string ahead of args when the next argument matches a flag name and there is no DefaultCommand.

Fixes 2257

## Testing

Ran tests at my organization that were failing due to the bug.

## Release Notes

```
Fix spurious empty string in arguments when the next argument matches a flag name and there is no DefaultCommand.
```

### DIFF
--- a/command_test.go
+++ b/command_test.go
@@ -1413,6 +1413,33 @@ func TestCommand_UseShortOptionAfterSliceFlag(t *testing.T) {
 	assert.Equal(t, expected, name)
 }
 
+func TestCommand_UseShortOptionWithArg(t *testing.T) {
+	var rootPath string
+	cmd := &Command{
+		UseShortOptionHandling: true,
+		Commands: []*Command{
+			{
+				Name:  "short",
+				Usage: "complete a task on the list",
+				Arguments: []Argument{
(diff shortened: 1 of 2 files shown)


bbadour:
> @bbadour can you add a test ?

Done.

command_test.go
```diff
@@ -5444,3 +5444,37 @@ func TestCommand_ParallelRun(t *testing.T) {
...
+			},
+			Action: func(ctx context.Context, c *Command) error {
+				gotArgs = c.Args()
+				return nil
+			},
+		},
+	}
+
+	// flag has to be set to execute code path
+	if err := app.Commands[0].Flags[0].Set("list", "true"); err != nil {
```
dearchap: Maybe set the default value of list to true in flag ? Or set an env variable so that it triggers the Set rather than us having to manually do it

PR #2280, opened 2026-03-15
Fix 2257 - Remove incorrect flag check

### LINKED ISSUE #2257: Problem with next argument matching flag name with no default command.
https://github.com/urfave/cli/blob/fd14621cbcd644aa35483480d40f7cdf7b69f4c0/command_run.go#L276

In the case where a command has no DefaultCommand and has a flag with the name of the next argument, the line above inserts a spurious empty string ahead of the flag name. This causes existing code expecting the flag name in element zero to fail because an empty string is there instead and the expected flag name is in element one.

MohitPanchariya:
## What type of PR is this?

_(REQUIRED)_

- bug

_(REQUIRED)_

### Current Behavior
- If the argument passed with a command isn't a sub command, there are two checks performed - check if the command has a default command, and check if the argument passed is a flag.
- In the case that the default command doesn't exist, and the argument passed is the name of a flag, the sub command resolves to nil, which it was previously as well. Implying the check for the flag name doesn't have affect the sub command resolution.

- Since a default command doesn't exist, the cmd.parsedArgs list ends up with an empty argument as its zeroth element. 

### Changed Behavior
- This PR removes the flag name check since it doesn't have an effect on the sub command resolution.

_(REQUIRED)_

Fixes #2257 

## Testing

Tested the changes locally with the following program
```
package main

import (
	"context"
	"fmt"
	"os"

	"github.com/urfave/cli/v3"
)

func main() {
	topCommand := cli.Command{
		Name:  "topCommand",
		Usage: "topCommand usage",
		Flags: []cli.Flag{
			&cli.StringFlag{
				Name: "flag1",
			},
		},
		Action: func(ctx context.Context, c *cli.Command) error {
			fmt.Printf("%#v\n", c.Args())
			return nil
		},
	}
	subCommand1 := cli.Command{
		Name:  "subCommand1",
		Usage: "subCommand1 description",
		Action: func(ctx context.Context, c *cli.Command) error {
			fmt.Println("I am subCommand1")
			return nil
		},
	}
	subCommand2 := cli.Command{
		Name:  "subCommand2",
		Usage: "subCommand2 description",
		Action: func(ctx context.Context, c *cli.Command) error {
			fmt.Println("I am subCommand2")
			return nil
		},
	}

	topCommand.Commands = append(topCommand.Commands, &subCommand1, &subCommand2)
	topCommand.Run(context.Background(), os.Args)
}
```

Current state (observe the empty string in the arguments list)
```
./main --flag1 arg1 flag1
&cli.stringSliceArgs{v:[]string{"", "flag1"}}
```

With the change,
```
./main --flag1 arg1 flag1
&cli.stringSliceArgs{v:[]string{"flag1"}}
```

## Release Notes

_(REQUIRED)_

```release-note
NONE
```


MohitPanchariya:
@dearchap Just saw the comments today because I kept wondering why the PR wasn't reviewed yet and decided to take a look. I likely missed the mail. I will try to address the comments over the weekend.

MohitPanchariya:
@dearchap Added a test case.

@yassinebenaid Since its not a change in the expected behavior, or interface, I put `NONE` in the release notes.

dearchap (approved):

PR #2288, opened 2026-03-20
flag: replace regexp use

kolyshkin:
## What type of PR is this?

- cleanup

Commit ec05a8d3 introduces the only (non-test) user of regexp package.
It is pretty easy to not use regexp here, so let's get rid of it.

References:
 - https://github.com/opencontainers/runc/pull/5184#issuecomment-4100226858
 - https://github.com/opencontainers/runc/pull/3460

None

None

## Testing

Assuming existing tests cover the change.

## Release Notes

```release-note
Remove regexp dependency.
```

### DIFF
--- a/flag.go
+++ b/flag.go
@@ -3,7 +3,6 @@ package cli
 import (
 	"context"
 	"fmt"
-	"regexp"
 	"slices"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


kolyshkin:
@dearchap PTAL

PR #2314, opened 2026-04-15
flag: prevent BoolWithInverseFlag.String from panicking without a tab

alliasgher:
## Summary

\`BoolWithInverseFlag.String()\` panics with \`slice bounds out of range [-1:]\` when the configured \`FlagStringer\` returns a string that does not contain a tab character.

\`\`\`go
out := FlagStringer(bif)
i := strings.Index(out, \"\\t\")  // -1 when no tab
// ...
out[i:]                            // panic: slice bounds out of range [-1:]
\`\`\`

This happens with custom stringers and with the default stringer when the flag doesn't implement \`DocGenerationFlag\`.

## Fix

Clamp the tab index to 0 when no tab is found. The full \`FlagStringer\` output is then used as the suffix portion of the formatted string, matching the intent of callers using a custom stringer without the usual tab layout.

Fixes #2303

## Checklist

- [x] Test added covering both multi-character and single-character flag names with a no-tab \`FlagStringer\`
- [x] Full test suite passes

### DIFF
--- a/flag_bool_with_inverse.go
+++ b/flag_bool_with_inverse.go
@@ -179,6 +179,14 @@ func (bif *BoolWithInverseFlag) String() string {
 		prefix = "-"
 	}
 
+	// Guard against a FlagStringer that returns a string without a tab (e.g.
+	// a custom stringer or the default stringer when the flag does not
+	// implement DocGenerationFlag). In that case treat the entire output as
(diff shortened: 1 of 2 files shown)


alliasgher:
Closing — this is covered by #2311 which already merged.

Issue #2260, opened 2026-02-06
A recent change breaks custom help Commands


bbadour:
The recent change at https://github.com/urfave/cli/blob/fd14621cbcd644aa35483480d40f7cdf7b69f4c0/command_run.go#L144C29-L144C53 assumes any command named "help" must be the default Help handler. This breaks applications with custom Help commands.

There are lots of reasons to want a custom Help command. e.g. to write to the cmd.Root()'s OutWriter or ErrWriter instead of to stdout.

PR #2319, opened 2026-04-19
fix: parse flags for help subcommand (#2271)

### LINKED ISSUE #2271: v3.6.2 regression: flag parsing skipped for help subcommands
*Filed by Claude on behalf of @silverwind*

## Description

PR #2245 introduced a regression in v3.6.2 where flag pre-parsing is skipped for any command named "help". This breaks applications that use flags with help subcommands, as the flags are never parsed.

The problematic change in `command_run.go`:

```diff
-	if cmd.isCompletionCommand {
+	if cmd.isCompletionCommand || cmd.Name == helpName {
		tracef("special command detected, skipping pre-parse (cmd=%[1]q)", cmd.Name)
		cmd.parsedArgs = args
		return ctx, cmd.Action(ctx, cmd)
	}
```

By matching on `cmd.Name == helpName`, this skips pre-parsing for ALL commands named "help", not just the internal completion command. This causes flags passed to help subcommands (e.g. `myapp help -c mycmd`) to be silently ignored.

## Steps to reproduce

1. Define a CLI app with a flag (e.g. `-c`) on the help command
2. Run `myapp help -c value`
3. The flag value is never parsed

## Expected behavior

Flags should be parsed normally for help commands, as they were in v3.6.1.

## Actual behavior

Flag pre-parsing is skipped entirely, flags are ignored.

## Workaround

Pin `urfave/cli/v3` to v3.6.1 or earlier via `replace` directive in `go.mod`.

## Related

- #2260 (same root cause, different symptom)
- #2245 (PR that introduced the regression)

barry3406:
## What type of PR is this?

- bug

PR #2245 introduced a regression in v3.6.2 by short-circuiting flag pre-parsing for any command named \`help\`:

```go
if cmd.isCompletionCommand || cmd.Name == helpName {
    cmd.parsedArgs = args
    return ctx, cmd.Action(ctx, cmd)
}
```

The original goal of #2245 was to let \`help\` and \`completion\` run without tripping required-flag enforcement, but skipping pre-parse entirely also drops flags and arguments on user-supplied help commands (#2271) and breaks custom help commands (#2260).

This PR keeps the pre-parse path intact for the help command and instead short-circuits \`checkAllRequiredFlags\` for \`help\` and completion, which was the actual intent. The completion command still bypasses pre-parse as it did before #2245.

- \`command_run.go\`: restore the pre-#2245 condition (\`isCompletionCommand\` only).
- \`command.go\`: exempt \`help\` and completion commands from the required-flag check.

Fixes #2271

## Testing

Added \`Test_HelpCommand_ParsesFlags\` in \`help_test.go\`, which fails on current main and passes with this change. The existing \`Test_HelpCommand_RequiredFlagsNoDefault\` added in #2245 still passes.

\`\`\`
make lint vet test
\`\`\`

## Release Notes

\`\`\`release-note
Fix v3.6.2 regression where flags and arguments on user-supplied help subcommands were silently dropped.
\`\`\`

### DIFF
--- a/command.go
+++ b/command.go
@@ -408,6 +408,12 @@ func (cmd *Command) checkRequiredFlag(f Flag) (bool, string) {
 }
 
 func (cmd *Command) checkAllRequiredFlags() requiredFlagsErr {
+	// The help and completion commands are allowed to run without
+	// enforcement of required flags, since they do not invoke user
+	// actions that depend on those flag values.
+	if cmd.Name == helpName || cmd.isCompletionCommand {
+		return nil
+	}
 	for pCmd := cmd; pCmd != nil; pCmd = pCmd.parent {
 		if err := pCmd.checkRequiredFlags(); err != nil {
 			return err
--- a/command_run.go
+++ b/command_run.go
@@ -141,8 +141,8 @@ func (cmd *Command) run(ctx context.Context, osArgs []string) (_ context.Context
 	var rargs Args = &stringSliceArgs{v: osArgs}
 	var args Args = &stringSliceArgs{rargs.Tail()}
 
-	if cmd.isCompletionCommand || cmd.Name == helpName {
-		tracef("special command detected, skipping pre-parse (cmd=%[1]q)", cmd.Name)
+	if cmd.isCompletionCommand {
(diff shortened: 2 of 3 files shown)


aryounce (approved):
Thanks for a concise fix.

PR #2330, opened 2026-05-16
v3: yield the version flag's -v alias to a user-defined flag

### LINKED ISSUE #2229: --verbose flag gets interpreted as "print version and exit"
## My urfave/cli version is

_**3.6.1**_

## Checklist

## Dependency Management

- My project is using vendoring.

## Describe the bug

After the upgrade from v2 to v3, our command starting always printing the version and exiting when passed `--verbose`.

We narrowed it down to the way `--verbose` is defined:

```go
&cli.BoolFlag{
	Name:    "verbose",
	Aliases: []string{"v"},
},
```

It looks like the `v` alias conflicts with [the default version flag](https://cli.urfave.org/v3/examples/flags/basics/#version-flags).

## To reproduce

* Define a boolean flag that has `v` as alias.
* Call the command, setting the flag to true.

[See in Go playground](https://go.dev/play/p/yfNGl3GTAer).

## Observed behavior

Command prints version and exits.

## Expected behavior

One of:

* The clash on alias `v` is detected and the library returns an error or panics.
* The custom flag definition completely replaces default version flag.

## Additional context

—

## Want to fix this yourself?

I'm open to it!

```
go version go1.25.4 darwin/arm64
```

```
AR='ar'
CC='clang'
CGO_CFLAGS='-O2 -g'
CGO_CPPFLAGS=''
CGO_CXXFLAGS='-O2 -g'
CGO_ENABLED='1'
CGO_FFLAGS='-O2 -g'
CGO_LDFLAGS='-O2 -g'
CXX='clang++'
GCCGO='gccgo'
GO111MODULE=''
GOARCH='arm64'
GOARM64='v8.0'
GOAUTH='netrc'
GOBIN=''
GOCACHE='/Users/tcard/Library/Caches/go-build'
GOCACHEPROG=''
GODEBUG=''
GOENV='/Users/tcard/Library/Application Support/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFIPS140='off'
GOFLAGS=''
GOGCCFLAGS='-fPIC -arch arm64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -ffile-prefix-map=/var/folders/l3/h28qjr993sl4y2cgmkl06v980000gn/T/go-build751984059=/tmp/go-build -gno-record-gcc-switches -fno-common'
GOHOSTARCH='arm64'
GOHOSTOS='darwin'
GOINSECURE=''
GOMODCACHE='/Users/tcard/go/pkg/mod'
GOOS='darwin'
GOPATH='/Users/tcard/go'
GOPROXY='https://proxy.golang.org,direct'
GOROOT='/Users/tcard/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.4.darwin-arm64'
GOSUMDB='sum.golang.org'
GOTELEMETRY='local'
GOTELEMETRYDIR='/Users/tcard/Library/Application Support/go/telemetry'
GOTMPDIR=''
GOTOOLCHAIN='auto'
GOTOOLDIR='/Users/tcard/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.4.darwin-arm64/pkg/tool/darwin_arm64'
GOVCS=''
GOVERSION='go1.25.4'
GOWORK=''
PKG_CONFIG='pkg-config'
```

c-tonneslan:
Fixes #2229.

The default version flag carries \`-v\` as an alias. A user-defined flag that also wants \`-v\` (canonical case: \`--verbose / -v\`) wasn't actually broken at parse time: \`-v\` ended up setting the user flag. But \`checkVersion\` went through \`cmd.Bool\`, which resolves aliases, so it asked for the value of \`\"v\"\` and got back the user flag's value. The end result: passing \`-v\` was silently treated as \"print version and exit.\"

Two narrow changes:

- During root setup, drop any aliases from the local copy of the version flag that are already claimed by a user-defined flag's name or alias. Keeps the user flag the sole owner of the short form.
- \`checkVersion\` now finds the actual version flag attached to the command (matching against the canonical primary name from the global VersionFlag) and checks \`IsSet\` directly, so it can't be fooled by a same-named alias on a different flag.

Added \`TestVersionFlagYieldsAliasToUserFlag\` covering the reporter's reproducer.

### DIFF
--- a/command_setup.go
+++ b/command_setup.go
@@ -99,6 +99,10 @@ func (cmd *Command) setupDefaults(osArgs []string) {
 			var localVersionFlag Flag
 			if globalVersionFlag, ok := VersionFlag.(*BoolFlag); ok {
 				flag := *globalVersionFlag
+				// Drop any alias a user flag already claims (e.g. -v
+				// for --verbose) so the user flag wins but --version
+				// still works. See #2229.
+				flag.Aliases = dropClashingAliases(flag.Aliases, cmd.allFlags(), flag.Name)
 				localVersionFlag = &flag
 			} else {
 				localVersionFlag = VersionFlag
@@ -254,4 +258,32 @@ func (cmd *Command) ensureHelp() {
 			}
 		}
 	}
+}
+
+// dropClashingAliases removes aliases from `aliases` that are already
+// claimed by a flag in `userFlags` (either as a primary name or as one
+// of its own aliases). Aliases equal to `selfName` are kept so the
+// flag's primary name doesn't accidentally remove itself.
+func dropClashingAliases(aliases []string, userFlags []Flag, selfName string) []string {
+	if len(aliases) == 0 || len(userFlags) == 0 {
+		return aliases
+	}
+	taken := map[string]struct{}{}
+	for _, f := range userFlags {
+		for _, n := range f.Names() {
+			taken[n] = struct{}{}
+		}
+	}
+	kept := aliases[:0:0]
+	for _, a := range aliases {
+		if a == selfName {
+			kept = append(kept, a)
+			continue
+		}
+		if _, ok := taken[a]; ok {
+			continue
+		}
+		kept = append(kept, a)
+	}
+	return kept
 }
--- a/command_test.go
+++ b/command_test.go
@@ -3019,6 +3019,54 @@ func TestCustomFlagsUsed(t *testing.T) {
 	assert.NoError(t, err, "Run returned unexpected error")
 }
 
+// Regression for #2229. When a user flag claims the -v alias, only that
+// alias is yielded: --version must still print the version.
+func TestVersionFlagWorksWhenAliasYieldedToUserFlag(t *testing.T) {
+	buf := new(bytes.Buffer)
+	called := false
+
+	cmd := &Command{
+		Name:    "boom",
+		Version: "0.1.0",
+		Writer:  buf,
+		Flags: []Flag{
+			&BoolFlag{Name: "verbose", Aliases: []string{"v"}},
+		},
+		Action: func(_ context.Context, _ *Command) error {
+			called = true
+			return nil
+		},
+	}
+
+	err := cmd.Run(buildTestContext(t), []string{"boom", "--version"})
+
+	assert.NoError(t, err)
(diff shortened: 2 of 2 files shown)


dearchap:
@c-tonneslan Can you rebase off main and fix test cov ?

dearchap:
## Review Summary

### 1. `dropClashingAliases` in `command_setup.go` — ✅ Merge with fix

Good improvement over the existing `flagNamesInUse` approach. Instead of dropping the entire `--version` flag when `-v` clashes, it only removes the clashing `-v` alias, preserving `--version`.

**Minor issue:** `dropClashingAliases` is called with `cmd.Flags` but `flagNamesInUse` just below uses `cmd.allFlags()` (which also includes `MutuallyExclusiveFlags`). If a mutually exclusive flag group claims `-v`, `dropClashingAliases` won't see it. Pass `cmd.allFlags()` instead of `cmd.Flags` for consistency.

### 2. `checkVersion` rewrite in `help.go` — ❌ Drop this change

Keep the current `cmd.versionFlag.IsSet()` implementation. The proposed name-based matching (`names[0] == primary`) is fragile — it could match a user-defined flag whose primary name happens to be `"version"`, causing `--version` to trigger `ShowVersion` based on the wrong flag's state.

### 3. New test — ⚠️ Redundant; consider expanding

The existing `TestCommand_Run_CustomFlagCanUseVersionAlias` already covers the same scenario. Consider testing that `--version` still works after `-v` is yielded to a user flag — neither test currently covers that.

c-tonneslan:
Rebased on main, conflict is gone.

Addressed the review:
- `dropClashingAliases` now takes `cmd.allFlags()` instead of `cmd.Flags`, so a mutually-exclusive group claiming `-v` is also seen.
- Dropped the `checkVersion` rewrite. Once the clashing alias is stripped in setup, `cmd.versionFlag.IsSet()` already does the right thing, so help.go is untouched now.
- Replaced my redundant test (the existing `TestCommand_Run_CustomFlagCanUseVersionAlias` already covers `-v`) with one that checks `--version` still prints after `-v` is yielded. That case fails on main today. Added a small table test for `dropClashingAliases` to cover the helper.

dearchap (approved):

PR #2331, opened 2026-05-17
docs: document v2 build tags removed in v3

wyf027:
Fixes #2304

Adds a migration section describing v2 build tags (`urfave_cli_no_docs`, `urfave_cli_no_suggest`) and notes that v3 removed these build constraints, with pointers to v2 releases and common binary-size alternatives.


abitrolly:
@leno23 why close?

PR #1938, opened 2024-06-29
fix: disable bash completion if double dash is included in arguments (v2)

suzuki-shunsuke:
## What type of PR is this?

_(REQUIRED)_

- bug

## What this PR does / why we need it:

_(REQUIRED)_

This pull request disables shell completion if double dash `--` is included in arguments.
Arguments after double dash `--` are not flags but positional arguments.

https://unix.stackexchange.com/a/11382

## Which issue(s) this PR fixes:

_(REQUIRED)_

Fixes #1932

## Special notes for your reviewer:

_(fill-in or delete this section)_

## Testing

_(fill-in or delete this section)_

## Release Notes

_(REQUIRED)_

```release-note

```

PR #2342, opened 2026-05-27
fix: prevent action execution during shell completion after double dash

### LINKED ISSUE #1993: Autocomplete after double dash (--) executing command action.
## My urfave/cli version is

v2.27.5

## Checklist

- [X] Did you check the manual for your release? The v2 manual is [here](https://cli.urfave.org/v2/getting-started/)

## Dependency Management

## Describe the bug

If Bash completion is enabled, and the user attempts to tab complete after a double dash (--), it will execute the normal command action.

## To reproduce

main.go
```go
package main

import (
	"os"

	"github.com/urfave/cli/v2"
)

func main() {
	app := &cli.App{
		Name:                 "cli_test",
		EnableBashCompletion: true,
		Action: func(*cli.Context) error {
			os.Create("/tmp/cli_test/TestFile.out")
			return nil
		},
		Flags: []cli.Flag{
			&cli.BoolFlag{
				Name: "verbose",
			},
		},
	}

		panic(err)
	}
}
```

using the v2 [zsh_autocomplete](https://github.com/urfave/cli/blob/v2-maint/autocomplete/zsh_autocomplete)

Sourcing it as such in my ~/.zshrc
```bash
PROG=cli_test source <path to project>/zsh_autocomplete
```

```bash
mkdir /tmp/cli_test/ && ls /tmp/cli_test
```
Dir is empty

Try autocomplete
```
cli_test --<TAB>
# or 
cli_test -- --generate-bash-completion
```
Check dir
```bash
ls /tmp/cli_test
```
TestFile.out is present

## Observed behavior

Due to the double dash (--), the `--generate-bash-completion` flag was treated as an argument, and the command was executed normally. This behavior is INCREDIBLY DANGEROUS as it can cause premature execution of a command, without even pressing enter!

## Expected behavior

There are several ways this could be handled. 

The ideal solution would be autocompleting the list of long flags, instead of executing the normal command action
Another solution would be preventing the command from being executed and not autocompleting.

## Additional context

This problem was introduced by this PR
https://github.com/urfave/cli/pull/1938

I personally disagree with this PR being marked as a bug instead of a feature, and think that this behavior should be toggleable instead of enabled for all cli applications.

This PR did disable bash autocomplete after a double dash, but introduced a much worse issue, which is executing the command normally...

## Want to fix this yourself?

If I find time, I may be willing to fix this myself, but I wanted to report this as soon as possible since this bug could cause severe damage in the worse case scenario.

```
go version go1.23.2 darwin/arm64
```

```
GO111MODULE=''
GOARCH='arm64'
GOBIN=''
GOCACHE='/Users/aarolieb/Library/Caches/go-build'
GOENV='/Users/aarolieb/Library/Application Support/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFLAGS=''
GOHOSTARCH='arm64'
GOHOSTOS='darwin'
GOINSECURE=''
GOMODCACHE='/Users/aarolieb/go/pkg/mod'
GONOPROXY=''
GONOSUMDB=''
GOOS='darwin'
GOPATH='/Users/aarolieb/go'
GOPRIVATE=''
GOPROXY='direct'
GOROOT='/opt/homebrew/Cellar/go/1.23.2/libexec'
GOSUMDB='sum.golang.org'
GOTMPDIR=''
GOTOOLCHAIN='local'
GOTOOLDIR='/opt/homebrew/Cellar/go/1.23.2/libexec/pkg/tool/darwin_arm64'
GOVCS=''
GOVERSION='go1.23.2'
GODEBUG=''
GOTELEMETRY='local'
GOTELEMETRYDIR='/Users/aarolieb/Library/Application Support/go/telemetry'
GCCGO='gccgo'
GOARM64='v8.0'
AR='ar'
CC='cc'
CXX='c++'
CGO_ENABLED='1'
GOMOD='/Users/aarolieb/Code/Scripts/cli_test/go.mod'
GOWORK=''
CGO_CFLAGS='-O2 -g'
CGO_CPPFLAGS=''
CGO_CXXFLAGS='-O2 -g'
CGO_FFLAGS='-O2 -g'
CGO_LDFLAGS='-O2 -g'
PKG_CONFIG='pkg-config'
GOGCCFLAGS='-fPIC -arch arm64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -ffile-prefix-map=/var/folders/_s/694rx7x90x96jkm795h4tk5h0000gr/T/go-build3647699756=/tmp/go-build -gno-record-gcc-switches -fno-common'
```

toller892:
## Description

When a shell autocomplete script appends `--generate-bash-completion` to a command line that already contains `--`, the completion flag is passed through to the action as a positional argument, causing the **action to execute during a completion attempt**.

This is particularly dangerous because tab-completion could trigger side effects (e.g., file creation, network calls) without the user pressing Enter.

Fixes #1993

## Root Cause

PR #1938 added a check in `checkShellCompleteFlag` to disable shell completion when `--` is present (correct behavior — after `--`, only positional args should be accepted). However, when this check fires, the function returns the **unmodified arguments** including `--generate-bash-completion`. The app then runs normally with `--generate-bash-completion` as a positional argument, executing the action.

### `help.go` — `checkShellCompleteFlag`

- Always strip `--generate-bash-completion` from arguments when it's detected (regardless of `--` presence), so it never leaks into the action's positional arguments
- Added a new return value `completionDetected bool` to indicate whether the completion flag was present
- Return `(false, true, stripped)` when `--` is present (completion disabled, but flag was detected and stripped)
- Return `(true, true, stripped)` in the normal case (completion enabled, flag stripped)

### `app.go` — `RunContext`

- Added early return (`nil`) when `completionDetected && !shellComplete` — the completion flag was present but shell completion is disabled (due to `--`). This prevents the action from executing during a completion attempt.

### `help_test.go`

- Updated `Test_checkShellCompleteFlag` to verify the new `completionDetected` return value
- Updated the "arguments include double dash" test case: expected args changed from `["--", "foo", "--generate-bash-completion"]` to `["--", "foo"]`

## Behavior Change

| Scenario | Before | After |
|----------|--------|-------|
| `app --<TAB>` | Action executes with `--generate-bash-completion` as arg | Returns nil, action not executed |
| `app -- foo <TAB>` | Action executes with `--generate-bash-completion` as arg | Returns nil, action not executed |
| `app foo <TAB>` | Completion works normally | No change |
| `app <TAB>` | Completion works normally | No change |

### DIFF
--- a/app.go
+++ b/app.go
@@ -319,7 +319,14 @@ func (a *App) RunContext(ctx context.Context, arguments []string) (err error) {
 	// flag name as the value of the flag before it which is undesirable
 	// note that we can only do this because the shell autocomplete function
 	// always appends the completion flag at the end of the command
-	shellComplete, arguments := checkShellCompleteFlag(a, arguments)
+	shellComplete, completionDetected, arguments := checkShellCompleteFlag(a, arguments)
+
+	// If the completion flag was present but shell completion is disabled
+	// (e.g., because "--" was on the command line), do not execute the
+	// action. The flag has already been stripped from arguments.
+	if completionDetected && !shellComplete {
+		return nil
+	}
 
 	cCtx := NewContext(a, nil, &Context{Context: ctx})
 	cCtx.shellComplete = shellComplete
--- a/help.go
+++ b/help.go
@@ -435,28 +435,34 @@ func checkHelp(cCtx *Context) bool {
 	return found
 }
 
-func checkShellCompleteFlag(a *App, arguments []string) (bool, []string) {
+func checkShellCompleteFlag(a *App, arguments []string) (bool, bool, []string) {
 	if !a.EnableBashCompletion {
-		return false, arguments
+		return false, false, arguments
 	}
 
 	pos := len(arguments) - 1
 	lastArg := arguments[pos]
 
 	if lastArg != "--generate-bash-completion" {
-		return false, arguments
+		return false, false, arguments
 	}
+
+	// The completion flag was detected — strip it from arguments regardless
+	// of "--" presence, so that it never leaks into the action's arguments.
+	stripped := make([]string, 0, len(arguments)-1)
+	stripped = append(stripped, arguments[:pos]...)
+	stripped = append(stripped, arguments[pos+1:]...)
 
 	for _, arg := range arguments {
 		// If arguments include "--", shell completion is disabled
 		// because after "--" only positional arguments are accepted.
 		// https://unix.stackexchange.com/a/11382
 		if arg == "--" {
-			return false, arguments
+			return false, true, stripped
 		}
 	}
 
-	return true, arguments[:pos]
+	return true, true, stripped
 }
 
 func checkCompletions(cCtx *Context) bool {
--- a/help_test.go
+++ b/help_test.go
@@ -1716,20 +1716,20 @@ GLOBAL OPTIONS:
 func Test_checkShellCompleteFlag(t *testing.T) {
 	t.Parallel()
 	tests := []struct {
-		name                string
-		app                 *App
-		arguments           []string
-		wantShellCompletion bool
+		name                   string
+		app                    *App
+		arguments              []string
+		wantShellCompletion    bool
 		wantCompletionDetected bool
-		wantArgs            []string
+		wantArgs               []string
 	}{
 		{
-			name:                "disable bash completion",
-			arguments:           []string{"--generate-bash-completion"},
-			app:                 &App{},
-			wantShellCompletion: false,
+			name:                   "disable bash completion",
+			arguments:              []string{"--generate-bash-completion"},
+			app:                    &App{},
+			wantShellCompletion:    false,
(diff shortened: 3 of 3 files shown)


dearchap (commented):
The fix looks correct. The key insight is that stripping `--generate-bash-completion` from args before the `--` loop, combined with the early return in `RunContext`, prevents the flag from leaking into the action.

One observation: the `stripped` slice is constructed unconditionally even in the fast path where `lastArg != "--generate-bash-completion"`. This is fine since that path returns early at line 443, but the allocation on line 447-448 is unnecessary for that case. Not a real concern though.

The test coverage is thorough — verifies both the new return value and the stripped args for the `--` case.

dearchap (commented):
The fix looks correct. The key change — stripping `--generate-bash-completion` before iterating for `--`, then returning early when `completionDetected && !shellComplete` — cleanly prevents action execution during completion attempts after `--`.

PR #2345, opened 2026-05-28
Issue#2323

### LINKED ISSUE #2323: Add a way to add additional topics to generated usage texts
## Checklist

I would like to be able to add additional topics to generated help messages.

## Solution description

A good way would look something like this:

```go
...
return &cli.Command{
  Name: "set",
  Topics: []cli.HelpTopics{
    &cli.HelpTopic{ // must be a map[string]string
      Reference: `https://example.com/api/overview`,
    }
  },
  ...
```

Which would generate:

```default
NAME:
   esctl cluster settings set - set|update cluster settings

REFERENCE:
   https://example.com/api/overview

..
```

I am currently doing this:

```go
	return &cli.Command{
		Name:      "set",
		Usage:     "set|update cluster settings",
		Aliases:   []string{"set", "update"},
		UsageText: "set [options] setting:value [setting:value ...]\nReference: " + SETTINGS,

		Flags: []cli.Flag{
			&cli.BoolFlag{
				Name:        "persistent",
				Usage:       "add persistent setting[s] (default)",
				Destination: &conf.Persistent,
				Aliases:     []string{"p"},
			},
			&cli.BoolFlag{
				Name:        "transient",
				Usage:       "add transient setting[s]",
				Destination: &conf.Transient,
				Aliases:     []string{"t"},
			},
		},
```

which outputs:

```default
NAME:
   esctl cluster settings set - set|update cluster settings

USAGE:
   set [options] setting:value [setting:value ...]
   Reference: https://www.elastic.co/docs/reference/elasticsearch/configuration-reference

OPTIONS:
   --persistent, -p  add persistent setting[s] (default)
   --transient, -t   add transient setting[s]
   --help, -h        show help
```

So, the reference is just part of the usage. I could add it to the NAME topic, but this would be wrong as well.

MohitPanchariya:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

* This PR introduces an option to include arbitrary help topics to the help message.
* There exists a way to do this for the Root command right now, through the `ExtraInfo` field,
but that requires using a custom template. Its a bit cumbersome to use custom templates from a
users perspective just to print additional help topics.
* Given the use case to have additional help topics seems generic, this PR changes the default
help templates to conditionally render additional help topics.

Fixes https://github.com/urfave/cli/issues/2323

_(REQUIRED)_

## Testing

Added unit tests to test additional help topics.

## Release Notes

_(REQUIRED)_

```release-note
The `AdditionalHelpTopics` field can be used to print arbitrary help topics, in addition to the existing
help topics, without the need for a custom help template.
```

### DIFF
--- a/template.go
+++ b/template.go
@@ -82,7 +82,7 @@ var CommandHelpTemplate = `NAME:
 
 USAGE:
    {{template "usageTemplate" .}}
-   
+
 {{- if .AdditionalHelpTopics}}{{ template "additionalHelpTopicsTemplate" .}}{{end}}{{if .Category}}
 
 CATEGORY:
@@ -106,7 +106,7 @@ var SubcommandHelpTemplate = `NAME:
 
 USAGE:
    {{if .UsageText}}{{wrap .UsageText 3}}{{else}}{{.FullName}}{{if .VisibleCommands}} [command [command options]]{{end}}{{if .ArgsUsage}} {{.ArgsUsage}}{{else}}{{if .Arguments}} [arguments...]{{end}}{{end}}{{end}}
-   
+
 {{- if .AdditionalHelpTopics}}{{ template "additionalHelpTopicsTemplate" .}}{{end}}{{if .Category}}
 
 CATEGORY:
--- a/template.go
+++ b/template.go
@@ -38,6 +38,12 @@ VERSION:
 
 var copyrightTemplate = `{{wrap .Copyright 3}}`
 
+var additionalHelpTopicsTemplate = `{{range $key, $value := .AdditionalHelpTopics}}
+
+{{$key}}:
+   {{$value}}
+{{- end}}`
+
 // RootCommandHelpTemplate is the text template for the Default help topic.
 // cli.go uses text/template to render templates. You can
 // render custom help text by setting this variable.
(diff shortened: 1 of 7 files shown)


template.go
```diff
@@ -38,14 +38,22 @@ VERSION:
 
 var copyrightTemplate = `{{wrap .Copyright 3}}`
 
+var additionalHelpTopicsTemplate = `{{range $key, $value := .AdditionalHelpTopics}}
+
+{{$key}}:
+   {{$value}}
+{{- end}}`
```
dearchap, line 45:
```suggestion
{{end}}`
```

dearchap (commented):
@MohitPanchariya I like this PR . Its very clean and backward compatible. However I am more worried about side effects of a special case like this. People will start adding more and more things like this throughout the template. Do you see lots of people using this or is this a one off for convenience ? Strictly speaking users can remap the template to whatever they'd like.

template.go
MohitPanchariya: Without the trim, there ends up being an additional new line after the additional help topics and before the next topic.

For example, with the trim
```
./main          
NAME:
   root - A new cli application

USAGE:
   root [global options] [command [command options]]

ADDITIONAL-HELP-1:
   ADDITIONAL-HELP-1

ADDITIONAL-HELP-2:
   ADDITIONAL-HELP-2

ADDITIONAL-HELP-3:
   ADDITIONAL-HELP-3

COMMANDS:
   sub      

GLOBAL OPTIONS:
   --help, -h  show help
```

Without the trim,
```
% ./main
NAME:
   root - A new cli application

USAGE:
   root [global options] [command [command options]]

ADDITIONAL-HELP-1:
   ADDITIONAL-HELP-1


ADDITIONAL-HELP-2:
   ADDITIONAL-HELP-2


ADDITIONAL-HELP-3:
   ADDITIONAL-HELP-3


COMMANDS:
   sub      

GLOBAL OPTIONS:
   --help, -h  show help
```

We can do this instead, to avoid the trim.
```
{{$key}}:
   {{$value}}{{end}}`
```
or collapse it into a single line, 
```
{{$key}}:{{"\n"}}{{"   "}}{{$value}}{{end}}`
```

MohitPanchariya:
> @MohitPanchariya I like this PR . Its very clean and backward compatible. However I am more worried about side effects of a special case like this. People will start adding more and more things like this throughout the template. Do you see lots of people using this or is this a one off for convenience ? Strictly speaking users can remap the template to whatever they'd like.

I agree with the sentiment, this PR may end up setting a precedent we may not want.

Given it is trivial to remap the template, I believe we can halt the effort here for now, and revisit later if we see more users asking for a mechanism to print arbitrary key-value pairs, without having to modify the template.

@dearchap Does this sound reasonable?

dearchap:
@MohitPanchariya Thanks. Yes that sounds reasonable.

PR #2351, opened 2026-06-05
Reject v1-style flag aliases

### LINKED ISSUE #1103: v2 bug: Use of obsolete v1 aliases goes undetected
## my urfave/cli version is

2.2.0

## Checklist

## Dependency Management

- [X] My project is using go modules.

## Describe the bug

There is no warning if old-style v1 aliases are used.  The alias simply doesn't exist.

## To reproduce

Create a flag using the v1 method:

```go
cli.StringFlag{
  Name: "config, cfg"
}
```

## Observed behavior

The flag --config works; --cfg does not.

## Expected behavior

The system should output an error or warning.

## Additional context

Ideally if `Name` includes a space or a comma, the code should output an
message about "Old style alias being used. Please see docs/migrate-v1-to-v2.md"
then the program should exit.

## Want to fix this yourself?

no

```
go1.14.1
```

```
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPRIVATE=""
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/qw/qp8v2j353wz7q57_jymyxj0h0000gn/T/go-build854616664=/tmp/go-build -gno-record-gcc-switches -fno-common"
```

ibobgunardi:
Fixes #1103.

This makes the v2 branch fail fast when a built-in flag is still written with the old v1-style Name value, for example Name: "config, cfg" or Name: "config cfg". The error points developers to Aliases instead of silently registering only the first name.

I also updated two existing tests that were still using the old form so they keep testing aliases through the supported API.

Tests:
- go test . -run 'TestAppRunRejectsV1StyleFlagNames|TestCommandRunRejectsV1StyleFlagNames|TestCheckRequiredFlags|ExampleApp_Run_bashComplete_withLongFlag' -count=1
- go test . -count=1
- go test ./... -count=1
- go vet ./...

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -283,7 +283,8 @@ func ExampleApp_Run_bashComplete_withLongFlag() {
 			Aliases: []string{"x"},
 		},
 		&StringFlag{
-			Name: "some-flag,s",
+			Name:    "some-flag",
+			Aliases: []string{"s"},
 		},
 		&StringFlag{
 			Name: "similar-flag",
--- a/context_test.go
+++ b/context_test.go
@@ -626,7 +626,7 @@ func TestCheckRequiredFlags(t *testing.T) {
 			expectedAnError:       true,
 			expectedErrorContents: []string{"Required flag \"names\" not set"},
 			flags: []Flag{
-				&StringSliceFlag{Name: "names, n", Required: true},
+				&StringSliceFlag{Name: "names", Aliases: []string{"n"}, Required: true},
 			},
 		},
 		{
--- a/flag.go
+++ b/flag.go
@@ -182,6 +182,14 @@ func flagSet(name string, flags []Flag, spec separatorSpec) (*flag.FlagSet, erro
 	return set, nil
 }
 
+func validateFlagName(name string) error {
+	if !commaWhitespace.MatchString(name) {
+		return nil
+	}
+
+	return fmt.Errorf("invalid flag name %q: use Aliases instead of v1-style comma or space separated names", name)
+}
+
 func copyFlag(name string, ff *flag.Flag, set *flag.FlagSet) {
 	switch ff.Value.(type) {
 	case Serializer:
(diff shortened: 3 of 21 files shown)


ibobgunardi:
Checked the red jobs after the update.

- Package tests pass locally with `go test ./... -count=1`, and `go vet ./...` passes.
- The lint job fails before reporting a repo file: `golangci-lint` v1.64.8 is typechecking a Go 1.26 stdlib file and exits with `file requires newer Go version go1.26 (application built with go1.24)`.
- The Ubuntu job gets through the tests/builds and fails at `make check-binary-size`. I also ran `go run internal/build/build.go check-binary-size` on a clean `origin/v2-maint` checkout with the current local Go toolchain, and that baseline fails the same threshold. I removed the reflection-based version of this patch; the current branch only adds about 1.5KB to `internal/example-cli` in my local byte comparison and leaves `internal/example-hello-world` unchanged.

I am leaving the binary-size threshold alone here so this PR stays focused on #1103.

dearchap:
@ibobgunardi v2 is in maint with only security patches.

ibobgunardi:
Thanks, that makes sense. This is not security-related, so I'm closing it rather than keeping a v2-maint PR open.

PR #2353, opened 2026-06-06
feat: add Command.Walk() and Command.Path() convenience methods

### LINKED ISSUE #2324: Add Command.Walk(fn) and Command.Path() convenience methods
## Summary

Add two small convenience methods to `*cli.Command` that several third-party extensions reimplement today:

```go
// Walk visits cmd and every descendant in pre-order. Returning a non-nil
// error from fn short-circuits the walk and is returned to the caller.
func (cmd *Command) Walk(fn func(*Command) error) error

// Path returns the path of command names from the root to cmd, inclusive.
// Each element is a Command.Name. Useful for tools that need to manipulate
// the segments (dot-joined for tool names, slash-joined for URLs, etc).
// FullName(), which already exists, is equivalent to strings.Join(cmd.Path(), " ").
func (cmd *Command) Path() []string
```

## Motivation

I maintain a family of small urfave/cli extensions where each repo independently reimplements both of these:

- [cli-mcp](https://github.com/coilysiren/cli-mcp) — projects a command tree as an MCP server. `registerTree` is a hand-rolled walker; tool names are dot-joined paths.
- [cli-web-docs](https://github.com/coilysiren/cli-web-docs) — emits static HTML docs. `buildNavList` and `renderPerPage` each walk the tree; nav slugs use dash-joined paths.
- [cli-web-ops](https://github.com/coilysiren/cli-web-ops) — mobile-first MCP-client web UI. Walks an MCP tool surface, but the underlying motivation is the same.

The pattern across all three is:

```go
var walk func(prefix []string, cmd *cli.Command)
walk = func(prefix []string, cmd *cli.Command) {
    if cmd.Hidden { return }
    path := append(prefix, cmd.Name)
    // ... do per-command work using `path` ...
    for _, sub := range cmd.Commands {
        walk(path, sub)
    }
}
walk(nil, root)
```

Three flavors of the same thing. `Walk` removes the recursion boilerplate; `Path()` removes the manual `append(prefix, cmd.Name)` accumulation. Both are non-breaking additions.

## Alternatives considered

- Just keep the local reimplementations. Has worked, but the third extension to need it (cli-web-ops) shipped before promotion. Two reasonable choices for ergonomic extension surface seems like the threshold.
- A separate `urfave/cli-walk` helper module. Possible but seems silly for ~15 lines of code that fit cleanly on `*Command`.

Happy to send a PR if the team is interested.

dearchap:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

_(REQUIRED)_

Fixes #2324 

## Testing

go test -run=TestCommand_Walk
go test -run=TestCommand_Path

## Release Notes

_(REQUIRED)_

```release-note
Add Command.Path() and Command.Walk() convenience functions
```

### DIFF
--- a/command.go
+++ b/command.go
@@ -181,8 +181,8 @@ func (cmd *Command) Path() []string {
 	return []string{cmd.Name}
 }
 
-// Walk visits cmd and every descendant in pre-order. Returning a non-nil
-// error from fn short-circuits the walk and is returned to the caller.
+// Walk visits cmd and every descendant. If fn returns a non-nil error, the
+// walk terminates and the error is returned to the caller.
 func (cmd *Command) Walk(fn func(*Command) error) error {
 	if fn == nil {
 		return nil
--- a/command.go
+++ b/command.go
@@ -166,8 +166,7 @@ type Command struct {
 }
 
 // FullName returns the full name of the command.
-// For commands with parents this ensures that the parent commands
-// are part of the command path.
+// Includes parent commands separated by space.
 func (cmd *Command) FullName() string {
 	return strings.Join(cmd.Path(), " ")
 }
--- a/command.go
+++ b/command.go
@@ -169,13 +169,34 @@ type Command struct {
 // For commands with parents this ensures that the parent commands
 // are part of the command path.
 func (cmd *Command) FullName() string {
-	namePath := []string{}
+	return strings.Join(cmd.Path(), " ")
+}
 
+// Path returns the path of command names from the root to cmd, inclusive.
+// Each element is a Command.Name. FullName() is equivalent to
+// strings.Join(cmd.Path(), " ").
+func (cmd *Command) Path() []string {
 	if cmd.parent != nil {
-		namePath = append(namePath, cmd.parent.FullName())
+		return append(cmd.parent.Path(), cmd.Name)
 	}
+	return []string{cmd.Name}
+}
 
-	return strings.Join(append(namePath, cmd.Name), " ")
+// Walk visits cmd and every descendant in pre-order. Returning a non-nil
+// error from fn short-circuits the walk and is returned to the caller.
+func (cmd *Command) Walk(fn func(*Command) error) error {
+	if fn == nil {
+		return nil
+	}
+	if err := fn(cmd); err != nil {
+		return err
+	}
+	for _, sub := range cmd.Commands {
+		if err := sub.Walk(fn); err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 func (cmd *Command) Command(name string) *Command {
--- a/command.go
+++ b/command.go
@@ -165,39 +165,6 @@ type Command struct {
 	isCompletionCommand bool
 }
 
-// FullName returns the full name of the command.
-// Includes parent commands separated by space.
-func (cmd *Command) FullName() string {
-	return strings.Join(cmd.Path(), " ")
-}
-
-// Path returns the path of command names from the root to cmd, inclusive.
-// Each element is a Command.Name. FullName() is equivalent to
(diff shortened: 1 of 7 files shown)


testdata/godoc-v3.x.txt
```diff
@@ -689,6 +689,11 @@ func (cmd *Command) Names() []string
 func (cmd *Command) NumFlags() int
     NumFlags returns the number of flags set
 
+func (cmd *Command) Path() []string
+    Path returns the path of command names from the root to cmd,
+    inclusive. Each element is a Command.Name. FullName() is equivalent to
```
abitrolly: `FullName()`?

testdata/godoc-v3.x.txt
```diff
@@ -801,6 +806,10 @@ func (cmd *Command) VisiblePersistentFlags() []Flag
     VisiblePersistentFlags returns a slice of LocalFlag with Persistent=true and
     Hidden=false.
 
+func (cmd *Command) Walk(fn func(*Command) error) error
+    Walk visits cmd and every descendant in pre-order. Returning a non-nil error
+    from fn short-circuits the walk and is returned to the caller.
```
abitrolly:
```suggestion
    Walk visits cmd and every descendant in pre-order. If an error is
    encountered, the walk terminates with the error returned to the caller.
```

abitrolly: `pre-order` is still unclear.

command.go
```diff
@@ -169,13 +169,34 @@ type Command struct {
 // For commands with parents this ensures that the parent commands
 // are part of the command path.
```
abitrolly:
```suggestion
// Includes parent commands separated by space.
```

abitrolly:
Strange that `Walk` didn't find the use inside current code. I thought that recursive walk over commands should happen during processing more than once.

dearchap (commented):
Addressed feedback on Walk godoc — removed the unclear 'pre-order' phrasing and simplified per suggestion. FullName comment was already updated in prior commit.

abitrolly:
Still not clear what FullName does in Path description.

meatballhat:
Missing a `make v3approve`?

testdata/godoc-v3.x.txt
dearchap: pre-order tree traversal

testdata/godoc-v3.x.txt
meatballhat: Is the intention here to point out the relationship between `Path` and `FullName`? That's how I was reading it.

command.go
meatballhat, line 1: I think this is similar to @abitrolly feedback, but I think I'd like to see `*Command.Path` and `*Command.Walk` located closer to `*Command.Lineage`, and for the documentation for each of them to include cross-references, and for the higher-level documentation under the `./docs/` tree to describe the relationship between them with some examples. None of this is blocking, imho.

meatballhat (approved):

command_setup.go
```diff
@@ -165,11 +165,13 @@ func (cmd *Command) setupDefaults(osArgs []string) {
...
 	tracef("setting up command graph (cmd=%[1]q)", cmd.Name)
 
-	for _, subCmd := range cmd.Commands {
-		subCmd.parent = cmd
-		subCmd.setupSubcommand()
-		subCmd.setupCommandGraph()
-	}
+	_ = cmd.Walk(func(sub *Command) error {
+		for _, subCmd := range sub.Commands {
+			subCmd.parent = sub
```
abitrolly, line 170: Strange that parent is injected only here and not when `subCmd` is added into `sub.Commands` array.

PR #2354, opened 2026-06-06
feat: Add Schema Type and Schema Items Type

### LINKED ISSUE #2326: Expose Flag.SchemaType() for typed flag introspection
## Summary

Expose typed-schema introspection on `cli.Flag`:

```go
// SchemaType returns the JSON Schema type name for the value this flag
// accepts: "boolean", "integer", "number", "string", "array".
// Returns "" if the flag does not map cleanly to a JSON Schema type.
//
// For slice flags, the returned type is "array"; callers can use
// SchemaItemsType for the element type.
func (f Flag) SchemaType() string
func (f Flag) SchemaItemsType() string // optional, for slice flags
```

## Motivation

[cli-mcp](https://github.com/coilysiren/cli-mcp) projects a urfave/cli command tree as a Model Context Protocol server. MCP tools carry a JSON Schema for their input, derived from the command's flags. Today this is a mechanical type-switch:

```go
switch f.(type) {
case *cli.BoolFlag:
    s.Type = "boolean"
case *cli.IntFlag, *cli.Int8Flag, *cli.Int16Flag, *cli.Int32Flag, *cli.Int64Flag:
    s.Type = "integer"
case *cli.UintFlag, *cli.Uint8Flag, ... :
    s.Type = "integer"
case *cli.FloatFlag:
    s.Type = "number"
case *cli.StringSliceFlag:
    s.Type = "array"  // items: string
case *cli.IntSliceFlag:
    s.Type = "array"  // items: integer
// ... etc
}
```

This is brittle - every new flag type urfave ships (or that consumers add via the third-party flag system) requires a code change in every extension. A `SchemaType()` method on Flag (or a `Type` constant) would let extensions stop pattern-matching on concrete types.

This same need exists for:

- **cli-web-ops** — rendering JSON Schema input as HTML form fields (`<input type="text|number|checkbox">`)
- **cli-web-docs** — rendering flag types in generated documentation
- **urfave/cli-docs** itself — currently maps flag types in Markdown output by similar type switching

## Alternatives considered

- A standalone `cli-schema-types` helper. Forces a hard-coded import dependency just for type strings.
- Reflection. Works but is fragile and slow.

The right home is the Flag interface itself, since the flag knows its own type.

Happy to send a PR if there's appetite. Could plausibly be hidden behind a feature flag or new `DocFlag` interface to preserve backwards compatibility with third-party flag types.

dearchap:
## What type of PR is this?

_(REQUIRED)_

- feature

_(REQUIRED)_

_(REQUIRED)_

Fixes #2326 

## Testing

## Release Notes

_(REQUIRED)_

```release-note
Add SchemaType and SchemaItemsType for Flag 
```

### DIFF
--- a/flag_ext.go
+++ b/flag_ext.go
@@ -61,3 +61,11 @@ func (e *extFlag) GetDefaultText() string {
 func (e *extFlag) GetEnvVars() []string {
 	return nil
 }
+
+func (e *extFlag) SchemaType() string {
+	return ""
+}
+
+func (e *extFlag) SchemaItemsType() string {
+	return ""
+}
--- a/flag_impl.go
+++ b/flag_impl.go
@@ -319,8 +319,6 @@ func (f *FlagBase[T, C, V]) SchemaItemsType() string {
 	t := reflect.TypeOf(zero)
 	if t.Kind() == reflect.Slice {
 		switch t.Elem().Kind() {
-		case reflect.Bool:
-			return "boolean"
 		case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64,
 			reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
 			return "integer"
--- a/flag_impl.go
+++ b/flag_impl.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
+	"time"
 )
 
 // Value represents a value as used by cli.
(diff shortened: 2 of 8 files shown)


coilysiren:
👀❕

I shall give this a look, no ETA promise though. Which is probably fine, I'm the one who asked for it.

coilysiren:
Also: thank you!

flag_ext.go
meatballhat, line 1: I don't love returning `""` for all discovered stdlib `Flag`, but I think that returning the types requires reading through [`Get() any`](https://pkg.go.dev/flag#Getter) and doing reflection. At least it would be in one place?

flag_impl.go
```diff
@@ -285,6 +286,51 @@ func (f *FlagBase[T, C, V]) RunAction(ctx context.Context, cmd *Command) error {
...
+	case int, int8, int16, int32, int64, uint, uint8, uint16, uint32, uint64:
+		return "integer"
+	case float32, float64:
+		return "number"
+	case string:
+		return "string"
+	case time.Duration:
+		return "string"
+	case time.Time:
+		return "string"
```
meatballhat, line 304: I would have thought [`"duration"` and `"date-time"`](https://json-schema.org/understanding-json-schema/reference/type#dates-and-times) would fit better, or at least provide more semantic information (?) even though Go durations aren't ISO-8601 durations, and library users could always "downgrade" to `"string"`.

meatballhat (approved):
Some questions, but none blocking ❤️

PR #2355, opened 2026-06-07
Honor env sources on duplicated flags

### LINKED ISSUE #2217: Environment variables not read when flag is defined on both root and subcommand
### My urfave/cli version is

v3.5.0 (latest as of 2025-10-30)

### Checklist

  - Are you running the latest v3 release? The list of releases is https://github.com/urfave/cli/releases. ✅ 
  - Did you check the manual for your release? The v3 manual is https://cli.urfave.org/v3/getting-started/ ✅ 
  - Did you perform a search about this problem? Here's the https://help.github.com/en/github/managing-your-work-on-github/using-search-to-filter-issues-and-pull-requests about searching. ✅ 

### Dependency Management

My project is using go modules.

### Describe the bug

  When the same flag (same flag object instance) is defined in both a root command's Flags array and a subcommand's Flags array, environment variables specified via
  Sources: cli.EnvVars() are not read correctly when executing the subcommand. However, CLI flag values passed via command-line arguments work fine.

  This appears to be a regression or unintended behavior in v3, as the flag duplication pattern was common in v2 for creating "global" flags.

### Minimal reproduction code

```go
  package main

  import (
        "context"
        "fmt"
        "os"

        "github.com/urfave/cli/v3"
  )

  func main() {
        // Define flag once
        flagTest := &cli.StringFlag{
                Name:     "test-flag",
                Sources:  cli.EnvVars("TEST_FLAG"),
                Usage:    "Test flag",
                Required: true,
        }

        // Subcommand that uses the flag
        subCmd := &cli.Command{
                Name: "subcmd",
                Flags: []cli.Flag{
                        flagTest,  // Flag on subcommand
                },
                        value := cmd.String(flagTest.Name)
                        fmt.Printf("Subcommand - Flag value: '%s'\n", value)
                        return nil
                },
        }

        // Root command that ALSO has the same flag
        rootCmd := &cli.Command{
                Flags: []cli.Flag{
                        flagTest,  // SAME flag on root
                },
                Before: func(ctx context.Context, cmd *cli.Command) (context.Context, error) {
                        value := cmd.String(flagTest.Name)
                        fmt.Printf("Root Before - Flag value: '%s'\n", value)
                        return ctx, nil
                },
                Commands: []*cli.Command{
                        subCmd,
                },
        }

        if err := rootCmd.Run(context.Background(), os.Args); err != nil {
                fmt.Fprintf(os.Stderr, "Error: %v\n", err)
                os.Exit(1)
        }
  }
```

### Steps to reproduce

  1. Save the code above as main.go
  2. Run with environment variable set:
  TEST_FLAG="my-value" go run main.go subcmd

  Observed behavior

  Root Before - Flag value: ''
  Subcommand - Flag value: ''

  The environment variable TEST_FLAG is not read, and the flag value is empty.

  However, if you pass the flag via CLI:
  TEST_FLAG="my-value" go run main.go subcmd --test-flag="cli-value"

  Output:
  Root Before - Flag value: 'cli-value'
  Subcommand - Flag value: 'cli-value'

  The CLI flag does work correctly.

  Expected behavior

  The environment variable should be read and the flag value should be:

  Root Before - Flag value: 'my-value'
  Subcommand - Flag value: 'my-value'

  This is the behavior when the flag is not duplicated on the root command (i.e., when flagTest is only in the subcommand's Flags array).

### Additional context

  Workaround: Remove the flag from the root command's Flags array if it's only used by subcommands. Only include flags on the root that are actually used by the root's
  Before/After hooks.

  This issue was discovered while migrating a large codebase from v2 to v3. In v2, it was common to define flags on both root and subcommands to make them "globally
  available." The migration guide doesn't mention this breaking change.

### Impact
This could affect many projects migrating from v2 to v3, as the duplication pattern was used for global flags. The issue is subtle because:
  1. CLI flags still work (masking the problem during development)
  2. Environment variables silently fail (causing issues in production/CI)

### Questions
  - Is this intended behavior?
  - Should the migration guide warn about this?
  - Is there a recommended pattern for "global" flags in v3 that work with both CLI args and env vars?

  Run go version and paste its output here

  go version go1.24.6 darwin/arm64

  Run go env and paste its output here

  GO111MODULE='on'
  GOARCH='arm64'
  GOOS='darwin'
  GOVERSION='go1.24.6'

ibobgunardi:
## What type of PR is this?

- bug

- Preserves env-source values when the same persistent flag instance is registered on an ancestor command and a subcommand.
- Skips re-preparing a shared persistent flag instance on descendant commands so the value loaded during the parent parse is not reset.
- Adds regression coverage for duplicated persistent flag env sources.

Fixes #2217

The main behavior to review is the exact flag-instance check for inherited persistent flags.

## Testing

- go test ./... -run TestDuplicatePersistentFlagUsesEnvSource -count=1 -v
- go test ./... -run "TestPersistentFlag|TestDuplicatePersistentFlagUsesEnvSource|TestRequiredPersistentFlag|TestCommand_IsSet_fromEnv|TestCheckRequiredFlags" -count=1 -v
- go test ./... -count=1

## Release Notes

```release-note
Fixed env-source values being reset when the same persistent flag instance is registered on a parent command and subcommand.
```

### DIFF
--- a/command.go
+++ b/command.go
@@ -373,6 +373,22 @@ func (cmd *Command) lFlag(name string) Flag {
 	return nil
 }
 
+func (cmd *Command) hasPersistentFlagOnAncestor(fl Flag) bool {
+	for pCmd := cmd.parent; pCmd != nil; pCmd = pCmd.parent {
+		for _, pFl := range pCmd.allFlags() {
+			if pFl != fl {
+				continue
+			}
+
+			pfl, ok := pFl.(LocalFlag)
+			if ok && !pfl.IsLocal() {
+				return true
+			}
+		}
+	}
+	return false
+}
+
 func (cmd *Command) lookupFlag(name string) Flag {
 	for _, pCmd := range cmd.Lineage() {
 		if f := pCmd.lFlag(name); f != nil {
--- a/command_run.go
+++ b/command_run.go
@@ -142,6 +142,9 @@ func (cmd *Command) run(ctx context.Context, osArgs []string) (_ context.Context
 	var args Args = &stringSliceArgs{rargs.Tail()}
 
 	for _, f := range cmd.allFlags() {
+		if cmd.hasPersistentFlagOnAncestor(f) {
+			continue
+		}
 		if err := f.PreParse(); err != nil {
 			return ctx, err
 		}
--- a/command_test.go
+++ b/command_test.go
@@ -3387,6 +3387,48 @@ func TestPersistentFlagIsSet(t *testing.T) {
 	require.True(t, resultIsSet)
 }
 
+func TestDuplicatePersistentFlagUsesEnvSource(t *testing.T) {
+	t.Setenv("APP_RESULT", "from-env")
+
+	resultFlag := &StringFlag{
+		Name:     "result",
+		Sources:  EnvVars("APP_RESULT"),
+		Required: true,
+	}
+
+	var beforeResult string
+	var actionResult string
+	app := &Command{
+		Name: "root",
+		Flags: []Flag{
+			resultFlag,
+		},
+		Before: func(_ context.Context, cmd *Command) (context.Context, error) {
+			beforeResult = cmd.String("result")
+			return nil, nil
+		},
+		Commands: []*Command{
+			{
+				Name: "sub",
+				Flags: []Flag{
+					resultFlag,
+				},
+				Action: func(_ context.Context, cmd *Command) error {
+					actionResult = cmd.String("result")
+					return nil
+				},
+			},
+		},
+	}
+
+	err := app.Run(context.Background(), []string{"root", "sub"})
+	require.NoError(t, err)
+	require.Equal(t, "from-env", beforeResult)
+	require.Equal(t, "from-env", actionResult)
+	require.True(t, app.IsSet("result"))
+	require.True(t, app.Command("sub").IsSet("result"))
+}
+
 func TestRequiredFlagDelayed(t *testing.T) {
 	sf := &StringFlag{
 		Name:     "result",


dearchap:
@ibobgunardi Can you follow the PR template when submitting a PR ? Thanks

dearchap (dismissed):
## Root Cause

When the **same flag instance** is registered on both a parent and subcommand, `FlagBase.PreParse()` re-initializes `f.value` to the default (`f.Value`) **without checking `f.applied`**. This overwrites the env-sourced value set by the parent`s `PostParse()`.

Meanwhile, `FlagBase.PostParse()` guards with `f.hasBeenSet`, which is already `true` (set by the parent`s PostParse), so it skips re-reading the env var. End result: the subcommand gets the zero/default value instead of the env value.

## Fix Validation

`hasPersistentFlagOnAncestor()` uses pointer equality to detect the shared instance, and the PreParse skip in `command_run.go:144-146` correctly:
1. Prevents the destructive re-initialization on the subcommand
2. Preserves the env (or CLI-arg) value already set on the parent
3. PostParse respects `hasBeenSet` and does not overwrite

## Test Coverage

The test `TestDuplicatePersistentFlagUsesEnvSource` covers:
- Env source read by the parent` s `Before` handler ✓  
- Env source value propagated to the subcommand`s `Action` ✓  
- `IsSet()` returning `true` on both parent and subcommand ✓  

LGTM.

ibobgunardi:
Updated the description to follow the PR template. Thanks for the reminder.

dearchap (commented):
Good question about `f.applied` in `PreParse`. I tested it — two regressions:

1. **TestSliceStringFlagParsing** — PreParse re-creates `f.value` from scratch (via `f.creator.Create`). Without it, `Set()` appends to the existing slice, causing cross-sub-test value leakage. e.g. expected `["x", "y"]`, got `["x", "x", "y"]`.

2. **TestRequiredFlagDelayed** — `applied` being `true` on a shared flag instance prevents PreParse from running on the leaf command, altering required-flag behavior that non-shared flags depend on.

`applied` serves dual roles (prevent recursion + track initialization), so adding it to PreParse is too broad. The pointer-equality/ancestor check is correctly scoped. LGTM.

Issue #2126, opened 2025-05-09
your bug title goes here


bigluck:
## My urfave/cli version is

v3.3.2

## Checklist

- [x] Are you running the latest v3 release? The list of releases is [here](https://github.com/urfave/cli/releases).
- [x] Did you check the manual for your release? The v3 manual is [here](https://cli.urfave.org/v3/getting-started/)
- [x] Did you perform a search about this problem? Here's the [GitHub guide](https://help.github.com/en/github/managing-your-work-on-github/using-search-to-filter-issues-and-pull-requests) about searching.

## Describe the bug

The `--help` argument causes an error when specified in a command with arguments.

## To reproduce

```go
package main

import (
	"context"
	"fmt"
	"os"

	"github.com/urfave/cli/v3"
)

func main() {
	app := &cli.Command{
		Name:  "My CLI",
		Usage: "My Usage",
		Flags: []cli.Flag{
			&cli.StringFlag{
				Name:    "profile",
				Aliases: []string{"p"},
				Usage:   "Name of the profile to use",
				Value:   "default",
			},
		},
		Commands: []*cli.Command{
			{
				Name: "command",
				Arguments: []cli.Argument{
					&cli.StringArg{
						Name:      "arg1",
						UsageText: "ARG1",
					},
				},
				Action: func(ctx context.Context, cmd *cli.Command) error {
					fmt.Printf("My profile: %s\n", cmd.String("profile"))
					fmt.Printf("My command: arg1=%s\n", cmd.StringArg("arg1"))
					return nil
				},
				Usage: "Show the version of My CLI",
			},
		},
	}

	err := app.Run(context.Background(), os.Args)
	if err != nil {
		fmt.Println("Error:", err)
		os.Exit(1)
	}
}
```

## Observed behavior

Both these commands work:

```bash
$ go run ./cli-error command --help
...
$ go run ./cli-error command --profile dev --help
...
```

But when I specify one of the command's arguments, it raises an error:

```bash
$ go run ./cli-error command ciao --help
Error: No help topic for 'ciao'
exit status 1

$ go run ./cli-error command --help ciao
Error: No help topic for 'ciao'
exit status 1
```

## Expected behavior

I'm not sure if that's intentional behavior, but with `nano` for example (just a random command), I can specify `--help` both before and after an argument.

```bash
$ nano ./poetry.lock --help
Usage: nano [OPTIONS] [[+LINE[,COLUMN]] FILE]...

$ nano --help poetry.lock
[...]

PR #2360, opened 2026-06-08
fix: let --help flag take precedence over parse errors

dearchap:
## What type of PR is this?
- bug

Fixes two help invocation inconsistencies related to issue #2126.

1. **`--help` flag now takes precedence over parse errors**: When `--help` is given alongside a bad flag (e.g. `app --help --undefined`), the help output is now shown instead of printing "Incorrect Usage".

2. **Updated stale comments** in `helpCommandAction` to list all actual invocation paths.

Related to #2126

## Release Notes
```release-note
`--help` flag works when given alongside invalid flags
```

### DIFF
--- a/command_run.go
+++ b/command_run.go
@@ -175,7 +175,6 @@ func (cmd *Command) run(ctx context.Context, osArgs []string) (_ context.Context
 		deferErr = err
 
 		cmd.isInError = true
-
 		if cmd.checkHelp() {
 			if cmd.parent == nil {
 				_ = ShowRootCommandHelp(cmd)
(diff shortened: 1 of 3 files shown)


opencode.json
meatballhat, line 1: Accidental addition?

meatballhat (commented):
If the addition of the opencode config in this change is intentional, I'd like to understand why 😁

opencode.json
dearchap: Yes accidental. Removing

meatballhat (approved):

PR #2363, opened 2026-06-08
fix: skip After hook when help is displayed on subcommand

### LINKED ISSUE #2250: Imbalance between Before and After calls at printing help
## My urfave/cli version is

_**3.6.2**_

## Checklist

## Dependency Management

## Describe the bug

For "help" action or "subcommand --help" "After" is called without calling "Before".

## To reproduce

## Observed behavior

What did you see happen immediately after the reproduction steps
above?

## Expected behavior

Callback "After" is called iff "Before" was called in the past.

## Additional context

If the issue relates to a specific open source GitHub repo, please
link that repo here.

If you can reproduce this issue with a public CI system, please
link a failing build here.

## Want to fix this yourself?

pull request for it.

```
# paste `go version` output in here
```

```
```

dearchap:
## Summary

Fixes #2250 — when `--help` is passed to a subcommand, the parent command's `After` hook was firing even though the parent's `Before` hook was never called. This violates the expected invariant that `After` should only be called when `Before` was also called.

## Root Cause

In `command_run.go`, the `After` hook is registered as a deferred call at line 227, before the subcommand dispatch (line 298) and the `Before` chain (line 312). When a subcommand receives `--help`, the subcommand's `run()` returns early at the help check (line 205), never reaching the `Before` chain. The parent's deferred `After` then fires without the corresponding `Before`.

## Fix

Track help invocation via a `helpShownKey` in the context. When `--help` is detected (both in the success path and the parse-error-then-help path), set the key in the context before returning. The deferred `After` hook checks this key and returns early if help was the reason for early exit.

## Tests Added

- `TestCommand_AfterNotCalledOnSubcommandHelp` — verifies `After` is NOT called when `app sub --help` is used
- `TestCommand_AfterStillCalledOnNormalSubcommand` — verifies `After` IS called for normal `app sub` execution

### DIFF
--- a/command_run.go
+++ b/command_run.go
@@ -9,6 +9,8 @@ import (
 	"unicode"
 )
 
+type helpShownKey struct{}
+
 func (cmd *Command) parseArgsFromStdin() ([]string, error) {
 	type state int
 	const (
@@ -176,6 +178,7 @@ func (cmd *Command) run(ctx context.Context, osArgs []string) (_ context.Context
 
 		cmd.isInError = true
 		if cmd.checkHelp() {
+			ctx = context.WithValue(ctx, helpShownKey{}, true)
 			if cmd.parent == nil {
 				_ = ShowRootCommandHelp(cmd)
 			} else {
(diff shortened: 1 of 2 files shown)


abitrolly:
I don't know. Maybe there should be a separate `helpMode` traversal|walk which is (docu|imple)mented as not triggering any hooks at all?

meatballhat (approved):
I tend to agree with @abitrolly but given that this is internal I think a more tidy implementation could be in a follow up.

Issue #2365, opened 2026-06-12
Documentation on optimizing binary size


abitrolly:
In https://github.com/urfave/cli/issues/2304 we've tried to trace what happened to build flags that helped to reduce binary size. The information in the thread appeared quite valuable to me, and I thought that `urfave/cli` is popular enough among Go beginners to share a tutorial on where the binary size comes from, and how to prove and optimize it (if possible).

I am not sure I am capable to write it myself in a concise and visually pleasing way with dumps, pictures and commands. So I leave this issue here to challenge other Go hackers.

abitrolly:
#2366 solves 80% of the issue. The rest is using https://github.com/Zxilly/go-size-analyzer to see where bloat comes from. Then explain some techniques like syncing `go.mod` files to have the same dependency versions for different modules. Maybe something else, like why and how introspection bloats, how to remove it. Also how to remove templates that trigger introspection.

imsumedhaa:
@abitrolly  Hi! I'd like to work on this issue.
I've started looking into the remaining parts from #2366, especially using go-size-analyzer to investigate binary size, and understanding the impact of dependency versions, reflection/introspection, and templates.

kolyshkin:
Related: #2441 / #2442

PR #2366, opened 2026-06-12
Document v3 binary size checks

rupayon123:
## Summary
- add a v3 docs page for measuring and reducing Go binary size
- document that v3 currently does not define `urfave_cli_no_*` build tags
- add practical commands for `go build`, `go version -m`, and `go tool nm`
- link the page from the v3 MkDocs nav

## Why
Addresses #2365 with a concise starting point for users who want to understand where binary size comes from before proposing build tags or package splits.

## Validation
- `git diff --check`
- inspected the generated Markdown content locally

I did not run the MkDocs build because `mkdocs` is not installed in my local Python environment.

### DIFF
--- a/docs/v3/binary-size.md
+++ b/docs/v3/binary-size.md
@@ -1,6 +1,6 @@
 # Binary Size
 
-Go already removes unreachable code during compilation, so the first step is to
+Go removes unreachable code during compilation, so the first step is to
 measure the binary you ship instead of assuming a specific feature is expensive.
 
 ```sh
@@ -8,8 +8,8 @@ go build -trimpath -o myapp ./cmd/myapp
 ls -lh myapp
 ```
 
-Build the same command with stripped symbol and debug information to compare the
-release artifact size:
+For a release-style build, combine reproducible paths with stripped symbol and
+debug information:
 
 ```sh
 go build -trimpath -ldflags="-s -w" -o myapp ./cmd/myapp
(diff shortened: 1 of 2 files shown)


rupayon123:
I reran local checks with Go installed and the pinned docs dependencies:

- `go test ./...` — passed
- `mkdocs build --strict` — passed

Marking this ready for review.

dearchap (commented):
## Review of PR #2366

This is a well-written, practical documentation page that addresses issue #2365. The content is accurate and useful. A few suggestions:

**1. Add a link to new-issue page** (line ~50) — The paragraph says "open an issue" but does not provide a URL. Add a link to `https://github.com/urfave/cli/issues/new` for convenience.

**2. Mention `make check-binary-size`** — The PR title says "binary size checks" but the page doesn't reference the existing `make check-binary-size` target already in the Makefile/scripts. This is the most relevant CI check and would be worth mentioning.

**3. Combine the two `go build` examples** — Showing `-trimpath` and `-ldflags="-s -w"` in separate code blocks is fine, but many readers will want the release recipe as one command. Consider adding a combined example like:

```sh
go build -trimpath -ldflags="-s -w" -o myapp ./path/to/main
```

**4. Remove "already"** — Line 1: `s/Go already removes/Go removes/` for tighter prose.

**5. Section order** — Consider moving "Current v3 Build Tags" after "Practical Checks". The reader benefits from learning *how to measure* before learning about v3-specific tag caveats.

The `mkdocs.yml` change is indented correctly and follows the existing nav pattern.

rupayon123:
Thanks for the detailed pass. I updated the page with the issue link, `make check-binary-size`, a combined release-build command, the wording cleanup, and moved the build-tag caveat after the practical checks.

I reran `go test ./...`, `make check-binary-size`, and `mkdocs build --strict`; all pass.

dearchap (approved):

rupayon123:
Thanks for the review and merge. Glad the docs update landed; I followed you here too so I can keep up with the project.

PR #2279, opened 2026-03-13
Improve the completion command to handle flags and arguments properly

suzuki-shunsuke:
## What type of PR is this?

_(REQUIRED)_

- bug
- feature

## What this PR does / why we need it:

_(REQUIRED)_

- Skip VisiblePersistentFlags in case of completion command
- Skip checking required flags in case of completion command
- Parse flags for completion command same with normal commands
- Change shells to sub commands to handle flags and arguments properly

## Which issue(s) this PR fixes:

_(REQUIRED)_

Fixes #2278

## Special notes for your reviewer:

_(fill-in or delete this section)_

## Testing

_(fill-in or delete this section)_

I'm using urfave/cli/v3 for my OSS project [pinact](https://github.com/suzuki-shunsuke/pinact).
I built pinact using this feature branch by `replace` directive and ran completion commands.

<details>

```console
$ pinact completion --help
NAME:
   pinact completion - Output shell completion script for bash, zsh, fish, or Powershell

USAGE:
   pinact completion [command [command options]]

DESCRIPTION:
   Output shell completion script for bash, zsh, fish, or Powershell.
   Source the output to enable completion.

   # .bashrc
   source <(pinact completion bash)

   # .zshrc
   source <(pinact completion zsh)

   # fish
   pinact completion fish > ~/.config/fish/completions/pinact.fish

   # Powershell
   Output the script to path/to/autocomplete/pinact.ps1 an run it.


COMMANDS:
   zsh   Output zsh completion script
   fish  Output fish completion script
   pwsh  Output pwsh completion script
   bash  Output bash completion script

OPTIONS:
   --help, -h  show help
```

```console
$ pinact completion --foo
Incorrect Usage: flag provided but not defined: -foo
```

```console
$ pinact completion bash --help
NAME:
   pinact completion bash - Output bash completion script

USAGE:
   pinact completion bash [options]

OPTIONS:
   --help, -h  show help
```

```console
$ pinact completion bash --foo
Incorrect Usage: flag provided but not defined: -foo
```

```console
$ pinact completion help bash
NAME:
   pinact completion bash - Output bash completion script

USAGE:
   pinact completion bash

OPTIONS:
   --help, -h  show help
```

</details>

## Release Notes

_(REQUIRED)_

```release-note
feat(completion): Improve the completion command to handle flags and arguments properly
```

PR #2374, opened 2026-06-26
fix: keep completion subcommand order deterministic in help output

### LINKED ISSUE #2373: completion subcommands appear in non-deterministic order in help output
## My urfave/cli version is

[v3.10.0](https://github.com/urfave/cli/releases/tag/v3.10.0) (the bug was introduced in [v3.9.1](https://github.com/urfave/cli/releases/tag/v3.9.1))

## Checklist

## Dependency Management

## Describe the bug

The order of the shell completion subcommands (`bash`, `zsh`, `fish`, `pwsh`) in help output is non-deterministic across process runs.

`completion` was changed in #2279 (commit e8afe93601d05d891ad44ad846288ee352142a0b, released in [v3.9.1](https://github.com/urfave/cli/releases/tag/v3.9.1)) to expose each shell as its own subcommand. Those subcommands are generated by ranging over the `shellCompletions` map in `buildCompletionCommand` (`completion.go`):

```go
for shell, render := range shellCompletions {
    cmd.Commands = append(cmd.Commands, buildShellCompletionSubcommand(shell, render, appName))
}
```

Go randomizes map iteration order per process, so the order of these subcommands — and therefore the `COMMANDS:` section of `completion --help` — changes between runs.

This breaks downstream projects that generate documentation from `--help` output and commit it (e.g. via CI). The generated docs flip-flop depending on which run produced them, creating spurious diffs and noise commits.

## To reproduce

```go
package main

import (
	"context"
	"os"

	cli "github.com/urfave/cli/v3"
)

func main() {
	cmd := &cli.Command{
		Name:                  "app",
		EnableShellCompletion: true,
	}
	_ = cmd.Run(context.Background(), []string{"app", "completion", "--help"})
	_ = os.Stdout
}
```

Build the binary and run `app completion --help` several times (each run is a separate process). The order of the commands under `COMMANDS:` varies.

It can also be reproduced directly with a small test that builds the completion command repeatedly across `go test` invocations; the subcommand name order is usually `bash,zsh,fish,pwsh` but occasionally `pwsh,bash,zsh,fish`, etc.

## Observed behavior

The completion subcommands appear in a random order that differs between process runs, e.g. one run prints:

```
COMMANDS:
   bash  Output bash completion script
   zsh   Output zsh completion script
   fish  Output fish completion script
   pwsh  Output pwsh completion script
```

and another run prints:

```
COMMANDS:
   pwsh  Output pwsh completion script
   bash  Output bash completion script
   zsh   Output zsh completion script
   fish  Output fish completion script
```

## Expected behavior

The completion subcommands should appear in a fixed, deterministic order (e.g. `bash, zsh, fish, pwsh`) on every run, so that help output and any documentation generated from it are stable.

## Additional context

Real-world impact (CI repeatedly committing reordered docs): https://github.com/suzuki-shunsuke/ghtkn/pull/454

The fix is to iterate over an explicitly ordered list of shells instead of ranging over the map.

Image

## Want to fix this yourself?

Yes — I'll open a PR.

```
go version go1.26.4 darwin/arm64
```

```
GOOS=darwin
GOARCH=arm64
GOVERSION=go1.26.4
CGO_ENABLED=1
```

suzuki-shunsuke:
## What type of PR is this?

- bug

The `completion` command exposes each shell as its own subcommand (`bash`, `zsh`, `fish`, `pwsh`), introduced in #2279 (released in v3.9.1). The subcommands were generated by ranging over the `shellCompletions` map in `buildCompletionCommand`:

```go
for shell, render := range shellCompletions {
    cmd.Commands = append(cmd.Commands, buildShellCompletionSubcommand(shell, render, appName))
}
```

Go randomizes map iteration order per process, so the order of these subcommands — and therefore the `COMMANDS:` section of `completion --help` — changed between runs. This breaks downstream projects that generate documentation from `--help` output and commit it via CI: the generated docs flip-flop depending on which run produced them, creating spurious diffs and noise commits. Real-world impact: https://github.com/suzuki-shunsuke/ghtkn/pull/454

Changes:

- `completion.go`: add `completionShells = []string{"bash", "zsh", "fish", "pwsh"}`, an explicitly ordered list of the supported shells, and iterate it (looking the renderer up in `shellCompletions`) instead of ranging over the map. The map stays as the renderer registry; the slice fixes the display order. The order is now deterministic on every run.
- `completion_test.go`: add `TestCompletionSubcommandOrder`, which asserts the subcommand order is always `bash, zsh, fish, pwsh` and that every shell in `shellCompletions` is also present in `completionShells` (so the two cannot silently drift). `TestCompletionShellRenderError` and `TestCompletionShellWriteError` inject a custom shell into `shellCompletions`; they now also register it in `completionShells` (with restore) so the injected shell still becomes a subcommand.

Fixes #2373

The display order is fixed to `bash, zsh, fish, pwsh`. If you would prefer alphabetical order instead, that is a one-line change to `completionShells`.

## Testing

- `go test ./...` passes.
- The completion tests are stable across repeated `go test` invocations (separate processes), where the subcommand order previously varied.

## Release Notes

```release-note
Fix non-deterministic ordering of `completion` shell subcommands in help output; they now always appear as bash, zsh, fish, pwsh.
```

### DIFF
--- a/completion.go
+++ b/completion.go
@@ -20,6 +20,12 @@ var (
 	//go:embed autocomplete
 	autoCompleteFS embed.FS
 
+	// completionShells defines the order in which the shell completion
+	// subcommands appear in help output. Iterating shellCompletions directly
+	// would use Go's randomized map order, making the listing nondeterministic.
+	// Keep this in sync with shellCompletions.
+	completionShells = []string{"bash", "zsh", "fish", "pwsh"}
+
 	shellCompletions = map[string]renderCompletion{
 		"bash": func(c *Command, appName string) (string, error) {
 			b, err := autoCompleteFS.ReadFile("autocomplete/bash_autocomplete")
@@ -65,8 +71,8 @@ func buildCompletionCommand(appName string) *Command {
 		isCompletionCommand: true,
 	}
 
-	for shell, render := range shellCompletions {
-		cmd.Commands = append(cmd.Commands, buildShellCompletionSubcommand(shell, render, appName))
+	for _, shell := range completionShells {
+		cmd.Commands = append(cmd.Commands, buildShellCompletionSubcommand(shell, shellCompletions[shell], appName))
 	}
 
 	return cmd
--- a/completion_test.go
+++ b/completion_test.go
@@ -519,6 +519,9 @@ func TestCompletionShellRenderError(t *testing.T) {
 		}
 		return "something", nil
 	}
+	// buildCompletionCommand only turns shells listed in completionShells into
+	// subcommands, so register the injected shell there too (restoring the
+	// original slice afterward) for it to be reachable.
 	defer func(orig []string) { completionShells = orig }(completionShells)
 	completionShells = append(completionShells, unknownShellName)
 	defer func() {
@@ -549,6 +552,9 @@ func TestCompletionShellWriteError(t *testing.T) {
 	shellCompletions[shellName] = func(c *Command, appName string) (string, error) {
 		return "something", nil
 	}
+	// buildCompletionCommand only turns shells listed in completionShells into
+	// subcommands, so register the injected shell there too (restoring the
+	// original slice afterward) for it to be reachable.
 	defer func(orig []string) { completionShells = orig }(completionShells)
 	completionShells = append(completionShells, shellName)
 	defer func() {
--- a/completion_test.go
+++ b/completion_test.go
@@ -118,6 +118,31 @@ func TestCompletionShell(t *testing.T) {
 	}
 }
 
+func TestCompletionSubcommandOrder(t *testing.T) {
+	// The completion subcommands must appear in a deterministic order so that
+	// help output (and docs generated from it) does not change between runs.
+	// Previously they were built by iterating a map, whose order Go randomizes.
+	want := []string{"bash", "zsh", "fish", "pwsh"}
+
+	// Build several times to guard against intra-process variation.
+	for range 10 {
+		cmd := buildCompletionCommand("foo")
+
+		got := make([]string, 0, len(cmd.Commands))
+		for _, sub := range cmd.Commands {
+			got = append(got, sub.Name)
+		}
+
+		assert.Equal(t, want, got)
+	}
+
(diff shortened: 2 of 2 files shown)


completion.go
```diff
@@ -20,6 +20,12 @@ var (
 	//go:embed autocomplete
 	autoCompleteFS embed.FS
 
+	// completionShells defines the order in which the shell completion
+	// subcommands appear in help output. Iterating shellCompletions directly
+	// would use Go's randomized map order, making the listing nondeterministic.
+	// Keep this in sync with shellCompletions.
+	completionShells = []string{"bash", "zsh", "fish", "pwsh"}
```
abitrolly, line 27: Is it possible to get sorted list instead of maintaining hardcoded indexes?

suzuki-shunsuke: In case of the dictionary order, the order is changed.

1. bash
2. fish
3. pwsh
4. zsh

Is that okay?

I think the current order is more natural.

1. bash
2. zsh
3. fish
4. pwsh

https://cli.urfave.org/v3/examples/completions/shell-completions/

image

dearchap:
## Review: PR #2374 — Deterministic completion subcommand order

**Verdict: Clean fix. No bugs.**

The root cause is correctly identified (Go map iteration randomization). The fix is minimal and correct: a `completionShells` slice to control iteration order, while keeping `shellCompletions` as the renderer registry.

### Minor observations (none blocking):

1. **10-iteration loop in the test** — The subcommand order is now purely slice-driven, so there is no possible intra-process variation. A single assertion would suffice. Not a problem, just redundant.

2. **`completionShells` is a package-level `var` mutated by two tests** — `TestCompletionShellRenderError` and `TestCompletionShellWriteError` both `append` to it and defer-restore the original. Correctly done, no race in practice since these are sequential tests. Worth noting for future maintenance.

3. **The guard test** (`assert.Len` + loop over map keys) is a nice touch — prevents the two data structures from drifting silently.

Ready to merge.

dearchap (approved):

PR #2376, opened 2026-06-29
fix: add HasHiddenParent() to skip hidden command subtrees in Walk()

### LINKED ISSUE #2372: cmd.Hidden is not recursive
## My urfave/cli version is

`v3@f980ca8`

## Checklist

## Dependency Management

## Describe the bug

I am using the [new Walk() feature](https://github.com/urfave/cli/pull/2353) by @dearchap to generate a tree of supported commands of my tool. This works great.

However, I want to exclude Hidden commands from the output. I can query the `Hidden` flag and do that, however, subcommands of a hidden command show up anyway.

## To reproduce

```go
func Tree(cmd *cli.Command) error {
	max := 20

	cmd.Walk(func(cmd *cli.Command) error {
		path := cmd.Path()
		command := path[len(path)-1]

		if len(path) == 1 || command == "help" || cmd.Hidden {
			return nil
		}

		indent := strings.Repeat("  ", len(path[1:])-1)
		space := strings.Repeat(" ", max-(len(command)+len(indent)))

		fmt.Printf("%s%s %s - %s\n", indent, command, space, cmd.Usage)

		return nil
	})

	return nil
}
```

## Observed behavior

This prints this at the end:
```default
[..]
version               - show esctl version information
debug                 - developer only
help-jsonpath         - show jsonpath help
  bash                - Output bash completion script
  zsh                 - Output zsh completion script
  fish                - Output fish completion script
  pwsh                - Output pwsh completion script
```
So, the command `completion` has been skipped, but its subcommands are not hidden and show up anyway.

## Expected behavior

I'd like to have a way to check the parent of a cmd if it's hidden, eg:

```go
if len(path) == 1 || command == "help" || cmd.Hidden || cmd.Parent.Hidden {
	return nil
}
```

## Additional context

Complete code can be found [here](https://codeberg.org/scip/esctl), to generate the tree, run: `esctl --show-command-tree`

```
go version go1.25.8 linux/amd64
```

```
AR='ar'
CC='gcc'
CGO_CFLAGS='-O2 -g'
CGO_CPPFLAGS=''
CGO_CXXFLAGS='-O2 -g'
CGO_ENABLED='1'
CGO_FFLAGS='-O2 -g'
CGO_LDFLAGS='-O2 -g'
CXX='g++'
GCCGO='gccgo'
GO111MODULE=''
GOAMD64='v1'
GOARCH='amd64'
GOAUTH='netrc'
GOBIN=''
GOCACHE='/home/scip/.cache/go-build'
GOCACHEPROG=''
GODEBUG=''
GOENV='/home/scip/.config/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFIPS140='off'
GOFLAGS=''
GOGCCFLAGS='-fPIC -m64 -pthread -Wl,--no-gc-sections -fmessage-length=0 -ffile-prefix-map=/tmp/go-build3184483892=/tmp/go-build -gno-record-gcc-switches'
GOHOSTARCH='amd64'
GOHOSTOS='linux'
GOINSECURE=''
GOMOD='/home/scip/dev/esctl/go.mod'
GOMODCACHE='/home/scip/go/pkg/mod'
GONOPROXY=''
GONOSUMDB=''
GOOS='linux'
GOPATH='/home/scip/go'
GOPRIVATE=''
GOPROXY='https://proxy.golang.org,direct'
GOROOT='/home/scip/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.8.linux-amd64'
GOSUMDB='sum.golang.org'
GOTELEMETRY='off'
GOTELEMETRYDIR='/home/scip/.config/go/telemetry'
GOTMPDIR=''
GOTOOLCHAIN='auto'
GOTOOLDIR='/home/scip/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.8.linux-amd64/pkg/tool/linux_amd64'
GOVCS=''
GOVERSION='go1.25.8'
GOWORK=''
PKG_CONFIG='pkg-config'
```

And many thanks for the cool `Walk()` and `Path()` feature, I'm loving it!

dearchap:
## What type of PR is this?

- feature

When using `Walk()` to traverse the command tree, users who want to skip hidden commands currently check `cmd.Hidden`. However, the subcommands of a hidden command are not themselves hidden and still appear in the walk output. There is no way to check whether a command belongs to a hidden subtree without manually walking parent pointers, which are unexported.

This PR adds `HasHiddenParent() bool` to `Command`, which returns `true` if any ancestor has `Hidden: true`. Users can combine it with `cmd.Hidden` in their `Walk` callback to skip entire hidden subtrees:

```go
cmd.Walk(func(c *cli.Command) error {
    if c.Hidden || c.HasHiddenParent() {
        return nil
    }
    // process visible commands only
})
```

Fixes #2372

## Testing

- `go test ./...` passes
- `go vet ./...` passes
- `make v3diff` shows only the new `HasHiddenParent` method
- `make v3approve` run to update API surface docs

## Release Notes

```release-note
Add Command.HasHiddenParent() method to help users skip subcommands of hidden commands during Walk().
```

### DIFF
--- a/testdata/godoc-v3.x.txt
+++ b/testdata/godoc-v3.x.txt
@@ -821,6 +821,17 @@ func (cmd *Command) Walk(fn func(*Command) error) error
     Walk visits cmd and every descendant. If fn returns a non-nil error,
     the walk terminates and the error is returned to the caller.
 
+    To skip subcommands of hidden commands in the walk, check both Hidden and
+    HasHiddenParent in the callback:
+
+        cmd.Walk(func(c *cli.Command) error {
+            if c.Hidden || c.HasHiddenParent() {
+                return nil
+            }
+            // process visible commands only
+            return nil
+        })
+
 type CommandCategories interface {
 	// AddCommand adds a command to a category, creating a new category if necessary.
 	AddCommand(category string, command *Command)
--- a/testdata/godoc-v3.x.txt
+++ b/testdata/godoc-v3.x.txt
@@ -618,6 +618,10 @@ func (cmd *Command) FullName() string
 func (cmd *Command) Generic(name string) Value
     Generic looks up the value of a local GenericFlag, returns nil if not found
 
+func (cmd *Command) HasHiddenParent() bool
+    HasHiddenParent returns true if any ancestor of this command has Hidden set
+    to true.
+
 func (cmd *Command) HasName(name string) bool
     HasName returns true if Command.Name matches given name
 
--- a/command.go
+++ b/command.go
@@ -596,6 +596,17 @@ func (cmd *Command) HasHiddenParent() bool {
 
 // Walk visits cmd and every descendant. If fn returns a non-nil error, the
 // walk terminates and the error is returned to the caller.
+//
+// To skip subcommands of hidden commands in the walk, check both Hidden and
+// HasHiddenParent in the callback:
+//
+//	cmd.Walk(func(c *cli.Command) error {
+//	    if c.Hidden || c.HasHiddenParent() {
+//	        return nil
+//	    }
+//	    // process visible commands only
+//	    return nil
+//	})
 func (cmd *Command) Walk(fn func(*Command) error) error {
 	if fn == nil {
 		return nil
--- a/command.go
+++ b/command.go
@@ -582,6 +582,18 @@ func (cmd *Command) Path() []string {
 	return []string{cmd.Name}
 }
 
+// HasHiddenParent returns true if any ancestor of this command has
+// Hidden set to true.
+func (cmd *Command) HasHiddenParent() bool {
+	if cmd.parent == nil {
+		return false
+	}
+	if cmd.parent.Hidden {
+		return true
+	}
+	return cmd.parent.HasHiddenParent()
+}
+
 // Walk visits cmd and every descendant. If fn returns a non-nil error, the
 // walk terminates and the error is returned to the caller.
 func (cmd *Command) Walk(fn func(*Command) error) error {
--- a/command_test.go
+++ b/command_test.go
@@ -6450,3 +6450,56 @@ func TestCommand_Walk_NilFn(t *testing.T) {
 	cmd := &Command{Name: "foo"}
 	assert.Nil(t, cmd.Walk(nil))
 }
+
+func TestCommand_HasHiddenParent(t *testing.T) {
+	t.Run("root command", func(t *testing.T) {
+		cmd := &Command{Name: "root"}
+		assert.False(t, cmd.HasHiddenParent())
+	})
+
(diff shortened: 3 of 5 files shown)


abitrolly:
Is it possible to add this to Walk examples?

dearchap:
@abitrolly Added

testdata/godoc-v3.x.txt
```diff
@@ -821,6 +821,17 @@ func (cmd *Command) Walk(fn func(*Command) error) error
...
+    To skip subcommands of hidden commands in the walk, check both Hidden and
+    HasHiddenParent in the callback:
+
+        cmd.Walk(func(c *cli.Command) error {
+            if c.Hidden || c.HasHiddenParent() {
+                return nil
+            }
+            // process visible commands only
+            return nil
+        })
```
abitrolly, line 833: If `c.Hidden` is set, then the function returns `nil`, the walk terminates, and `c.HasHiddenParent` is never reached, right?

dearchap: yeah

abitrolly: @dearchap so what's the point in having `c.HasHiddenParent` condition here? It could be that immediate parent is not hidden, but the grandfather is. Maybe `c.AnyHiddenParent` method would be better? If the logic is `if any([c.Hidden for c in [c] + c.Parents])`.

dearchap: See the implementation of HasHiddenParent. If any command in the chain has Hidden set it will return true

abitrolly: @dearchap I feel like the better API would be to add a generic way to check some property in current and all parent commands. Otherwise there has to be a plenty of method for each property.

And also, how to check that only the parent command has hidden status? If I want to hide intermediate command, but make all its children visible? This `HasHiddenParent` method seems too specific as for me.

dearchap: @abitrolly do you have a design in mind ?

abitrolly: @dearchap yes, I think so.

Pseudocode.
```
func (cmd *Command) Parents(withself=False) []cmd {
  if cmd.parent == nil {
    return []
  }
  parents = []
  curcmd = cmd
  while curcmd.parent {
    parents.append(curcmd.parent)
    curcmd = curcmd.parent
  }
  if withself {
    parents += cmd
  }
  return parents
}
```

And then everybody can do this by passing parents.

Pseudocode.
```
func IsBranchHidden(cmdbranch []cmd) {
  for c in cmdbranch {
    if c.Hidden {
      return True
    }
  }
  return False
}
```

Sry. Typing from the browser, or else there won't be a chance for reply.

PR #2380, opened 2026-07-02
Handle Before errors with OnUsageError

### LINKED ISSUE #707: Inconsistent error handling between App.OnUsageError and App.Before during App.Run

cyphercodes:
## What type of PR is this?

- bug

- Calls `OnUsageError` when a `Before` hook returns an error, matching the existing usage-error path and allowing applications to customize the returned error/output.
- Preserves the existing fallback behavior when no `OnUsageError` handler is configured.
- Adds a regression test for app-level `Before` errors and updates the operation-order expectations.

Fixes #707

This is intentionally limited to the `Before` error path discussed in the issue.

## Testing

- `go test ./...`
- `go vet ./...`
- `git diff --check`

## Release Notes

```release-note
`Before` hook errors now invoke `OnUsageError` when configured.
```

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -1346,6 +1346,37 @@ func TestApp_BeforeFunc(t *testing.T) {
 	}
 }
 
+func TestApp_OnUsageErrorHandlesBeforeError(t *testing.T) {
+	beforeError := errors.New("before failed")
+	usageError := errors.New("usage handled")
+	var called bool
+	var gotErr error
+	var gotIsSubcommand bool
+
+	app := &App{
+		Before: func(_ *Context) error {
+			return beforeError
+		},
+		OnUsageError: func(_ *Context, err error, isSubcommand bool) error {
+			called = true
+			gotErr = err
+			gotIsSubcommand = isSubcommand
+			return usageError
+		},
+		Action: func(_ *Context) error {
+			t.Fatal("Action should not be called when Before fails")
+			return nil
+		},
+	}
+
+	err := app.Run([]string{"command"})
+
+	expect(t, called, true)
+	expect(t, gotErr, beforeError)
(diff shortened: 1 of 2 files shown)


dearchap:
@cyphercodes v2 is in maint. Want to port this to v3 ?

PR #2381, opened 2026-07-02
fix: error on legacy v1-style flag alias syntax

### LINKED ISSUE #1103: v2 bug: Use of obsolete v1 aliases goes undetected

happysnaker:
Fixes #1103

This changes legacy v1-style flag alias definitions like `Name: "config, cfg"` from a silent partial-accept path into an explicit error.

What this patch does:
- validates `Name` / `Aliases` before registering each flag
- rejects commas or spaces inside the primary `Name` field
- rejects malformed aliases that still contain commas or spaces
- adds a regression test that asserts `App.Run()` errors with a migration hint
- updates the long-flag completion example to use v2-style `Aliases` explicitly

Why this shape:
- today the old syntax looks half-working: the primary flag still exists, but the alias silently disappears
- failing fast is easier to debug and matches the issue request for an error/warning rather than hidden misconfiguration
- validating in each flag `Apply()` path keeps the behavior consistent for direct flag registration and wrappers like `altsrc` that delegate to the same `Apply()` methods

I intentionally kept the error message migration-oriented rather than trying to preserve the old behavior.

### DIFF
--- a/app_test.go
+++ b/app_test.go
@@ -283,7 +283,8 @@ func ExampleApp_Run_bashComplete_withLongFlag() {
 			Aliases: []string{"x"},
 		},
 		&StringFlag{
-			Name: "some-flag,s",
+			Name:    "some-flag",
+			Aliases: []string{"s"},
 		},
 		&StringFlag{
 			Name: "similar-flag",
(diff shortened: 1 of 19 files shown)


dearchap:
@happysnaker We arent updating v2 anymore other than for security related fixes.

PR #2393, opened 2026-07-20
Feat:(issue_2392) Single-value arguments can be marked as required

### LINKED ISSUE #2392: No way to mark a single-value argument as required
## Checklist

* [x] Did you check the manual for your release? The v3 manual is [here](https://github.com/urfave/cli/blob/main/docs/v3/index.md).

Version: urfave/cli v3.10.1 (also on latest main at commit 4937c16)

There is no way to require a single-value argument (`{Type}Arg`). `ArgumentBase.Parse` falls back to the default value when no argument is given, so a command like `app delete` runs with an empty ID instead of failing.

The closest workaround is the plural type with `Min`/`Max`:

```go
var ids []string

&cli.StringArgs{Name: "id", Min: 1, Max: 1, Destination: &ids}
```

which needs a slice destination for a single value, and shows this when the argument is missing:

```
sufficient count of arg id not provided, given 0 expected 1
```

## Solution description

The same thing flags already have (`FlagBase.Required`):

```go
&cli.StringArg{Name: "id", Required: true, Destination: &id}
```

With `Required: true`, `Parse` returns `Required argument "id" not set` when the argument is missing.

Help output could also follow the bracket convention that `{Type}Args` already uses: `[name]` when optional, `name` when required. Today single args always render as `name` even though they are optional. This changes the expectation in `TestArgUsage` from `ia` to `[ia]`; no other test encodes the current rendering.

Making `ArgumentBase.Parse` error when the argument is missing, without a new field. That would be a breaking change — the fallback to `Value` is intentional (`TestSingleOptionalArg`).

A `Min` field like the plural type has. On a single value it can only ever be 0 or 1, so a bool seemed like a better fit.

idelchi:
## What type of PR is this?

- feature

- `args.go`: adds a `Required` field to `ArgumentBase`, same as flags have. A missing required argument returns `required argument "name" not set`. `Usage()` now renders optional single args as `[name]`, matching what `{Type}Args` already does.
- `args_test.go`: adds `TestSingleRequiredArg` and `TestChainedRequiredArgs`. The `TestArgUsage` expectation for an optional arg changes from `ia` to `[ia]`.
- `docs/v3/examples/arguments/advanced.md`: short section on required arguments.
- `godoc-current.txt` / `testdata/godoc-v3.x.txt`: regenerated.

A default `Value` does not satisfy `Required`, same as with flags.

Fixes #2392

The help output change (optional single args rendering as `[name]`) is a separate commit and can be dropped if unwanted.

## Testing

- `go test ./...`
- `make gfmrun` for the docs example
- `make generate` + `make v3approve` for the godoc diff check

## Release Notes

```release-note
Single-value arguments can be marked as `Required`. Optional single-value arguments now render as `[name]` in help output.
```

### DIFF
--- a/args.go
+++ b/args.go
@@ -104,7 +104,10 @@ func (a *ArgumentBase[T, C, VC]) Usage() string {
 		return a.UsageText
 	}
 
-	usageFormat := "%[1]s"
+	usageFormat := "[%[1]s]"
+	if a.Required {
+		usageFormat = "%[1]s"
+	}
 	return fmt.Sprintf(usageFormat, a.Name)
 }
 
--- a/args.go
+++ b/args.go
@@ -89,6 +89,7 @@ type ArgumentBase[T any, C any, VC ValueCreator[T, C]] struct {
 	Value       T      `json:"value"`     // the default value of this argument
 	Destination *T     `json:"-"`         // the destination point for this argument
 	UsageText   string `json:"usageText"` // the usage text to show
+	Required    bool   `json:"required"`  // whether the argument is required or not
 	Config      C      `json:"config"`    // config for this argument similar to Flag Config
 
 	value *T
@@ -109,6 +110,10 @@ func (a *ArgumentBase[T, C, VC]) Usage() string {
 
 func (a *ArgumentBase[T, C, VC]) Parse(s []string) ([]string, error) {
 	tracef("calling arg%[1] parse with args %[2]", a.Name, s)
+
+	if a.Required && len(s) == 0 {
+		return s, fmt.Errorf("required argument %q not set", a.Name)
+	}
 
 	var vc VC
 	var t T
--- a/args.go
+++ b/args.go
@@ -77,6 +77,11 @@ type Argument interface {
 	Get() any
 }
 
+type requiredArgument interface {
+	name() string
+	required() bool
+}
+
 // AnyArguments to differentiate between no arguments(nil) vs aleast one
 var AnyArguments = []Argument{
 	&StringArgs{
@@ -99,6 +104,14 @@ func (a *ArgumentBase[T, C, VC]) HasName(s string) bool {
 	return s == a.Name
 }
 
+func (a *ArgumentBase[T, C, VC]) name() string {
+	return a.Name
+}
+
+func (a *ArgumentBase[T, C, VC]) required() bool {
+	return a.Required
+}
+
 func (a *ArgumentBase[T, C, VC]) Usage() string {
 	if a.UsageText != "" {
 		return a.UsageText
@@ -115,7 +128,7 @@ func (a *ArgumentBase[T, C, VC]) Parse(s []string) ([]string, error) {
 	tracef("calling arg%[1] parse with args %[2]", a.Name, s)
 
 	if a.Required && len(s) == 0 {
-		return s, fmt.Errorf("required argument %q not set", a.Name)
+		return s, &errRequiredArguments{missingArguments: []string{a.Name}}
 	}
 
 	var vc VC
--- a/args_test.go
+++ b/args_test.go
@@ -552,6 +552,81 @@ func TestSingleOptionalArg(t *testing.T) {
 	}
 }
 
+func TestSingleRequiredArg(t *testing.T) {
+	tests := []struct {
+		name     string
+		args     []string
+		argValue string
+		exp      string
+		expErr   string
+	}{
+		{
+			name:   "no args",
+			args:   []string{"foo"},
+			expErr: `required argument "sa" not set`,
+		},
+		{
+			name:     "no arg with def value",
+			args:     []string{"foo"},
+			argValue: "bar",
+			expErr:   `required argument "sa" not set`,
+		},
+		{
+			name: "one arg",
+			args: []string{"foo", "zbar"},
+			exp:  "zbar",
+		},
+		{
+			name: "empty string arg",
+			args: []string{"foo", ""},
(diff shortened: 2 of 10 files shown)


dearchap:
Thanks for the PR @idelchi! Nice, well-scoped feature. The semantics correctly mirror flags (a default `Value` doesn't satisfy `Required`, matching `checkRequiredFlag`'s `IsSet()` check in `command.go:410`), the tests cover the interesting cases (default-not-satisfying, empty-string-provided, chained required args), and the docs gfmrun error spec matches the existing pattern. CI is green.

A few suggestions, in rough priority order:

1. **Error UX inconsistent with flags (main point).** A missing required *flag* prints `Incorrect Usage: ...` plus help via `checkAllRequiredFlags` (`command_run.go:346`), returns a typed error (`requiredFlagsErr`), and reports *all* missing flags at once. A missing required *arg* returns a bare `fmt.Errorf` mid-parse (`command_run.go:369-377`): no help is shown, it isn't programmatically detectable, and it only reports the first missing arg. Consider pre-scanning arguments like the flags path (or at least wrapping in a typed error) so the behavior is parallel. (Minor: `required argument "x" not set` vs the flags' capitalized `Required flag "x" not set`.)

2. **Required-after-optional ordering is silently unsatisfiable.** With `[optional, required]`, the optional arg always consumes the token, so the required one can never be satisfied. This is inherent to positional args (same as multi-value `Min`), but a one-line note in the docs — or an ordering test — would help, since the docs currently don't mention it.

3. **Docs import ordering.** The new example has `"os"` before `"context"` (pre-existing in the file too); goimports would reorder. gfmrun only compiles it, so it passes, but it's untidy.

4. **Minor test suggestion.** Consider asserting `Destination`/`Get()` are untouched on the missing-required path, and a help-output test that renders `[sa]` vs `sa` for the required case — `TestArgUsage` covers `Usage()` but not the full help line.

I'd be happy to merge once #1 is decided on (align with the flags error path, or explicitly accept the divergence).

dearchap (changes requested):
## Issues to address

### 1. Duplicate enforcement with divergent error handling (moderate)
`Required` is checked in two places:

- `checkRequiredArguments` (command.go:457) — pre-parse path that prints `Incorrect Usage: ...` + help and sets `isInError`.
- `ArgumentBase.Parse` (args.go) — the arg-parsing loop only calls `OnUsageError` + `handleExitCoder`, with no help output and no `isInError`.

The `Parse`-level check is effectively dead code for the documented (required-first) ordering and only fires when a multi-value/variadic arg precedes a required arg. When it does fire, the error handling differs from the primary path (no help, different exit handling). Consider dropping one, or unifying the handling so both paths behave identically.

### 2. Index-based counting breaks with multi-value args (command.go:465)
`index >= providedArguments` assumes each declared argument consumes exactly one positional. I reproduced with:

```go
Arguments: []Argument{
    &StringArgs{Name: "rest", Min: 0, Max: -1},
    &StringArg{Name: "required", Required: true},
}
cmd.Run(ctx, []string{"foo", "a", "b", "c"}) // → error: Required argument "required" not set
```

3 args are provided but the variadic consumes them all, so the arg genuinely receives no value — but the error is misleading. The docs do tell users to declare required args before optional/multi-value args, and the `Parse` backstop catches it, so acceptable — but worth a comment or an error message that hints at the ordering rule.

### 3. Missing `builtInHelp || isCompletionCommand` guard (command.go:457)
`checkAllRequiredFlags` (command.go:424) skips both; `checkRequiredArguments` doesn't. Currently harmless because `--help` short-circuits earlier via `checkHelp()` (command_run.go:215) and the built-in completion subcommands have no `Arguments`, but it's fragile if help/completion flow changes. Add the same guard for parity.

## Minor / nits

- `requiredArgument` interface uses unexported methods (`name()`, `required()`), diverging from flags' public `RequiredFlag` pattern (flag.go:116). Custom third-party `Argument` implementations can never participate — fine, but inconsistent.
- `Usage()` now renders optional single args as `[name]` — a visible change to existing help output. It's consistent with `ArgumentsBase` and covered by the updated `TestArgUsage`, so I'd keep it, but it deserves the explicit release-note mention (which it has).
- Empty string (`""`) counts as "provided" — consistent with `--flag ""`, fine.

Net: reasonable to merge after addressing #1 and considering #2/#3.

idelchi:
Thanks for the review @dearchap Take a look if it's okay now

dearchap (approved):
LGTM. All feedback addressed and verified locally: full test suite passes,  and  clean, and all 9 new tests pass on the PR head (e.g. , , ).

The duplicate-enforcement concern is resolved by routing both the pre-parse  and the  backstop through , so error handling is identical on both paths. The count-based precheck is correct under the now-documented required-first ordering, with  as the fallback for unsupported orderings. CI is green across the full matrix.

dearchap (approved):
LGTM. All feedback addressed and verified locally: full test suite passes, go vet and gofmt clean, and all 9 new tests pass on the PR head (e.g. TestMissingRequiredArgDoesNotMutateValue, TestRequiredArgAfterMultiValueArgUsesRequiredErrorHandling, TestCheckRequiredArgumentsSkipsHelpAndCompletionCommands).

The duplicate-enforcement concern is resolved by routing both the pre-parse checkRequiredArguments and the ArgumentBase.Parse backstop through handleRequiredError, so error handling is identical on both paths. The count-based precheck is correct under the now-documented required-first ordering, with Parse as the fallback for unsupported orderings. CI is green across the full matrix.

PR #2395, opened 2026-07-26
fix: inherit HideHelpCommand in subcommands

### LINKED ISSUE #2129: HideHelpCommand issues
## My urfave/cli version is

v3.3.3

## Checklist

- [ ] Did you perform a search about this problem? Here's the [GitHub guide](https://help.github.com/en/github/managing-your-work-on-github/using-search-to-filter-issues-and-pull-requests) about searching.

## Dependency Management

## Describe the bug

I have noticed strange results produced by the default help template when my application is using `HideHelpCommand`:

Using `HideHelpCommand` in the root command and in a subcommand that has only one command (e.g. `program config generate`), the output is displayed without "COMMANDS" section.

https://go.dev/play/p/v7DRNLiVRzN

Using `HideHelpCommand` in the root only affects subcommands in the sense that "help" command is not shown in "COMMANDS" section, however the "help" command can still be invoked when it is specified as an argument.

https://go.dev/play/p/cGhEL-usWcW

## To reproduce

I have attached Go playground links in the description.

## Observed behavior

See the description.

## Expected behavior

HideHelpCommand should hide help command from the usage output as well as disable the command when specified. Ideally, the behavior should be inherited from the parent command.

## Additional context

None.

## Want to fix this yourself?

pull request for it.

```
# paste `go version` output in here
```

```
```

utkarshalpha:
## What type of PR is this?

- bug

- Makes `HideHelpCommand` inherit through the command hierarchy, matching the existing behavior of `HideHelp`.
- Prevents nested commands from re-adding the built-in `help` command when an ancestor disabled it.
- Extends the existing test through two nested command levels while confirming `--help` remains available.

Fixes #2129

## Testing

- `go test ./... -count=1`
- `go run scripts/build.go generate`
- `go run scripts/build.go vet`
- `go run scripts/build.go check-binary-size`
- `go run scripts/build.go gfmrun`

## Release Notes

```release-note
Make HideHelpCommand apply to nested subcommands.
```

### DIFF
--- a/command_setup.go
+++ b/command_setup.go
@@ -227,6 +227,17 @@ func (cmd *Command) hideHelp() bool {
 	return false
 }
 
+func (cmd *Command) hideHelpCommand() bool {
+	tracef("hide help command (cmd=%[1]q)", cmd.Name)
+	for c := cmd; c != nil; c = c.parent {
+		if c.HideHelpCommand {
+			return true
+		}
+	}
+
+	return false
+}
+
 func (cmd *Command) ensureHelp() {
 	tracef("ensuring help (cmd=%[1]q)", cmd.Name)
 
@@ -234,7 +245,7 @@ func (cmd *Command) ensureHelp() {
 
 	if !cmd.hideHelp() {
 		if cmd.Command(helpCommand.Name) == nil {
-			if !cmd.HideHelpCommand {
+			if !cmd.hideHelpCommand() {
 				tracef("appending helpCommand (cmd=%[1]q)", cmd.Name)
 				cmd.appendCommand(helpCommand)
 			}
--- a/command.go
+++ b/command.go
@@ -46,6 +46,9 @@ type Command struct {
 	Flags []Flag `json:"flags"`
 	// Boolean to hide built-in help command and help flag
 	HideHelp bool `json:"hideHelp"`
+	// Boolean to hide the built-in help command. Applies to this command and
+	// all of its subcommands: as with HideHelp, a true value is inherited and
+	// a subcommand cannot turn it back off.
 	// Ignored if HideHelp is true.
 	HideHelpCommand bool `json:"hideHelpCommand"`
 	// Boolean to hide built-in version flag and the VERSION section of help
--- a/godoc-current.txt
+++ b/godoc-current.txt
@@ -464,6 +464,9 @@ type Command struct {
 	Flags []Flag `json:"flags"`
 	// Boolean to hide built-in help command and help flag
 	HideHelp bool `json:"hideHelp"`
+	// Boolean to hide the built-in help command. Applies to this command and
+	// all of its subcommands: as with HideHelp, a true value is inherited and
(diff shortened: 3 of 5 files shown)


command_setup.go
```diff
@@ -227,14 +227,25 @@ func (cmd *Command) hideHelp() bool {
...
+
 func (cmd *Command) ensureHelp() {
 	tracef("ensuring help (cmd=%[1]q)", cmd.Name)
 
 	helpCommand := buildHelpCommand(true)
 
 	if !cmd.hideHelp() {
 		if cmd.Command(helpCommand.Name) == nil {
-			if !cmd.HideHelpCommand {
+			if !cmd.hideHelpCommand() {
```
dearchap, line 248: Heads-up: this is a behavior change for nested trees, not just a bug fix. `hideHelpCommand()` walks the whole ancestor chain, so once any ancestor sets `HideHelpCommand: true`, a descendant cannot re-enable the built-in `help` command — even by explicitly setting `HideHelpCommand: false` on itself. Before this change, `app config help` still worked when only the root hid the command, because each command's own field value controlled the append. If that was intentional (matching `HideHelp` inheritance semantics), consider noting it explicitly in the release notes.

dearchap (commented):
Reviewed the change. Looks correct — the new test fails without the fix and passes with it, and test suite/vet/gofmt are clean. Two notes below (one behavior change to confirm, one minor doc touch-up).

command_setup.go
```diff
@@ -227,14 +227,25 @@ func (cmd *Command) hideHelp() bool {
 	return false
 }
 
+func (cmd *Command) hideHelpCommand() bool {
```
dearchap, line 230: Minor: with this change `HideHelpCommand` now propagates to all descendants, but its doc comment in `command.go` still reads `// Ignored if HideHelp is true.` only. A short note that the value applies to this command and its subcommands would prevent confusion about why a nested command lost its `help` command.

utkarshalpha:
Yes, that's intentional. HideHelp already works this way, it walks up the parent chain, so a subcommand setting HideHelp: false under a hidden root doesn't get help back either. This just makes HideHelpCommand behave the same as HideHelp.

I added two tests for it so it doesn't get flipped back by accident later.

Doc comment updated as you asked. I also had to update godoc-current.txt and testdata/godoc-v3.x.txt, since CI compares those against the source.

For the release note, maybe something like: HideHelpCommand is now inherited by subcommands, same as HideHelp. If a command sets it, the help command is hidden for that command and everything under it.

dearchap (approved):
LGTM. The fix is correct: hideHelpCommand() mirrors hideHelp() by walking the parent chain, and the parent pointers are guaranteed to be set before ensureHelp() runs for any subcommand (setupDefaults sets direct children parents before ensureHelp; setupCommandGraph uses a pre-order Walk). Verified locally: full suite + race pass, vet/gofmt clean, generated godoc matches the updated godoc-current.txt and testdata/godoc-v3.x.txt.

PR #2401, opened 2026-08-10
Fix index out of range panic when Run is given no arguments

mrueg:
`Command.Run` indexes into the argument slice in two places without first checking that it is non-empty, so passing an empty slice panics instead of returning.

### `setupDefaults`

Reads `osArgs[0]` to derive a default command name:

```go
cmd := &cli.Command{Action: func(context.Context, *cli.Command) error { return nil }}
cmd.Run(context.Background(), []string{})
// panic: runtime error: index out of range [0] with length 0
```

Affects any root command that does not set `Name` explicitly.

### `checkShellCompleteFlag`

Reads `arguments[len(arguments)-1]`:

```go
cmd := &cli.Command{Name: "x", EnableShellCompletion: true}
cmd.Run(context.Background(), nil)
// panic: runtime error: index out of range [-1]
```

Affects any command with `EnableShellCompletion` set, named or not.

## Fix

Guard both index accesses. A command invoked with no arguments now runs its `Action` as it would with any other argument list, and simply keeps an empty `Name` when there is no `argv[0]` to derive one from.

This is a pure robustness fix: in practice callers pass `os.Args`, which is never empty, so no working behaviour changes — the only reachable states previously were the two panics. It matters mostly for tests and for library consumers that build an argument slice programmatically.

## Testing

Adds `TestRunWithNoOsArgs` (named / shell-completion-enabled / unnamed) and a `no arguments at all` case to the existing `Test_checkShellCompleteFlag` table. Both fail on `main` without the change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/command_setup.go
+++ b/command_setup.go
@@ -24,7 +24,7 @@ func (cmd *Command) setupDefaults(osArgs []string) {
 		cmd.ShellComplete = DefaultCompleteWithFlags
 	}
 
-	if cmd.Name == "" && isRoot {
+	if cmd.Name == "" && isRoot && len(osArgs) > 0 {
 		name := filepath.Base(osArgs[0])
 		tracef("setting cmd.Name from first arg basename (cmd=%[1]q)", name)
 		cmd.Name = name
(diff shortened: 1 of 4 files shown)


dearchap (approved):
Fix is correct and verified: both guards in command_setup.go and help.go resolve the index-out-of-range panics; existing tests, vet, and lint pass. Test-coverage improvements tracked separately in #2406.

PR #2403, opened 2026-08-13
fix: honor BoolWithInverseFlag DefaultText

### LINKED ISSUE #2402: `BoolWithInverseFlag` does not honor explicitly set `DefaultText`
## My urfave/cli version is

v3.10.1

## Checklist

## Dependency Management

## Describe the bug

[`BoolWithInverseFlag`](https://github.com/urfave/cli/blob/main/flag_bool_with_inverse.go) does NOT honor its property [`DefaultText`](https://github.com/urfave/cli/blob/c6f4cf7e9223793478cfcde9b8f135cc8f86e78f/flag_bool_with_inverse.go#L15), which, if I am not mistaken, should override the `(default: […])` output in the help message regardless of any actual value.

## To reproduce

1. Create a new Go module
2. Add urfave/cli
3. Copy/paste the following snippet and run it

```go
package main

import (
	"context"
	"os"

	"github.com/urfave/cli/v3"
)

var cmd = &cli.Command{
	Name: "test",
	Flags: []cli.Flag{
		&cli.BoolFlag{
			Name:        "debug",
			Usage:       "set log level to DEBUG",
			Value:       true,
			DefaultText: "blabla", // for comparison
		},
                # control output color
		&cli.BoolWithInverseFlag{
			// accepts (true|false|auto) where "auto" is tty-aware
			// --color (not set) = auto
			// --color (set)     = ture
			// --no-color        = false
			Name:  "color",
			Usage: "colorise text output",
			// BUG: default text is not shown but the actual default value "false"
			DefaultText: "auto",
		},
	},
}

func main() {
	cmd.Run(context.Background(), os.Args)
}
```

## Observed behavior

```bash
$ go run main.go
NAME:
   test - A new cli application

USAGE:
   test [global options]

GLOBAL OPTIONS:
   --debug           set log level to DEBUG (default: blabla)
   --[no-]color    colorise text output (default: false)
   --help, -h        show help
```

## Expected behavior

I specified `DefaultText: "auto"` for the `--[no-]color` flag and expected the help message to print this value, regardless of the actual default value.

## Additional context

I was basically trying to achieve a ternary state flag using [`BoolWithInverseFlag`](https://github.com/urfave/cli/blob/main/flag_bool_with_inverse.go) and [`Flag.IsSet`](https://github.com/urfave/cli/blob/c6f4cf7e9223793478cfcde9b8f135cc8f86e78f/flag.go#L111), where the flag is not required and defaults to "auto":

1. default (user did nothing) → "auto" (i.e., `IsSet` evaluates to false)
2. user set `--color` → true
3. user set `--no-color` → false

This flag would allow to explicitly control colour printing in the application's output. If left untouched (=auto), it would automatically check whether the output stream (usually STDERR) is a TTY, and then set a proper value (TTY→true, no TTY→false).

Reading https://github.com/urfave/cli/pull/2214#issuecomment-3443192541, the current behaviour I perceived as a bug could be intentional though, as boolean flags usually suggest to have a binary state only.

## Want to fix this yourself?

I am not certain, but it appears that this method is the culprit:

https://github.com/urfave/cli/blob/c6f4cf7e9223793478cfcde9b8f135cc8f86e78f/flag_bool_with_inverse.go#L221-L226

```go
	if bif.Required {
		return bif.DefaultText
	}
```

is also a dead branch, because 

https://github.com/urfave/cli/blob/main/docs.go#L108-L117

prevents the `(default: […])` text from being rendered at all if the flag is required.

So the obvious solution could simply be:

```go
func (bif *BoolWithInverseFlag) GetDefaultText() string {
	if bif.DefaultText != "" {
		return bif.DefaultText
	}
	return boolValue{}.ToString(bif.Value)
}
```

I changed that locally and modified/extended the tests accordingly. They pass but

- with `Required`=true and no `DefaultText`, `GetDefaultText()` now returns "false" (was "" before)
- the suppressing of the default text now depends on external factors (`GetDefaultText` probably should not handle that anyway) 

```
go version go1.26.5 darwin/arm64
```

```
n/a
```

fzlzjerry:
## What type of PR is this?

- bug

- Make `BoolWithInverseFlag.GetDefaultText` honor an explicitly configured `DefaultText` before falling back to the flag's boolean value.
- Cover both the direct default-text behavior and the generated help output while preserving the existing `true`/`false` fallback.

Fixes #2402

## Testing

- `go test -race ./... -count=1`
- `go vet ./...`
- `make lint`
- `make check-binary-size`
- `make generate v3diff`
- `make diffcheck`

## Release Notes

```release-note
`BoolWithInverseFlag` now honors `DefaultText` in generated help output.
```

### DIFF
--- a/flag_bool_with_inverse.go
+++ b/flag_bool_with_inverse.go
@@ -219,7 +219,7 @@ func (bif *BoolWithInverseFlag) Count() int {
 
 // GetDefaultText returns the default text for this flag
 func (bif *BoolWithInverseFlag) GetDefaultText() string {
-	if bif.Required {
+	if bif.DefaultText != "" {
 		return bif.DefaultText
 	}
 	return boolValue{}.ToString(bif.Value)
--- a/flag_bool_with_inverse_test.go
+++ b/flag_bool_with_inverse_test.go
@@ -338,6 +338,12 @@ func TestBoolWithInverseNames(t *testing.T) {
 	require.Equal(t, "bool", d.TypeName())
 }
 
+func TestBoolWithInverseDefaultText(t *testing.T) {
+	require.Equal(t, "false", (&BoolWithInverseFlag{}).GetDefaultText())
+	require.Equal(t, "true", (&BoolWithInverseFlag{Value: true}).GetDefaultText())
+	require.Equal(t, "auto", (&BoolWithInverseFlag{DefaultText: "auto"}).GetDefaultText())
+}
+
 func TestBoolWithInverseString(t *testing.T) {
 	tcs := []struct {
 		testName      string
--- a/flag_test.go
+++ b/flag_test.go
@@ -483,6 +483,11 @@ func TestFlagStringifying(t *testing.T) {
 			expected: "--vividly, --no-vividly\t(default: false)",
 		},
 		{
+			name:     "bool-inv-flag-with-default-text",
+			fl:       &BoolWithInverseFlag{Name: "vividly", DefaultText: "auto"},
+			expected: "--vividly, --no-vividly\t(default: auto)",
+		},
+		{
 			name:     "duration-flag",
 			fl:       &DurationFlag{Name: "scream-for"},
 			expected: "--scream-for duration\t(default: 0s)",


Juneezee (approved):

lsg551:
thanks for the quick fix, appreciate that!

PR #2409, opened 2026-08-16
fix: report the flag name as typed in mutually exclusive flag errors

### LINKED ISSUE #2404: flag_mutex_test.go: ineffective error-string assertion masks wrong error message
Found while reviewing #2386.

**Bug in test harness (flag_mutex_test.go:107):** the assertion checking the error string uses `t.Logf` instead of `t.Errorf`/`require.Contains`:

```go
if !strings.Contains(err.Error(), test.errStr) {
    t.Logf("Invalid error string %v", err)
}
```

So a wrong error message never fails the test. This is currently masking a real discrepancy: the `"set both flags"` case passes `--i 11 --ai 12` and expects `"option i cannot be set along with option ai"`, but the actual error is `"option i cannot be set along with option t"`. `MutuallyExclusiveFlags.check` uses `f.Names()[0]`, which returns the primary name (`t`), not the alias used on the CLI (`ai`).

**Fix:** change the assertion to `t.Errorf` (or `assert.Contains(t, err.Error(), test.errStr)`) so wrong messages fail, and decide whether the message should reference the flag name as typed on the command line (alias `ai`) rather than the primary name.

**Test coverage gap (related):** the regression test added in #2386 (`--i --q`) only covers "first member of group 0 × second member of group 1". The mirror cases — second member of group 0 (`--s --q`), and the issue-reproduction layout where the conflicting flag is the second member of both groups — are untested. Recommend adding at least one such case to pin the cross-group fix.

chjnett:
Closes #2404.

The mutually exclusive flag check reported `flg.Names()[0]` — the *primary* name — so a
conflict involving a flag set via an alias referenced the primary name (`option i
cannot be set along with option t`) instead of the alias the user actually typed
(`option i cannot be set along with option ai`).

Changes:

- Track the name used to set each flag (`setFlags` is now `map[Flag]string` instead of
  `map[Flag]struct{}`), populated in `Command.set` from the parsed flag name and, for
  env-var-set flags, the primary name.
- `MutuallyExclusiveFlags.findSetFlag` reports that tracked name, falling back to
  `Names()[0]` when no name was recorded.
- Fix the test assertion that used `t.Logf`, which never fails on a wrong message and
  was masking this discrepancy — it now uses `t.Errorf`.

The existing `errStr` expectations (`option ai`, `option q`) now actually pass: the
`--i 11 --ai 12` case reports `ai` (the alias), and `--i 11 --q` reports `q`.

I was not able to run `go test` locally (no Go toolchain on this machine), so the test
is reasoned through rather than executed — please let CI judge it.

### DIFF
--- a/command.go
+++ b/command.go
@@ -147,8 +147,8 @@ type Command struct {
 	flagCategories FlagCategories
 	// flags that have been applied in current parse
 	appliedFlags []Flag
-	// flags that have been set
-	setFlags map[Flag]struct{}
+	// flags that have been set, mapped to the name used to set them
+	setFlags map[Flag]string
 	// The parent of this command. This value will be nil for the
 	// command at the root of the graph.
 	parent *Command
@@ -349,7 +349,7 @@ func (cmd *Command) Root() *Command {
 }
 
 func (cmd *Command) set(fName string, f Flag, val string) error {
-	cmd.setFlags[f] = struct{}{}
+	cmd.setFlags[f] = fName
 	cmd.setMultiValueParsingConfig(f)
 	if err := f.Set(fName, val); err != nil {
 		return fmt.Errorf("invalid value %q for flag -%s: %v", val, fName, err)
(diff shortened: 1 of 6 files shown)


chjnett:
Upstream has already resolved #2404 on `main` in the opposite direction from this PR: the assertion now uses `assert.Contains` (same intent as my `t.Errorf` change), and the `errStr` expectations were corrected to the primary flag name (`option t`) rather than changing the code to report the alias. That's a valid resolution of the issue's open question, so this PR is superseded — closing it.

PR #2414, opened 2026-08-18
Support custom FlagStringer for MutuallyExclusiveFlags

### LINKED ISSUE #2220: Mutually exclusive flags in help output

- needing to have the opportunity to change FlagStringer for mutually exclusive flags

## Solution description

For example, we have mutually exclusive flags m1 & m2, they are required. The fact that one of them (m1 or m2) is required must be displayed to the user. With simple flags we can override cli.FlagStringer func for our custom, where can add some special information if flag is required. But with MutuallyExclusiveFlags it doesn't work.
Maybe add some func to override here too? So we can add some inside logic for output this groups of flags

SHIVANSHGARG07:
## What type of PR is this?

- feature

Adds a `Stringer` field to `MutuallyExclusiveFlags` that lets callers
override how flags within the group are rendered in help output.
Previously, flags inside a mutex group could only use the package-level
`FlagStringer`, so there was no way to add group-specific context (e.g.
marking one of the flags as "required") to the help text.

- `flag.go`: added the `FlagStringerOverrider` interface
  (`SetStringer(FlagStringFunc)`) that flags implement to opt in to a
  custom stringer.
- `flag_mutex.go`: added `Stringer FlagStringFunc` field to
  `MutuallyExclusiveFlags`, plus `propagateStringer()` which pushes the
  group's stringer down to member flags (mirrors the existing
  `propagateCategory()` pattern used for `Category`).
- `flag_impl.go`: `FlagBase` now stores an optional per-flag `stringer`
  and uses it in `String()` when set; implements `SetStringer`.
- `flag_bool_with_inverse.go`: same treatment for `BoolWithInverseFlag`,
  which doesn't embed `FlagBase` and needed a separate implementation.
- `command_setup.go`: calls `grp.propagateStringer()` alongside the
  existing `grp.propagateCategory()` call during command setup.
- `godoc-current.txt`: regenerated to reflect the new public API.

Fixes #2220

Followed the existing `propagateCategory` pattern for consistency.
`BoolWithInverseFlag` needed a standalone implementation since it
doesn't embed `FlagBase`.

## Testing

Added `TestMutuallyExclusiveFlags_PropagateStringer` and
`TestMutuallyExclusiveFlags_PropagateStringerNil` in
`flag_mutex_test.go`. Ran the full suite with `go test ./...` — all
passing.

## Release Notes

```release-note
Added a `Stringer` field to `MutuallyExclusiveFlags`, allowing a custom
`FlagStringFunc` to override how flags within a mutually exclusive
group are displayed in help output.

### DIFF
--- a/command.go
+++ b/command.go
@@ -165,6 +165,10 @@ type Command struct {
 	didSetupDefaults bool
 	// whether in shell completion mode
 	shellCompletion bool
+	// whether the shell completion request came after a "--" separator,
+	// after which only positional arguments are accepted and nothing is
+	// suggested. The request is still a completion, never a command run.
+	shellCompletionPastDoubleDash bool
 	// whether global help flag was added
 	globaHelpFlagAdded bool
 	// whether global version flag was added
@@ -306,12 +310,25 @@ func (cmd *Command) VisiblePersistentFlags() []Flag {
 		return nil
 	}
 	var flags []Flag
-	for _, fl := range cmd.Root().Flags {
-		pfl, ok := fl.(LocalFlag)
-		if !ok || pfl.IsLocal() {
-			continue
+	lineage := cmd.Lineage()
+	for i := len(lineage) - 1; i > 0; i-- {
+		for _, fl := range lineage[i].allFlags() {
+			pfl, ok := fl.(LocalFlag)
+			if !ok || pfl.IsLocal() {
+				continue
+			}
+			applies := true
+			for _, name := range fl.Names() {
+				if cmd.lookupFlag(name) != fl {
+					applies = false
+					break
+				}
+			}
+			if !applies {
+				continue
+			}
+			flags = append(flags, fl)
 		}
-		flags = append(flags, fl)
 	}
 	return visibleFlags(flags)
 }
--- a/command_parse.go
+++ b/command_parse.go
@@ -114,7 +114,9 @@ func (cmd *Command) parseFlags(args Args) (Args, error) {
 				return &stringSliceArgs{posArgs}, nil
 			}
 
-			posArgs = append(posArgs, firstArg)
+			// firstArg is a trimmed copy that classifies the argument; the
+			// argument itself is what the action receives, byte for byte.
+			posArgs = append(posArgs, rargs[0])
 			continue
 		}
 
--- a/command_run.go
+++ b/command_run.go
@@ -125,6 +125,11 @@ func (cmd *Command) run(ctx context.Context, osArgs []string) (_ context.Context
 		// note that we can only do this because the shell autocomplete function
 		// always appends the completion flag at the end of the command
 		tracef("checking osArgs %v (cmd=%[2]q)", osArgs, cmd.Name)
+		// completion request state is per-run: a Command answering several
+		// requests (tests, REPL, embedded use) must not carry one request
+		// into the next
+		cmd.shellCompletion = false
+		cmd.shellCompletionPastDoubleDash = false
 		cmd.shellCompletion, osArgs = checkShellCompleteFlag(cmd, osArgs)
 
 		tracef("setting cmd.shellCompletion=%[1]v from checkShellCompleteFlag (cmd=%[2]q)", cmd.shellCompletion && cmd.EnableShellCompletion, cmd.Name)
--- a/command_setup.go
+++ b/command_setup.go
@@ -153,6 +153,7 @@ func (cmd *Command) setupDefaults(osArgs []string) {
 	tracef("setting category on mutually exclusive flags (cmd=%[1]q)", cmd.Name)
 	for _, grp := range cmd.MutuallyExclusiveFlags {
 		grp.propagateCategory()
+		grp.propagateStringer()
 	}
 
 	tracef("setting flag categories (cmd=%[1]q)", cmd.Name)
@@ -196,6 +197,7 @@ func (cmd *Command) setupSubcommand() {
 	tracef("setting category on mutually exclusive flags (cmd=%[1]q)", cmd.Name)
 	for _, grp := range cmd.MutuallyExclusiveFlags {
 		grp.propagateCategory()
+		grp.propagateStringer()
 	}
 
 	tracef("setting flag categories (cmd=%[1]q)", cmd.Name)
--- a/command_test.go
+++ b/command_test.go
@@ -1156,6 +1156,22 @@ func TestCommand_CommandWithDash(t *testing.T) {
 	require.Equal(t, "-", args.Get(1))
 }
 
+func TestCommand_PositionalArgsKeepWhitespace(t *testing.T) {
+	var args Args
+
+	cmd := &Command{
+		Name: "prog",
+		Action: func(_ context.Context, cmd *Command) error {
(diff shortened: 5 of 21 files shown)

### CODE REVIEW

**Review by dearchap — COMMENTED** [review]
Overall: Clean implementation following existing patterns. A few suggestions below.

**Comment by dearchap** [comment]
### Suggestion 1: Rename `FlagStringerOverrider` to `StringerSetter`

The name `FlagStringerOverrider` is a tongue-twister. Consider `StringerSetter` — it is shorter, more idiomatic, and still clearly conveys intent. This would affect:

- `flag.go`: interface definition
- `flag_mutex.go`: type assertion in `propagateStringer()`
- `flag_mutex_test.go`: no direct reference but indirectly
- `godoc-current.txt` / `godoc-v3.x.txt` (after regeneration)

```go
// Before
type FlagStringerOverrider interface {
    SetStringer(FlagStringFunc)
}

// After
type StringerSetter interface {
    SetStringer(FlagStringFunc)
}
```

This is a public API addition, so if you agree, now is the time to rename before merge.

dearchap:
### Suggestion 2: Add test for non-implementing flags

The `propagateStringer()` loop uses a type assertion to silently skip flags that do not implement the interface. This is correct, but there is no test that verifies this behavior. Worth adding a test to lock in the skip behavior:

```go
func TestMutuallyExclusiveFlags_PropagateStringerSkipsNonImplementor(t *testing.T) {
    customStringer := func(f Flag) string {
        return "custom:" + f.Names()[0]
    }

    // plainStringFlag is a minimal Flag impl that does NOT implement StringerSetter
    type plainStringFlag struct {
        StringFlag
    }

    grp := MutuallyExclusiveFlags{
        Stringer: customStringer,
        Flags: [][]Flag{
            {
                &plainStringFlag{StringFlag{Name: "plain"}},
            },
        },
    }

    // should not panic
    grp.propagateStringer()

    // stringer should NOT have been set — verify default behavior
    assert.NotEmpty(t, grp.Flags[0][0].String())
}
```

This ensures that the interface assertion skip path is covered and will not regress.

dearchap:
### Suggestion 3: Verify `make lint` and `make vet` pass

The PR mentions `go test ./...` passes, but does not mention linting. Per the repo CI workflow, `make lint` (goimports check) and `make vet` also run. Please run these locally before merge:

```bash
make lint
make vet
```

This avoids surprises in CI, especially since the godoc files and source files were all modified.

SHIVANSHGARG07:
Thanks for the review! Addressed all three suggestions in the latest commit:

1. Renamed FlagStringerOverrider → StringerSetter across flag.go, flag_mutex.go, and regenerated godoc-current.txt / testdata/godoc-v3.x.txt. Since this API hasn't been released yet, the rename is safe.
2. Added TestMutuallyExclusiveFlags_PropagateStringerSkipsNonImplementor. Note: I used a standalone minimal Flag implementation instead of embedding StringFlag/FlagBase, since FlagBase now implements StringerSetter itself (that's the whole point of this PR), so embedding it would have made the test pass for the wrong reason.
3. Re-ran make lint and make vet locally — both clean, along with make test, make generate, and make v3diff.

Also synced the branch with the latest main since it was flagged out-of-date.

SHIVANSHGARG07:
Hi @dearchap Can you please look into this !!

dearchap:
## Review

Approach is sound and follows the existing `propagateCategory` pattern. Signature-wise: `go vet`, `go build`, full test suite (`go test ./...`), `make lint` (goimports), and `make v3diff` all pass locally. The `json:"-"` tag keeps the `Command` JSON marshal tests green. A few concerns below, none blocking.

### 1. `BoolWithInverseFlag` only honors the stringer partially
`BoolWithInverseFlag.String()` (flag_bool_with_inverse.go) always discards the `names` segment of the stringer output and recomputes it from `bif.Name`/`Aliases`; only the details after the first tab are kept. So a group stringer cannot control how a `BoolWithInverseFlag` renders its names, unlike every `FlagBase`-based flag. Consider:
- documenting this asymmetry on `BoolWithInverseFlag.SetStringer` (the current doc says it overrides the `FlagStringFunc` used by `String`), or
- making the behavior consistent.

The test currently papers over it with `assert.Contains` (flag_mutex_test.go) rather than asserting exact output.

### 2. Test coverage gap — no end-to-end test
The three new tests call `grp.propagateStringer()` directly. Nothing verifies the stringer actually reaches rendered `--help` output through `Run` → setup → template (`$e.String`). That path also goes through `flagCategories.AddFlag` (category.go), which keys flags by `String()`. Recommend an integration test in `help_test.go` (e.g. next to the existing `TestMutuallyExclusiveFlags`) that runs a command with a group `Stringer` and asserts the help output.

### 3. Category-map collision footgun
`flagCategories.AddFlag` uses `fl.String()` as the map key for help flags. If a group `Stringer` returns identical text for multiple `FlagBase` flags in the same category (e.g. `"(required: pick one)"` for every member), the later flag silently clobbers the earlier one in help output. The default `stringifyFlag` is unique per flag because it embeds the name. Suggest documenting a uniqueness requirement on `MutuallyExclusiveFlags.Stringer`.

### Nits
- `TestMutuallyExclusiveFlags_PropagateStringerNil` only covers the early-return when the group has no stringer; it doesn't test nil *resetting* (stringer set, then `SetStringer(nil)` restoring default behavior).
- `didSetupDefaults`/`setupSubcommand` short-circuit means mutating `Stringer` between runs of a reused `Command` won't re-propagate — same behavior as `Category`, so not new, but worth a one-line doc note.
- No docs update in `docs/v3/examples/flags/advanced.md` (which already documents mutex groups) for the new field.

SHIVANSHGARG07:
Thanks for the detailed review! Addressed all the points in the latest commit:

1. BoolWithInverseFlag stringer asymmetry — documented this on SetStringer's doc
   comment: the names segment (before the first tab) is always recomputed from
   Name/Aliases/InversePrefix and can't be overridden, only the details after it
   come from the stringer. Kept the assert.Contains in the existing test since
   the exact-match behavior didn't change, just documented it.

2. Test coverage gap — added TestMutuallyExclusiveFlags_StringerInHelpOutput in
   help_test.go. It runs a real Command through Run(..., "--help") with a group
   Stringer set and asserts the custom output appears in the rendered help text,
   covering the full Run → setup → template path (including flagCategories.AddFlag).

3. Category-map collision footgun — documented the uniqueness requirement
   directly on MutuallyExclusiveFlags.Stringer's doc comment, explaining that
   help rendering keys flags by String() output and identical strings will
   silently clobber each other.

Nits:
- Added TestMutuallyExclusiveFlags_PropagateStringerNilResetsToDefault, covering
  SetStringer(nil) restoring default FlagStringer-based output.
- Added a doc note on propagateStringer about the didSetupDefaults short-circuit
  (mirrors the existing Category behavior/limitation).
- Added a short section to docs/v3/examples/flags/advanced.md documenting the
  new Stringer field.

Re-ran go build, go vet, go test ./..., make lint, make generate, and
make v3diff (via make v3approve to promote the godoc changes) locally — all
clean.

dearchap (approved):

PR #2419, opened 2026-08-26
Keep arguments after a bare dash

### LINKED ISSUE #2418: A bare "-" argument silently discards all following arguments in parseFlags
## Summary

In `parseFlags` (v3 `main`), the branch handling a lone `-` argument appends it to `posArgs` and then executes a plain `break`, which exits the entire rearrangement loop. Every remaining command-line argument after the `-` is dropped without error: `myapp - foo bar` ends up with `cmd.Args()` == `["-"]`.

Static-analysis finding based on reading `main`; not executed here.

## Location

- File: `command_parse.go`
- Function: `(*Command).parseFlags`
- Relevant code:

```go
	numMinuses := 1
	// this is same as firstArg == "-"
	if len(firstArg) == 1 {
		posArgs = append(posArgs, firstArg)
		break
	}
```

## Problem

Compare this branch with every sibling path in the same loop:

- empty argument ? `append(posArgs, rargs[0])` + `continue`
- `"--"` ? appends the arg (if completion mode) plus **all** of `rargs[1:]`, then returns
- positional (`firstArg[0] != '-'`) ? `append(posArgs, firstArg)` + `continue`; if it names a subcommand, `append(posArgs, rargs...)` and return
- non-letter after `-` (e.g. `-3`) ? `append(posArgs, rargs...)` + return

The lone `-` is the only hyphen-path that stops parsing *without preserving the unconsumed remainder*. Since the loop's post-statement only advances `rargs` between iterations, `break` here throws away `rargs[1:]` entirely.

## Trigger / Reproduction

Based on source reading:

```console
$ myapp - foo bar
```

Iteration 1: `firstArg == "-"` ? `len(firstArg) == 1` matches ? `posArgs = ["-"]` ? `break`.
Result: `cmd.Args().Len() == 1`, containing only `-`. `foo` and `bar` never reach the action.

Expected: the arguments are retained - either `["-", "foo", "bar"]` (matching Go's stdlib `flag`, which treats a bare `-` like an ordinary operand) or at minimum everything from `-` onward, mirroring the `--` and non-letter branches.

## Expected Behavior

A single hyphen is conventionally used as an operand (stdin placeholder) or a terminator; in both interpretations the arguments that follow must survive into `cmd.Args()`.

## Actual Behavior

They are silently discarded; no error is returned.

## Impact

Any CLI invocation that passes `-` (very common for "read stdin") followed by further operands loses those operands. Because nothing errors, callers observe truncated `Args()` with no diagnostic, which is difficult to debug downstream.

## Suggested Direction

Mirror the neighboring branches: replace `break` with appending the rest of `rargs` and returning (terminator semantics), or `continue` after appending just `firstArg` (operand semantics). Either preserves user data; which one matches intent should be confirmed against the v2 behavior.

## Evidence

- Loop structure at the top of `parseFlags` shows no other mechanism that would revisit discarded args.
- All four alternative branches in the same function explicitly preserve the remainder, demonstrating that dropping it here is anomalous rather than intended.

official-burak:
## What type of PR is this?

- bug

`parseFlags` treated a lone `-` as a positional, then `break` out of the rearrangement loop. Everything after `-` was dropped with no error.

Neighboring paths use `continue` (empty arg, other positionals) or keep the remainder (`--`). This matches Go's `flag` package: `-` is an operand, parsing continues.

Fixes #2418

## Testing

`TestCommand_BareDashKeepsFollowingArgs` failed with `Args() == ["-"]` before the change and passes after. Full `go test -count=1` passes.

## Release Notes

```release-note
Keep arguments that follow a bare `-` instead of silently discarding them.
```

### DIFF
--- a/command_parse.go
+++ b/command_parse.go
@@ -122,7 +122,7 @@ func (cmd *Command) parseFlags(args Args) (Args, error) {
 		// this is same as firstArg == "-"
 		if len(firstArg) == 1 {
 			posArgs = append(posArgs, firstArg)
-			break
+			continue
 		}
 
 		shortOptionHandling := cmd.useShortOptionHandling()
--- a/command_test.go
+++ b/command_test.go
@@ -1156,6 +1156,26 @@ func TestCommand_CommandWithDash(t *testing.T) {
 	require.Equal(t, "-", args.Get(1))
 }
 
+func TestCommand_BareDashKeepsFollowingArgs(t *testing.T) {
+	var args Args
+
+	cmd := &Command{
+		Commands: []*Command{
+			{
+				Name: "cmd",
+				Action: func(_ context.Context, cmd *Command) error {
+					args = cmd.Args()
+					return nil
+				},
+			},
+		},
+	}
+
+	require.NoError(t, cmd.Run(buildTestContext(t), []string{"", "cmd", "-", "foo", "bar"}))
+	require.NotNil(t, args)
+	require.Equal(t, []string{"-", "foo", "bar"}, args.Slice())
+}
+
 func TestCommand_CommandWithNoFlagBeforeTerminator(t *testing.T) {
 	var args Args
 
--- a/command_test.go
+++ b/command_test.go
@@ -1176,6 +1176,56 @@ func TestCommand_BareDashKeepsFollowingArgs(t *testing.T) {
 	require.Equal(t, []string{"-", "foo", "bar"}, args.Slice())
 }
 
+func TestCommand_BareDashThenFlags(t *testing.T) {
+	cases := []struct {
+		name   string
+		argv   []string
+		option string
+		args   []string
+	}{
+		{
+			name:   "flags after a bare dash still parse",
+			argv:   []string{"", "cmd", "-", "--option", "my-option", "foo"},
+			option: "my-option",
+			args:   []string{"-", "foo"},
+		},
+		{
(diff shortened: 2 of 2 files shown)


dearchap:
Consider adding a test case for  followed by flags (e.g., ) to ensure the  behavior interacts correctly with the existing  terminator handling.

official-burak:
Good catch. `-` is an operand so a flag after it still parses; `--` is what stops the rest. I added `TestCommand_BareDashThenFlags` for both `- --option` and `- -- --option`.

thaJeztah:
Looks containerd was impacted by this as well after updating to v3; https://github.com/containerd/containerd/pull/14217

We're applying a workaround, but looking forward to see this land 😅

command_parse.go
```diff
@@ -122,10 +122,9 @@ func (cmd *Command) parseFlags(args Args) (Args, error) {
 		}
 
 		numMinuses := 1
-		// this is same as firstArg == "-"
-		if len(firstArg) == 1 {
-			posArgs = append(posArgs, firstArg)
-			break
+		if firstArg == "-" {
+			posArgs = append(posArgs, rargs[0])
```
Juneezee: The purpose of my commit df51685d1fee17fb948aab42a816eb624b9a0fe0 is to apply the fix submmited in PR https://github.com/urfave/cli/pull/2423

Juneezee (approved):

PR #2422, opened 2026-08-29
fix: show inherited persistent flags in command help

### LINKED ISSUE #2420: Persistent flags defined on non-root command, do not show up in help output of leaf command
## My urfave/cli version is

`v3.10.0` (also verified in `v3.11.0`)

## Checklist

## Dependency Management

## Describe the bug

When a flag is defined as persistent (`Local == false`), this flag is not displayed in the help output of a child command that inherits it. Only the persistent flags defined on root command are displayed under `GLOBAL OPTIONS:`. Note that the flag gets parsed as expected, just isn't displayed in the help output.

## To reproduce

```go
package main

import (
	"context"
	"fmt"
	"os"

	"github.com/urfave/cli/v3"
)

func main() {
	cmd := &cli.Command{
		Name: "root",
		Flags: []cli.Flag{
			&cli.StringFlag{Name: "root-persistent", Usage: "persistent flag on root", Local: false},
		},
		Commands: []*cli.Command{
			{
				Name: "mid",
				Flags: []cli.Flag{
					&cli.StringFlag{Name: "mid-persistent", Usage: "persistent flag on mid", Local: false},
					&cli.StringFlag{Name: "mid-local", Usage: "local flag on mid", Local: true},
				},
				Commands: []*cli.Command{
					{
						Name:  "leaf",
						Flags: []cli.Flag{&cli.StringFlag{Name: "leaf-local", Usage: "local flag on leaf"}},
						Action: func(_ context.Context, c *cli.Command) error {
							fmt.Println("leaf action; mid-persistent =", c.String("mid-persistent"))
							return nil
						},
					},
				},
			},
		},
	}

	fmt.Println("=========== root mid leaf --help ===========")
	_ = cmd.Run(context.Background(), []string{"root", "mid", "leaf", "--help"})
	fmt.Println("=========== root mid --help ===========")
	_ = cmd.Run(context.Background(), []string{"root", "mid", "--help"})
	fmt.Println("=========== parse check: root mid leaf --mid-persistent=X ===========")
	if err := cmd.Run(context.Background(), []string{"root", "mid", "leaf", "--mid-persistent", "X"}); err != nil {
		fmt.Println("ERR:", err)
		os.Exit(1)
	}
}
```

## Observed behavior

Persistent flags defined on non-root commands are absent in the help of a child command inheriting said flag.

Example output where `--mid-persistent` flag is declared on `mid` subcommand:

```
NAME:
   root mid leaf

USAGE:
   root mid leaf [options]

OPTIONS:
   --leaf-local string  local flag on leaf
   --help, -h           show help

GLOBAL OPTIONS:
   --root-persistent string  persistent flag on root
```

## Expected behavior

All non-hidden inherited flags are displayed in help output of the commands inheriting them.

## Additional context

I assume this is a bug, not intended behavior.

The root cause seems to be in the `VisiblePersistentFlags()` method, which only loops over `cmd.root.Flags()` and [does not traverse the whole command chain](https://github.com/urfave/cli/blob/0ed457cf64de75990ce1779f0b9632796f1a4962/command.go#L309-L315).

This is in contrast to `ParseFlags()` method which [walks the full command lineage](https://github.com/urfave/cli/blob/0ed457cf64de75990ce1779f0b9632796f1a4962/command_parse.go#L33).

## Want to fix this yourself?

I'd be happy to make a contribution. The obvious solution would be to replicate similar loop as that in `ParseFlags()`. However, that would populate the persistent interim-command flags under the existing "GLOBAL OPTIONS:" header, which might not be entirely accurate for flags not defined on root command.

I think it would be more accurate to rename the heading to "INHERITED OPTIONS:", but would like confirmation whether this would be acceptable, and whether this is truly a bug and not intended behavior.

amtbsl:
Fixes #2420.

Include persistent flags inherited from parent commands when rendering subcommand help, while avoiding duplicates. Adds regression coverage for inherited flag visibility.

Validation: regression test added and `git diff --check` passed. Go tests were unavailable locally because the Go toolchain is not installed.

### DIFF
--- a/command.go
+++ b/command.go
@@ -300,18 +300,21 @@ func (cmd *Command) appendFlag(fl Flag) {
 	}
 }
 
-// VisiblePersistentFlags returns a slice of [LocalFlag] with Persistent=true and Hidden=false.
+// VisiblePersistentFlags returns inherited persistent flags with Hidden=false.
 func (cmd *Command) VisiblePersistentFlags() []Flag {
 	if cmd.isCompletionCommand {
 		return nil
 	}
 	var flags []Flag
-	for _, fl := range cmd.Root().Flags {
-		pfl, ok := fl.(LocalFlag)
-		if !ok || pfl.IsLocal() {
-			continue
+	lineage := cmd.Lineage()
+	for i := len(lineage) - 1; i > 0; i-- {
+		for _, fl := range lineage[i].allFlags() {
+			pfl, ok := fl.(LocalFlag)
+			if !ok || pfl.IsLocal() {
+				continue
+			}
+			flags = append(flags, fl)
 		}
-		flags = append(flags, fl)
 	}
 	return visibleFlags(flags)
 }
--- a/help_test.go
+++ b/help_test.go
@@ -858,6 +858,52 @@ GLOBAL OPTIONS:
 	assert.Contains(t, output.String(), expected, "expected output to include global options")
 }
 
+func TestShowNestedSubcommandHelp_InheritedPersistentFlags(t *testing.T) {
+	cmd := &Command{
+		Flags: []Flag{
+			&StringFlag{
+				Name:  "root-persistent",
+				Local: false,
+			},
+		},
+		Commands: []*Command{
+			{
+				Name: "mid",
+				Flags: []Flag{
+					&StringFlag{
+						Name:  "mid-persistent",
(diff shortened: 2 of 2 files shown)


dearchap:
@amtbsl Please fix the PR checks. Thanks

Issue #2169, opened 2025-07-12
TrimSpace Config has no effect on StringSliceFlag


ayoisaiah:
## My urfave/cli version is

v3.3.8

## Checklist

- [x] Are you running the latest v3 release? The list of releases is [here](https://github.com/urfave/cli/releases).
- [x] Did you check the manual for your release? The v3 manual is [here](https://cli.urfave.org/v3/getting-started/)
- [x] Did you perform a search about this problem? Here's the [GitHub guide](https://help.github.com/en/github/managing-your-work-on-github/using-search-to-filter-issues-and-pull-requests) about searching.

## Describe the bug

Strings arguments to a StringSliceFlag are always trimmed regardless of the `TrimSpace` configuration

## To reproduce

```go
// See flag_test.go

// Existing test
		{
			name:   "StringSliceFlag valid with TrimSpace",
			input:  "foo , bar ",
			output: []string{"foo", "bar"},
			fl:     &StringSliceFlag{Name: "names", Sources: EnvVars("NAMES"), Config: StringConfig{TrimSpace: true}},
		},

// Add a new test with TrimSpace set to false
		{
			name:   "StringSliceFlag valid without TrimSpace",
			input:  "foo , bar ",
			output: []string{"foo ", " bar "},
			fl:     &StringSliceFlag{Name: "names", Sources: EnvVars("NAMES"), Config: StringConfig{TrimSpace: false}},
		},
```

## Observed behavior

Spaces are trimmed regardless:

```go
[]string{"foo", "bar"}
```

## Expected behavior

I expected it to preserve spaces with `TrimSpace` set to false

```go
 []string{"foo ", " bar "}
```

PR #2425, opened 2026-09-07
fix: preserve whitespace in equals-form flag values

BenYang12:
## What type of PR is this?
  bug

  `parseFlags` trimmed the entire argument before extracting the value after `=`. Consequently, `--text=" hello "` became `" hello"` even with `TrimSpace: false`, while `--text
  " hello "` preserved the value.

 Extract the value from the original argument, keeping flag-name classification unchanged. Each flag's value parser now controls trimming and validation.

  Related to #2423, which inspired this investigation and fixed positional whitespace preservation, and #2169, which addressed string-slice trimming.

  ## Testing
The regression tests fail in 25 cases with the original parser and all 60 cases pass with the fix. Coverage includes string, slice, and map flags; aliases; empty values; Unicode whitespace; embedded equals; and numeric/boolean validation.

  ## Release Notes
 Preserve whitespace in equals-form flag values, respecting each flag's trimming and validation rules. Numeric and boolean values with trailing whitespace now return parse
 errors instead of being silently trimmed.

### DIFF
--- a/command_parse.go
+++ b/command_parse.go
@@ -147,7 +147,9 @@ func (cmd *Command) parseFlags(args Args) (Args, error) {
 		valFromEqual := false
 		tracef("flagName:1 (fName=%[1]q)", flagName)
 		if index := strings.Index(flagName, "="); index != -1 {
-			flagVal = flagName[index+1:]
+			// Classify the flag using the trimmed token, but leave value
+			// whitespace handling to the flag's value parser.
+			_, flagVal, _ = strings.Cut(rargs[0], "=")
 			flagName = flagName[:index]
 			valFromEqual = true
 		}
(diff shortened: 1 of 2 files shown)


dearchap (approved):
LGTM. Verified manually:

- Change is minimal and correct: value extracted from raw token via `strings.Cut(rargs[0], "=")` while flag-name classification stays on the trimmed token. Raw prefix before the flag's `=` can only contain whitespace/`-` (never `=`), so the first `=` always maps correctly (embedded equals in values unaffected).
- Makes equals-form consistent with the separate-arg form, which already took the value raw.
- Focused behavior change for `--int=42 ` / `--bool= ` (parse errors instead of silent trim) is consistent with the separate-arg form and documented in release notes.
- Ran go build/vet/goimports (clean); new tests + full suite pass on PR head; re-applying tests onto base reproduces the ~20 equals-form whitespace failures, confirming they're meaningful regression tests.
- No API surface change; the docs/v3diff gate is unaffected.

Non-blocking nits: comment could note that `firstArg` is already trimmed (why raw arg is used); subtest name `whitespace=%q` prints `_` for a space.

PR #2426, opened 2026-09-08
fix: complete subcommand flags after positional arguments

### LINKED ISSUE #2248: Auto-completion with unfinished partial flags (`--XXX<TAB>` returns no completions
## My urfave/cli version is

v3.6.1

## Checklist

## Dependency Management

## Describe the bug

When passing in uncompleted flags (`--XXX<TAB>`), no auto-completion is available. 

## To reproduce

In a project with autocompletions enabled with command and arguments run:
```
$ command <TAB>
subcommand
$ command subcommand -<TAB>
--par1
--par2
--flag
$command subcommand --pa<TAB>
* empty
```
Underlying I see respectively begin passed to the binary:
- `--generate-shell-completion`, returns list of commands
- `- --generate-shell-completion` returns list of flags
- `--pa --generate-shell-completion`, but this one is empty

## Observed behavior

No autocompletions given

## Expected behavior

Correctly complete autocompletion of partial flags

## Additional context

I worked around this by using a bash wrapper around the `command` binary, which rewrites calls in this form `command --XXX  --generate-shell-completion` to `command -  --generate-shell-completion`, which seems to work for bash/zsh.
Even though the command returns all possible flags, it's completion suggestions are filtered to relevant completions only.

## Want to fix this yourself?

pull request for it.

```
go version go1.25.3 darwin/arm64
```

```
Rather not, but available upon requests. Very simple default goenv setup. Go version 1.24.4
```

fzlzjerry:
## What type of PR is this?

- bug

Fix the remaining partial-flag completion case after a positional argument in a child command. For example, current main completes `app sub --pa`, but `app sub value --pa` suggests child commands instead of matching flags.

`DefaultCompleteWithFlags` reads raw `os.Args` for the root command and parsed arguments for a child. Only the raw arguments still contain the trailing completion marker, so always selecting the penultimate argument skips the flag being completed in the child case.

- `help.go`: start with the final argument and step back only when that argument is the completion marker.
- `completion_partial_regression_test.go`: cover root, child, and nested commands with zero, one, and two positional arguments, plus hidden flags, custom-completer arguments, disabled completion, and requests after `--`.

The completion request format, flag parsing, and custom completion handlers are unchanged.

Fixes #2248.

## Testing

- The completed regression suite fails for the four child/nested positional cases on unchanged `5081218` and passes with this change. The already-working root and no-positional cases stay unchanged.
- Go 1.27.0: `make all`, `make lint`, `make v3diff`, and post-commit `make diffcheck` pass.
- Go 1.26.5: `make vet test check-binary-size` passes.
- Built the same CLI probe against the baseline and this branch and compared nine completion inputs. Matching flags replace the incorrect command suggestions; both binaries preserve ordinary action execution and produce no output/action for completion after `--`.

The README `gfmrun` target found zero executable blocks; the compiled CLI probe supplies the execution check above. macOS and Windows are left to CI.

## Release Notes

```release-note
Fix flag-name completion after positional arguments in subcommands and nested commands.
```

### DIFF
--- a/completion_partial_regression_test.go
+++ b/completion_partial_regression_test.go
@@ -1,0 +1,121 @@
+package cli
+
+import (
+	"bytes"
+	"context"
+	"os"
+	"reflect"
+	"strings"
+	"testing"
+)
+
+func partialCompletionCommand(output *bytes.Buffer, ran *bool) *Command {
+	command := func(name string) *Command {
+		return &Command{
+			Name: name,
+			Flags: []Flag{
+				&StringFlag{Name: "param-" + name + "-one"},
+				&StringFlag{Name: "param-" + name + "-two"},
+				&StringFlag{Name: "param-" + name + "-hidden", Hidden: true},
+				&BoolFlag{Name: "unrelated-" + name},
+			},
+			Action: func(context.Context, *Command) error { *ran = true; return nil },
+		}
+	}
+	root, child, nested := command("app"), command("sub"), command("nested")
+	root.Writer, root.ErrWriter = output, output
+	root.EnableShellCompletion = true
+	root.Commands = []*Command{child}
+	child.Commands = []*Command{nested}
+	return root
+}
+
+func TestPartialFlagCompletionAfterPositionalArgument(t *testing.T) {
+	originalArgs := os.Args
+	t.Cleanup(func() { os.Args = originalArgs })
+	for _, scope := range []struct {
+		name string
+		path []string
+	}{
+		{"app", nil},
+		{"sub", []string{"sub"}},
+		{"nested", []string{"sub", "nested"}},
+	} {
+		for _, positional := range [][]string{nil, {"value"}, {"first", "second"}} {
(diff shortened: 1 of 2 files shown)


dearchap (approved):
## LGTM

Approve. Small, targeted fix for the remaining partial-flag completion case after a positional argument in child/nested commands.

### Verified locally
- Confirmed the bug on base `50812189`: new tests fail for all four child/nested positional cases (output suggests commands/help instead of flags).
- Fix passes: `TestPartialFlagCompletion*` all green on head `816e14f1`.
- No regressions: full `go test -count=1 .`, `go vet ./...`, `goimports -l` all clean.

### Minor non-blocking notes
- `TestPartialFlagCustomCompletionArgs` asserts current behavior where `cmd.Args()` holds the partial `--pa` as positional — good to keep explicit.
- No issues found with root-vs-child args asymmetry handling.

fzlzjerry:
Thanks, @dearchap, for the detailed review and the base/head regression check. I’ll keep the fix and the explicit custom-completion argument test as reviewed.

PR #2427, opened 2026-09-08
Fix a year-less timestamp layout losing January 1

youdie006:
## What type of PR is this?

- bug

A `TimestampFlag` whose layout carries a month and day but no year returns **today's date** instead of the parsed date, but only when the value is January 1.

```
Config: TimestampConfig{Layouts: []string{"01-02"}, Timezone: time.UTC}

--time 01-01   before: 2026-09-08T00:00:00Z   after: 2026-01-01T00:00:00Z
--time 01-02   2026-01-02T00:00:00Z  (both)
--time 03-04   2026-03-04T00:00:00Z  (both)
--time 12-31   2026-12-31T00:00:00Z  (both)
```

`timestampValue.Set` decides "this layout has no date" by comparing the parsed instant against `defaultTS`, a probe parsed from `time.TimeOnly`, which lands on year-0 January 1 (`flag_timestamp.go:90-93`). A layout that *does* carry a date parses `01-01` onto that exact same day, so the two cases are indistinguishable and the date-less branch fires, overwriting month and day with `time.Now()`.

The fix asks the layout rather than the parsed value. `layoutHasDate` round-trips a reference date that is deliberately not January 1 (`0002-03-04`) through `Format`/`Parse`: a date-less layout drops it and comes back as January 1, a layout with a date returns March 4.

- `flag_timestamp.go` — capture the layout that matched in the parse loop, add the `layoutHasDate` guard to the first branch, add the helper.
- `flag_test.go` — `TestTimestampFlagApply_YearlessLayoutJanuaryFirst`, three rows.

**The one-sentence version:** the very next branch (`flag_timestamp.go:104`, `else if timestamp.Year() == 0`) already handles "layout has a date but no year" correctly — January 1 is the single input that never reaches it.

`matchedLayout` is load-bearing: the loop otherwise discards the layout and the check cannot be made without it. The helper is unexported, so `make generate` produces no change to `godoc-current.txt`.

None — found by reading `flag_timestamp.go`. I did not open one first since #2400 and #2401 landed the same way, but happy to file one if you would prefer.

`time.Stamp`-style layouts (`"Jan _2 15:04:05"`, i.e. syslog and journal timestamps) are the canonical reason to use a year-less layout, and they are affected too — that row is in the test.

One honest limitation: `flag_timestamp.go:92` calls `time.Now()` directly, so on January 1 itself the buggy code returns the right answer by coincidence and the two January rows would pass against the bug. That day is unobservable by construction, not a gap in the test. The March 4 control row and the mutation checks below still discriminate on every other day.

I also considered detecting date components by scanning the layout string for `1`/`01`/`Jan`/`2`/`02`/`_2` tokens, and rejected it — the round-trip probe is shorter and does not have to enumerate Go's layout vocabulary. If a layout is lossy enough that `Format`/`Parse` does not round-trip, the `err == nil` guard makes `layoutHasDate` return false, which is exactly today's behaviour, so an exotic layout degrades to the status quo rather than regressing.

## Testing

`TestTimestampFlagApply_YearlessLayoutJanuaryFirst` drives the production chain — `cmd.Run` → `Action` → `cmd.Timestamp("time")` — with two January 1 rows and a March 4 control.

- On `main` the two January rows fail (`expected: 2026-01-01`, `actual: 2026-09-08`) and the control passes.
- Mutation-checked both directions, failing disjoint tests: forcing `layoutHasDate` to `false` (restoring current behaviour) fails only the new test, and forcing it to `true` fails only the pre-existing `TestTimestampFlagApply_ShortenedLayouts`. Both sides of the boundary are pinned.
- `make test` passes (`-race`, coverage 99.3%). `make vet`, `make lint`, `make ensure-goimports`, `make generate` + `make diffcheck`, `make v3diff`, `make check-binary-size` (1.8MB) all clean, and `golangci-lint run ./...` reports 0 issues.
- Existing timestamp coverage still passes, including `time.Kitchen`, `time.Stamp`, `StampMilli/Micro/Nano`, `TimeOnly`, `"15:04"` and `time.RFC3339`.

## Release Notes

```release-note
Fixed `TimestampFlag` returning the current date instead of the parsed date when the layout has no year and the value is January 1.
```

---

AI assistance disclosure: this patch was found and written with Claude Code. Every number above is verbatim from running it against `main` and against this branch.

### DIFF
--- a/flag_test.go
+++ b/flag_test.go
@@ -2718,6 +2718,36 @@ func TestTimestampFlagApply_ShortenedLayouts(t *testing.T) {
 	}
 }
 
+func TestTimestampFlagApply_YearlessLayoutJanuaryFirst(t *testing.T) {
+	year := time.Now().UTC().Year()
+
+	for _, tc := range []struct {
+		layout   string
+		value    string
+		expected time.Time
+	}{
+		{layout: "01-02", value: "01-01", expected: time.Date(year, time.January, 1, 0, 0, 0, 0, time.UTC)},
+		{layout: "01-02", value: "03-04", expected: time.Date(year, time.March, 4, 0, 0, 0, 0, time.UTC)},
+		{layout: "Jan _2 15:04:05", value: "Jan  1 06:07:08", expected: time.Date(year, time.January, 1, 6, 7, 8, 0, time.UTC)},
+	} {
+		t.Run(tc.layout+" "+tc.value, func(t *testing.T) {
+			var got time.Time
+			cmd := &Command{
+				Flags: []Flag{
+					&TimestampFlag{Name: "time", Config: TimestampConfig{Layouts: []string{tc.layout}}},
+				},
+				Action: func(_ context.Context, cmd *Command) error {
+					got = cmd.Timestamp("time")
+					return nil
+				},
+			}
+
+			assert.NoError(t, cmd.Run(buildTestContext(t), []string{"", "--time", tc.value}))
+			assert.Equal(t, tc.expected, got)
+		})
+	}
+}
+
 func TestTimestampFlagApplyValue(t *testing.T) {
 	expectedResult, _ := time.Parse(time.RFC3339, "2006-01-02T15:04:05Z")
 	fl := TimestampFlag{Name: "time", Aliases: []string{"t"}, Config: TimestampConfig{Layouts: []string{time.RFC3339}}, Value: expectedResult}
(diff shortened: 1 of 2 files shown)


dearchap (approved):
Approving. Verified on the PR branch: the new test fails on `main` (bug reproduced) and passes with the fix; full `go test -race .`, `go vet`, gofmt, and `make v3diff` are clean. The round-trip probe (`layoutHasDate`) is exact and only changes behavior in the formerly buggy case — edge probing (year-only, year-month, day-of-year, time-only, multi-layout layouts) shows no regressions. Minor non-blocking notes: Hoist `layoutHasDate` out of the per-`Set` path if you want to avoid the extra Format/Parse on each parse; the test's `time.Now().UTC().Year()` has a narrow NYE UTC rollover race, consistent with existing tests.

PR #2428, opened 2026-09-12
docs: add go-size-analyzer, module version, and reflection

imsumedhaa:
## What type of PR is this?
- Documentation

Extends `docs/v3/binary-size.md` with the remaining topics requested in
#2365 that weren't covered by #2366:

- Added a section on using `go-size-analyzer` to get a package-level
  breakdown of binary size (install steps + example output).
- Added a section on checking selected dependency versions across
  multi-module repos with `go list -m all` / `go list -m -json`, and
  clarified that differing `go.mod` files don't automatically cause
  duplicate module versions in a build.
- Added a section explaining why `reflect` and `text/template` show up as
  significant contributors to binary size in `urfave/cli/v3` apps (help
  rendering uses templates internally).
- Added a section testing whether `HideHelp` / `HideHelpCommand` reduce
  binary size. They don't — verified with two otherwise identical builds
  producing the same size and identical `reflect`/`text/template` symbol
  footprint via `go-size-analyzer`. Documented this as a known limitation
  since these are runtime, not compile-time, options.

#2365

This is my first contribution to this project. I'd appreciate feedback on
whether the level of detail/tone fits the rest of the docs, and whether the
"Hiding Built-in Help" finding (that it doesn't reduce size) is presented
clearly — I wanted to document it as a tested result rather than a guess.

## Testing

All example output and size numbers were captured from real local builds
of a minimal `urfave/cli/v3` test app (one command, one string flag),
using `go build`, `go tool nm -size`, and `go-size-analyzer`/`gsa`, rather
than estimated or copied from elsewhere.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/docs/v3/binary-size.md
+++ b/docs/v3/binary-size.md
@@ -51,3 +51,127 @@ only import `github.com/urfave/cli/v3` do not pull in that package.
 Shell completion support is part of the core package. Leave
 `EnableShellCompletion` disabled unless the application needs shell completion,
 then measure the result with the commands above.
+
+
+## Deeper Analysis with go-size-analyzer
+ 
+The commands above show symbol-level sizes, but they do not clearly show which
+packages contribute most to the binary. [go-size-analyzer](https://github.com/Zxilly/go-size-analyzer)
+provides a package-level breakdown.
+ 
+Install it with:
+ 
+ 
+```sh-session
+go install github.com/Zxilly/go-size-analyzer/cmd/gsa@latest
+```
+ 
+Then analyze a binary:
+ 
+```sh-session
+gsa myapp
+```
+ 
+For example, a minimal `urfave/cli/v3` binary showed:
+ 
+```
+21.58%  runtime                  1.0 MB
+ 6.57%  reflect                  315 kB
+ 6.25%  .rodata                  300 kB
+ 5.87%  github.com/urfave/cli/v3 282 kB
+ 5.74%  text/template            275 kB
+```
+ 
+This helps identify which packages are worth investigating. Packages such as
+`runtime` are fundamental to a Go binary and are not normally removable.
+For an interactive view, use:
+ 
+```sh-session
+gsa --tui myapp
+```
+ 
+> **Note:** analyze an unstripped binary when possible. Stripping debug
+> information can make package attribution less accurate.
+ 
+## Dependency Versions and Multiple Modules
+ 
+A repository may contain multiple Go modules, each with its own `go.mod`.
+Check the modules and their selected versions with:
+ 
+```sh-session
+go list -m all
+```
+ 
+For a specific dependency:
+ 
+```sh-session
+go list -m -json github.com/urfave/cli/v3
+```
+ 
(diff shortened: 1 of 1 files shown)


dearchap:
## Review

Thanks — nice, well-scoped doc that fills in the remaining binary-size topics from #2365. I verified the claims against the repo and a live build rather than trusting them.

### Content accuracy — confirmed
- `docs/go.mod` does contain `replace github.com/urfave/cli/v3 => ../`, so the "documentation module" example is correct.
- I built a minimal one-flag `cli/v3` app against the current checkout: ~4.4 MB unstripped vs the ~4.6 MB stated (toolchain variance). The `reflect` ~315 kB / `text/template` ~275 kB numbers also line up with `go tool nm -size`.
- `HideHelp`/`HideHelpCommand` genuinely do not change binary size — I built both variants and got near-identical sizes (only the build-id differs). The "runtime option, not compile-time" explanation is accurate.
- `go-size-analyzer`, `gsa --tui`, and the unstripped-binary caveat are all accurate.

### CI safety
- New `sh-session` fences are not executed by `gfmrun` (only `go`/`sh`/`bash` are), and the new `go` blocks lack `package main`, so they are ignored. `check-binary-size` is unaffected (docs only).

### Issues worth fixing before merge

1. **Trailing whitespace** — 34 added lines are blank lines containing a single trailing space, plus 4 extra blank lines at EOF. CI won't catch it, but it adds noise to the diff; please trim.
2. **Section ordering (minor)** — the new sections are appended *after* "Current v3 Build Tags", which reads like a concluding/status section. The deep-dive sections would flow better inserted before it, and "Hiding Built-in Help" pairs naturally with the build-tags section.
3. **Fence language (minor)** — the `replace github.com/urfave/cli/v3 => ../` snippet is a `go.mod` directive but fenced as `go`; a plain fence (or non-`go` tag) would be less misleading.
4. **Redundancy (nit)** — "Reflection and Templates" repeats the exact sizes already shown in the `go-size-analyzer` section; could tighten.
5. **Precision (nit)** — "Go normally selects a single version ... for a build" is fine, but could note this is per-main-module MVS resolution, so the `go list -m` check matters mainly when building the same binary from different modules.

### Verdict
Solid, well-researched content — approve after the trailing-whitespace cleanup (1). Items 2–5 are optional polish.

imsumedhaa:
@dearchap I have done those changes. Could you please review it.

Juneezee (approved):

kolyshkin:
I _think_ the biggest issue with using `reflect` is it might disable DCE (dead code elimination) optimization. So, this PR mentions reflection but does not says why it's important (it actually disables DCE).

To re-enable DCE, we need to stop using text/template. It's a big change...

kolyshkin:
> I _think_ the biggest issue with using `reflect` is it might disable DCE (dead code elimination) optimization. So, this PR mentions reflection but does not says why it's important (it actually disables DCE).
> 
> To re-enable DCE, we need to stop using text/template. It's a big change...

Done in #2441

PR #2432, opened 2026-09-18
Fix inverse bool flag counting with an external counter

jakezwang:
## What type of PR is this?

- bug

With `BoolWithInverseFlag.Config.Count` configured, `Count()` stays at zero and `OnlyOnce` accepts repeated flags because only the external counter is incremented.

- Track the flag's invocation count independently of the optional external counter.
- Add `Command.Run` regressions for positive/inverse names and aliases, duplicate rejection, and a pre-populated external counter.

No existing issue.

## Testing

Go 1.27 and 1.26: vet, full race tests, and binary-size checks. Go 1.27: generated docs check, README examples, goimports, API diff check, and golangci-lint 2.13.2 (0 issues). The regression fails against the unchanged implementation.

## Release Notes

```release-note
Fix BoolWithInverseFlag.Count and OnlyOnce when an external count is configured.
```


flag_bool_with_inverse.go
```diff
@@ -83,18 +83,14 @@ func (bif *BoolWithInverseFlag) inversePrefix() string {
 }
 
 func (bif *BoolWithInverseFlag) PreParse() error {
-	count := bif.Config.Count
```
dearchap, line 86: I'm not clear why this needs to be removed. In the absence of user providing a count pointer we use one internally

flag_bool_with_inverse.go
```diff
@@ -153,6 +149,7 @@ func (bif *BoolWithInverseFlag) Set(name, val string) error {
 		}
 		bif.nset = true
 	}
+	bif.count++
```
dearchap, line 152: This is going to increase count irrespective of whether the flag is set or not?

dearchap (approved):
Approved: fix correctly tracks internal invocation count independently of external Config.Count, matching FlagBase semantics. Minimal change with good regression coverage.

PR #2434, opened 2026-09-19
fix: parse single-rune Unicode flags

### LINKED ISSUE #2433: Single-rune Unicode flags can be emitted by completion but not parsed
## My urfave/cli version is

v3.12.0 (`9b981a04295de5121091c479f8a7a0476d2e8d38`)

## Checklist

- [x] I am running the latest v3 release.
- [x] I checked the v3 manual.
- [x] I searched existing issues and pull requests for this problem.

## Dependency Management

- My project is using Go modules.

## Describe the bug

Single-rune Unicode flag names are handled inconsistently. Shell completion counts runes and emits a single-dash flag such as `-ש`, but help formatting counts bytes and displays `--ש`. More importantly, the parser examines `rune(firstArg[1])`, which converts only the first UTF-8 byte rather than decoding the first rune.

For `ש`, the first byte is `0xD7`; converting that byte to a rune produces `×`, which is not a letter. The parser therefore classifies `-ש` as a positional argument and never sets the registered flag.

## To reproduce

```go
package main

import (
    "context"
    "fmt"

    "github.com/urfave/cli/v3"
)

func main() {
    const name = "ש"
    cmd := &cli.Command{
        Name:  "app",
        Flags: []cli.Flag{&cli.StringFlag{Name: name}},
        Action: func(_ context.Context, cmd *cli.Command) error {
            fmt.Printf("value=%q args=%q\n", cmd.String(name), cmd.Args().Slice())
            return nil
        },
    }
    _ = cmd.Run(context.Background(), []string{"app", "-" + name, "value", "operand"})
}
```

## Observed behavior

```text
value="" args=["-ש" "value" "operand"]
```

Help formatting also prefixes this single-rune name with `--`, while flag completion uses `-`.

## Expected behavior

A one-rune Unicode flag should be treated consistently with a one-byte ASCII flag:

```text
value="value" args=["operand"]
```

Help and completion should both render it as `-ש`.

## Additional context

`prefixFor` uses `len(name) == 1`, while completion already uses `utf8.RuneCountInString`. In `parseFlags`, decoding the first rune after the leading dash instead of converting `firstArg[1]` would preserve existing ASCII behavior and correctly classify Unicode letters.

## Want to fix this yourself?

Yes. The fix is small and can be covered by a focused regression test.

```text
go version go1.26.5 windows/amd64
```

## Relevant `go env` output

```text
GOOS=windows
GOARCH=amd64
GOVERSION=go1.26.5
CGO_ENABLED=0
```

znnnnnnn-wil:
## What type of PR is this?

- bug

- Count Unicode code points, rather than UTF-8 bytes, when selecting the prefix for a flag name.
- Decode the first rune after `-` before deciding whether an argument is a flag.
- Add a regression test showing that a single-rune Hebrew flag is displayed and parsed like a single-character ASCII flag.

Some code and tests were assisted by Codex. I reviewed, understood, and tested the complete change and take responsibility for it.

Fixes #2433

The parsing change only affects non-ASCII input. ASCII arguments decode to the same rune value as the byte previously inspected.

## Testing

- `go test . -run '^TestCommand_SingleRuneUnicodeFlag$' -count=1`
- `go test ./... -count=1`
- `go run scripts/build.go vet`
- `go run scripts/build.go check-binary-size`
- `go run scripts/build.go generate`
- `go run scripts/build.go gfmrun`
- Equivalent public API diff using `git diff --no-index --ignore-all-space` (no differences)

The repository's `go run scripts/build.go test` command requires `-race`; it could not run on this Windows host because CGO is disabled and no C compiler is installed. The same packages passed without `-race`, and CI will run the race-enabled suite on supported runners.

## Release Notes

```release-note
Fix parsing and help formatting for single-rune Unicode flag names.
```


dearchap (approved):
Approved: minimal, correct rune-based fix for prefix and flag classification with end-to-end regression test.

